Issue #449, opened 2026-05-21
Fn::GetStackOutput cross-account RoleArn support


go-to-k:
## Problem

`Fn::GetStackOutput` rejects the optional `RoleArn` argument with a clear "cross-account references via RoleArn are not yet supported" error. CDK apps that consume outputs from a producer stack in a sibling AWS account (the canonical multi-account pattern: shared-services account producing platform outputs, workload account consuming them) cannot reference those outputs through cdkd today.

The strategy is documented in CLAUDE.md but unimplemented: assume the producer-account's role via STS, then read the producer account's separate cdkd state bucket (`cdkd-state-{producerAccountId}`).

## Current behavior

`src/deployment/intrinsic-function-resolver.ts:1567-1574`:

```typescript
if (roleArn) {
  throw new Error(
    `Fn::GetStackOutput: cross-account references via RoleArn are not yet supported by cdkd ` +
    `(StackName=${stackName}, Region=${region}, RoleArn=${roleArn}). ` +
    `cdkd reads outputs from S3 state instead of CloudFormation DescribeStacks, ` +
    `so cross-account requires assuming the role and reading the producer account's ` +
    `state bucket — not yet implemented.`
  );
}
```

Even though the argument is validated (`src/deployment/intrinsic-function-resolver.ts:1557-1564`) and resolved to a string, it always rejects.

## Expected behavior

`Fn::GetStackOutput` with `RoleArn` set:

1. Resolves `RoleArn` to a literal string (already done at lines 1557-1564).
2. Calls `sts:AssumeRole` to obtain temporary credentials in the producer account.
3. Resolves the producer's state bucket name via the cdkd convention (`cdkd-state-{producerAccountId}` — `accountId` extracted from the role ARN).
4. Calls `s3:GetObject` against the producer's bucket at `cdkd/{stackName}/{region}/state.json` using the assumed credentials.
5. Reads the requested output from the producer's `state.outputs`.
6. Caches the assumed credentials per `(roleArn, region)` for the duration of the deploy (STS credentials are valid for 1 hour by default).

## Implementation strategy

1. Extract account ID from `RoleArn` via the standard `arn:aws:iam::{accountId}:role/{name}` regex.
2. In `src/utils/aws-region-resolver.ts` (or a new sibling helper), add `resolveCrossAccountStateBucket(accountId, optionalRegion)` returning the cdkd-convention bucket name; auto-detect region via `GetBucketLocation` after the STS hop succeeds.
[...]

PR #482, opened 2026-05-21
feat(intrinsic): Fn::ImportValue cross-region resolution (opt-in) (#451)

### LINKED ISSUE #451: Fn::ImportValue cross-region resolution (opt-in)
## Problem

`Fn::ImportValue` resolution in `cdkd deploy` assumes same-region: the persistent exports index lives at `s3://{bucket}/cdkd/_index/{region}/exports.json`, scoped to one region per file. A consumer stack deployed to `us-east-1` that does `Fn::ImportValue: 'ProducerStack-BucketName'`, where the producer is deployed to `us-west-2`, cannot resolve the import — the consumer's `us-east-1` index has no entry, and cdkd has no signal that the producer might live elsewhere.

## Current behavior

`src/state/export-index-store.ts:159` and `src/deployment/intrinsic-function-resolver.ts:resolveImportValue` are single-region by construction. The resolver scans the configured region's index and falls back to per-stack state scan in the SAME region; it never crosses regions.

Note that `Fn::GetStackOutput` already supports cross-region — it takes an explicit `Region` argument. `Fn::ImportValue` (the CFn standard) does not encode region in its argument, which is why CFn itself is same-region-only for ImportValue too.

## Expected behavior

A configurable opt-in mechanism lets `cdkd deploy` resolve `Fn::ImportValue` against indices in other regions when the same-region lookup misses. The user explicitly opts in because CFn semantics are same-region; cdkd should not silently diverge from CFn behavior by default.

## Implementation strategy

1. New CLI flag `--import-value-cross-region <regions>` (comma-separated list, e.g. `us-west-2,eu-west-1`) on `cdkd deploy`. Off by default.
2. Equivalent `cdk.json` setting `context.cdkd.importValueCrossRegion: ["us-west-2", ...]` and env var `CDKD_IMPORT_VALUE_CROSS_REGION=us-west-2,eu-west-1`.
3. In `src/state/export-index-store.ts`, add a `crossRegionFallback?: string[]` config option. When set, the index store keeps one `ExportIndexStore` instance per region.
4. In `src/deployment/intrinsic-function-resolver.ts:resolveImportValue`, on a same-region miss, iterate the configured `crossRegionFallback` regions and look up each. First match wins; if multiple regions have the same export, error out with a clear ambiguity message naming each region.
5. Once resolved cross-region, record the resolution in `state.imports[]` with the producer's region (the existing v4 schema field already carries `sourceRegion`).
6. `cdkd destroy`'s strong-reference check already iterates every region's state via `listStacks`; no change needed there.
7. Document the trade-off in `docs/cross-stack-references.md`: cross-region `Fn::ImportValue` is non-standard CFn behavior and a deliberate cdkd opt-in.

## Test strategy

- Unit test the multi-region resolver: same-region miss + cross-region hit succeeds; same-region hit short-circuits (no cross-region read).
- Unit test the ambiguity case: same export name in two configured regions errors with the per-region detail.
- Unit test the deliberate-skip case: cross-region disabled → same-region miss errors normally (no fallback scan).
- Integ test: extend `tests/integration/import-value-strong-ref/` with a 2-region branch (`CDKD_INTEG_MULTI_REGION=1`), producer in one region, consumer in another, opt-in via the new flag.

## Acceptance criteria

- [ ] New flag + env var + cdk.json setting accepted; same-region behavior unchanged when flag is off.
- [ ] Same-region miss + configured-region hit resolves successfully.
- [ ] Multi-region match (same export name in two regions) errors with a clear ambiguity message.
- [ ] `state.imports[].sourceRegion` records the cross-region resolution (already supported by the v4 schema).
- [ ] `docs/cross-stack-references.md` documents the trade-off and the opt-in flag.

## Related

- Memory rule `feedback_default_on_unless_semantic_change.md` — cross-region `Fn::ImportValue` IS a semantic change vs CFn; deliberately keep off by default.
- Memory rule `feedback_design_alternatives_full_palette.md` — full scan / configured-allowlist / cdk.json-only — the configured-allowlist option is the right trade-off for blast radius.
- PR (#343) — persistent exports index, the foundation this issue extends.
- W2-9 (`--from-state` cross-stack intrinsics) — sibling work; once both ship the local emulator gets cross-region for free.
- W2-11 (`Fn::GetStackOutput` cross-account) — sibling cross-stack work on a different axis.

go-to-k:
## Summary

Opt-in cross-region resolution for `Fn::ImportValue` within the same
AWS account. Off by default — CloudFormation's `Fn::ImportValue` is
same-region by construction and cdkd does not silently diverge.

When opted in, the resolver consults a per-region `ExportIndexStore`
for each configured foreign region after the same-region index AND
`state.json` scan both miss. Ambiguity (same export name in two
configured regions) fails with a clear error pointing at
`Fn::GetStackOutput` as the region-explicit alternative.

Resolution chain (highest wins):

- `--import-value-cross-region us-west-2,eu-west-1` CLI flag
- `CDKD_IMPORT_VALUE_CROSS_REGION` env var
- `cdk.json context.cdkd.importValueCrossRegion` (string or JSON array)
- default empty (same-region only — pre-PR behavior)

Same-account only; cross-account is tracked separately under (#449).

## Behavior

- Same-region resolves short-circuit cross-region entirely — no
  perf or blast-radius change when the flag is off.
- Strong-reference bookkeeping records the producer's actual region
  in `state.imports[].sourceRegion` (schema v4 unchanged), so
  destroy-time refusal scans the right state file regardless of
  whether the import resolved same-region or cross-region.
- The consumer's own region is stripped from the resolved list
  automatically (listing it would no-op and false-fire ambiguity).
- Foreign-region lookup errors degrade gracefully (WARN + skip);
  remaining configured regions are still walked.

## Test plan

- [x] `vp run typecheck`
- [x] `vp run test` — 306 files / 4005 tests, 0 failures
  - 7 new unit tests in `tests/unit/deployment/intrinsic-import-value-cross-region.test.ts`
  - 11 new tests in `tests/unit/cli/config-loader.test.ts` (resolveImportValueCrossRegion resolution chain)
- [x] `vp check --fix` (lint + format)
- [x] `vp run build`
- [x] `vp run cli-flag-coverage` regen (new flag declared)
- [x] `vp run integ-coverage` (no drift)
- [x] `vp run scenario-coverage` (no drift)
- [ ] Real-AWS integ deferred (per task constraint: do not run real-AWS integ)

Closes #451


go-to-k:
Closed because the design diverges from CloudFormation's `Fn::ImportValue` semantics.

CFn's `Fn::ImportValue` is structurally same-region-only — the argument carries no region encoding, and CFn exports are region-scoped. Extending the cdkd resolver to do cross-region lookups (even as an opt-in flag) would silently change the behavior of templates that real CFn would reject.

cdkd already exposes the right cross-region path: `Fn::GetStackOutput` takes an explicit `Region` argument and resolves through the same S3 state backend. Users who need cross-region cross-stack references should use `Fn::GetStackOutput`, not a CFn-incompatible `Fn::ImportValue` extension.

Per the project's CDK-CLI / CFn parity rule, cdkd should not invent semantics CFn doesn't have. Issue #451 closed with the same rationale.

PR #841, opened 2026-06-13
test(integ): multi-asset failure-seeking integration test

go-to-k:
## Summary

A new failure-seeking integration test, `tests/integration/multi-asset/`, that
stresses cdkd's asset-publishing layer when MANY assets of MULTIPLE kinds
publish concurrently in a single `cdkd deploy`. This is TEST-ONLY: no `src/`
changes.

Where the existing `docker-image-asset` fixture exercises the ECR build+push
path alone and `s3-asset-deploy` exercises the S3 zip path alone, this fixture
forces `FileAssetPublisher` and `DockerAssetPublisher` to run together in one
run, exercising the concurrent multi-asset path that neither narrow fixture
covers.

## Coverage

The `CdkdMultiAssetExample` stack publishes 1 ECR image + 4 S3 objects in one
deploy:

- **ECR / Docker image asset** — a `lambda.DockerImageFunction` built from a
  local `Dockerfile` (`public.ecr.aws/lambda/nodejs:20` base), routed through
  `DockerAssetPublisher` (build + push to ECR).
- **3 Python Lambda ZIP assets** — three `lambda.Function`s (Python 3.12), each
  `Code.fromAsset('<distinct local dir>')`. Each directory is a genuine
  multi-file tree (handler + `pkg/` sub-package) with distinct content, so each
  produces a distinct asset hash and a separate `FileAssetPublisher` S3 upload.
- **1 generic S3 file asset** — a generic `s3_assets.Asset` (not Lambda code),
  consumed by one Lambda via env, exercising the asset-ref intrinsics for a
  non-code asset.

## Cross-architecture Docker pin fix

The fixture's Docker setup pins BOTH the build `platform`
(`ecr_assets.Platform.LINUX_ARM64`) AND the Lambda `architecture`
(`lambda.Architecture.ARM_64`) so they match.

Lesson: a Docker image asset must be built for the Lambda's target
architecture. A default `DockerImageFunction` emits no `source.platform` and no
`Architectures`, so on an Apple-Silicon (arm64) host cdkd builds for the host
arch and pushes an arm64 image to an x86_64 Lambda, which fails at invoke time
with `Runtime.InvalidEntrypoint: ProcessSpawnFailed`. Pinning
`--platform` / `Architecture` to a matched pair avoids the trap.

## Validation

This integration test was already validated GREEN against real AWS (deploy +
destroy, no leftover resources). The live-test requirement is satisfied.

Coverage matrices (`docs/_generated/integ-coverage.json`,
`docs/_generated/scenario-coverage.json`) and the scenario list were
regenerated and are up to date.


go-to-k:
Review note: pr-review-gate classifies this as 3-axis purely by file count (29 files, but all are fixture boilerplate: Lambda code, Dockerfile, asset data, configs + generated coverage matrices). Per maintainer go-ahead, this fixture-only PR (no src logic) was reviewed by 1 pr-code-reviewer; the one blocker found (EXIT-trap exit-code masking) is fixed in a2a9e4b69efb07687e2d0367df16eb8d7b494106. Setting the pr-review marker manually per the gate's documented escape hatch for fixture-only PRs where 3-axis is disproportionate.

PR #842, opened 2026-06-13
test(integ): sg-circular-dependency failure-seeking integration test

go-to-k:
## Summary

Test-only PR (no `src/` change). Adds a failure-seeking integration test
`tests/integration/sg-circular-dependency/` that stresses cdkd's create/destroy
DAG ordering with the classic CloudFormation "circular Security Group" shape:
SG-A allows ingress from SG-B AND SG-B allows ingress from SG-A.

The cycle is broken the CloudFormation-safe way: each ingress rule is emitted as
a STANDALONE `AWS::EC2::SecurityGroupIngress` resource (not inline), so the two
SGs can exist before the cross-references are added. In CDK,
`sgA.addIngressRule(sgB, ...)` + `sgB.addIngressRule(sgA, ...)` against two
distinct SG constructs makes CDK emit standalone ingress resources (each
`Fn::GetAtt`s `GroupId` on the SG it attaches to and `SourceSecurityGroupId` on
the OTHER SG). The stack is a `natGateways: 0` single-AZ VPC + SG-A + SG-B + the
two cross-referencing ingress resources (no EC2 instances, so no instance cost).

## What it stresses

- **Create ordering**: the DAG builder (`src/analyzer/dag-builder.ts`) must NOT
  raise a false `DependencyError` -- the standalone ingress resources break what
  would otherwise be a genuine SG-to-SG cycle.
- **Destroy ordering (the key test)**: the cross-referencing ingress rules must
  be revoked / deleted BEFORE the SGs. If cdkd deletes an SG while its
  cross-referencing ingress rule is still live, AWS rejects
  `DeleteSecurityGroup` with
  `DependencyViolation: resource sg-xxx has a dependent object`, so a wrong
  delete order fails or orphans here. This exercises the existing
  `AWS::EC2::SecurityGroup -> AWS::EC2::SecurityGroupIngress` implicit-delete-dep
  edge in `src/analyzer/implicit-delete-deps.ts` end-to-end on real AWS.

`verify.sh` (BSD/macOS-portable, captures real exit codes, prints PASS only on
full success): Phase 0 synth-asserts the cycle-breaking shape (>= 2 standalone
ingress resources each carrying a `SourceSecurityGroupId`); Phase 1 deploys and
asserts both SGs exist with the live cross-reference; Phase 2 destroys and
asserts 0 errors plus both SGs + the VPC + the state file gone. Resources are
located by the `cdkd:integ-fixture=sg-circular-dependency` tag, and the EXIT-trap
cleanup revokes-then-deletes both SGs directly so a destroy-ordering bug never
leaks billing resources.

## ASCII-description fix

While authoring the fixture, AWS rejected the deploy because the SecurityGroup
`GroupDescription` strings contained a non-ASCII em-dash (U+2014). AWS only
accepts ASCII characters in `GroupDescription`, so both descriptions are now
ASCII-only (plain hyphen). Lesson recorded: integ fixture strings that AWS
validates (such as SecurityGroup descriptions) must be ASCII.

## Validation

Validated GREEN against real AWS: `cdkd deploy` and `cdkd destroy` both
completed cleanly with 0 orphan resources. The live-test requirement is
satisfied.

New scenario tag `sg-circular-dependency` added to the canonical taxonomy
(`scripts/build-scenario-coverage-matrix.ts`); the integ-coverage and
scenario-coverage matrices were regenerated. Typecheck / lint / build / unit
tests (5755 tests) all pass.


go-to-k:
Review note: pr-review-gate classifies this 3-axis by file count, but it is a fixture-only PR (no src logic). Per maintainer go-ahead, reviewed by 1 pr-code-reviewer; the one minor finding (missing zero-inline-ingress assertion) is fixed and the assertion now enforced. Setting pr-review marker manually per the gate's documented fixture-only escape hatch. (rebased on main: 8a29a1c502ceaeff15a5a300e4a75be0052bb565)

Issue #381, opened 2026-05-15
Bug: IntrinsicFunctionResolver fails to resolve nested attribute paths (Endpoint.Port etc.) for AWS::RDS::DBCluster


go-to-k:
## Summary

`Fn::GetAtt: [<RDSCluster>, 'Endpoint.Port']` resolves to the cluster's physical ID instead of the actual port (3306). This breaks any downstream resource that consumes the cluster's port via `Fn::GetAtt`, including CDK's auto-generated `AWS::EC2::SecurityGroupIngress` "IndirectPort" rules emitted by `Connections.allowDefaultPortFrom(...)`.

## Reproduction

Deploy `tests/integration/rds-aurora` (any cdkd version, observed on v0.102.6 / commit 523cb15):

```
node dist/cli.js deploy --region us-east-1 --state-bucket <bucket> --verbose
```

Verbose log shows:

```
[DeployEngine] [13/19] ✅ AuroraCluster23D869C0 (AWS::RDS::DBCluster) created
[IntrinsicFunctionResolver] Resolved Fn::GetAtt: AuroraCluster23D869C0.Endpoint.Port -> rdsaurorastack-auroracluster23d869c0
[EC2Provider] Creating SecurityGroupIngress AuroraSecurityGroupfromRdsAuroraStackAuroraSecurityGroupBEF0D396IndirectPort7A3A2446
[DeployEngine] Failed to create AuroraSecurityGroupfromRdsAuroraStackAuroraSecurityGroupBEF0D396IndirectPort7A3A2446:
  Failed to create SecurityGroupIngress ...: Invalid integer value rdsaurorastack-auroracluster23d869c0
```

The "Resolved Fn::GetAtt: …" (no "from attributes:" prefix) means the resolver fell through to `constructAttribute` which returned the bare physicalId as the default fallback for unknown attribute paths.

## Root cause (best guess)

`src/provisioning/providers/rds-provider.ts:createDBCluster` returns:

```typescript
attributes: {
  'Endpoint.Address': described?.Endpoint ?? '',
  'Endpoint.Port': String(described?.Port ?? ''),
  ...
}
```

These are flat string keys. `IntrinsicFunctionResolver.resolveGetAtt` (`src/deployment/intrinsic-function-resolver.ts:494`) reads `resource.attributes[attributeName]` where `attributeName === 'Endpoint.Port'`. The flat-key lookup SHOULD work — but the log shows it falls through to `constructAttribute`. Possibilities:

[...]

PR #844, opened 2026-06-13
fix(provisioning): enrich AWS::RDS::DBInstance Endpoint attributes on the Cloud Control path

go-to-k:
## Summary

Enriches an `AWS::RDS::DBInstance`'s computed endpoint attributes on the Cloud Control provisioning path so a downstream `Fn::GetAtt(<db>, 'Endpoint.Address')` resolves to the real endpoint hostname instead of the raw physical id.

## Root cause

`CloudControlProvider.enrichResourceAttributes` (`src/provisioning/cloud-control-provider.ts`) had a per-type case for `AWS::RDS::DBCluster` (added per #381) but **none** for `AWS::RDS::DBInstance`.

When a `DBInstance` template sets even one #614 silent-drop top-level property (e.g. `BackupRetentionPeriod` / `CopyTagsToSnapshot` / `MultiAZ` / `PubliclyAccessible` / `StorageType`), the #614 routing rule sends the entire resource through the Cloud Control API and bypasses `RDSProvider.create`. The SDK provider would normally populate the flat-key `Endpoint.Address` / `Endpoint.Port` attributes after create; on the CC-API path those attributes were never set, so a downstream consumer doing `Fn::GetAtt(<DBInstance>, 'Endpoint.Address')` fell through the resolver's `constructAttribute` branch to the `physicalId` (the DB identifier, not the endpoint hostname).

## Fix

Add a parallel `AWS::RDS::DBInstance` case to `enrichResourceAttributes` that issues `DescribeDBInstances` after create and flattens the result into the SDK provider's flat-key attribute shape:

- `Endpoint.Address`
- `Endpoint.Port`
- `Endpoint.HostedZoneId`
- `Arn`

Shape note: unlike the DBCluster case (whose `Endpoint` / `ReaderEndpoint` are flat scalars), `DescribeDBInstances` returns `Endpoint` as a NESTED object `{ Address, Port, HostedZoneId }`, so the case flattens it. The lookup is best-effort: a failed `Describe` leaves the CC-API attribute shape unchanged and never fails the deploy (the resolver's nested-path walk is the second line of defence), exactly matching the DBCluster branch.

## Tests

- Unit tests in `tests/unit/provisioning/cloud-control-provider.test.ts` cover the new DBInstance case (endpoint/port/hosted-zone/ARN flattening from the nested `Endpoint` object, plus the best-effort Describe-failure path).
- New `rds-full-stack` failure-seeking integ (`tests/integration/rds-full-stack/**`): a VPC (no NAT) + explicit `rds.SubnetGroup` + explicit `rds.ParameterGroup` (Postgres, non-default `application_name`) + SecurityGroup + a small `rds.DatabaseInstance` (`db.t3.micro`, 20 GiB, CDK-managed Secrets Manager credentials, no final snapshot) + an `ssm.StringParameter` whose value is `Fn::GetAtt(<Database>, Endpoint.Address)`. The integ asserts the SSM parameter value equals the LIVE `DescribeDBInstances` endpoint address (proving the computed `Fn::GetAtt` resolved post-create), then destroys and asserts the instance, subnet group, parameter group, SSM parameter, and state file are all gone with 0 orphans. New scenario tag `rds-full-stack`; coverage matrices regenerated.

## Validation

Validated GREEN end-to-end against real AWS (`/run-integ rds-full-stack`): deploy clean, the computed-endpoint `Fn::GetAtt` resolved to the live endpoint address, destroy reported 15 deleted / 0 errors, and the post-destroy orphan sweep was empty (0 orphans).


go-to-k:
Independent review complete: pr-code-reviewer + pr-test-reviewer ran (3-axis; spec-reviewer N/A for a bug fix without a design doc). No blockers. One minor nit accepted as known cost: the rds-full-stack verify.sh post-destroy DBInstance assertion accepts `deleting` as well as `gone` while the adjacent comment notes cdkd waits for terminal NotFound. The actual real-AWS run reported `status: gone` (15 deleted / 0 errors / 0 orphans) and the EXIT-trap cleanup is the backstop, so tightening it would only add flakiness risk if AWS is briefly slow. Setting pr-review marker bound to 4659e90ed9209168cfb9a0e6ec81133212a2da4f.

PR #846, opened 2026-06-13
fix(deployment): resolve composite conditions + prune Condition-gated resources on update (#840)

### LINKED ISSUE #840: fix(deployment): resource whose Condition flips to false on update may not be deleted (needs confirm vs test artifact)
## Bug (needs confirmation: cdkd vs test artifact)
The new `conditions-and-if` integ deploys with `-c tier=premium` then redeploys `-c tier=basic`. A resource gated on a `Condition:` that is TRUE in premium and FALSE in basic is expected to be DELETED on the basic redeploy. Observed:
```
FAIL: PremiumOnlyParam ... STILL EXISTS but should be removed in basic tier
```
Either (a) cdkd's diff/deploy does not delete a resource whose `Condition:` evaluated true on the prior deploy and false on this one (a real semantic gap — CloudFormation removes it), OR (b) the `-c tier=basic` context flip did not actually change the synthesized template (test-side). Needs investigation: synth both ways + confirm the condition truly flips + check whether cdkd's diff marks the now-condition-excluded resource for DELETE.

If (a) is real: cdkd must treat "resource present in prior state but condition-excluded from the new template" as a DELETE in the diff phase.

go-to-k:
## Summary

Fixes two related cdkd defects around CloudFormation Conditions (issue #840), plus a failure-seeking integ and unit tests.

### 1. Composite conditions referencing other named conditions resolve correctly and order-independently

`src/deployment/intrinsic-function-resolver.ts` + `src/analyzer/template-parser.ts`. A `Conditions` section entry can compose other named conditions via `{Condition: X}` inside `Fn::And` / `Fn::Or` / `Fn::Not` (for example `IsPremiumPrimary: { Fn::And: [ { Condition: IsPremium }, { Condition: IsPrimaryRegion } ] }`). cdkd previously did not resolve the `{Condition: X}` reference form when evaluating composite conditions, and the result also depended on the order conditions appeared in the template. Conditions are now evaluated so that a `{Condition: X}` reference resolves to the already-evaluated boolean of condition `X` regardless of declaration order, matching CloudFormation semantics.

### 2. A resource whose `Condition:` flips `true -> false` on a redeploy is now pruned / deleted

`src/analyzer/template-parser.ts` (new `filterResourcesByCondition`) + `src/deployment/deploy-engine.ts`. CloudFormation does not strip condition-gated resources at synth time -- CDK emits a resource carrying a `Condition:` key into `Resources` regardless of the condition value. cdkd evaluated the `Conditions` section for `Fn::If` resolution but never consulted the resource-level `Condition:` key, so a condition-false resource was created on first deploy and never deleted when its condition flipped (it stayed in the desired set and diffed as `NO_CHANGE`). The deploy engine now prunes every resource whose `Condition:` resolved to `false` (via `TemplateParser.filterResourcesByCondition`) immediately after evaluating the `Conditions` section, so the whole downstream pipeline (type/property validation, DAG build, diff, provisioning) operates on the CFn-effective resource set: a condition-false resource is never created, and one present in prior state but condition-excluded from the effective template falls through the diff's existing "present in state, absent from desired -> DELETE" path exactly as CloudFormation removes it. A resource whose `Condition:` names an unknown/unevaluated condition is kept (treated as present, not silently dropped).

## Tests

- New `conditions-and-if` integ (`tests/integration/conditions-and-if/**`): a cheap stack (3x `AWS::SSM::Parameter` + 1x `AWS::SNS::Topic`, no VPC / NAT) whose `Conditions` section combines `Fn::Equals` on a `CfnParameter` with `Fn::And` / `Fn::Or` / `Fn::Not`, carries two resources with a `Condition:` key, an always-created parameter whose `Value` is an `Fn::If` branch, and an SNS topic whose `DisplayName` is `Fn::If(IsPremium, ..., AWS::NoValue)`. `verify.sh` runs three phases against real AWS: Phase 1 (`-c tier=premium`) asserts the `Fn::If` branch, both condition-gated parameters present, and the SNS `DisplayName` set; Phase 2 (`-c tier=basic`, in-place redeploy) asserts the `Fn::If` branch flipped, both condition-gated parameters now absent (resource removed because its `Condition:` went false), and the `DisplayName` genuinely omitted (`AWS::NoValue`); Phase 3 destroys and asserts every named resource is gone and the state file removed.
- New `filterResourcesByCondition` unit tests (`tests/unit/analyzer/template-parser.test.ts`): false -> removed, true -> kept, no-Condition -> kept, unknown-condition -> kept, preserves Conditions+Outputs and does not mutate input.
- New diff-calculator test (`tests/unit/analyzer/diff-calculator.test.ts`): a resource in state but absent from the pruned template diffs as `DELETE`.
- New composite-condition `{Condition: X}` resolution unit tests covering the order-independent evaluation.

## Validation

- Validated GREEN end-to-end against real AWS: `/run-integ conditions-and-if` passed (deploy + redeploy flip + destroy clean, 0 orphans).
- The cross-cutting `bench-cdk-sample` broad integ ran clean in this worktree (deploy + destroy, 0 orphans), so the `integ-broad` and `integ-destroy` gates are satisfied.
- Full local check passes: typecheck, lint (0 errors), build, and the unit suite (5776 tests). Coverage matrices regenerated.

Closes #840


go-to-k:
Independent 3-axis review complete (code + test; spec N/A for a bug fix). Test review: clean, with the prune-on-UPDATE path integ-asserted (deploy condition=true -> redeploy condition=false -> resource DELETED on AWS). Code review found one latent hazard: the {Condition: X} reference branch fired for every property, so a resource property literally {Condition: "string"} could be coerced to false. Fixed in d97cf5ac5823a244cadfc79202b006eccc6336ff by gating that branch on context.conditionResolver presence (set only during evaluateConditions, the sole valid context) + a regression unit test. Setting pr-review bound to d97cf5ac5823a244cadfc79202b006eccc6336ff.

PR #847, opened 2026-06-13
fix(deployment): Fn::Join over list-returning intrinsics + AWS::NotificationARNs in Fn::Sub (#838)

### LINKED ISSUE #838: fix(deployment): Fn::Join with a list-returning intrinsic (Fn::Cidr/GetAZs/Split) as its 2nd arg crashes 'values.map is not a function'
## Bug
`IntrinsicFunctionResolver.resolveJoin` (src/deployment/intrinsic-function-resolver.ts ~L1182) destructures `[delimiter, values]` and calls `values.map(...)` — assuming the 2nd arg is always a literal array. But CloudFormation allows `Fn::Join`'s second argument to be a SINGLE intrinsic that RETURNS a list: `Fn::Cidr`, `Fn::GetAZs`, `Fn::Split`, or a `Ref` to a `CommaDelimitedList` parameter.

So `{ "Fn::Join": [",", { "Fn::Cidr": ["10.0.0.0/16", 8, 8] }] }` fails at deploy with:
```
Failed to create <X>: values.map is not a function
```

## Repro
The new `intrinsics-torture` integ's `cidr-join` SSM parameter (`Fn::Join(',', Fn::Cidr(...))`) fails deploy on this.

## Fix
In `resolveJoin`, if `values` is not an array, first `resolveValue(values, context)` (it may be a list-returning intrinsic) and only then `.map`. Hard-error if the resolved value is still not an array. CloudFormation parity.

go-to-k:
## What

Two real fixes to cdkd's hand-rolled intrinsic-function resolver
(`src/deployment/intrinsic-function-resolver.ts`), both surfaced by a new
failure-seeking `intrinsics-torture` integration test:

### (a) issue #838 -- `Fn::Join` over a list-returning intrinsic

`Fn::Join` with a list-returning intrinsic (`Fn::Cidr` / `Fn::GetAZs` /
`Fn::Split`) as its second argument crashed with `values.map is not a
function`. The resolver assumed the second argument was already an array and
called `.map()` on it, but a list-returning intrinsic is an object
(`{ "Fn::Cidr": [...] }`) until it is resolved. The fix resolves the second
argument FIRST, then maps over the resulting list. Closes #838.

### (b) `AWS::NotificationARNs` in `Fn::Sub` / `Ref`

`AWS::NotificationARNs` resolved to `undefined`, which left the literal
`${AWS::NotificationARNs}` placeholder in an `Fn::Sub` output (the pseudo
branch is skipped on `undefined`, and the subsequent `Ref` attempt throws and
keeps the placeholder). cdkd has no stack-notification-ARN concept, so the
list is always empty -- and CloudFormation resolves an empty
`AWS::NotificationARNs` list to an empty string in a string context. The
resolver now returns `''` (empty string) for `AWS::NotificationARNs` in both
`Fn::Sub` and bare `Ref` contexts, matching CloudFormation's own behavior
instead of leaking an unresolved placeholder downstream.

## Tests

- **Unit** (`tests/unit/deployment/intrinsic-functions.test.ts`): `Fn::Join`
  over `Fn::Cidr` / `Fn::GetAZs` / `Fn::Split`, and `AWS::NotificationARNs`
  resolving to `''` in both `Fn::Sub` and `Ref`.
- **Integration** (`tests/integration/intrinsics-torture/**`): a new
  failure-seeking fixture (an SNS topic + SQS queue + ten
  `AWS::SSM::Parameter`s, no VPC / NAT / Lambda) that stress-tests the
  less-common + deeply-nested intrinsics the existing `intrinsic-functions`
  fixture never exercised -- `Fn::Cidr`, `Fn::FindInMap`, `Fn::GetAZs` +
  `Fn::Select`, `Fn::Base64`, nested `Fn::Split` + `Fn::Select` + `Fn::Join`,
  a deeply-nested two-arg `Fn::Sub`, and ALL pseudo-parameters. Each SSM
  parameter is built with the raw CFn escape hatch
  (`new ssm.CfnParameter` + `addPropertyOverride('Value', <intrinsic>)`) so the
  synth template carries the EXACT intrinsic shape under test. `verify.sh`
  reads each parameter back and asserts it equals a value computed
  independently from the account / region, so a wrong resolution pinpoints
  which intrinsic cdkd got wrong; it then destroys and asserts clean (state
  gone + zero orphan SSM parameters).
- New scenario tag `intrinsics-torture`; coverage matrices regenerated;
  testing-guide entry added to `docs/testing.md`.

## Validation

- **Real AWS**: the `intrinsics-torture` integ passed GREEN end-to-end
  (`/run-integ intrinsics-torture`) -- deploy clean, every intrinsic resolved
  to the expected value, destroy clean with 0 orphans.
- **Broad**: the `bench-cdk-sample` broad integ passed clean in the same
  worktree (`integ-broad` + `integ-destroy` markers fresh).
- Local checks: typecheck / lint / build / full test suite (5771 tests) all
  green.

Closes #838


go-to-k:
Independent 3-axis review complete (code re-review + test review; spec N/A for a bug fix). No blockers. Code review confirmed the Fn::Join 2nd-arg list-intrinsic resolution (no literal-array regression, throws on non-list) and AWS::NotificationARNs -> '' parity (strictly better: the bare-Ref path previously threw). Test review confirmed exact-value unit + integ assertions for both fixes incl. real-AWS readback. Setting pr-review bound to c4fdf1a1a4c13aafb5f480926a5d66bf4ebdc7d1.

PR #848, opened 2026-06-14
fix(provisioning): retry KMS fresh-principal policy + bind EC2 IAM instance profile post-launch

go-to-k:
## Summary

This branch fixes two real cdkd bugs that surface when a resource consumes a freshly-created IAM principal/profile, plus adds the `propagation-races-2` integ fixture and unit tests.

## Bug 1: KMS Key create fails on fresh-IAM-principal key policy (propagation race)

Creating a KMS Key whose key policy references a just-created IAM principal failed with `Policy contains a statement with one or more invalid principals`. The IAM principal exists but has not yet propagated to the region KMS validates against, so the first `CreateKey` is rejected.

Fix: added the fresh-principal propagation pattern to `src/deployment/retryable-errors.ts` so the create is retried until the principal is visible to KMS.

## Bug 2: EC2 instance silently launches without its fresh IAM instance profile

`IamInstanceProfile` in the template is a NAME string, not a `{Arn, Name}` object. The prior code passed the raw value through in a shape the SDK ignored, so the instance launched with no instance profile bound and no error was raised.

Fix in `src/provisioning/providers/ec2-provider.ts`:
- normalize the `IamInstanceProfile` shape (accept a name string),
- after launch, describe-then-associate the profile and poll the association until it reports associated, so a fresh profile that has not yet propagated still binds deterministically.

## Tests

- `propagation-races-2` integ fixture (4 fresh-principal/consumer race edges) with `verify.sh`.
- Unit tests: `tests/unit/deployment/retryable-errors.test.ts` (KMS fresh-principal retry classification) and `tests/unit/provisioning/providers/ec2-provider-instance-profile-bind.test.ts` (shape normalization + post-launch associate/poll).

## Verification

Validated green against real AWS (deploy + destroy clean) on the `propagation-races-2` fixture.


go-to-k:
Independent code review complete (pr-code-reviewer): both fixes clean, no blockers. KMS + IAM-instance-profile retry patterns well-anchored; normalizeIamInstanceProfile handles all 3 shapes; post-launch describe-associate-poll is bounded (10 attempts each), idempotent, and correctly treats 'associating' as not-yet-bound. 3 minor nits accepted as known cost: a dead defensive branch (ec2-provider.ts:2521), worst-case ~90s propagation wall-time (within the per-resource path), and a slightly-broad 'not authorized' substring (bounded by maxAttempts). Plus unit tests + real-AWS integ PASS. Setting pr-review bound to c7bddabdc0058b252f6d01d16f1a0e57f04781d1.

PR #851, opened 2026-06-14
fix(deployment): resolve CompositeAlarm/EC2 GetAtt attributes + CompositeAlarm delete ordering

go-to-k:
## Summary

Fixes two real cdkd bugs surfaced by deep `Fn::GetAtt` computed-attribute chains, plus adds the failure-seeking `deep-getatt-chains` integ fixture and unit tests.

## Bug 1: Fn::GetAtt returned the physicalId for several computed attributes

The intrinsic-function resolver fell through to returning the resource's `physicalId` for attributes it had no per-type case for, silently producing wrong values:

- `AWS::CloudWatch::CompositeAlarm` `Arn` -- now constructs the `arn:<partition>:cloudwatch:<region>:<account>:alarm:<name>` ARN instead of returning the alarm name (physicalId).
- `AWS::EC2::Instance` `PrivateIp` / `PublicIp` / `PrivateDnsName` / `PublicDnsName` / `AvailabilityZone` -- now resolved via a live `DescribeInstances` call instead of returning the instance id (physicalId).

Files: `src/deployment/intrinsic-function-resolver.ts`, `src/deployment/deploy-engine.ts`.

## Bug 2: CompositeAlarm deleted after the metric Alarms it references

A `CompositeAlarm`'s `AlarmRule` names its child metric Alarms as plain strings (no `Ref` / `Fn::GetAtt`), so the analyzer built no dependency edge and could delete the child Alarms before the CompositeAlarm. AWS rejects deleting an Alarm still referenced by a CompositeAlarm's rule.

Added an implicit delete-ordering rule so a `CompositeAlarm` is deleted BEFORE the metric Alarms its `AlarmRule` references (parsed out of the rule string).

Files: `src/analyzer/implicit-delete-deps.ts`, `src/deployment/deploy-engine.ts`, `src/cli/commands/destroy-runner.ts`.

## Tests

- New integ fixture `tests/integration/deep-getatt-chains/` exercises long computed-attribute chains through CompositeAlarm and EC2 Instance.
- Unit tests added for the new resolver cases and the new implicit delete-ordering rule (`tests/unit/deployment/intrinsic-functions.test.ts`, `tests/unit/analyzer/implicit-delete-deps.test.ts`).
- Integ-coverage / scenario-coverage matrices regenerated.

## Verification

Validated end-to-end (deploy + destroy clean, 0 orphans) against real AWS on this branch.

## Note

The EC2 `PrivateIp` GetAtt fix in this PR is also required by the upcoming `deletion-ordering-complex` fixture; merging this PR first lands that fix on main.


go-to-k:
Independent code review complete (pr-code-reviewer): both fixes clean, no blockers. CompositeAlarm ARN shape correct; EC2 live-lookup bounded+cached+falls-back; AlarmRule parsing handles bare/quoted/ARN + ignores TRUE/FALSE + self-edge-guarded; delete-edge wired into BOTH deploy-engine and destroy-runner with correct direction. No regressions (purely additive constructAttribute cases). 2 minor nits accepted: duplicate-alarm-name last-wins (AWS names are unique, harmless) + absent-attr physicalId fallback (narrow acceptable case). Plus unit tests + real-AWS integ PASS. Setting pr-review bound to c6ca6235371c79fdf42fb9b07a0ecc0011becc46.

PR #854, opened 2026-06-14
test(integ): custom-resource-getatt-data failure-seeking integration test

go-to-k:
## Summary

Test-only, failure-seeking integration fixture that exercises `Fn::GetAtt` of
a Custom Resource's `Data` keys resolving through to dependent resources during
a cdkd deploy. No `src/**` changes.

cdkd resolves CloudFormation intrinsic functions itself (no CloudFormation in
the loop), so the path where a Custom Resource returns `Data` attributes and a
downstream resource consumes them via `Fn::GetAtt(<CR>, 'Data.<key>')` is a
distinct resolver case that deserves dedicated real-AWS regression coverage.

## What the fixture proves

- A Custom Resource (`AWS::CloudFormation::CustomResource` backed by a Lambda)
  returns one or more `Data` keys from its provider.
- A downstream resource references those keys via `Fn::GetAtt` of the CR's
  `Data.<key>` attribute.
- cdkd's intrinsic-function resolver materializes the CR `Data` keys at deploy
  time and propagates the resolved values to the dependent resource.
- A wrong or unresolved CR `Data` value would fail the dependent's create /
  the post-deploy assertion, so the fixture is genuinely failure-seeking rather
  than a happy-path smoke test.

`verify.sh` is BSD/macOS-portable (real exit-code capture plus an explicit
`[verify] PASS` summary line) and the fixture deploys with no VPC to keep the
run cheap.

## Validation

Validated green against real AWS: deploy + destroy completed cleanly with no
orphan resources. Regenerated integ-coverage and scenario-coverage matrices
(0 orphans) and added a changelog entry.

## Scope

- `tests/integration/custom-resource-getatt-data/**` (new fixture)
- `docs/changelog-cdkd.md`, `docs/_generated/*`, `docs/*coverage*.md`
  (regenerated matrices + changelog)
- `scripts/build-scenario-coverage-matrix.ts` (matrix regeneration)

Do not merge yet.


go-to-k:
Independent code review (pr-code-reviewer): clean, test-only, no src. verify.sh proves CR Data-key GetAtt resolved into all 3 SSM dependents on real AWS (exact value match) + state.outputs cross-check; portable, 0-orphan. 1 cosmetic README count typo accepted. pr-review override (test-only, validated PASS) bound to 95edc0d19cfac36000ea08779547b8c7b6299dbf.

Issue #807, opened 2026-06-12
fix(deployment): replacement of a referenced resource does not propagate to dependents marked NO_CHANGE (ECS Service keeps running old TaskDefinition revision)


sakurai-ryo:
## Summary

When a resource that gets a new physical ID on replacement (e.g. `AWS::ECS::TaskDefinition`) is updated, the replacement itself works — but dependents that reference it via `Ref` / `Fn::GetAtt` are diffed as `NO_CHANGE` and skipped entirely. For ECS this means: a task definition change registers a new revision, but `UpdateService` is never called with the new revision ARN, so no new deployment starts and tasks keep running on the old (now deregistered / INACTIVE) revision.

CloudFormation propagates the new physical ID to dependents (the Service's `TaskDefinition` property changes → `UpdateService` → new deployment rolls out new tasks). cdkd currently lacks this semantic.

## Reproduction

1. Deploy a stack with an `AWS::ECS::Cluster` + `AWS::ECS::TaskDefinition` + `AWS::ECS::Service` (service references the task definition via `Ref`).
2. Change a task definition property (e.g. container image tag) and run `cdkd deploy` again.
3. Observed:
   - TaskDefinition is replaced (new revision registered, old revision deregistered) ✅
   - Service is reported `NO_CHANGE` and skipped — `UpdateService` is never issued ❌
   - The service keeps running tasks on the old revision and now references an INACTIVE task definition ❌

## Root cause

- The diff is calculated once, **before** any provisioning, with intrinsics resolved against the **current state** (`src/deployment/deploy-engine.ts:740`, `src/analyzer/diff-calculator.ts:110`). The Service's `TaskDefinition: {Ref: TaskDef}` resolves to the **old** revision ARN, which equals the stored state value → `NO_CHANGE` (`src/analyzer/diff-calculator.ts:143-153`).
- `NO_CHANGE` resources are excluded from the execution DAG up front (`src/deployment/deploy-engine.ts:1012`).
- When a resource is replaced, the engine only logs the physical-ID swap (`src/deployment/deploy-engine.ts:2168`) — there is no cascade to dependents.

Note: if the dependent happens to have any *other* property change, it is marked `UPDATE` and provisioning-time re-resolution picks up the new ARN correctly. The bug only manifests when the dependency change is the *only* change.

## Scope

[...]

PR #855, opened 2026-06-14
test(integ): replacement-fanout failure-seeking integration test

go-to-k:
## What

Adds a failure-seeking integration fixture, `tests/integration/replacement-fanout/`, that stresses replacement propagation (issue #807) at FAN-OUT scale. Test-only: no `src/` changes. Validated PASS against real AWS.

## Why

#807 fixed the basic replacement-propagation case (`promoteReplacementDependents` in `src/analyzer/diff-calculator.ts`: a dependent whose only "change" is a `Ref`/`Fn::GetAtt`/`Fn::Sub` to a to-be-REPLACED resource is promoted `NO_CHANGE` -> `UPDATE` so it re-points at the new physical id). It was verified end-to-end with a SINGLE dependent (the ECS `Service` -> replaced `TaskDefinition` case in `ecs-fargate`). A single-dependent test cannot surface a PARTIAL-propagation gap, where the new physical ref reaches some dependents but not all.

## How

The fixture deploys one base resource referenced by many dependents:

- `BaseTopic` is an `AWS::SNS::Topic` with an explicit, phase-keyed `TopicName` (`cdkd-replacement-fanout-{region}-a`).
- 10 `AWS::SSM::Parameter`s set `Value` to `Fn::Sub("arn=<topicArn>|idx=N", { arn: Ref(BaseTopic) })` (SNS `Ref` resolves to the topic ARN).
- An `AWS::SNS::TopicPolicy` whose policy `Resource` is the topic ARN.

A `-c phase=a|b` flip (read at synth time) renames the topic `-a` -> `-b` on phase b. `TopicName` is in the SNS entry of the replacement-rules registry, so the rename forces delete + recreate -> a NEW topic ARN. The 10 parameters are auto-named, so they keep their physical id and take an in-place `Value` update; only the embedded ARN changes.

`verify.sh` (BSD/macOS-portable: no `grep -P` / `date -d`, real exit codes captured to variables, explicit `[verify] PASS`):

1. Deploy `-c phase=a`; capture the base ARN and every dependent's AWS-resolved `Value` (baseline asserts all embed the phase-a ARN, TopicPolicy references it).
2. Redeploy `-c phase=b`; assert the base ARN CHANGED (old topic gone, new present) AND every dependent re-resolved to the NEW ARN on AWS. A dependent left on the stale phase-a ARN FAILS naming its index (the #807 fan-out gap); each parameter keeps its physical id; the TopicPolicy re-points at the new ARN with the stale one absent.
3. Destroy; assert the state file, base topic, and all 10 parameters are gone (per-index orphan attribution).

New scenario tag `replacement-fanout-propagation` in `KNOWN_SCENARIOS` (`scripts/build-scenario-coverage-matrix.ts`); coverage matrices regenerated.

## Test

- typecheck / lint / build / full unit suite: pass (387 files, 5839 tests).
- Coverage matrices fresh.
- `verify.sh` validated PASS against real AWS.


go-to-k:
Code review clean (pr-code-reviewer): test-only, no src; verify.sh proves real-AWS fan-out re-resolution across all 10 dependents on replacement (phase-a->phase-b, old topic gone, per-index orphan attribution), portable, 0-orphan. 1 nit accepted (unguarded env.region, /run-integ sets it). pr-review override (test-only, validated PASS) bound to 3dc79d98973cfbc942ea69fd5e71d703c56a8fc8.

PR #856, opened 2026-06-14
test(integ): tags-propagation failure-seeking integration test

go-to-k:
## Summary

Adds a failure-seeking integration test fixture, `tags-propagation`, that verifies
stack-level Tags propagate across 9 taggable resource types on both the SDK
provider path and the Cloud Control API path, and that the deployed stack is
drift-clean.

This is a TEST-ONLY change. There are no `src/**` modifications. The fixture
sets tags at the stack level and asserts each taggable type carries the
propagated tag values after deploy, exercising the routing gradient
(SDK-registered providers and CC-API fallback) so a regression in tag
propagation on either path is caught against real AWS.

## What it covers

- Stack-level `Tags.of(stack).add(...)` propagation to 9 taggable types.
- Both provisioning paths: SDK providers and the Cloud Control API fallback.
- Drift-clean assertion on the deployed stack (no false-positive tag drift).

## Validation

- Validated PASS against real AWS (deploy + verify + destroy clean).
- Local: typecheck, lint, build, and full unit test suite pass.
- Coverage matrices (integ-coverage / scenario-coverage) regenerated and committed.

## Notes

- No `Closes #` reference.
- Squash-merge only per repo policy.


go-to-k:
Code review clean (pr-code-reviewer): test-only, no src; 9-type tag readback via type-specific APIs (correct {Key,Value}[] vs map shapes), routing split asserted (Athena cc-api / 8 sdk), drift exit 0 (#802 no false positive), portable, 0-orphan. Minor accepted (absent-type false-OK in routing loop — deploy fails first). pr-review override (test-only, validated PASS) bound to 865fb67d9116d7e531d731663e53454e3ac42b32.

PR #857, opened 2026-06-14
test(integ): getstackoutput-crossregion failure-seeking integration test

go-to-k:
## Summary

Adds a failure-seeking, test-only integration fixture
(`tests/integration/getstackoutput-crossregion/**`) that exercises
cdkd's same-account / CROSS-REGION `Fn::GetStackOutput` read path.

A producer stack publishes an output in one region; a consumer stack in
a DIFFERENT region resolves that output value at deploy time via
`Fn::GetStackOutput`. This is a cdkd-unique capability (CloudFormation
itself has no cross-region output read), so the fixture is a regression
guard for the cross-region output-resolution code path.

The fixture is failure-seeking: the consumer wires the resolved
cross-region value into a real resource so that a broken or stale
resolution would surface as a deploy failure or a wrong-value assertion
in `verify.sh`, not a silent pass.

## Scope

- TEST-ONLY. No `src/**` changes.
- New fixture under `tests/integration/getstackoutput-crossregion/**`
  (CDK app, stack, `verify.sh`, `.scenarios.json`, README).
- Regenerated integ-coverage and scenario-coverage matrices to include
  the new fixture (`multi-stack-getstackoutput`,
  `getstackoutput-cross-region` scenarios).
- Changelog entry recorded.

## Validation

- Validated PASS against real AWS (deploy + cross-region resolution +
  destroy).
- Local check: typecheck / lint / build pass; full unit suite green
  (one unrelated environmental timeout in `version.test.ts` that passes
  in isolation).
- Coverage matrices fresh (0 orphan).

Do NOT merge yet.


go-to-k:
Code review clean (pr-code-reviewer): test-only, no src; verify.sh deploys producer (us-west-2) + consumer (us-east-1) as separate deploys, asserts the consumer resolved the producer's REAL us-west-2 ARN cross-region (belt-and-suspenders :us-west-2: check rejects same-region fallback), cleans BOTH regions' params+state. 1 cosmetic nit accepted. pr-review override (test-only, validated PASS) bound to 6e993bba29584c40273508e45fa55a165430d4f0.

PR #859, opened 2026-06-14
test(integ): update-policy-mutations failure-seeking integration test

go-to-k:
## Summary

Adds a failure-seeking, real-AWS integration fixture
`tests/integration/update-policy-mutations/**` that exercises cdkd's
UPDATE-time handling of CloudFormation template-level ATTRIBUTES (not
properties) that change between two deploys. Test-only: no `src/**`
changes. Validated PASS against real AWS.

The two deploys are driven by a CDK context flip (`-c phase=a` then
`-c phase=b`), read at synth time, so the second deploy synthesizes a
mutated template with no source change. Cheap resources only (S3 / SSM /
SNS), so each assertion isolates the policy / replacement behavior.

## What it tests

1. `UpdateReplacePolicy: Retain` orphan-on-replace - an S3 `BucketName`
   change forces replacement; the OLD physical bucket must be RETAINED on
   AWS (not deleted) while the new one is created.
2. `DeletionPolicy` flip DESTROY -> RETAIN on an SSM Parameter - the final
   destroy must honor the CURRENT (Retain) policy and leave it on AWS.
3. `DependsOn` add / remove between SNS topics - a metadata-only change
   that must update successfully without replacing either topic (both keep
   their ARNs).
4. metadata-only / no-op identical redeploy reporting `No changes
   detected`.

## Why

The existing fixtures under-cover this surface: `update-replace` covers
property mutations and `deletion-policy-retain` covers a single-deploy
Retain skip, but neither exercises a template-level attribute that
CHANGES across two deploys. This fixture is a regression net for
`diff-calculator.ts` attribute diff plus the `deploy-engine.ts`
Retain-on-replace and DeletionPolicy destroy-skip paths.

## Orphans

The test creates intentional orphans by design (the Retain-replaced old
bucket and the phase-b Retain SSM parameter + Retain bucket). The
`verify.sh` trap deletes every captured / deterministic physical id, and
the final step asserts 0 leftovers via direct AWS API checks.

## Verification

- Local: typecheck, lint, build, full unit test suite (5839 passing) all
  green; integ-coverage and scenario-coverage matrices regenerated.
- Real AWS: `verify.sh` validated PASS end-to-end (deploy phase-a ->
  deploy phase-b -> identical redeploy -> destroy -> 0 leftovers).


go-to-k:
Code review clean (pr-code-reviewer): test-only, no src; verify.sh asserts all 4 update-policy mutation cases vs real AWS (Retain-replace bucket survives, Retain vs Delete param positive/negative pair, no spurious SNS replacement on DependsOn update, no-op redeploy 'No changes detected'), 0-orphan. 2 cosmetic nits accepted (stale prose). pr-review override (test-only, validated PASS) bound to 2dde079000af75656e3ea2891ac4bc5f98ba623b.

Issue #809, opened 2026-06-12
Cloud Control UPDATE silently drops write-only properties (read-modify-write patch): ECS Service loses VolumeConfigurations and the deploy fails


sakurai-ryo:
## Problem

`CloudControlProvider.update` (`src/provisioning/cloud-control-provider.ts`) generates the JSON Patch from **cdkd's stored previous properties vs the new desired properties**, so the patch contains only the paths that changed. Cloud Control API applies that patch with read-modify-write semantics: it calls the type's read handler to get the current model, applies the patch, and hands the result to the update handler as the desired state.

Read handlers cannot return **write-only properties**, so any write-only property that is not explicitly present in the patch document vanishes from the desired state on every CC-routed UPDATE. This is a documented Cloud Control API caveat — write-only properties must be re-included in every patch — that cdkd's minimal-diff patch generation does not honor.

## Real-world failure

`AWS::ECS::Service` declares `writeOnlyProperties: [/properties/ServiceConnectConfiguration, /properties/VolumeConfigurations, /properties/ForceNewDeployment]`. On a service with a managed EBS volume (`VolumeConfigurations`, CDK's `ServiceManagedVolume`), a deploy whose only service change was the task definition ARN produced this patch:

```
Generated 2 patch operations for RunnerBackendRunnerServiceA291189C: [
  {"op":"replace","path":"/TaskDefinition","value":"arn:aws:ecs:...:task-definition/dev-main-CI-Runner-backend-runner-task:9"},
  {"op":"add","path":"/ServiceName","value":"dev-main-CI-Runner-RunnerBackendRunnerServiceA291189C"}
]
```

No `VolumeConfigurations` in the patch -> not in the desired state the handler receives -> `UpdateService` starts a new deployment without volume configurations:

```
UPDATE failed for RunnerBackendRunnerServiceA291189C: Invalid request provided: Task definition has configuredAtLaunch volume but no volume configuration provided at runtime. (Service: Ecs, Status Code: 400, ...) (SDK Attempt Count: 1)
```

The stack is then wedged: the same UPDATE is re-attempted and fails on every subsequent deploy (state still records the old service properties), and there is no CLI workaround.

ECS happens to hard-error here. The scarier variant is **silent config loss**: for types whose handler accepts the write-only-less desired state (e.g. a secret value, a password, a provisioning token), the property is removed from the live resource without any error.

CloudFormation does not hit this because it passes the full template-derived desired model to the handler directly — no read-modify-write.

[...]

PR #900, opened 2026-06-21
test(integ): add lambda-destinations fixture (EventInvokeConfig via Cloud Control)

go-to-k:
## Summary

Adds an integration fixture `lambda-destinations` for Lambda async destinations
(`onSuccess` / `onFailure` -> SQS), which CDK synthesizes as an
`AWS::Lambda::EventInvokeConfig`. cdkd has no SDK provider for that type, so it
routes via **Cloud Control**. Its `DestinationConfig` is a write-only property
Cloud Control read handlers cannot return, so an in-place UPDATE that changes
`MaximumRetryAttempts` must re-include it in the patch or the destinations are
silently dropped (the `write-only-properties.ts` re-include path; issue #809).

Found by a `/hunt-bugs` sweep that confirmed the path works end-to-end but had
**zero integ coverage** — a regression in the write-only re-include would
otherwise be silent. `AWS::Lambda::EventInvokeConfig` is now covered in the
integ-coverage matrix.

## What `verify.sh` checks (3 phases)

1. **Deploy** (retryAttempts=2): assert the EventInvokeConfig reached AWS with
   `MaximumRetryAttempts=2`, `MaximumEventAgeInSeconds=300`, both `OnSuccess` +
   `OnFailure` destinations wired, and routed via `cc-api` (state
   `provisionedBy`). Functional: async-invoke with a success payload and confirm
   the `onSuccess` record lands in the success SQS queue (`condition=Success`).
2. **UPDATE** (retryAttempts 2 -> 1): assert the change reached AWS AND the
   write-only `DestinationConfig` **survived** the in-place patch (the
   regression this fixture pins).
3. **Destroy**: assert the function / EventInvokeConfig and cdkd state are gone;
   sweep `/aws/lambda/<fn>*` log groups.

## Test plan

- `/run-integ lambda-destinations` against real AWS (`us-east-1`): PASS in 80s,
  all 3 phases, destroy 0 errors / 0 orphans. Ledger updated.
- `vp check` (typecheck/lint/format) clean; `vp run build` OK; `vp run test`
  6030/6030 pass.
- Coverage matrices regenerated (`integ-coverage`, `scenario-coverage`).

No `src/` changes — fixture + auto-generated docs only.


go-to-k:
Code review (1-reviewer tier, no design doc → downgraded from 3-axis): **clean, no blockers**. Three minor nits, all accepted as known cost:

1. `onFailure` destination is asserted-wired (Phase 1 + Phase 2 string-match on both OnSuccess/OnFailure) but not functionally driven by a failing invoke — accepted because onFailure delivery requires async retries to exhaust (minutes) and would balloon the integ runtime; the issue-#809 write-only-survival regression is already covered by the Phase-2 match on **both** destinations, and the success path is functionally exercised.
2. `/tmp/cdkd-lambda-dest-out.json` (the `aws lambda invoke` sink) is left behind — harmless, overwritten on re-run.
3. The `state show --json` parser is first-match-only for the EventInvokeConfig resource — correct for this single-config fixture; emits a clear FAIL (not a crash) if `provisionedBy` is absent.

PR #905, opened 2026-06-21
fix(deploy): resolve Ref of AppConfig compound-id resources to the trailing id segment

go-to-k:
## Summary

Fixes a compound-id `Ref` bug (same class as #897 / #898) for AWS AppConfig.
AppConfig resources are all Cloud-Control-provisioned with COMPOUND CC primary
identifiers, while CFn `Ref` returns only the trailing id:

| Type | CC physical id | `Ref` returns |
|------|----------------|---------------|
| `AWS::AppConfig::Environment` | `<appId>\|<envId>` | env id |
| `AWS::AppConfig::ConfigurationProfile` | `<appId>\|<profileId>` | profile id |
| `AWS::AppConfig::HostedConfigurationVersion` | `<appId>\|<profileId>\|<ver>` | version number |
| `AWS::AppConfig::Deployment` | `<appId>\|<envId>\|<deployNum>` | deployment number |

None were in `REF_RETURNS_SEGMENT_AFTER_PIPE`, so `Ref` returned the raw
compound. **Found by a `/hunt-bugs` sweep**: a `HostedConfigurationVersion`
wiring `ConfigurationProfileId: { Ref: <Profile> }` received the compound
`<appId>|<profileId>` and the CREATE failed with `Configuration Profile ... could
not be found`, rolling back the whole chain.

## Fix

- Add the four compound AppConfig types to `REF_RETURNS_SEGMENT_AFTER_PIPE`
  (`Application` + `DeploymentStrategy` stay out — simple pipe-free ids).
- Change the extraction from `indexOf('|')` (after-first) to `lastIndexOf('|')`
  (after-last) so it is correct for the **3-segment** AppConfig compounds; for
  the existing 2-segment ApiGateway/Cognito types after-first == after-last, so
  no behavior change.

## Test plan

- **Unit** (+3): ConfigurationProfile + Environment (2-segment), and
  HostedConfigurationVersion (3-segment, proves the last-pipe extraction). Full
  suite 6037 pass.
- **Integ** `appconfig`: deploys the full chain (the regression proof — pre-fix
  the Version CREATE failed), UPDATE bumps the hosted config version, destroy
  clean. Real AWS (us-east-1): **PASS 63s, 0 orphans**.
- **Broad integ** `lambda` (re-run because `intrinsic-function-resolver.ts` is
  cross-cutting): **PASS, 9 deleted / 0 errors**.
- `vp check` clean; build OK.


go-to-k:
Code review (1-reviewer, downgraded from 3-axis file-count — no design doc): **no blockers**. Reviewer verified all 6 AppConfig primaryIdentifiers against live CC `describe-type` and confirmed the trailing-segment Ref + the `indexOf`→`lastIndexOf` change are correct (identical for the existing 2-segment types). Nits addressed in 79459da:
1. verify.sh cleanup deleted the DeploymentStrategy by NAME (always 400s; dead fallback that would leak the auto-id'd strategy on an interrupted run) → now lists + deletes by Id. Re-ran integ: PASS 61s, strategy/app/state all gone.
2. Added a unit test for `AWS::AppConfig::Deployment` (4th type, 3-segment) so the test set matches the type set 1:1.
3. Accepted: Deployment.Ref isn't consumed by the fixture, but shares the exact 3-segment code path as HostedConfigurationVersion (whose Ref IS consumed + asserted live).

PR #909, opened 2026-06-21
fix(provisioning): map DynamoDB SSESpecification.SSEEnabled to the SDK Enabled field

go-to-k:
## Summary

Fixes a **silent security downgrade** in the DynamoDB table provider. CFn's
`AWS::DynamoDB::Table.SSESpecification` names the enable flag `SSEEnabled`, but
the DynamoDB SDK `CreateTable` field is `Enabled`. The provider passed the CFn
shape verbatim, so the SDK **silently ignored** the unknown `SSEEnabled` key and
the table was created with **AWS-owned (default) encryption** instead of the
requested AWS-managed / customer-managed KMS encryption — no error, no warning.

Found by a `/hunt-bugs` sweep: `TableEncryption.AWS_MANAGED` (synthesizes
`SSESpecification: { SSEEnabled: true }`) deployed "successfully" but
`describe-table` showed no `SSEDescription` (AWS-owned), not the expected
`Status=ENABLED, SSEType=KMS`. CFn maps this field internally; cdkd bypasses CFn
so must map it itself.

## Fix

`mapSSESpecification()` maps `SSEEnabled` → `Enabled`, passes `SSEType` /
`KMSMasterKeyId` through (same names), tolerates the stringified-boolean form,
and omits the field for an absent / non-object value. The readback already emits
the correct CFn shape, so state round-trips unchanged.

## Test plan

- **Unit** (+2): CreateTable input carries `{ Enabled: true }` (not
  `SSEEnabled`); `mapSSESpecification` maps all three subfields + edge cases.
  Full suite 6040 pass.
- **Integ** `dynamodb-sse`: deploy a table with `AWS_MANAGED` encryption, assert
  `describe-table` SSEDescription.Status=ENABLED + SSEType=KMS, destroy clean.
  Real AWS (us-east-1): **PASS 43s, 0 orphans**.

Note: `update()` does not yet apply SSESpecification changes on `UpdateTable` —
a separate pre-existing limitation, not addressed here.


go-to-k:
Code review (1-reviewer, provider security path): **no blockers**. Reviewer verified: `'false'`->`Enabled:false` (explicit equality, not Boolean-coercion), no create/readback phantom drift (state stays CFn-cased), unit+integ pin the fix at both layers. Dispositions:
- Nit (empty-object SSESpecification -> `{}` set; harmless SDK no-op): accepted as-is per reviewer ("could fix later").
- Pre-existing gap: `update()` does not apply SSESpecification changes on UpdateTable (encryption change on an existing table is a silent no-op) — out of scope for this CREATE-path fix; will file a follow-up issue.

PR #914, opened 2026-06-21
fix(provisioning): apply DynamoDB SSESpecification changes on update (UpdateTable)

### LINKED ISSUE #910: DynamoDB: update() does not apply SSESpecification changes on UpdateTable
Follow-up from PR #909 (which fixed the CREATE-path `SSEEnabled`->`Enabled` mapping).

`DynamoDBTableProvider.update()` never wires `SSESpecification` into `UpdateTableCommandInput`, so changing encryption on an **existing** table — e.g. AWS_MANAGED -> CUSTOMER_MANAGED, or rotating `KMSMasterKeyId` — is a silent no-op (the same class of silent security gap #909 fixed, on the UPDATE path).

The readback Class-1 guard comment in dynamodb-table-provider.ts already anticipates 'a future SSE-aware update()'.

Scope: map + apply SSESpecification via UpdateTable on update() when it changes, with a unit test + an integ UPDATE-phase assertion (extend `dynamodb-sse`). Found by /hunt-bugs sweep 2026-06-21.

go-to-k:
## Summary

Follow-up to #909 (closes #910). #909 fixed the DynamoDB **create-path**
`SSEEnabled`→`Enabled` mapping, but `update()` never wired `SSESpecification`
into `UpdateTable` — so changing encryption on an **existing** table
(AWS_MANAGED → CUSTOMER_MANAGED, or rotating the KMS key) was a silent no-op,
the same class of silent security gap on the UPDATE path.

## Fix

`update()` now fires its own `UpdateTable` with the mapped `SSESpecification`
when the value changed (reusing `mapSSESpecification`), then waits for ACTIVE so
the GSI block doesn't race an UPDATING table. A removal is a deliberate no-op
(CFn has no clean "drop SSE to AWS-owned" mapping; mirrors WarmThroughput).
Also tightens the create-path guard to `Object.keys(sse).length > 0` (the #909
review nit) so an SSESpecification object with no recognized keys doesn't send
an empty `{}`.

## Test plan

- **Unit** (+2): update() with a changed SSESpecification fires `UpdateTable`
  with the mapped `{ Enabled, SSEType, KMSMasterKeyId }`; an unchanged
  SSESpecification fires no UpdateTable. Full suite 6042 pass.
- **Integ**: the existing `dynamodb-sse` fixture (create AWS_MANAGED + destroy)
  exercises this provider end-to-end and satisfies integ-destroy. **PASS 41s,
  0 orphans.**
- **Why no integ for the key-flip:** switching a DynamoDB table's KMS key
  triggers a slow background re-encryption (`SSEDescription.Status=UPDATING` for
  many minutes) that **blocks table deletion**, making such a fixture's destroy
  hang and flaky. The update() SSE path is covered by the unit tests; the live
  key-flip was **manually verified** (a real table's `KMSMasterKeyArn` switched
  to the customer CMK after a `CDKD_TEST_UPDATE` deploy).

### DIFF
--- a/docs/_generated/integ-last-run.tsv
+++ b/docs/_generated/integ-last-run.tsv
@@ -1,164 +1,164 @@
 # integ-last-run ledger (update-type: one row per test). cols: test  last_run_iso  result  duration_s  flow  note
-acm-certificate	2026-06-21T11:00:28Z	PASS	20	verify.sh	sweep-4..8 staleness re-run (ledger-restore; rc=0)
-alb	2026-06-15T06:14:28Z	PASS		verify.sh	rc ok, orph clean; #609 ListenerAttributes assert
-alb-advanced	2026-06-21T11:00:28Z	PASS	54	standard	sweep-4..8 staleness re-run (ledger-restore; rc=0)
-api-cognito	2026-06-21T11:00:28Z	PASS	66	standard	sweep-4..8 staleness re-run (ledger-restore; rc=0)
-apigateway	2026-06-21T05:45:25Z	PASS	51	verify.sh	Model/RequestValidator Ref fix; 200/400 asserted; 0 orph
-appconfig	2026-06-21T08:22:45Z	PASS	61	verify.sh	AppConfig compound-id Ref fix: chain creates, UPDATE v2, destroy clean
-appsync	2026-06-21T11:00:28Z	PASS	25	standard	sweep-4..8 staleness re-run (ledger-restore; rc=0)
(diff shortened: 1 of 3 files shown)


go-to-k:
Code review (1-reviewer, provider path): **no blockers**. Verified: SSE update branch correctness (mapSSESpecification reuse, Object.keys guard, own UpdateTable + waitForTableActiveAfterUpdate so GSI doesn't race UPDATING), branch ordering (billing/throughput before, GSI after — no concurrent UpdateTable), removal-no-op consistent with WarmThroughput, no phantom drift vs the CFn-shape readback, and the create-path guard tightening is a safe superset. 2 nits accepted as-is (serial ACTIVE-wait latency note; an unreachable same-key edge).

PR #970, opened 2026-07-02
test(integ): update-replace fixture passes --force-stateful-recreation for the ReplaceBucket rename

go-to-k:
## Summary

The `update-replace` fixture's Phase 1b (ReplaceBucket rename) is a property-driven S3 replacement. Since the stateful-replace guard shipped (2026-06-29, the createOnly-fallback PR), that phase requires an explicit `--force-stateful-recreation` — without it the deploy hard-fails on `STATEFUL_REPLACE_BLOCKED`, so the fixture has been red on main. Mirrors the `replacement-immutable-name` fixture's rename phase, which was updated in the guard PR itself.

Note: this change was originally drafted directly in the main worktree by an earlier session and left uncommitted; it was preserved into this branch, comment-polished, and live-verified before landing.

## Verification

- `bash verify.sh` against real AWS: **PASS** (77s, all 16 assertions, clean destroy, 0 orphans). Ledger updated in the same commit.
- Tests-only change (fixture script + ledger) — no src surface.


go-to-k:
Superseded: PR #968 (merged while this was in flight) already landed the identical --force-stateful-recreation fixture fix plus a verified update-replace ledger row. The only residual delta here was a comment-grammar nit — not worth a PR. This branch's own verify.sh run also passed (77s, 0 orphans), independently confirming the fix.

PR #1005, opened 2026-07-15
feat(local): recognize cdkd-owned + custom-qualifier container-assets images in run-task

### LINKED ISSUE #1002: cdkd-published assets are deleted by `cdk gc` — publish to cdkd-owned asset bucket / ECR repo created by `cdkd bootstrap`
# Problem

cdkd currently publishes file assets and Docker image assets to the **CDK bootstrap** resources (`cdk-<qualifier>-assets-<account>-<region>` S3 bucket / `cdk-<qualifier>-container-assets-<account>-<region>` ECR repo), exactly as the synthesized `*.assets.json` instructs.

`cdk gc` (aws-cdk-cli garbage collection, stable since CDK 2.165.0) determines "in use" by **scanning the CloudFormation stack templates in the environment** for asset hashes. Stacks deployed via `cdkd deploy` have **no CloudFormation stack**, so every cdkd-published asset looks unused and gets tagged isolated → deleted.

Verified against the gc implementation (`aws-cdk-cli` `garbage-collector.ts`):

- In-use detection is CFn-template-scan only (`ActiveAssetCache` built from `ListStacks` + `GetTemplate`).
- There is **no protection / do-not-delete tag mechanism** — only the isolation markers `aws-cdk:isolated` (S3) / `aws-cdk.isolated` (ECR).
- With default `--rollback-buffer-days 0`, deletion is immediate on the first `cdk gc full` run.
- gc discovers the bucket/repo from the **bootstrap stack** (default `CDKToolkit`) outputs — storage not referenced by that stack is structurally out of gc's reach. Only default-bootstrap-shaped destinations are at risk; user-chosen storage (custom `fileAssetsBucketName`, `AppStagingSynthesizer`) is never touched by `cdk gc`.

## Severity

- **ECS**: image deleted from ECR → every new task launch (scale-out, replacement) fails to pull.
- **Lambda container images**: existing functions keep invoking (internal copy), but updates / rollbacks that re-resolve the image fail.
- **Lambda ZIP / S3 assets**: deploy self-heals via re-upload, but runtime readers (`s3.Asset` URLs in env vars, SFN definitions, BucketDeployment sources) break immediately.

# Design — CANONICAL DOC

**The single source of truth is [`docs/design/1002-cdkd-asset-storage.md`](https://github.com/go-to-k/cdkd/blob/claude/cdk-asset-cleanup-design-i1ji2a/docs/design/1002-cdkd-asset-storage.md)** on branch `claude/cdk-asset-cleanup-design-i1ji2a` (all open questions are decided; §13 has the suggested PR phasing). The implementation's first PR must move this doc onto `main` (cherry-pick or copy from that branch). Summary:

## Chosen: cdkd-owned asset storage (option D), transparently gated

- Storage: `cdkd-assets-{accountId}-{region}` (per-region — Lambda same-region code constraint) + `cdkd-container-assets-{accountId}-{region}` (IMMUTABLE tags). Deliberately **not** the state bucket (region constraint, versioning mismatch, lifecycle isolation, blast radius — doc §3).
- **Transparent upgrade (hard requirement)**: upgrading the binary alone changes nothing. `cdkd bootstrap` (new version) additionally creates the asset bucket + ECR repo and writes a marker `s3://cdkd-state-{accountId}/cdkd-bootstrap/{region}.json`. Deploy reads the marker per (account, region): absent → **legacy mode, byte-identical to today** (one info line per invocation); present → cdkd-assets mode; present-but-deleted-resources → hard error, never silent fallback. No state schema bump; old-binary rollback safe both directions; `--use-cdk-bootstrap-assets` per-app pin; `cdkd bootstrap --no-assets` opt-out. (doc §4)
- Bucket-squatting defense: bootstrap refuses owned-elsewhere buckets; every asset-bucket S3 call passes `ExpectedBucketOwner`. (doc §5)
- **Redirection + rewrite are destination-driven, not naming-heuristic**: one mapping table per deploy built from each stack's `assets.json` destinations (the source of truth for ANY synthesizer), consumed by both the publishers and the template rewrite so they cannot diverge. `objectKey` / `imageTag` unchanged (`bucketPrefix` flows through; existence-skip works as-is). (doc §6)
- Template rewrite (pre-resolution, analyzer entry): string nodes + `Fn::Sub` bodies + pseudo-param-folded `Fn::Join` segments, **boundary-aware matching**; **post-resolution audit** fails loudly on any unrewritten mapped name (no split-brain deploys). (doc §7)
- Command coverage: `deploy`/`diff`/`import` rewrite; `synth`/`export` intentionally unrewritten; `publish-assets` reads the marker (legacy fallback + info when no state bucket resolves); `destroy`/`state *`/`drift`/`events` unchanged. (doc §7.1)
- Scope rule: redirect **only default-bootstrap-shaped destinations, any qualifier** (`cdk-[a-z0-9]+-(container-)?assets-…`) — exactly the population `cdk gc` can delete. Custom `fileAssetsBucketName` / `AppStagingSynthesizer` / cross-region (`dest.region ≠ deploy region`) destinations left verbatim. (doc §8)
- Migration: first deploy after re-bootstrap re-publishes (same hash keys) + in-place UPDATE repoints properties; one-time "everything with assets shows an update"; no replacement, no state migration. Accepted edges (rollback-after-gc, mixed binary versions, IAM additions, partition parity) documented in §9.
- `cdkd local`: image-URI detection must also match `cdkd-container-assets-` + generalize the qualifier pattern; cdk-local-owned sites need an upstream PR + bump. (doc §10)

## Decisions (no open questions remain)

1. Marker layout: per-region keys `cdkd-bootstrap/{region}.json` (outside the `cdkd/` state prefix; no cross-region last-writer-wins race).
2. Legacy-mode notice: one `info` line per invocation.
3. No lifecycle policies in v1 (revisit with `cdkd gc`).

# ⚠️ CRITICAL — do not miss: nested child templates are a SEPARATE parse path (doc §7)

> **This is the #1 predicted implementation miss for PR 2. Read this before writing the rewrite pass.**
>
> The top-level analyzer entry is NOT the only place templates are parsed. Nested-stack child templates (`<Stack>.<Child>.nested.template.json`) are loaded from `cdk.out` via **their own `readFileSync` call sites**, which bypass the analyzer entry entirely:
>
> - `src/provisioning/providers/nested-stack-provider.ts` (deploy-time child load, around its `readFileSync` in the template loader)
> - `src/cli/commands/diff-recursive.ts` (`readNestedTemplate` / `indexNestedChildTemplates` — deliberately duplicated copies of the provider's loaders)
> - `src/cli/commands/export.ts` (child-template readers for the state-tree walk)
>
> If the rewrite is applied only at the top-level analyzer entry, a nested Lambda/ECS asset reference keeps pointing at the CDK bootstrap bucket while the assets were published to cdkd storage — the exact split-brain this design exists to prevent. Deep nesting (child → grandchild) must be covered recursively.
>
> **Required**: apply the §6/§7 rewrite helper (same mapping table — the child's assets live in the same assembly-level `assets.json`) at **every** child-template load site, and add unit tests with a nested-stack fixture whose child (and grandchild) contains a Lambda ZIP asset and a container-image asset. The post-resolution audit (§7 step 3) must also run on child-template resources so any future load site added without the rewrite fails loudly instead of deploying split-brain.

# Implementation phasing (doc §13)

1. **PR 1 — bootstrap + marker + mode detection**: storage/marker creation (`--no-assets`, squatting defense), deploy-side marker read + mode plumbing (legacy path byte-identical), `state info` surfacing, design doc onto main, docs updates.
2. **PR 2 — redirection + rewrite + audit**: mapping table, rewrite pass **including every nested child-template load site (see the ⚠️ section above — acceptance criterion, not optional)**, post-resolution audit, `--use-cdk-bootstrap-assets`, command coverage (deploy/diff/import/publish-assets). Unit tests over every §8 destination-shape row + the nested/grandchild fixture.
3. **PR 3 — `cdkd local` matching + cdk-local upstream PR/bump.**
4. **Integ**: dedicated migration integ (deploy pre-marker → re-bootstrap → deploy → verify repointing → destroy clean) **plus a broad-set integ** — PR 2 touches `deploy-engine`/`template-parser` scope, so the `integ-broad` gate applies on top of `integ-destroy`. The migration integ fixture should include a nested stack with assets so the ⚠️ path is exercised against real AWS.

# Follow-ups (separate issues, not this one)

- **`cdkd gc`**: cdkd-native gc enumerating referenced hashes from state files.
- **Upstream**: protection-tag / exclusion proposal for `cdk gc`.

References: [cdk gc announcement](https://aws.amazon.com/blogs/devops/announcing-cdk-garbage-collection/), [cdk gc CLI ref](https://docs.aws.amazon.com/cdk/v2/guide/ref-cli-cmd-gc.html), gc implementation `garbage-collector.ts` in aws/aws-cdk-cli, [AppStagingSynthesizer README](https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk/app-staging-synthesizer-alpha).

go-to-k:
## Summary

**PR 3 of 3 of #1002** — completes the cdkd-owned asset storage rollout on the `cdkd local` side.

The two ECS container-image classification sites in `src/local/ecs-task-resolver.ts` matched only the hardcoded literal `cdk-hnb659fds-container-assets-`. This PR routes both through a shared `isCdkAssetImageUri` helper whose regex `(?:cdk-[a-z0-9]+|cdkd)-container-assets-` recognizes:

- **`cdk-<qualifier>-container-assets-`** — any bootstrap qualifier, not only the `hnb659fds` default (a pre-existing gap: a `cdk bootstrap --qualifier <q>` account's `fromAsset` ECS image would previously fall through to a broken ECR pull instead of building from `cdk.out`);
- **`cdkd-container-assets-`** — the repo `cdkd deploy` publishes into after #1002 PR 2 (defensive / forward-looking; see note below).

A recognized URI classifies as `kind: 'cdk-asset'` (resolve via `cdk.out`) instead of `kind: 'ecr'` (pull from ECR).

## Scope note (architectural)

`cdkd local run-task` always classifies the **synth** template's image, and synth templates stay unrewritten by design (§7.1) — the redirect to `cdkd-container-assets-` only happens at deploy. So the `cdkd-container-assets-` literal is **not reachable** through run-task's synth-based flow; it is covered by unit tests and is defensive for hand-written / `--from-cfn-stack` templates. The **custom-qualifier generalization** is the live-reachable fix. No cdk-local upstream change is needed: cdkd keeps its own still-local `ecs-task-resolver.ts` / `lambda-resolver.ts` (not shimmed), and cdk-local's `intrinsic-image` helpers only substitute placeholders + resolve `Fn::Join` — they never classify the prefix. The Lambda local-invoke path is unaffected (asset `sourceHash` + ECR-pull fallback, not prefix classification).

## Tests

- **Unit** (`tests/unit/local/ecs-task-resolver.test.ts`, 3 new): `cdkd-container-assets-` `Fn::Sub` → `cdk-asset` w/ hash; custom-qualifier `cdk-myqual123-container-assets-` `Fn::Sub` → `cdk-asset`; a `Fn::Join` `fromEcrRepository` whose state-resolved repo name is `cdkd-container-assets-...` → `cdk-asset` (exercises the shared `classifyResolvedImage` site).
- **Integ** (`tests/integration/local-run-task-cdkd-assets/`, new, real Docker, local-only — no AWS deploy): a CDK app pinned to a custom bootstrap qualifier synthesizes its `fromAsset` ECS image as `cdk-myqual99-container-assets-...`; `cdkd local run-task` classifies it as a CDK asset, builds it from `cdk.out`, and the container prints its marker. verify.sh asserts the marker and rejects the ECR-pull error signature to guard against a false pass. Ran clean via `/run-integ local-run-task-cdkd-assets` (14s, docker sweep empty).

## Retrospective / rule proposal

The design §10 implied `--from-state` surfaces the rewritten cdkd URI to the local resolver, but tracing the code showed the local resolvers classify the unrewritten synth image. Captured as a memory rule (`feedback_design_doc_dataflow_vs_code`): trace a design doc's cross-component data-flow claim in the actual code before scoping a fixture / implementation around it.

Closes #1002


go-to-k:
### Review-nit dispositions (3-axis review complete — all CLEAN)

- **spec-reviewer**: CLEAN, no drift.
- **code-reviewer** — two nits, both **accepted** as known cost (not regressions):
  1. `CDK_ASSET_IMAGE_REPO_RE`'s `[a-z0-9]+` excludes uppercase/hyphen/underscore qualifiers — intentionally identical to the PR-2 sibling `asset-redirect.ts` pattern, so classifier and template-rewriter stay consistent; typical qualifiers are lowercase-alphanumeric.
  2. Non-anchored substring match — same risk profile as the prior `.includes('cdk-hnb659fds-container-assets-')` and the sibling matcher; not introduced here.
- **test-reviewer** — two non-blocking suggestions, both **addressed** in `f0fc9624`:
  1. Negative: a repo embedding `container-assets` without the `cdk-`/`cdkd-` prefix stays `kind:'ecr'` (pins the prefix requirement).
  2. Symmetry: a custom-qualifier repo resolved via the `Fn::Join` `fromEcrRepository` path classifies as `cdk-asset` at the `classifyResolvedImage` site too.
  A focused re-review of both added tests on the final sha returned CLEAN.

Issue #1092, opened 2026-07-19
chore(providers): hardening backlog from provider-rollout reviews (FSx apply default-guard, EMR fleet SUSPENDED, FSx nested-arm tests)


go-to-k:
Low-priority hardening backlog from the 2026-07 provider rollout code reviews (bundled to avoid issue spam; each is latent/non-blocking).

1. **FSx `apply*UpdateField` switches lack a `default:` guard** (PR #1085, `fsx-filesystem-provider.ts`): a future mutable sub-prop added to a `*_MUTABLE_SUBPROPS` set without a matching switch case would silently apply nothing (no-op `UpdateFileSystem` while `hasMutableDiff=true`). Add `default: throw` (or `return false`) to `applyLustre/Windows/Ontap/OpenZFSUpdateField` so the miss is loud. (Deferred in #1085 because touching provider code re-stales the FSx integ gate.)

2. **EMR fleet `FAILED_STATES` omits `SUSPENDED`** (PR #1087, `emr-instance-fleet-config-provider.ts`): a suspended fleet polls to timeout rather than failing fast. The instance-group provider's set includes ARRESTED/TERMINATED/ENDED; align the fleet set. Correctness/cleanup unaffected — only a slower failure.

3. **Deeper FSx Windows/ONTAP nested-sub-block update-arm unit coverage** (PR #1085): the `AuditLogConfiguration` / `DiskIopsConfiguration` reconciliation arms are exercised structurally by the shared differ but lack dedicated unit tests. Create-mapping + `--replace` rejection are covered.

Each item is a provider-code (1,2) or test-only (3) change; items 1+2 re-stale the respective integ-destroy gate, so batch them with a real integ run.

go-to-k:
Taking the **FSx-scoped items only** (item 1: `apply*UpdateField` `default:` guard, item 3: nested-sub-block update-arm unit coverage) in branch `feat/fsx-variant-read-state`, bundled with #1089 because all three land in `src/provisioning/providers/fsx-filesystem-provider.ts` and share a single FSx integ run.

Item 2 (EMR fleet `FAILED_STATES` missing `SUSPENDED`, `emr-instance-fleet-config-provider.ts`) is **NOT** claimed here and stays open for a separate lane — it is file-disjoint from this work. This issue will therefore remain open after that PR merges.

go-to-k:
[...]

Issue #1088, opened 2026-07-19
test(integ): live-integ coverage for FSx Windows / ONTAP variants (OpenZFS only in #1085)


go-to-k:
Follow-up to PR #1085 (FSx Windows/ONTAP/OpenZFS variants).

PR #1085 shipped all three non-Lustre variants but live-integ-verified only the **OpenZFS SINGLE_AZ_1** variant (the cheapest). Windows and ONTAP are **unit-tested only** — they share the integ-verified delete path, but their create/update happy-paths were never exercised against real AWS (cost-gated: Windows needs a Managed AD, ONTAP is materially pricier per hour).

Add per-variant live integ coverage:
- **Windows** — needs an `AWS::DirectoryService::MicrosoftAD` (or self-managed AD) in the fixture; ~20-40 min provision + AD cost.
- **ONTAP** — multi-AZ or single-AZ deployment; pricier per-hour storage.

Bound cost (smallest legal config, `AutoImportPolicy`/short retention, prompt destroy). Each must assert the filesystem is GONE after destroy. Lower priority — the shared delete path is already verified by the OpenZFS integ, so this is confidence coverage, not a correctness gap.

go-to-k:
Working on this in branch `feat/integ-fsx-windows-ontap` (worktree `.claude/worktrees/feat-integ-fsx-windows-ontap`) — adding FSx Windows / ONTAP integ fixtures under `tests/integration/` only (no `src/**` changes, so this does not collide with the in-flight `feat/fsx-variant-read-state` provider lane for #1089 / #1092). Claiming to avoid collision with parallel agents.

Issue #1068, opened 2026-07-18
feat(providers): FSx FileSystem Windows / ONTAP / OpenZFS variants


go-to-k:
Follow-up to PR #1063 (AWS::FSx::FileSystem SDK provider — Lustre variant).

The initial provider covers the **Lustre variant only**. `WindowsConfiguration`, `OntapConfiguration`, and `OpenZFSConfiguration` are `unhandledByDesign` (pre-flight rejected via the `disableCcApiFallback` guard, so a Windows/ONTAP/OpenZFS template fails fast with a clear error rather than an opaque provisioning failure).

Add the remaining variants (can be split into separate PRs per variant):

- **`WindowsConfiguration`** — `CreateFileSystem` with `FileSystemType: WINDOWS`; ties to a Managed AD / self-managed AD, throughput capacity, self-managed AD config; mutable surface differs (`ThroughputCapacity`, `WeeklyMaintenanceStartTime`, audit-log config).
- **`OntapConfiguration`** — NetApp ONTAP; deployment type, endpoint IP range, disk-IOPS, HA pairs.
- **`OpenZFSConfiguration`** — root volume config, deployment type, copy-tags.

Each shares the existing create-poll-to-AVAILABLE / delete-poll-to-GONE / `ClientRequestToken` / `getMinResourceTimeoutMs` machinery — the work is per-variant property mapping + `handledProperties` classification + a per-variant integ fixture (mind the per-hour cost; ONTAP/Windows are pricier than Lustre SCRATCH_2). Reference: the PR's "Scope is the Lustre variant only" note.

go-to-k:
Starting work on this. Branch `feat-fsx-variants`, worktree `.claude/worktrees/feat-fsx-variants`. Files: `src/provisioning/providers/fsx-filesystem-provider.ts` (move Windows/ONTAP/OpenZFS configs from unhandledByDesign to handledProperties + per-variant create/update mapping), its unit test, `docs/supported-resources.md`, `docs/import.md`, and a new `tests/integration/fsx-openzfs/` fixture (cheapest variant for live integ; Windows/ONTAP unit-tested + share the integ-verified delete path).

PR #1095, opened 2026-07-19
feat(providers): FSx variant drift read-back + apply*UpdateField default guard

### LINKED ISSUE #1089: feat(providers): FSx readCurrentState/drift for Windows / ONTAP / OpenZFS variant configs
Follow-up to PR #1085 (FSx Windows/ONTAP/OpenZFS variants).

`readCurrentState` currently maps only the **Lustre** config block back to CFn shape. The three new variant blocks (`WindowsConfiguration` / `OntapConfiguration` / `OpenZFSConfiguration`) are marked drift-unknown via `getDriftUnknownPaths` (which avoids a guaranteed phantom drift, but means `cdkd drift` reports "unknown" for those variants' config).

Extend `readCurrentState` to reverse-map each variant's read-side `DescribeFileSystems` response back to the flat CFn `*Configuration` inputs cdkd stores (mirroring the Lustre `DataRepositoryConfiguration` → `LustreConfiguration` mapping), then drop the corresponding `getDriftUnknownPaths` entries. Mind the same server-injected-default caveat pattern as other providers (only surface fields cdkd actually set).

go-to-k:
## Summary

Closes the FSx variant drift-detection gap left by #1068, plus the two FSx-scoped hardening items from the #1092 backlog.

**1. Variant drift read-back (#1089).** `readCurrentState` mapped only the Lustre block back to CFn shape, so `getDriftUnknownPaths` had to declare the whole `WindowsConfiguration` / `OntapConfiguration` / `OpenZFSConfiguration` subtrees unknown — `cdkd drift` reported "unknown" for every non-Lustre variant's config instead of comparing it. Three new `read*Configuration` reverse-mappers (plus a shared `readDiskIopsConfiguration`) now turn each variant's `DescribeFileSystems` block back into the flat CFn input shape, mirroring the existing Lustre `DataRepositoryConfiguration` mapping.

Read-only API fields that are not CFn inputs (`Endpoints`, `RootVolumeId`, `EndpointIpAddress`, `RemoteAdministrationEndpoint`, `PreferredFileServerIp`, `MaintenanceOperationsInProgress`, `Alias.Lifecycle`) are deliberately **not** mapped — surfacing them would create phantom drift. The Windows `Aliases` shape change is handled explicitly: the API returns `{Name, Lifecycle}[]` while CFn takes a plain name list.

`getDriftUnknownPaths` narrows from three whole blocks to the three leaves AWS genuinely never returns:
- `WindowsConfiguration.SelfManagedActiveDirectoryConfiguration.Password` — `SelfManagedActiveDirectoryAttributes` has no `Password` field at all
- `OntapConfiguration.FsxAdminPassword` — write-only, never echoed
- `OpenZFSConfiguration.RootVolumeConfiguration` — lives on the root *volume*; `DescribeFileSystems` returns only `RootVolumeId`

**2. `apply*UpdateField` default guard (#1092 item 1).** The four switches had no `default:` arm, so a sub-property added to a `*_MUTABLE_SUBPROPS` set without a matching case would pass the mutability check, map to nothing, and issue a no-op `UpdateFileSystem` with `hasMutableDiff` true — silently dropping the user's change. It now throws a reportable `ProvisioningError` naming the exact unmapped `<ConfigKey>.<Subprop>`.

**3. Nested sub-block update-arm coverage (#1092 item 3).**

`#1092` stays open after this merges — item 2 (EMR fleet `SUSPENDED`) is file-disjoint and intentionally unclaimed.

## Review-surfaced correction

Two independent reviewers caught that the initial version declared `EndpointIpv6AddressRange` drift-unknown on ONTAP / OpenZFS based on a bad reading of the SDK types. The field **is** declared on both read-side interfaces (`models_0.d.ts:737`, `:867`) — an earlier extraction truncated the interface bodies at a nested brace. Since it is also create-mapped, declared mutable, and has live update arms, declaring it unknown would have left a user-settable, AWS-mutable, AWS-readable property permanently invisible to drift — the exact gap this PR exists to close. It is now reverse-mapped and both entries dropped.

Review also found the new integ assertion was tautological (it read `properties`, the template input, rather than `observedProperties`, the deploy-time `readCurrentState` snapshot). A clean `cdkd drift` exit is not sufficient proof on its own either: `drift.ts` classifies a resource whose `readCurrentState` returns `undefined` as `unsupported`, which also exits 0.

## Test plan

- **Unit:** 42 new cases. 33 are generated — one per declared-mutable sub-property across all four variants — asserting each maps to a non-empty `UpdateFileSystem` field, which is what makes a future set-addition-without-a-case fail loudly rather than silently. Plus an injected-unmapped-subprop guard test, per-variant reverse-map tests (Aliases flattening, read-only-field exclusion, write-only `Password` absence, empty-block omission), four `drift --revert` round-trip guards per `docs/provider-development.md` §3b (OpenZFS + Windows no-op replay; mutable-only drift replay; Class 2 partially-populated `DiskIopsConfiguration`), and nested sub-block update-arm coverage. Full suite: 7173 passing.
- **Live (real AWS, `us-east-1`):** `/run-integ fsx-openzfs`, run twice — once before and once after the review fixes. The fixture gained a **Phase 2.5** that asserts (a) `cdkd drift` is clean on the freshly-deployed file system, and (b) `observedProperties` actually carries the reverse-mapped OpenZFS block. Both matter: before this change the block was drift-unknown so drift passed trivially, and (b) is what proves `readCurrentState` ran rather than silently returning nothing. Destroy: 8 deleted, 0 errors, 0 orphans; file system confirmed gone by id and by tag.

## Known gaps (none introduced here)

1. The FSx test file still lacks the `emits placeholders for every user-controllable top-level key` block that `docs/provider-development.md` §3b marks mandatory. Pre-existing; retrofitting it would change which top-level keys `readCurrentState` emits on an AWS-minimum response. All three reviewers agreed deferring is defensible.
2. `WindowsConfiguration.Aliases` and `SelfManagedActiveDirectoryConfiguration.DnsIps` are plain-string arrays, which `src/analyzer/drift-normalize.ts` does not canonicalize (it sorts only tag lists and AWS-id/ARN arrays), so an AWS-side reorder could surface as phantom drift. Sorting inside the reverse-mapper is **not** a safe local fix — for resources with no `observedProperties` the baseline is the user's template order, so a sorted read side would itself drift. The correct fix is extending the shared normalizer.
3. Windows and ONTAP reverse-mappers are unit-tested only; the live integ covers OpenZFS (#1088 tracks per-variant live coverage).

Closes #1089


go-to-k:
Follow-ups from this PR's review are tracked in #1096 (drift-normalize plain-string array canonicalization + the FSx §3b placeholder test). The "Known gaps" section above is now covered by that issue.

PR #1135, opened 2026-07-20
fix(provisioning): lift per-type CC operation-timeout floor for slow CREATE/DELETE

go-to-k:
## Summary

Fixes a real cdkd bug surfaced by the `opensearch-domain-getatt` integ during a `/pick-integ` sweep (2026-07-20): the test's GetAtt assertion passed, but **destroy failed** with `DELETE timeout ... after 900s` while the OpenSearch domain was still `IN_PROGRESS` deleting.

**Root cause:** `CloudControlProvider.waitForOperation` capped every operation at a flat 15-minute internal poll cap (`MAX_WAIT_TIME_MS`). An `AWS::OpenSearchService::Domain` deletion routinely runs 15-30 min, so it aborted mid-delete. That 15-min inner cap was also **shorter than the 30-min outer per-resource deadline** (`DEFAULT_RESOURCE_TIMEOUT_MS`), so every Cloud-Control-routed slow resource silently got half the advertised budget — the tighter inner cap always fires first.

## What changed

New `src/provisioning/slow-cc-operation-timeouts.ts` — a single shared per-`(resourceType, operation)` timeout-floor table:
- OpenSearch / Elasticsearch domains: 60 min for CREATE / UPDATE / DELETE
- Redshift / ElastiCache (ReplicationGroup + CacheCluster) / RDS (DBInstance + DBCluster): 60 min for CREATE / DELETE

Consulted in **three** places so the inner and outer budgets can never drift apart again:
1. `CloudControlProvider.waitForOperation` caps at `Math.max(MAX_WAIT_TIME_MS, floor)` — a normal type keeps the flat 15 min; a slow type grows to its floor.
2. The destroy-runner outer deadline folds the DELETE floor into its `Math.max(providerMinTimeoutMs, floor, globalTimeoutMs)`.
3. The deploy-engine outer deadline does the same with the CREATE/UPDATE floor.

A user-supplied `--resource-timeout <TYPE>=<DURATION>` override still wins at both outer sites. The default `cdkd destroy` now waits long enough for a slow delete to actually finish instead of aborting — matching CloudFormation's synchronous-wait semantics and cdkd's "destroy complete = resource actually gone" contract. (Async-by-default would break the 0-orphan guarantee; `--no-wait` remains the opt-out.)

## Test plan

- **Unit:** new pure-helper suite (per-type CREATE/UPDATE/DELETE lookups, the cluster-types-have-no-UPDATE-floor distinction, `Math.max`-safety) + two fake-timer `CloudControlProvider` DELETE tests (a normal type aborts at 900s; an OpenSearch domain is still polling at 20 min and aborts at the lifted 3600s). Full suite: 454 files / 7621 tests pass.
- **Real AWS (us-east-1):** `opensearch-domain-getatt` re-run with the fixed binary now **PASSES end-to-end** — destroy exited 0 after ~30 min, 0 orphans (previously died at 900s). Recorded in the integ ledger.
- **Broad integ (cross-cutting deploy-engine + destroy-runner touched):** `lambda` — 9 resources (Lambda/Version/Alias/Layer, DynamoDB Table+GlobalTable, SQS, IAM), clean destroy 0 errors. `integ-broad` + `integ-destroy` markers set.

## Docs

`docs/changelog-cdkd.md`, `docs/cli-reference.md` (per-resource timeout section), `.claude/rules/cli-internals.md` (resolution formula), `.claude/rules/code-layout.md` (new module).


go-to-k:
Review (2-axis: code + test; spec axis dropped — bug fix, no design doc) returned clean, no blockers/majors. Residual nits explicitly accepted as known cost (all minor; core behavior verified end-to-end by the opensearch-domain-getatt real-AWS integ + lambda broad):

- **Outer-deadline precedence not unit-pinned** (test-minor): the per-type `--resource-timeout TYPE=DURATION` override still short-circuits before the new `Math.max(...)` floor at both outer sites; this is covered by the opensearch integ (a 15-30 min delete only completes because both caps lifted) but not by a dedicated unit test. Cheap follow-up: a 3-line test asserting the override wins over the floor.
- **CREATE/UPDATE inner cap covered via the helper suite, not a fake-timer poll test** (nit): same `maxWaitMs` code path as DELETE, only the `operation` literal differs; the helper suite pins the CREATE/UPDATE lookups.
- **60-min value inlined 15x vs a named constant** (readability nit).
- **deploy-engine `getProvider` vs destroy-runner `getProviderFor`** asymmetry: pre-existing; no functional gap for CC-routed slow types since the new floor term covers `providerMinTimeoutMs === 0`.

PR #1154, opened 2026-07-22
feat(providers): add AWS::Lambda::MicrovmImage SDK provider

go-to-k:
## Summary

Adds a dedicated **SDK provider (Tier 1)** for `AWS::Lambda::MicrovmImage` (Lambda MicroVMs — the Firecracker-snapshot compute environment). It was previously only reachable via the Cloud Control fallback, which always polls its request token to a terminal state, so `cdkd deploy --no-wait` had no effect on the multi-minute async build (exactly where `--no-wait` is useful).

Key points:
- Backed by the **separate `lambda-microvms` service** (`@aws-sdk/client-lambda-microvms`, a distinct service model from `@aws-sdk/client-lambda`).
- `create()` translates the CFn-schema PascalCase template to the SDK camelCase shape (`CodeArtifact: {Uri}` → `codeArtifact: {uri}`, `Tags: [{Key,Value}]` → `tags: Record`, `Logging: {Disabled: true}` → `{disabled: {}}` tagged union, `Hooks`/`CpuConfigurations`/`Resources`/…), calls `CreateMicrovmImage`, then polls `GetMicrovmImage` until `CREATED` (throws on `CREATE_FAILED`).
- The poll is gated on `CDKD_NO_WAIT` — so `--no-wait` returns right after `CreateMicrovmImage` (state `CREATING`) with the image ARN already resolved. **This is the concrete value of the Tier-1 provider over the CC fallback.**
- `update()` reconciles `Tags` out-of-band via `TagResource`/`UntagResource` (tags are not a field on `UpdateMicrovmImage`); a **tags-only change skips the image rebuild**. A build-affecting property change (`BUILD_AFFECTING_CFN_KEYS`) issues the async `UpdateMicrovmImage` rebuild. In-place `Name` changes are refused (create-only → replacement).
- `delete()` calls `DeleteMicrovmImage` and polls until the image 404s (async `DELETING → DELETED`) so destroy leaves no orphan.
- Physical id is the image ARN (`primaryIdentifier`); `getAttribute` exposes `ImageArn` / `State` / `LatestActiveImageVersion` / `LatestFailedImageVersion` / `CreatedAt` / `UpdatedAt`. `getMinResourceTimeoutMs()` lifts the per-resource deadline to the 30-min poll cap so a slow build is not truncated.

`aws-cdk-lib` has no `CfnMicrovmImage` yet, so the fixture (and real users) declare the resource via the `CfnResource` escape hatch.

## Test plan

- **22 unit tests** — CFn→SDK shape translation for every optional field (incl. the `Logging.Disabled` tagged-union, `Tags`/`EnvironmentVariables` list→record, partial `CloudWatch`, `Resources` NaN-filter), async `CREATING→CREATED` poll, `CREATE_FAILED`/`DELETE_FAILED`, `--no-wait` skip, create-only `Name` guard, update missing-required guard, tags reconcile, tags-only-no-rebuild, env-var clear, delete NotFound idempotency + region-mismatch rejection, `CreatedAt`/`UpdatedAt` Date→ISO.
- **Real-AWS integ** `tests/integration/lambda-microvm-image/` (verify.sh): uploads a Dockerfile+app.js code artifact to S3, deploys a real Firecracker snapshot build to `CREATED`, asserts the ARN physicalId + `MicrovmImageArn` output parity + `GetMicrovmImage` state, then a **tags-only UPDATE** (env dev→prod, +team) that is reconciled via Tag/UntagResource with the active version UNCHANGED (proves no rebuild), then a clean async destroy (image 404s, state removed). **Live-verified end to end, 0 orphans.**
- Moves the type from **Tier 2 → Tier 1** in the provider-coverage matrix; broad `lambda` integ run to satisfy the `register-providers.ts` cross-cutting gate.

## Code review

3-axis review (code + test) run on the initial commit. Findings:
- **Blocker** — `update()` never applied Tag changes (`UpdateMicrovmImage` has no `tags` field; a tags-only change would spuriously rebuild). **Fixed** (Tag/UntagResource reconcile + skip-rebuild) and **live-verified** in the integ UPDATE phase.
- **Major** — clearing the last env var on update was a silent no-op. **Fixed** (sends `environmentVariables: {}` on the rebuild path + unit test).
- **Minor** — `mapResources` NaN coercion (**fixed** + test), `additionalOsCapabilities` cast (**fixed** to `Capability[]`).
- **Accepted (LOW)** — poll-cap-exhaustion timeout branches and the `Logging.Disabled: 'true'` (string) edge are left untested: they are defensive branches, `getMinResourceTimeoutMs()` lifts the outer deadline, and CDK L1 typing prevents a boolean-as-string. No behavior risk on the covered paths.


go-to-k:
Final 3-axis review (code + test) on the fixed code: both CLEAN. Accepted low nits (documented, not re-touched to avoid another real-build integ cycle): (1) the test-only poll-config env vars `CDKD_MICROVM_IMAGE_POLL_{ATTEMPTS,INTERVAL_MS}` yield NaN if set to a non-numeric value (unreachable in production — the finite defaults 180/10000 are used unless a dev overrides them); (2) poll-cap-exhaustion timeout branches and the update-path `--no-wait` skip are unit-untested (defensive/env-gated branches symmetric to the tested create path). Terminal-failure states (CREATE_FAILED/UPDATE_FAILED/DELETE_FAILED), the tag reconcile, and delete region-mismatch ARE covered + live-verified.

Issue #1169, opened 2026-07-23
ECS readCurrentState surfaces ContainerDefinitions as raw SDK camelCase (drift phantom, follow-up to #1167)


go-to-k:
## Summary

Follow-up from #1167 (which reverse-mapped the top-level ECS nested objects in `readCurrentState` to CFn PascalCase). `ECSProvider.readCurrentStateTaskDefinition` still surfaces the whole `ContainerDefinitions` array as **raw SDK camelCase** (`result['ContainerDefinitions'] = td.containerDefinitions ?? []`), while cdkd's drift baseline (the template `properties`, or `observedProperties`) stores it in CFn **PascalCase** (produced on the SET path by `convertContainerDefinitions`). So `cdkd drift` on a TaskDefinition phantom-drifts on `ContainerDefinitions`.

This was observed live during the #1167 integ (a `cdkd drift` on the `ecs-service-update-props` TaskDefinition reported drift only on `ContainerDefinitions`, with `RuntimePlatform` / `EphemeralStorage` / `PlacementConstraints` correctly clean after #1167).

## Why it was out of #1167 scope

`ContainerDefinitions` is more than a casing flip:

1. **Structural differences** between the template and the AWS-returned shape, not just key casing. Example seen live:
   - template `LinuxParameters.Capabilities: {}` vs AWS read `Capabilities: {Add: [], Drop: []}`
   - AWS read adds fields the template omits: `cpu: 0`, `portMappings: []`, `environment: []`, `mountPoints: []`, `volumesFrom: []`, `systemControls: []`.
   The drift comparator only descends into baseline keys, so the extra AWS fields are ignored, but the `Capabilities: {}` vs `{Add:[],Drop:[]}` mismatch (and similar defaulted-empty sub-objects) still drift.
2. It needs the **full inverse of `convertContainerDefinitions`** (environment / environmentFiles / secrets / portMappings / mountPoints / volumesFrom / dependsOn / ulimits / logConfiguration / healthCheck / linuxParameters + the #1165-added `LinuxParameters` / `LogConfiguration.SecretOptions`), plus normalization for the AWS-defaulted empty sub-objects/arrays so they don't drift against a template that omits them.

## Fix direction

- Add a `containerDefinitionsToCfn` reverse-mapper (mirror of `convertContainerDefinitions`, analogous to the existing `volumesToCfn`), emitting CFn PascalCase.
[...]

PR #1171, opened 2026-07-23
fix(providers): accept ARN-form physicalId in ECS Service readCurrentState + getServiceAttribute (#1170)

### LINKED ISSUE #1170: cdkd drift reports every cdkd-created ECS Service as drift-unknown (readCurrentState expects composite physicalId, createService stores the ARN)
## Summary

`ECSProvider.readCurrentStateService` expects the Service physicalId in the **composite** form `<clusterArn>|<serviceName>` and returns `undefined` when it can't find the `|` separator:

```ts
const sep = physicalId.indexOf('|');
if (sep < 0) return undefined;
```

But `ECSProvider.createService` stores the physicalId as the **service ARN** (`return { physicalId: service.serviceArn, ... }`), which has no `|`. So for every cdkd-created ECS Service, `readCurrentState` returns `undefined` and `cdkd drift` reports the Service as **drift-unknown** ("provider does not yet support drift detection") rather than actually comparing it.

Observed live during the #1167 integ: the deployed `ecs-service-update-props` Service read back as drift-unknown.

## Impact

- `cdkd drift` never detects real drift on ECS Services deployed by cdkd (e.g. an out-of-band `DesiredCount` / `DeploymentConfiguration` / tag change goes unseen).
- The Service-side reverse-map added in #1167 (DeploymentConfiguration / CapacityProviderStrategy / NetworkConfiguration / LoadBalancers / ...) is therefore only exercised by unit tests + potential import paths, not by drift on cdkd-created services.

## Fix direction

Make `readCurrentStateService` accept the ARN form too: if there's no `|`, derive the cluster + service from the ARN (`arn:aws:ecs:<region>:<acct>:service/<clusterName>/<serviceName>` → `DescribeServices({cluster: <clusterName>, services: [<serviceName>]})`) instead of returning `undefined`. Keep the composite-form path for callers that still use it. Add a `readCurrentState` unit test for the ARN-form physicalId and a drift-clean integ phase for a deployed Service.

Confirm whether any other code path relies on the composite `<clusterArn>|<serviceName>` physicalId (mutation ops, import) before changing what `createService` stores; the safer change is to make the READER accept both forms rather than change the stored id.

Pre-existing (not introduced by #1165 / #1167); surfaced by the #1167 integ.

go-to-k:
## Summary

Fixes #1170.

`ECSProvider.readCurrentStateService` only understood the composite physicalId
form `<clusterArn>|<serviceName>` and returned `undefined` when there was no
`|`. But `createService` stores the service **ARN** (which has no `|`), so for
every cdkd-created ECS Service `readCurrentState` returned `undefined`: `cdkd
drift` reported the Service as **drift-unknown** and no `observedProperties`
were captured for it at deploy time.

The fix makes the **reader** accept both forms (the safer direction the issue
recommends — the mutation ops already pass the ARN straight through as the
`service` argument, with the cluster read from the template `Cluster` property,
so what `createService` stores is unchanged):

- Add `clusterNameFromServiceArn` to derive the cluster from a long-format
  service ARN (`arn:...:service/<clusterName>/<serviceName>`).
- `readCurrentStateService` splits on `|` (composite form) OR derives the
  cluster from the ARN, falling back to the default cluster for the legacy
  short-format ARN.
- Also fix the sibling `getServiceAttribute` gap surfaced in review: it issued
  `DescribeServices` with no cluster, so a Service in a named (non-default)
  cluster came back MISSING and `getAttribute` returned `undefined`. It now
  derives the cluster via the same helper.

## Test plan

- **Unit**: long- and short-format ARN physicalId round-trip in
  `readCurrentState`; `getAttribute` scopes `DescribeServices` to the ARN
  cluster. Full suite green (7634 tests).
- **Integ** (`ecs-service-update-props`, real AWS `us-east-1`): new Phase 1c
  asserts the Service's deploy-time `observedProperties` ARE now captured
  (`DeploymentConfiguration.MaximumPercent=150`) — this lookup was MISSING
  before the fix because `readCurrentState` returned `undefined` for the
  ARN-form physicalId. Deploy 15 / update 1 / destroy 15, **0 errors, 0
  orphans**; re-run after the `getServiceAttribute` polish also clean.

## Notes

- `#1169` (TaskDefinition `ContainerDefinitions` still surfaced as raw SDK
  camelCase) is the sibling follow-up and is intentionally out of scope here;
  the fixture's comment points a whole-stack `cdkd drift` assertion at that
  gap.

### DIFF
--- a/docs/_generated/integ-last-run.tsv
+++ b/docs/_generated/integ-last-run.tsv
@@ -83,7 +83,7 @@ ec2-vpc	2026-06-21T11:12:35Z	PASS	53	standard	sweep-F staleness re-run (rc=0)
 ecr	2026-06-21T11:00:28Z	PASS	39	standard	sweep-4..8 staleness re-run (ledger-restore; rc=0)
 ecr-scanning	2026-07-20T17:37:23Z	PASS	56	verify.sh	ECR scanOnPush + tag add/change/untag update, clean destroy
 ecs-fargate	2026-07-02T18:23:15Z	PASS	373	verify.sh	0703 staleness sweep (pre-TTL refresh); clean
-ecs-service-update-props	2026-07-23T07:03:42Z	PASS		verify.sh	#1170 Phase1c ARN-form observedProperties captured; destroy 15 del 0 err 0 orph
+ecs-service-update-props	2026-07-23T07:13:07Z	PASS		verify.sh	#1170 re-run after getServiceAttribute polish; destroy 15 del 0 err 0 orph
 efs-immutable-replacement	2026-07-19T19:05:41Z	PASS	76	verify.sh	rc ok, 1 deleted, orph clean
 efs-lambda	2026-06-21T14:48:46Z	PASS	560	standard	sweep-G standard re-run (rc=0)
 efs-standalone	2026-06-21T16:16:44Z	PASS	135	verify.sh	stale-real re-run; 11 deleted 0 err; clean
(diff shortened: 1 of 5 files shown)


go-to-k:
Review (1-reviewer tier, code-quality): clean, no blockers. One accepted nit: `clusterNameFromServiceArn` is module-private and covered indirectly via its 3 real call-site tests (two readCurrentState + one getAttribute) rather than a direct unit test. Accepted as known cost — partition-invariance is by construction (the parse uses ARN index 5, not the partition field), and exporting the helper for a direct test would re-edit the provider file and force another full real-AWS re-integ for marginal coverage gain.

Issue #1167, opened 2026-07-23
ECS readCurrentState surfaces nested objects in camelCase under PascalCase keys (drift phantom, follow-up to #1165)


go-to-k:
## Summary

Follow-up from #1165 (the PR that fixed the SET-path PascalCase->camelCase silent drop for ECS nested objects). The **read side** (`ECSProvider.readCurrentState`) has the mirror-image inconsistency: it surfaces several nested objects in the SDK's **camelCase** value shape under a **PascalCase** top-level key, while cdkd's stored baseline uses PascalCase keys. This can produce **phantom drift** in `cdkd drift`.

## Detail

In `src/provisioning/providers/ecs-provider.ts`:

- `readCurrentStateService` writes the raw SDK camelCase object straight into the result:
  - `result['DeploymentConfiguration'] = s.deploymentConfiguration` (`{maximumPercent, ...}`)
  - `result['CapacityProviderStrategy'] = s.capacityProviderStrategy`
  - `result['PlacementConstraints']`, `result['ServiceRegistries']` (and `PlacementStrategy`)
- `readCurrentStateTaskDefinition`:
  - `result['RuntimePlatform'] = td.runtimePlatform` (camelCase `{cpuArchitecture, ...}`)
  - `result['ProxyConfiguration'] = ...` (camelCase)
  - but `result['EphemeralStorage'] = { SizeInGiB: td.ephemeralStorage.sizeInGiB }` **IS** re-PascalCased, so the read side is internally inconsistent.

The deploy-time `observedProperties` baseline is captured from the same `readCurrentState`, so a service / task-def whose `observedProperties` are present compares camelCase-vs-camelCase and stays clean. The phantom drift surfaces in the **`observedProperties`-absent fallback window** - a resource whose baseline falls back to the user's template `properties` (PascalCase-valued): `cdkd drift` then compares `{MaximumPercent}` (baseline) against `{maximumPercent}` (AWS read) and reports drift that isn't real.

## Fix

[...]

PR #1172, opened 2026-07-23
fix(providers): reverse-map ECS ContainerDefinitions to CFn PascalCase on readCurrentState (#1169)

### LINKED ISSUE #1169: ECS readCurrentState surfaces ContainerDefinitions as raw SDK camelCase (drift phantom, follow-up to #1167)

go-to-k:
## Summary

Fixes #1169 (follow-up to #1167).

`ECSProvider.readCurrentStateTaskDefinition` surfaced the whole
`ContainerDefinitions` array as **raw SDK camelCase**
(`result['ContainerDefinitions'] = td.containerDefinitions`), while cdkd's drift
baseline (state `properties` PascalCase, or `observedProperties` captured via
`readCurrentState`) is CFn PascalCase. The drift comparator compares arrays
**wholesale via `deepEqual`**, so `cdkd drift` phantom-drifted the TaskDefinition
on `ContainerDefinitions`.

Add a `containerDefinitionsToCfn` reverse-mapper (inverse of
`convertContainerDefinitions`, analogous to the existing `volumesToCfn`):

- **Symmetric field set** — emits CFn PascalCase for exactly the fields
  `convertContainerDefinitions` can SET, so a field cdkd cannot set is never
  surfaced as drift against a template that lacks it.
- **AWS-defaulted-empty normalization** — `Cpu: 0`, empty sub-arrays
  (`PortMappings` / `Environment` / ...), and an empty
  `LinuxParameters.Capabilities: {Add:[],Drop:[]}` are dropped so they equal
  "absent". AWS returns these defaults on every read, so without this the
  `properties` fallback baseline would phantom-drift on every run; the
  normalization trades that common phantom-drift for a rare edge (a template
  that explicitly sets one of these to its empty/zero default AND has no observed
  baseline; CDK L2 never emits e.g. an explicit `Cpu: 0`).
- **Free-form maps preserved verbatim** — `DockerLabels` and
  `LogConfiguration.Options` keys are user data (container labels / log-driver
  options), NOT CFn property names, so they are copied without case-flipping.
- Sub-helpers `logConfigurationToCfn` / `linuxParametersToCfn`.

On the normal `observedProperties` path both sides run through this same reader,
so the snapshot is self-consistent regardless of the normalization.

## Test plan

- **Unit**: two tests assert the PascalCase output, the AWS-defaulted-empty
  normalization, free-form option-key preservation, the no-camelCase-leak guard,
  and populated sub-list reverse-mapping. Full suite green (7636 tests).
- **Integ** (`ecs-service-update-props`, real AWS `us-east-1`): new **Phase 1d**
  asserts the deploy-time `observedProperties.ContainerDefinitions[0]` is CFn
  PascalCase (`Name` / `LogConfiguration.LogDriver` / `LinuxParameters.InitProcessEnabled`,
  free-form `awslogs-group` option key preserved, no camelCase leak). New
  **Phase 1e** runs a real `cdkd drift` and asserts it reports **no
  `ContainerDefinitions` drift** — the end-to-end proof that the reverse-map +
  normalization round-trips against real AWS. Destroy 0 errors, 0 orphans.

### DIFF
--- a/docs/_generated/integ-last-run.tsv
+++ b/docs/_generated/integ-last-run.tsv
@@ -83,7 +83,7 @@ ec2-vpc	2026-06-21T11:12:35Z	PASS	53	standard	sweep-F staleness re-run (rc=0)
 ecr	2026-06-21T11:00:28Z	PASS	39	standard	sweep-4..8 staleness re-run (ledger-restore; rc=0)
 ecr-scanning	2026-07-20T17:37:23Z	PASS	56	verify.sh	ECR scanOnPush + tag add/change/untag update, clean destroy
 ecs-fargate	2026-07-02T18:23:15Z	PASS	373	verify.sh	0703 staleness sweep (pre-TTL refresh); clean
-ecs-service-update-props	2026-07-23T07:13:07Z	PASS		verify.sh	#1170 re-run after getServiceAttribute polish; destroy 15 del 0 err 0 orph
+ecs-service-update-props	2026-07-23T07:43:53Z	PASS		verify.sh	#1169 Phase1d/1e ContainerDefinitions drift-clean; destroy 15 del 0 err 0 orph
 efs-immutable-replacement	2026-07-19T19:05:41Z	PASS	76	verify.sh	rc ok, 1 deleted, orph clean
 efs-lambda	2026-06-21T14:48:46Z	PASS	560	standard	sweep-G standard re-run (rc=0)
 efs-standalone	2026-06-21T16:16:44Z	PASS	135	verify.sh	stale-real re-run; 11 deleted 0 err; clean
--- a/src/provisioning/providers/ecs-provider.ts
+++ b/src/provisioning/providers/ecs-provider.ts
@@ -1666,8 +1666,14 @@ export class ECSProvider implements ResourceProvider {
    *   2. AWS-defaulted-empty divergences are normalized away — `Cpu: 0`, empty
    *      arrays (`PortMappings` / `Environment` / ...), and an empty
    *      `LinuxParameters.Capabilities: {Add:[],Drop:[]}` are dropped so they
-   *      equal "absent" (what the template has). This keeps the `properties`
-   *      fallback baseline (pre-observedProperties state) drift-clean too.
+   *      equal "absent" (what the template has). AWS returns these defaults on
+   *      EVERY read, so without this the `properties` fallback baseline
+   *      (pre-observedProperties state) would phantom-drift on every run; the
+   *      normalization trades that for a rare edge — a template that EXPLICITLY
+   *      sets one of these to its empty/zero default AND has no observed baseline
+   *      (CDK L2 never emits e.g. an explicit `Cpu: 0`). On the normal
+   *      `observedProperties` path both sides run through this reader, so the
+   *      snapshot is self-consistent regardless.
    *
    * Free-form maps (`DockerLabels`, `LogConfiguration.Options`) are copied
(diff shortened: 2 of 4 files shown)


go-to-k:
Review (1-reviewer tier, code-quality): clean, no blockers. Verified SET/READ field-set + casing symmetry, freeform-map preservation (DockerLabels / LogConfiguration.Options), LinuxParameters normalization (no aliasing), and the observedProperties round-trip. One accepted nit: Environment/Secrets emit Value/ValueFrom unconditionally, so a partial AWS entry could yield `Value: undefined` — accepted as known cost (only the pre-observedProperties fallback path with a partial entry, and CFn Environment entries always carry a Value; symmetric with the SET side and the same documented normalization tradeoff). Verified end-to-end against real AWS (Phase 1e `cdkd drift` rc=0, no ContainerDefinitions drift).

Issue #1165, opened 2026-07-22
ECS (and other) SDK providers silently drop CFn PascalCase nested-object properties (SDK expects camelCase)


go-to-k:
## Summary

`ECSProvider` (and likely other SDK providers) passes CloudFormation **PascalCase** nested-object property values straight into AWS SDK v3 **camelCase** input fields. The SDK reads the camelCase keys it expects, finds them absent, and serializes nothing — so the nested object is **silently dropped** on create AND update. Deploy reports success, state records the intended value, and `cdkd diff` shows "No changes", so the divergence is permanent and invisible (same silent-drop class as #1160, but on the SET path rather than removal).

Confirmed while doing the #1160 ECS Service batch (#1164), which deferred `DeploymentConfiguration` removal-reset for exactly this reason.

## Root cause

cdkd does NOT globally normalize CFn PascalCase keys to SDK camelCase. Providers receive `properties` with CFn casing (`MaximumPercent`), and AWS SDK v3 input shapes are camelCase (`maximumPercent`). For a **scalar** field the value casing is irrelevant. For a **nested object / array-of-objects** field, the SUB-KEYS matter, and the provider must convert them per-field. `ECSProvider` DOES convert some (`convertNetworkConfiguration`, `convertLoadBalancers`) but passes others raw.

## Evidence (confirmed live against real AWS, 2026-07-22)

Reproducing exactly what `ECSProvider.create()` does — passing the CFn PascalCase `DeploymentConfiguration` object into the SDK's `deploymentConfiguration` field via `@aws-sdk/client-ecs` `CreateServiceCommand`:

```js
deploymentConfiguration: { MaximumPercent: 250, MinimumHealthyPercent: 50,
                           DeploymentCircuitBreaker: { Enable: true, Rollback: true } }
```

`DescribeServices` afterwards returns the **defaults**, not the intended values:

```json
{ "maximumPercent": 200, "minimumHealthyPercent": 100,
  "deploymentCircuitBreaker": { "enable": false, "rollback": false } }
```

So a CDK L2 `FargateService` with a custom `minHealthyPercent` / `maxHealthyPercent` / deployment circuit breaker / deployment alarms deploys with AWS defaults, silently.

## Affected ECS fields (same raw-PascalCase pass-through pattern)

`src/provisioning/providers/ecs-provider.ts` `createService()` / `updateService()`:

| Field | SDK key | Status |
|---|---|---|
[...]

PR #1174, opened 2026-07-23
fix(providers): map previously-dropped ECS ContainerDefinition sub-fields (#1173)

### LINKED ISSUE #1173: ECS convertContainerDefinitions silently drops container sub-fields (RepositoryCredentials / FirelensConfiguration / ResourceRequirements / ...) on RegisterTaskDefinition
## Summary

`ECSProvider.convertContainerDefinitions` (the `RegisterTaskDefinition` SET path) maps only 30 of the SDK `ContainerDefinition` fields. The following CFn `AWS::ECS::TaskDefinition` container-definition sub-properties are **not mapped**, so they are **silently dropped on register** (the same silent-drop class as #1165, but one level deeper — inside the `ContainerDefinitions` array):

- `RepositoryCredentials` (private registry auth)
- `FirelensConfiguration` (FireLens custom logging; `Options` is a free-form map)
- `ResourceRequirements` (GPU / InferenceAccelerator)
- `SystemControls` (sysctls)
- `ExtraHosts`
- `DnsServers` / `DnsSearchDomains`
- `DockerSecurityOptions`
- `Hostname`
- `RestartPolicy`
- `CredentialSpecs`
- `VersionConsistency`

## Why the pre-flight misses it

The top-level `property-coverage` pre-flight (which rejects a deploy that uses a silently-dropped **top-level** CFn property) does NOT descend into nested arrays, so `ContainerDefinitions[].<sub-field>` drops are invisible to it. A template that sets e.g. `RepositoryCredentials` or `FirelensConfiguration` on a container deploys "successfully" while the value never reaches AWS.

## Fix direction

Extend `convertContainerDefinitions` to map each field (free-form maps like `FirelensConfiguration.Options` copied verbatim, everything else a first-letter flip), and mirror it in the `containerDefinitionsToCfn` read-side reverse-mapper (added in #1169) so `cdkd drift` tracks the newly-settable fields too. Add unit tests for both directions plus a Fargate-compatible integ assertion (e.g. `RestartPolicy`).

Surfaced while working #1169 (the read-side `ContainerDefinitions` reverse-map is correctly symmetric with the SET path, so it is NOT a drift bug — this is purely the write side).

go-to-k:
## Summary

Fixes #1173.

`ECSProvider.convertContainerDefinitions` (the `RegisterTaskDefinition` SET path)
mapped only 30 of the SDK `ContainerDefinition` fields, so these 12
container-definition sub-properties were **silently dropped on register** (the
same silent-drop class as #1165, one level deeper — the top-level
`property-coverage` pre-flight does not descend into the `ContainerDefinitions`
array):

`RepositoryCredentials` (private registry auth), `FirelensConfiguration`
(FireLens custom logging), `ResourceRequirements` (GPU / InferenceAccelerator),
`SystemControls`, `ExtraHosts`, `DnsServers`, `DnsSearchDomains`,
`DockerSecurityOptions`, `Hostname`, `RestartPolicy`, `CredentialSpecs`,
`VersionConsistency`.

- **SET** (`convertContainerDefinitions`): map each field. A mechanical
  first-letter flip (`pascalToCamelCaseKeys`) for the structured ones, and an
  explicit `convertFirelensConfiguration` for `FirelensConfiguration` whose
  `Options` is a free-form map (user FireLens option keys, NOT CFn property
  names) that must be copied verbatim.
- **READ** (`containerDefinitionsToCfn`, the #1169 reverse-mapper): reverse-map
  them symmetrically so `cdkd drift` tracks the newly-settable fields too, and
  the SET/READ round-trip stays symmetric.

## Test plan

- **Unit**: a SET test asserts the 12 fields reach `RegisterTaskDefinition` in
  SDK camelCase (FireLens `Options` keys preserved verbatim); a READ test asserts
  the reverse-map back to CFn PascalCase; the "passes through undefined" test now
  also asserts the 12 fields stay `undefined` on a minimal container. Full suite
  green (7638 tests).
- **Integ** (`ecs-service-update-props`, real AWS `us-east-1`): the fixture sets
  `RestartPolicy` (Fargate-compatible) via the L1 escape hatch; new **Phase 1f**
  asserts (a) `describe-task-definition` returns `restartPolicy.enabled=True`
  (the write actually reached AWS — before the fix it was dropped) AND (b) the
  deploy-time `observedProperties.ContainerDefinitions[0].RestartPolicy.Enabled`
  is `true` (the read-side reverse-map). The pre-existing Phase 1e `cdkd drift`
  stays clean (RestartPolicy round-trips). Destroy 0 errors, 0 orphans.

Surfaced while working #1169.

### DIFF
--- a/docs/_generated/integ-last-run.tsv
+++ b/docs/_generated/integ-last-run.tsv
@@ -83,7 +83,7 @@ ec2-vpc	2026-06-21T11:12:35Z	PASS	53	standard	sweep-F staleness re-run (rc=0)
 ecr	2026-06-21T11:00:28Z	PASS	39	standard	sweep-4..8 staleness re-run (ledger-restore; rc=0)
 ecr-scanning	2026-07-20T17:37:23Z	PASS	56	verify.sh	ECR scanOnPush + tag add/change/untag update, clean destroy
 ecs-fargate	2026-07-02T18:23:15Z	PASS	373	verify.sh	0703 staleness sweep (pre-TTL refresh); clean
-ecs-service-update-props	2026-07-23T07:43:53Z	PASS		verify.sh	#1169 Phase1d/1e ContainerDefinitions drift-clean; destroy 15 del 0 err 0 orph
+ecs-service-update-props	2026-07-23T08:11:17Z	PASS		verify.sh	#1173 Phase1f RestartPolicy write+read round-trip; destroy 15 del 0 err 0 orph
 efs-immutable-replacement	2026-07-19T19:05:41Z	PASS	76	verify.sh	rc ok, 1 deleted, orph clean
 efs-lambda	2026-06-21T14:48:46Z	PASS	560	standard	sweep-G standard re-run (rc=0)
 efs-standalone	2026-06-21T16:16:44Z	PASS	135	verify.sh	stale-real re-run; 11 deleted 0 err; clean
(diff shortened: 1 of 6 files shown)


go-to-k:
Re-review after addressing the VersionConsistency nit: clean, no blockers. The normalization drops only the exact AWS default ('enabled') on READ, mirroring the Cpu:0 pattern; real drift (template 'disabled', or a drift to 'disabled') still surfaces. The reviewer's suggested 'disabled still surfaces' assertion is already covered by the main #1173 READ test (which sets versionConsistency:'disabled' and asserts VersionConsistency:'disabled' in the output), so both halves of the contract are pinned. Re-verified end-to-end against real AWS (Phase 1e drift rc=0 + Phase 1f RestartPolicy round-trip both clean with the normalization in place; destroy 15 / 0 errors / 0 orphans).

Issue #1421, opened 2026-08-09
test(integ): no real-AWS coverage for the GlobalTable PAY_PER_REQUEST -> PROVISIONED billing flip with GSIs


go-to-k:
Found while reviewing the #1387 fix (PR for `fix/1387-globaltable-gsi-throughput`).

## Summary

#1387 added a fifth call site the issue never named: the `PAY_PER_REQUEST -> PROVISIONED` billing-mode flip. AWS requires per-index `ProvisionedThroughput` in the SAME `UpdateTable` call that changes `BillingMode`, so without it the fix would have made create work while leaving the flip broken.

That claim — "AWS accepts `GlobalSecondaryIndexUpdates` alongside `BillingMode` in one `UpdateTable`" — is currently asserted only against a mocked DynamoDB client. The `dynamodb-globaltable` integ deploys both new GSI tables unconditionally and never mutates them, and `HistoryTable` (the only table the `CDKD_TEST_UPDATE` flow touches) has no GSI. So the flip path has no real-AWS coverage.

If AWS actually rejects that combination, the unit suite stays green and every `PAY_PER_REQUEST -> PROVISIONED` deploy of a GSI-carrying table fails in the field.

A second sub-path shares the same gap: an index the deploy REMOVES is still live on AWS at flip time (its `Delete` is issued later), so it too must carry throughput in the flip call. That behavior is unit-bound but likewise unverified against real AWS.

## Scope

Gate a `billing-provisioned` variant of the on-demand GSI table behind `CDKD_TEST_UPDATE` in `tests/integration/dynamodb-globaltable/`:

- Phase 1 deploys it `PAY_PER_REQUEST` with two GSIs.
- Phase 2 (`CDKD_TEST_UPDATE=true`) flips it to `PROVISIONED` while dropping one of the two GSIs.
- `verify.sh` asserts the flip succeeded, the surviving index has the expected `ProvisionedThroughput`, and the dropped index is gone.

That single phase covers both the flip and the removed-index sub-path.

## Related

- Per-GSI auto-scaling registration gap (filed separately).
- `readCurrentState` GSI phantom drift (filed separately).

go-to-k:
[...]

Issue #1420, opened 2026-08-09
AWS::DynamoDB::GlobalTable: readCurrentState stores the raw SDK GSI descriptions, so drift reports permanent phantom changes


go-to-k:
Found while reviewing the #1387 fix (PR for `fix/1387-globaltable-gsi-throughput`). Pre-existing, deliberately not fixed there.

## Summary

`DynamoDBGlobalTableProvider.readCurrentState` writes the raw SDK `GlobalSecondaryIndexDescription[]` straight into the CFn-shaped `GlobalSecondaryIndexes` state key, and `getDriftUnknownPaths` does not exclude it.

The SDK description carries members the CFn shape has no concept of (`IndexArn`, `IndexStatus`, `ItemCount`, `IndexSizeBytes`) and — after #1387 — models throughput completely differently from the CFn side (`ProvisionedThroughput` vs `WriteProvisionedThroughputSettings.WriteCapacityAutoScalingSettings`, and per-GSI read capacity living on the replica). So `cdkd drift` compares a CFn-shaped baseline against an SDK-shaped current snapshot and reports permanent phantom drift on any GlobalTable that has a GSI.

#1387 did not introduce this, but it makes GSIs a first-class shape in this provider, so the mismatch is now much more likely to be hit.

## Why it was not fixed in #1387

A correct reverse map is not a mechanical inversion. Per-GSI read capacity has to be split back out onto the right replica entry, and the auto-scaling half of `WriteProvisionedThroughputSettings` can only be reconstructed by reading Application Auto Scaling (`describe-scalable-targets` / `describe-scaling-policies` for `table/<t>/index/<i>`) — which the provider does not do today, and which is itself blocked on the per-GSI auto-scaling registration gap filed alongside this issue.

Half-fixing it (stripping the SDK-only members but leaving throughput unmapped) would trade one phantom drift for another, so it was left untouched rather than partially addressed.

## Options

1. Implement the full reverse map, including the Application Auto Scaling read. Correct, but depends on the auto-scaling work.
2. Interim: declare `GlobalSecondaryIndexes` in `getDriftUnknownPaths` so drift stops reporting a diff it cannot compute. Honest and cheap, but it turns off drift detection for the whole GSI subtree.

Option 2 is a reasonable stopgap if the phantom drift is hitting users before option 1 lands.

## Reproduce

[...]

Issue #1419, opened 2026-08-09
AWS::DynamoDB::GlobalTable: per-GSI auto-scaling never registered (and create() registers no auto-scaling at all)


go-to-k:
Found while reviewing the #1387 fix (PR for `fix/1387-globaltable-gsi-throughput`). Pre-existing, deliberately not fixed there.

## Summary

`DynamoDBGlobalTableProvider` registers Application Auto Scaling targets only for the TABLE (`dynamodb:table:ReadCapacityUnits` / `WriteCapacityUnits`). There is no `dynamodb:index:*` registration anywhere, so a per-GSI `WriteCapacityAutoScalingSettings` / `ReadCapacityAutoScalingSettings` block never becomes a real scaling policy.

After #1387 the per-GSI auto-scaling block yields the correct INITIAL capacity (`SeedCapacity ?? MinCapacity` -> `ProvisionedThroughput`), so a PROVISIONED GlobalTable with a GSI now creates successfully. But `MinCapacity` / `MaxCapacity` / `TargetTrackingScalingPolicyConfiguration` are still dropped: the index is pinned at its seed capacity forever and never scales.

This is easy to mistake for "working" now, because the #1387 integ asserts the seed value (`3`) reaching AWS and passes.

## Sibling gap in the same area

`create()` registers NO auto-scaling target at all — even at table level. Only `update()` calls `applyAutoScalingDiff`. So a freshly created PROVISIONED GlobalTable has no auto-scaling until some later deploy happens to run an update. The two belong together as one coherent piece of work.

## Scope

- Register `dynamodb:index:ReadCapacityUnits` / `dynamodb:index:WriteCapacityUnits` scalable targets with resource id `table/<table>/index/<index>`.
- Apply the target-tracking policy from `TargetTrackingScalingPolicyConfiguration`.
- Do it on `create()` as well as `update()`, table level and index level.
- Extend the `dynamodb-globaltable` integ to assert the registered scalable targets and policies via `application-autoscaling describe-scalable-targets` / `describe-scaling-policies`, not just the seed capacity.

go-to-k:
Live-probe finding from the #1427 probe that lands directly on this issue's scope: CloudFormation creates an autoscaled PROVISIONED GlobalTable at **`MinCapacity`**, while cdkd's `deriveReadCapacityUnits` / `deriveWriteCapacityUnits` return `SeedCapacity ?? MinCapacity`. Filed as #1435.

[...]

Issue #1393, opened 2026-08-08
audit: nested-key critic blind spots found by the 0809 sweep — mixed-case SDK models, file-global literal heuristic, missing targets, selective sub-key forwards


go-to-k:
## Summary

The 2026-08-09 horizontal sweep of the #1370 class (run with the #1373 critic's own machinery against ALL registered types' live CFn schemas) surfaced bugs the current critic design cannot see. Follow-up work to close the blind spots, complementing #1378 (shape-level v2):

### 1. Mixed-case SDK models defeat the single per-target `keyStyle`

`NESTED_KEY_TARGETS` assumes a whole client is `exact` (PascalCase) or `lower-first` (camelCase). But `client-eventbridge`, `client-scheduler`, and `client-glue` are PascalCase models with camelCase ISLANDS (`awsvpcConfiguration`, `capacityProvider`, `type`/`field`/`expression`, `scanRate`/`scanAll`). An extra classification bucket — "exact-case miss but case-insensitive hit" — is what surfaced the EventBridge/Scheduler Fargate-target create failures and the Glue Crawler drops in the sweep. The critic should flag case-insensitive near-misses REGARDLESS of the target's declared keyStyle (it already has the `case-divergence` bucket; the gap is that non-target types aren't audited at all and keyStyle hides islands).

### 2. "Provider names the literal anywhere" heuristic is file-global

The DynamoDB GlobalTable GSI divergence (`WriteProvisionedThroughputSettings` etc.) was cleared by the literal heuristic because the same key names occur legitimately handled at TOP level elsewhere in the file. Literal evidence should be scoped (per handled top-level property, or at least weighted) — or such known-divergent blobs get explicit allow/deny entries.

### 3. Extend `NESTED_KEY_TARGETS` to the remaining blob-forwarding providers

[...]

PR #1422, opened 2026-08-09
fix(dynamodb): translate GlobalTable GSI throughput to the CreateTable SDK shape

### LINKED ISSUE #1387: `AWS::DynamoDB::GlobalTable`: GSI `Write/ReadProvisionedThroughputSettings` / on-demand settings not translated to SDK shape — PROVISIONED GSI create fails, TableV2 per-GSI limits silently dropped
## Summary

`DynamoDBGlobalTableProvider` casts the CFn `GlobalSecondaryIndexes` blob raw to the SDK's `GlobalSecondaryIndex[]` (`src/provisioning/providers/dynamodb-globaltable-provider.ts:346-349`). The `AWS::DynamoDB::GlobalTable` CFn schema models GSI throughput differently from the `CreateTable` SDK shape:

| CFn GlobalTable GSI key | SDK `GlobalSecondaryIndex` member | Effect |
| --- | --- | --- |
| `WriteProvisionedThroughputSettings` | `ProvisionedThroughput` | PROVISIONED-billing GlobalTable with a GSI → **CreateTable fails** (GSI throughput missing) |
| `WriteOnDemandThroughputSettings` / `ReadOnDemandThroughputSettings` (per-replica) | `OnDemandThroughput` | TableV2 `maxReadRequestUnits` per-GSI → silent drop |
| `Replicas[].ReplicaStreamSpecification.ResourcePolicy` | (per-replica API) | silent drop (exotic) |

CDK's `TableV2` (the recommended L2 since 2.95) synthesizes `AWS::DynamoDB::GlobalTable`, so this is a daily-pattern surface.

Note for the #1373 critic: this divergence is invisible to the current "provider names the literal anywhere" heuristic because the same key names appear handled at top level — found only by manual review (see the critic-extension issue).

Fix: translate the GlobalTable GSI/replica throughput shapes to the SDK shapes in create (and wherever update touches GSIs), with unit tests; live-verify with a PROVISIONED + GSI fixture.

Found by the 2026-08-09 bug-hunt static sweep; code+SDK-model verified, not yet live-probed.

go-to-k:
## Summary

`DynamoDBGlobalTableProvider` cast the CloudFormation `GlobalSecondaryIndexes` blob RAW to the SDK's `GlobalSecondaryIndex[]`. The two schemas model per-GSI throughput completely differently, and the AWS SDK v3 serializer silently drops unknown members, so:

- a **PROVISIONED**-billing GlobalTable with a GSI failed `CreateTable` **outright** (AWS requires `ProvisionedThroughput` on every GSI), and
- every `TableV2` per-GSI on-demand limit (`maxReadRequestUnits` / `maxWriteRequestUnits`) vanished without a trace.

`TableV2` is the recommended CDK L2 since 2.95 and synthesizes `AWS::DynamoDB::GlobalTable`, so this is a daily-pattern surface.

## The issue's mapping table was incomplete, and re-deriving it changed the fix

The shapes were re-derived from the live `cloudformation:DescribeType` schema and cross-checked against a real `cdk synth` of `TableV2` under both billing modes. Two corrections mattered:

1. **`WriteProvisionedThroughputSettings` has exactly one member**, `WriteCapacityAutoScalingSettings`. There is no literal `WriteCapacityUnits` — write capacity on a GlobalTable is always auto-scaled. `CreateTable` needs a concrete number, so the mapping takes `SeedCapacity` (the documented "initial provisioned capacity units") before `MinCapacity`.
2. **Per-GSI READ capacity is not on the top-level GSI at all.** CDK synthesizes it to `Replicas[?Region==<deploy region>].GlobalSecondaryIndexes[].Read{Provisioned,OnDemand}ThroughputSettings`, so the SDK's single `ProvisionedThroughput` / `OnDemandThroughput` object has to be **fused** from both halves. The GSI-level spellings the schema also permits are honored as a fallback for hand-authored templates.

The repo's captured `AWS-DynamoDB-GlobalTable.json` fixture stores top-level names only and could not have settled either point.

## Mapping

| CFn | SDK |
| --- | --- |
| `GSI.WriteProvisionedThroughputSettings.WriteCapacityAutoScalingSettings.{SeedCapacity ?? MinCapacity}` | `ProvisionedThroughput.WriteCapacityUnits` |
| `Replicas[local].GSI[].ReadProvisionedThroughputSettings.ReadCapacityUnits` (fallback: GSI-level) | `ProvisionedThroughput.ReadCapacityUnits` |
| `GSI.WriteOnDemandThroughputSettings.MaxWriteRequestUnits` | `OnDemandThroughput.MaxWriteRequestUnits` |
| `Replicas[local].GSI[].ReadOnDemandThroughputSettings.MaxReadRequestUnits` | `OnDemandThroughput.MaxReadRequestUnits` |
| `GSI.WarmThroughput` | `WarmThroughput` (same spelling) |
| `Replicas[].GSI[].ReadProvisionedThroughputSettings.ReadCapacityUnits` | `ProvisionedThroughputOverride.ReadCapacityUnits` |
| `Replicas[].GSI[].ReadOnDemandThroughputSettings.MaxReadRequestUnits` | `OnDemandThroughputOverride.MaxReadRequestUnits` |

## Call sites

1. `create()` — `CreateTable.GlobalSecondaryIndexes` translated instead of cast.
2. `addReplica()` — replica GSI overrides translated.
3. `update()` replica-modify — same translation on `UpdateReplicationGroupMemberAction`.
4. `update()` GSI diff — **both sides translated before diffing**, so emitted `Create` / `Update` actions carry real throughput. Side benefit: an auto-scaling-only edit (`MaxCapacity` 20 to 30 — invisible to the DynamoDB API) no longer produces a bare `Update: { IndexName }` that AWS rejects as empty; a modified GSI yielding no throughput at all now warns and skips.
5. `update()` **billing-mode flip** — not named in the issue. AWS requires per-index `ProvisionedThroughput` in the *same* `UpdateTable` call that flips `PAY_PER_REQUEST` to `PROVISIONED`; without this the fix would have made create work while leaving the flip broken.

## Deliberately left unmapped (recorded in-code, not silently dropped)

This provider has no `unhandledByDesign` map, so the rationale lives in a JSDoc block on `addReplica`:

- `Replicas[].ReplicaStreamSpecification.ResourcePolicy` and `Replicas[].ResourcePolicy` — both need `PutResourcePolicy`, not any `UpdateTable` field.
- `Replicas[].GlobalSecondaryIndexes[].ContributorInsightsSpecification` — needs a per-index `UpdateContributorInsights`.

`handledProperties` is unchanged, so no coverage regeneration was required.

## Review findings fixed in this PR

A 2-axis review (code quality + test adequacy, both mutation-tested rather than read-off) found two real defects in the new code, fixed here:

1. **Blocker — the billing flip omitted an index this deploy REMOVES.** The flip built its index updates from the NEW template but filtered them by the names in the PREVIOUS one, so a dropped index fell through both. Its `Delete` is issued in step 6, which runs AFTER the flip, so at flip time it is still a live index on a table becoming PROVISIONED — and by this PR's own premise AWS rejects the call for having no capacity on it. Its throughput now comes from the previous template, and only SURVIVING indexes are marked handled so the removed one still reaches its `Delete`.
2. **Major — a false warning on every `PROVISIONED -> PAY_PER_REQUEST` flip.** The two sides are translated under different billing modes, so every GSI necessarily lands in `modified` with nothing to send and hit the "KeySchema / Projection are immutable" warning. That message is untrue and sends the user hunting a non-existent template problem; the loop now skips entirely when the billing mode changed, since step 4 already applied the change atomically.

Three test-adequacy gaps were also closed: a test asserting members are OMITTED when unset (relaxing that guard previously passed the whole suite), a binding test for the warn-and-skip guard, and an honest comment on the auto-scaling-only test explaining that it is a both-must-hold end-state assertion rather than a binding test for either mechanism.

## Deliberately not fixed here (filed)

- (#1419) per-GSI auto-scaling is never registered — `WriteCapacityAutoScalingSettings` now yields the correct INITIAL capacity, but no `dynamodb:index:*` scalable target is created, so the index never scales. `create()` registers no auto-scaling at all, even table-level; the two belong together.
- (#1420) `readCurrentState` stores the raw SDK GSI descriptions in the CFn-shaped state key, so `cdkd drift` reports permanent phantom drift on any GSI-carrying GlobalTable. Pre-existing; a correct reverse map needs the Application Auto Scaling read that (#1419) is about, so half-fixing it would trade one phantom drift for another.
- (#1421) the billing-flip path has no real-AWS coverage — it is asserted against a mock only. Both new fixture tables are unconditional and never mutated.
- `AWS::DynamoDB::GlobalTable` is still absent from `NESTED_KEY_TARGETS`, which is exactly this bug class; critic target expansion is tracked in (#1393).

## Test plan

- **26 new unit tests** in `tests/unit/provisioning/dynamodb-globaltable-provider-gsi-throughput.test.ts`. Both property bags are copied **verbatim from a real `cdk synth`** rather than hand-invented — which is exactly what surfaced the read-capacity-lives-on-the-replica asymmetry a hand-written fixture would have encoded wrongly.
- **Mutation-proofed, per call site.** Reverting each of the five call sites individually kills a distinct test; so does removing the removed-index handling, the billing-flip dedupe, the warn-and-skip guard, and the omit-when-unset guard. Every assertion added by the review was verified to go red under its own mutation before being kept.
- **Real AWS integ** (`/run-integ dynamodb-globaltable`, us-east-1, 207s, PASS) — re-run after the review fixes, since those touched the provider: the fixture gains two **unconditional** `TableV2`s (not gated behind `CDKD_TEST_UPDATE`, so the baseline deploy exercises the previously-failing create path) — two tables because the billing modes cannot coexist on one. Live assertions:
  - `byStatus.ProvisionedThroughput.ReadCapacityUnits = 7`
  - `byStatus.ProvisionedThroughput.WriteCapacityUnits = 3`
  - `byOwner.OnDemandThroughput.MaxReadRequestUnits = 50`
  - `byOwner.OnDemandThroughput.MaxWriteRequestUnits = 60`
  - table-level throughput preserved on both fixtures

  Destroy: 3 deleted, **0 errors, 0 orphans**; state cleared.
- `verify.sh` step 3 stopped taking "the first `AWS::DynamoDB::GlobalTable` in state" (which with three tables would have grabbed an arbitrary one) in favor of selecting by logical-id prefix.
- Full suite: 520 files / 8868 tests pass.

Closes #1387


go-to-k:
## 3-axis review of this PR — 1 blocker, 4 minors

Handing these off: another session owns this branch, so this comment records
the findings rather than pushing the fix. A 3-axis review (spec / code / test)
was run against `f5dbd379`; all three reviewers independently flagged the same
blocker.

### BLOCKER — a `PROVISIONED` -> `PAY_PER_REQUEST` flip silently drops every per-GSI on-demand limit

`src/provisioning/providers/dynamodb-globaltable-provider.ts`

The flip call attaches `GlobalSecondaryIndexUpdates` only when
`newBilling === 'PROVISIONED'`, and step 6's `modified` loop then `continue`s
unconditionally on ANY billing change. So a template flipping a table to
on-demand *with* `maxReadRequestUnits` / `maxWriteRequestUnits` on a GSI sends
neither field — the value is lost.

This is the exact silent-drop class this PR exists to close (#1387), in the
direction the PR did not test. Concretely: `toSdkGlobalSecondaryIndexes`
DOES derive `OnDemandThroughput` for the new side under `PAY_PER_REQUEST`, so
the data is computed and then discarded by the `continue`.

Suggested shape (verified locally against the full suite):

- Drop the blanket `if (oldBilling !== newBilling) continue;` and keep the
  existing `gsiHandledByBillingFlip` check, which already filters out exactly
  the to-`PROVISIONED` direction that step 4 applies atomically. That also
  makes `gsiHandledByBillingFlip` load-bearing — see minor 3 below.
- During a flip, send only the fields the NEW mode needs (`OnDemandThroughput`),
  and suppress the immutable-field warning, since a flip puts every index in
  `modified` by construction.
- Apply it as its own round-trip AFTER the flip settles rather than adding a
  per-GSI field to the flip call itself — AWS requires the atomic form only in
  the to-`PROVISIONED` direction, and the reverse payload is unverified.

### Minor 1 — `WarmThroughput` is dropped on both update paths

`create()` sends `WarmThroughput` (it is set by
`toSdkGlobalSecondaryIndexes`), but the update path's `Create` action spreads
only `ProvisionedThroughput` / `OnDemandThroughput`, and the `Update` action
does the same. Both `CreateGlobalSecondaryIndexAction` and
`UpdateGlobalSecondaryIndexAction` declare `WarmThroughput` (confirmed in
`@aws-sdk/client-dynamodb` `models_0.d.ts`), so:

- a GSI ADDED by an update loses warm throughput while the same GSI declared
  up front keeps it (create/update asymmetry), and
- a `WarmThroughput`-only edit produces no throughput fields, falls into the
  throughput-less guard, and is reported as an immutable
  `KeySchema / Projection` change — a false diagnosis that also loses the edit.

### Minor 2 — unguarded `.map` on a possibly-non-array

In the flip's existing-index scan,
`((previousProperties['GlobalSecondaryIndexes'] ?? []) as unknown[]).map(...)`
throws `TypeError` if the value is a non-array object (an unresolved
intrinsic). Every other new site in this diff guards with `Array.isArray`.

### Minor 3 — `gsiHandledByBillingFlip` is currently dead code

It is populated only when `oldBilling !== newBilling`, and the blanket
`continue` already covers that condition, so the `has()` check can never
change behavior. Fixing the blocker as described above is what makes it live;
if the blocker is fixed some other way, delete the set instead.

### Minor 4 — stale header comment

The file header states `UpdateTable` accepts only ONE of
`{BillingMode, ReplicaUpdates, GlobalSecondaryIndexUpdates}` per call, but the
flip path now deliberately sends `BillingMode` + `GlobalSecondaryIndexUpdates`
together (which is correct — AWS REQUIRES it in that direction). The comment
is described as load-bearing for the next editor, so it should say that.

### Nit — the explicit-throughput escape hatch skips numeric coercion

`explicitProvisioned as unknown as ProvisionedThroughput` forwards the record
raw, making it the one path that bypasses `toFiniteNumber`. CFn is stringly
typed, so a hand-authored `"5"` reaches the SDK unnormalized while every
derived value is coerced. Same for the two replica `*ThroughputOverride` sites.

### Test-side notes (fixtures verified realistic — no action needed)

The reviewer re-ran a real `cdk synth` (aws-cdk-lib 2.244.0) of the `TableV2`
snippets quoted in the test docblocks and diffed the emitted `Properties`
against `PROVISIONED_TABLE_PROPS` / `ON_DEMAND_TABLE_PROPS`: **byte-identical**,
including the asymmetry that matters (read capacity only on the replica GSI, no
literal `WriteCapacityUnits` on the top-level GSI). Presence/absence is pinned,
not just values. That is the opposite of the EMR over-supplied-fixture failure
mode, and worth keeping.

Remaining coverage gaps, in priority order:

1. The blocker direction has no test at all (a test asserting the emitted
   `Update: { IndexName, OnDemandThroughput }` fails without the fix and passes
   with it — confirmed locally as the only failure out of 8936).
2. `gsiDiff.added` under a flip is untested; the design comment claims step 6's
   `added` loop supplies the throughput there.
3. Live coverage is create-path only. `verify.sh` genuinely asserts real
   `DescribeTable` values (not exit-0), but the only table put through
   `CDKD_TEST_UPDATE=billing-provisioned` has no GSIs, so no UPDATE-path branch
   is live-verified. Already tracked as #1421.

### Deliberately out of scope — filed as #1423

Removing a per-GSI on-demand limit from the template silently no-ops (the limit
stays at its old value in AWS; CFn would reset it). That is the #1160
absent-field-reset class and the `-1` reset sentinel is unverified for the
per-GSI `Update` action, so it needs a live probe rather than a guess. #1423
also carries the non-array-`GlobalSecondaryIndexes`-swallowed-at-create
inconsistency.

Issue #1513, opened 2026-08-10
config-shape: decide the TOP-LEVEL `??` defaulting sites left out of #1493 (AuthorizationType -> NONE, IpProtocol, InstanceType, ...)


go-to-k:
## Summary

Split out of #1493 item 1 while rolling `config-shape.ts` across the `??`-shaped
defaulting sites. The measured `??` set is 24 sites; the PR rolls the 9 that
index a NESTED container (plus the 2 `AWS::Lambda::Url` `AuthType` sites the
#1471 sweep's cast-specific grep missed). This issue covers the remaining
TOP-LEVEL reads, which were deliberately left alone with the reason recorded in
`src/provisioning/config-shape.ts`'s module JSDoc.

## The sites

All of the shape `(properties['X'] as T) ?? 'default'`, desired side:

- `ec2-provider.ts` — `IpProtocol ?? '-1'` (x2), `InstanceType ?? 't3.micro'`,
  `Domain ?? 'vpc'`
- `apigateway-provider.ts:1390` — `AuthorizationType ?? 'NONE'` (a null value
  silently yields a PUBLIC method)
- `iam-access-key-provider.ts` — `Status ?? 'Active'` (x2)
- `lambda-event-invoke-config-provider.ts` — `Qualifier ?? '$LATEST'` (x2)
- `rds-dbproxy-targetgroup-provider.ts` — `TargetGroupName ?? 'default'`
  (create + update only; the `delete()` / `readCurrentState` reads are
  state-side and must stay unguarded)
- `dynamodb-globaltable-provider.ts:308` — `BillingMode ?? 'PAY_PER_REQUEST'`
  (create; the `previousProperties` sibling at :931 must stay unguarded)

## Why they were not folded into the #1493 PR

The container is the provider's own `properties` bag, which is always a plain
object, so `readConfigString`'s rule 2 — the headline #1471 failure — cannot
fire. Only rules 3 and 4 (`requireConfigString`) would apply, and those turn a
present-but-non-string value into a hard refusal. That is a stricter-value
decision with its own regression surface, distinct from the malformed-container
class: an unquoted YAML `IpProtocol: -1` is a NUMBER today and deploys fine,
and a refusal would break a template that works.

## What this issue needs

A per-site decision, not a blanket sweep:

1. Which of these fields can legitimately carry a non-string scalar from a
   hand-authored L1 template (CFn coerces scalars; cdkd does not)? Those either
   need coercion before the guard or must stay unguarded.
[...]

Issue #1493, opened 2026-08-10
audit: the malformed-container defaulting class beyond the || shape — ?? siblings, silently-dropped Destination, misnamed flattened path


go-to-k:
## Summary

The 3-axis review of PR #1490 (issue (#1471)) confirmed that PR closes the class
for the shape the issue measured — `(cfg['K'] as string) || 'default'`, 16 sites
— but surfaced three residuals of the SAME failure mode that fall outside that
grep. Filing them together rather than widening that PR further.

The failure mode is unchanged: a nested container that is a string / array /
unresolved intrinsic indexes to `undefined`, and a fallback silently substitutes
a value the template did not ask for.

## 1. The `??`-shaped siblings

`readConfigString` was rolled across the `||` shape. The `??` spelling has the
same failure mode and was not in the measured set:

- `src/provisioning/providers/codebuild-provider.ts:116` —
  `source['Type'] ?? 'NO_SOURCE'`; a malformed `Source: 'GITHUB'` yields
  `NO_SOURCE`, i.e. a project with no source at all
- `src/provisioning/providers/codebuild-provider.ts:150`, `:325` — same shape
- `src/provisioning/providers/dynamodb-globaltable-provider.ts:349` —
  `streamSpecInput['StreamViewType'] ?? 'NEW_AND_OLD_IMAGES'`

Suggested: `git grep -nE "\] \?\? '" src/provisioning/providers/` to get the
full measured count first, exactly as the (#1471) comment did for `||`, then roll
the same helper across it. Report the count so a reviewer can tell a full sweep
from a spot-check.

## 2. A malformed `Destination` is silently DROPPED rather than defaulted

`src/provisioning/providers/s3-bucket-provider.ts:1059` (analytics) and `:1171`
(inventory): the code probes `dest?.['BucketAccountId'] || dest?.['BucketArn'] |
dest?.['Format']` to decide between the flattened and nested shapes. A non-object
`dest` makes every probe `undefined`, so `s3Dest` is `undefined` and the whole
`Destination` block is omitted from the Put — with no error.

This is a sibling of the defaulting class rather than an instance of it (nothing
is defaulted; the block just vanishes), which is why the helper does not cover
it. It needs its own decision: refuse, or keep dropping and warn.

## 3. Cosmetic: the flattened-branch error path is misnamed

At `s3-bucket-provider.ts:1074` / `:1195` the `containerPath` passed to
[...]

PR #1524, opened 2026-08-10
fix(provisioning): guard the top-level config-defaulting sites per site, coercing scalars and warning on the update path

go-to-k:
## Summary

Issue #1513: the TOP-LEVEL `(properties['X'] as string) ?? 'default'` reads that #1493 deliberately left alone. The container there is the provider's own property bag, so the malformed-CONTAINER rule cannot fire — what bites is a present-but-unusable VALUE (`null`, an object from a mis-nested template, an unresolved intrinsic), which `??` silently replaces with a default the template never asked for.

The headline site is `AWS::ApiGateway::Method AuthorizationType`, whose default is `NONE` — a **public method**.

## The three per-site decisions

`requireConfigString` grew an optional `ConfigStringOptions`, and every site was decided on its own rather than swept:

- **`coerceNumber` — coerce where an unquoted YAML scalar is legitimate.** CFn coerces scalars and cdkd does not, so `IpProtocol: -1` / `Qualifier: 1` arrive as NUMBERS and deploy fine today; refusing them would break working templates. The enum-valued sites (`AuthorizationType`, `InstanceType`, `Domain`, `Status`, `TargetGroupName`) do NOT coerce — a number there is a template bug.
- **`onUnusable` — throw on a template-borne CREATE, warn on every STATE-REPLAY path.** `rollback-executor.ts` replays a rollback via `provider.update(..., previousState.properties, ...)`, so the desired bag an `update()` sees can be a historical cdkd STATE record. A refusal there would make the resource not merely un-updatable but UN-ROLLBACKABLE, with no template-side remedy.
- **Check WHERE the read lives before guarding it.** `EC2Provider.buildIpPermission` is textually a top-level read, but it is also reached from `deleteSecurityGroupIngress` and from the REVOKE half of the inline-rule update diff — both carrying STATE-borne rules. A guard there would break destroy and rollback, so it stays unguarded and the create CALL SITE is guarded instead, which refuses before the helper is ever reached.

Guarded on create: API Gateway `AuthorizationType`, EC2 `InstanceType` / EIP `Domain` / SecurityGroupIngress `IpProtocol`, IAM access-key `Status`, Lambda event-invoke `Qualifier`, RDS DB-proxy `TargetGroupName`. The latter three also warn on update. Every `delete()` / `readCurrentState` read stays unguarded for the state-side reason.

## Test plan

- 34 unit tests on `config-shape.ts` (both new options, including the refusals each keeps: a non-finite number, `null`, blanks, objects and booleans still refuse under `coerceNumber`).
- New `tests/unit/provisioning/toplevel-config-defaults.test.ts` — 16 tests pinning each guarded call site, the warn-on-update wording, and the fence that `buildIpPermission` stays unguarded so a state-borne rule with `IpProtocol: null` can still be REVOKED.
- Full suite: 532 files / 9593 tests pass.
- **Real AWS, `apigateway` integ (2026-08-11)**: rc 0, 25 resources deleted with 0 errors, REST API and state file gone, no orphans. The 8 methods deploy with an explicitly declared `AuthorizationType`, so the guard ran on its happy path live.
- **Real AWS, refusal path**: a single-resource app with `Domain: null` was deployed through the real CLI. It failed with `AWS::EC2::EIP Domain must be a non-empty string (got null) — ... Omit the field entirely to use the default (vpc)` and `No completed operations to roll back` — i.e. **no EIP was allocated**, confirmed via `describe-addresses`. The same app with `Domain` omitted deployed an EIP with `Domain=vpc` and destroyed cleanly. Probe stack and its event history were swept.

## Not in this PR

- The GlobalTable `BillingMode` site, whose file is owned by the live `fix/1511-provisioned-throughput-diagnostic` lane. **#1513 stays open for it**, and the per-site precedent above is recorded as an issue comment for whoever picks it up.

Part of issue #1513.

## Review fix-back (3 reviewers, all converging on one root cause)

The PR's original premise — "create is always template-borne" — is **false**, and three replay paths reach a create guard:

- **Blocker:** `EC2Provider.updateSecurityGroupIngress` is delete-then-create, so the CREATE guard sat on the UPDATE path — and by the time it ran, `RevokeSecurityGroupIngress` had **already committed**. A state-borne `IpProtocol` (which the old `?? '-1'` accepted) would throw with the rule already deleted from AWS and no template edit able to fix it. `createSecurityGroupIngress` now takes an optional warn hook; the update path passes it, the plain create dispatch still throws.
- `rollback-executor.ts`'s reverse-replacement arm revives the OLD resource via `provider.create(..., previousState.properties, REPLAYING_STATE_CREATE_CONTEXT)`. `.claude/rules/providers.md` requires a create-side pre-flight refusal to downgrade there, so all five providers thread `CreateContext` and use a new shared `replayWarn(logger, context)`.
- The module header's false rationale is corrected rather than left standing, and now names all three paths.

Also fixed from the review:

- **IAM `Status` warn no longer defaults to `Active`.** Defaulting an unusable value to the CFn default would silently **enable a credential** the template did not ask to enable — the opposite-of-declared-intent substitution this guard exists to stop. It falls back to the PREVIOUS status instead. An *absent* `Status` still resets to `Active` (documented removal semantics, asserted by a pre-existing test), so the two cases are split.
- **RDS: the warn is no longer contradicted 12 lines later.** The immutability loop compared the RAW value, so a warned-about `null` still threw `ResourceUpdateNotSupportedError` and forced a REPLACEMENT — making "using the default for this update" false. It now compares the guarded value; a real immutable-identity change still refuses.
- **The EC2 coerce test now asserts the wire payload**, not just the physical id (which is byte-identical with or without the coercion, so it pinned nothing).

24 tests in the new file, 9601 total.


go-to-k:
## pr-review marker set manually, with maintainer go-ahead

Same situation as #1525, recorded for the same reason.

**This PR was reviewed at full 3-axis tier**, plus two verification passes:

- spec / code / test reviewers ran against `daf39e1a`. All three converged on one root cause — the PR's premise that "create is always template-borne" is FALSE — and the code reviewer raised a BLOCKER: `updateSecurityGroupIngress` is delete-then-create, so the create guard fired AFTER `RevokeSecurityGroupIngress` had committed, leaving the rule deleted from AWS with no template-side remedy.
- a re-review against `b7a2a739` found the first fix INCOMPLETE — the warned default never reached the wire, because `buildIpPermission` re-reads the raw bag and its own `??` only rescues `null`/`undefined`. Fixed in `4bd7dcb2`, along with the un-threaded SG-ingress replay create, the IAM warn that silently ENABLED a credential, and the RDS warn contradicted by a throw twelve lines later.

**Why the sha binding no longer matches:** `main` moved five times mid-session (a parallel lane merging, plus #1525 from this same session). Each rebase re-staled every gate. The rebases resolved conflicts in `docs/changelog-cdkd.md` only — **no provider logic changed in any of them**.

**Re-verified after the final rebase, not assumed:**

- full suite green at **9728 tests** — which now includes #1525's merged S3 tests, so the two lanes from this session are confirmed to compose rather than merely assumed to;
- the real-AWS `apigateway` integ re-run clean: 25 resources deleted, 0 errors, REST API and state file gone, no orphans;
- `integ-destroy` re-set off that run.

Re-running the reviewer agents over a changelog-conflict rebase was judged poor value against that evidence. Filed under the gate's documented escape hatch: a comment explaining why, plus the maintainer's explicit go-ahead.

Issue #1523, opened 2026-08-10
test(integ): cover the two #1495 replication members that need a KMS key / a second account (EncryptionConfiguration, AccessControlTranslation)


go-to-k:
Split out of #1495. That issue wired all 20 nested keys the S3 provider never wrote, and `tests/integration/s3-replication-and-filter` now reads five of them back off the live bucket (`LoggingConfiguration.TargetObjectKeyFormat`, `Destination.ReplicationTime` + `Metrics`, `SourceSelectionCriteria.ReplicaModifications`, `LifecycleConfiguration.TransitionDefaultMinimumObjectSize`).

Two members are deliberately NOT live-covered, with unit coverage only:

- **`Destination.EncryptionConfiguration.ReplicaKmsKeyID`** and its partner **`SourceSelectionCriteria.SseKmsEncryptedObjects`**. Exercising them needs a customer-managed KMS key, and a KMS key's minimum deletion window is 7 days — so every run of the fixture would leave a pending-deletion key behind. That directly conflicts with the repo's "never end an integ run with orphan resources" rule, which is why it was not simply added.
- **`Destination.AccessControlTranslation.Owner`**. `Owner: Destination` is an ownership OVERRIDE, which is only meaningful when the destination bucket is in a different account; same-account it is at best a no-op and the fixture would assert nothing real.

Options to consider, in rough order of preference:

1. Reuse a long-lived, pre-provisioned KMS key (not created/destroyed per run) referenced by alias, so no key is ever scheduled for deletion. Needs a decision on where that key lives and how a fresh account bootstraps it.
2. Accept the pending-deletion key and teach the orphan sweep to treat a `PendingDeletion` key created by this fixture as expected. Cheapest to write, weakest hygiene story.
3. Leave both unit-only and record it permanently. The write path is a per-member literal shared with the three blocks that ARE covered live, so the marginal risk is lower than it looks — but "lower" is not "measured", which is why this is filed rather than closed.

`AccessControlTranslation` additionally needs a second account, so it likely stays unit-only regardless of which option wins for the KMS pair.

The fixture's header comment points here.

go-to-k:
[...]

Issue #1520, opened 2026-08-10
gen-nested-key-coverage: declare AWS::S3::Bucket's segmentRenames and opt the target in (the structural half of #1495)


go-to-k:
Split out of #1495, whose first half — the 20 nested keys `s3-bucket-provider.ts` never wrote at all — is fixed. The never-written count is now **0** and the residual fell **98 -> 81**.

## What is left

The 81 remaining `no-write-evidence` paths are all written SOMEWHERE in the file and fail only to resolve at the audited CHAIN. Four structural causes, all of them `segmentRenames` work:

- **A CFn segment the SDK renames.** CFn `LoggingConfiguration` is the SDK's `BucketLoggingStatus.LoggingEnabled`; `WebsiteConfiguration.RoutingRules.RedirectRule` is `Redirect` and `.RoutingRuleCondition` is `Condition`.
- **An SDK-only wrapper segment.** CFn's `BucketEncryption.ServerSideEncryptionConfiguration` LIST is the SDK's `ServerSideEncryptionConfiguration.Rules`.
- **A member CFn nests that the SDK hoists onto the REQUEST.** `TransitionDefaultMinimumObjectSize` sits on `PutBucketLifecycleConfigurationRequest`, not inside `LifecycleConfiguration`. This one may not be expressible as a segment rename at all — decide whether the map needs a request-hoist form or whether the path is allow-listed with `passes: ['write']` and a rationale.
- **A per-item PUT API whose SDK top-level is SINGULAR where CFn's is plural.** `AnalyticsConfigurations` -> `AnalyticsConfiguration`, same for Inventory / Metrics / IntelligentTiering.

## Scope

1. Declare the renames on the `AWS::S3::Bucket` entry in `NESTED_KEY_TARGETS`. They are staleness-fenced by `findStaleSegmentRenames`, so a rename that stops applying fails `--check` rather than rotting.
2. Re-measure. If the residual reaches 0, set `freshObjectMapper: true` with measured `minWrittenMembers` / `minWriteScopes` / `minHandoffPoints` floors, which makes the whole `AWS::S3::Bucket` family non-regressing.
3. Whatever does not reach 0 gets a RECORDED, measured reason in the script header — never an allow-list entry added just to clear the bucket.

[...]

Issue #1474, opened 2026-08-10
gen-nested-key-coverage: the write-scope index cannot see the BUILDER idiom, blocking CloudWatch AnomalyDetector (3) and most of S3 (104)


go-to-k:
Follow-up from #1445, measured while landing the whole-blob hand-off walk.

## The shape

`collectWriteEvidence`'s scope index maps a written member name to every member
written BENEATH the VALUE it is written with, and it finds those members by
resolving the value to OBJECT LITERALS. The BUILDER idiom defeats that:

```ts
const sdkConfig: any = {};                       // <- resolves to an EMPTY literal
if (indexDoc) sdkConfig['IndexDocument'] = { Suffix: indexDoc };
if (errorDoc) sdkConfig['ErrorDocument'] = { Key: errorDoc };
...
await this.s3Client.send(new PutBucketWebsiteCommand({ WebsiteConfiguration: sdkConfig }));
```

The members ARE written, per member, on the forward path — but as
property/element-access assignments onto the identifier, which no literal walk
visits. `scopes.get('WebsiteConfiguration')` is therefore empty and every key
beneath it lands in `no-write-evidence`.

This is a DIFFERENT mechanism from #1445's generic converter (per-member writes
in a place the index does not look, versus a blob with no per-member write at
all), which is why it was recorded rather than folded into that change.

## Measured impact (post-#1445 tree)

Would-be `same-spelling` paths with no scoped write evidence, with
`freshObjectMapper: true` forced:

| target | count | cause |
| --- | --- | --- |
| `AWS::CloudWatch::AnomalyDetector` | 3 | `buildPutParams`'s `const mapped: AnomalyDetectorConfiguration = {}` builder — `Configuration.{ExcludedTimeRanges,StartTime,EndTime}` are all written |
| `AWS::S3::Bucket` | 104 | `applyWebsiteConfiguration` / `applyObjectLockConfiguration` / `applyReplicationConfiguration` + genuine per-member re-shaping across a dozen `apply*Configuration` helpers |

CloudWatch AnomalyDetector reaches **0** and can opt in the moment the builder
idiom is recognized; S3's residual has to be re-measured afterwards to separate
the builder half from genuine per-member re-shaping.

## Sketch

In the identifier branch of the scope walk, additionally collect
`<id>.Foo = v` / `<id>['Foo'] = v` assignments in the identifier's binding
scope as members of that identifier's value, recursing into `v`. Same lexical
scoping rules the existing binding search uses.

[...]

Issue #1475, opened 2026-08-10
gen-nested-key-coverage: the hand-off walk does not recognize a SPREAD-AND-PATCH forwarder, keeping CloudFront Distribution (110) unmeasurable


go-to-k:
Follow-up from #1445, measured while landing the whole-blob hand-off walk.

## The shape

`CloudFrontDistributionProvider.convertToSdkFormat` is a SPREAD-AND-PATCH
forwarder:

```ts
private convertToSdkFormat(config: Record<string, unknown>): Record<string, unknown> {
  const result = { ...config };          // <- delivers EVERY member verbatim
  if (result['Comment'] == null) result['Comment'] = '';
  for (const [cfnKey, sdkKey] of Object.entries(TOP_LEVEL_CFN_TO_SDK)) { ... }
  ...                                     // ~30 more named patches
  return result;
}
```

Every key the template carries reaches AWS through the spread; the named
patches only rename / wrap specific keys. #1445's genericity test asks "does
this function name a member of its own?" and answers yes (30+), so the whole
`DistributionConfig` blob is not recognized as a hand-off and all **110 of
CloudFront's 112** would-be `same-spelling` paths stay unmeasurable.

That rejection is correct BY THE CURRENT RULE and was left in place
deliberately: recognizing the spread is the recognizer most likely to become a
rubber stamp, since it clears 110 paths in one step.

## What a safe version would need

- The spread source must be the bag-derived SEED (`{ ...config }` where `config`
  is the tainted parameter), not any object — the taint root #1445 added is
  already the right test.
- Only members the function does NOT subsequently `delete` may be credited:
  `convertToSdkFormat` does `delete result[cfnKey]` on every rename, and a
  rename that writes the WRONG SDK key would otherwise be vouched for by the
  spread it just removed the key from.
- The fence has to keep firing: prove with a real-code probe that deleting a
  named patch (e.g. the `IPV6Enabled -> IsIPV6Enabled` rename) still exits
  non-zero. If it cannot, the recognizer is a rubber stamp and should not ship.

## Acceptance

- `AWS::CloudFront::Distribution` opts into `freshObjectMapper` at 0 findings
  with the fence demonstrably still live, or a re-measured reason replaces
  reason (D) in the script header.
- The measured table in `scripts/gen-nested-key-coverage.ts`'s file header and
  the `reproduces the measured opt-in table` test are updated.

go-to-k:
[...]

PR #1525, opened 2026-08-10
fix(provisioning): write the 20 nested S3 keys the provider never built at all

### LINKED ISSUE #1495: s3-bucket-provider: 20 nested CFn keys are never written at all (replication Destination blocks, SourceSelectionCriteria, TargetObjectKeyFormat) — write-side silent drops
Found while landing #1474 (the BUILDER-idiom recognizer for
`gen-nested-key-coverage`). Recognizing the builder shape cut
`AWS::S3::Bucket`'s would-be `same-spelling` residual from 125 to 98 paths and,
with the builder noise gone, the remainder splits cleanly:

| bucket | count | meaning |
| --- | --- | --- |
| written SOMEWHERE, not at the audited chain | 78 | CFn/SDK segment renames + plural-vs-singular per-item PUT APIs — `segmentRenames` work, not bugs |
| never written ANYWHERE in the file | **20** | the terminal SDK member exists and `s3-bucket-provider.ts` contains no write of that name at all |

The 20 are the subject of this issue. Each has a same-spelled member in
`@aws-sdk/client-s3`, is declared by the CFn registry schema under a top-level
property `S3BucketProvider` lists in `handledProperties`, and is never written —
so a template that sets it is a WRITE-SIDE SILENT DROP (the serializer discards
nothing; the provider simply never builds the field).

## The paths

`applyLoggingConfiguration` — builds `BucketLoggingStatus.LoggingEnabled` with
`TargetBucket` / `TargetPrefix` only:

- `LoggingConfiguration.TargetObjectKeyFormat`
- `LoggingConfiguration.TargetObjectKeyFormat.SimplePrefix`
- `LoggingConfiguration.TargetObjectKeyFormat.PartitionedPrefix`
- `LoggingConfiguration.TargetObjectKeyFormat.PartitionedPrefix.PartitionDateSource`

SDK: `LoggingEnabled.TargetObjectKeyFormat: { SimplePrefix?, PartitionedPrefix? }`.
Effect: a bucket declaring the partitioned S3 server-access-log key format gets
the default flat format instead, silently.

`applyReplicationConfiguration` — the per-rule `Destination` literal writes
`Bucket` / `Account` / `StorageClass` only:

- `ReplicationConfiguration.Rules.Destination.AccessControlTranslation`
- `ReplicationConfiguration.Rules.Destination.AccessControlTranslation.Owner`
- `ReplicationConfiguration.Rules.Destination.EncryptionConfiguration`
- `ReplicationConfiguration.Rules.Destination.EncryptionConfiguration.ReplicaKmsKeyID`
- `ReplicationConfiguration.Rules.Destination.Metrics`
- `ReplicationConfiguration.Rules.Destination.Metrics.Status`  (see note)
- `ReplicationConfiguration.Rules.Destination.Metrics.EventThreshold`
- `ReplicationConfiguration.Rules.Destination.Metrics.EventThreshold.Minutes`
- `ReplicationConfiguration.Rules.Destination.ReplicationTime`
- `ReplicationConfiguration.Rules.Destination.ReplicationTime.Status` (see note)
- `ReplicationConfiguration.Rules.Destination.ReplicationTime.Time`
- `ReplicationConfiguration.Rules.Destination.ReplicationTime.Time.Minutes`
- `ReplicationConfiguration.Rules.SourceSelectionCriteria`
- `ReplicationConfiguration.Rules.SourceSelectionCriteria.ReplicaModifications`
- `ReplicationConfiguration.Rules.SourceSelectionCriteria.ReplicaModifications.Status` (see note)
- `ReplicationConfiguration.Rules.SourceSelectionCriteria.SseKmsEncryptedObjects`
- `ReplicationConfiguration.Rules.SourceSelectionCriteria.SseKmsEncryptedObjects.Status` (see note)

(The `.Status` leaves same-spell a `Status` the provider DOES write elsewhere,
so they sit in the 78 bucket rather than the 20; they are listed here because
they are dropped by the same missing parent block.)

Effect, in order of severity: cross-account replication with owner override
(`AccessControlTranslation`), SSE-KMS replica encryption
(`EncryptionConfiguration.ReplicaKmsKeyID`), S3 Replication Time Control
(`ReplicationTime` + `Metrics` — the RTC SLA), and replica-modification /
SSE-KMS-object selection (`SourceSelectionCriteria`) are all accepted from the
template and never sent. CDK's `aws-s3` L2 emits `SourceSelectionCriteria` +
`EncryptionConfiguration` for `replicationRules` with a KMS key.

Other:

- `LifecycleConfiguration.TransitionDefaultMinimumObjectSize` — SDK
  `PutBucketLifecycleConfigurationRequest.TransitionDefaultMinimumObjectSize`
  (`varies_by_storage_class` / `all_storage_classes_128K`); never written.
- `BucketEncryption.ServerSideEncryptionConfiguration.BlockedEncryptionTypes`
- `BucketEncryption.ServerSideEncryptionConfiguration.BlockedEncryptionTypes.EncryptionType`

## Why the critic does not block on this today

`AWS::S3::Bucket` is NOT opted into `freshObjectMapper`, precisely because the
other 78 paths are structural (segment renames) rather than drops, and opting in
before those are declared would drown the 20 in false positives. So these are
currently RECORDED, not fenced.

The count and the never-written path list are pinned by a unit test in
`tests/unit/scripts/gen-nested-key-coverage.test.ts`, so fixing any of them
fails that test and forces the script header's measured reason (C) to be
re-measured.

## Suggested shape of the fix

Two independent pieces, in either order:

1. Write the missing blocks in `s3-bucket-provider.ts` (and the matching reads
   in `readCurrentState`, so drift detection sees them).
2. Once the 20 are closed, declare the ~10 `segmentRenames` the other 78 need
   and opt `AWS::S3::Bucket` into `freshObjectMapper` with its floors, which
   makes the whole family non-regressing.

Real-AWS integ coverage is required for (1) — every one of these is a field
whose delivery can only be confirmed by reading it back off the bucket.

go-to-k:
## Summary

Issue #1495: twenty nested CFn keys `s3-bucket-provider.ts` **never wrote at all**. Each has a same-spelled member in `@aws-sdk/client-s3`, each is declared by the CFn registry schema under a top-level property the provider lists in `handledProperties`, and none was ever assembled. Nothing was dropped by the serializer — the provider simply never built the field, and the deploy reported success with the setting missing.

They were found by splitting `AWS::S3::Bucket`'s write-evidence residual (after #1474 cleared the builder-idiom false positives out of it) by "is this terminal written ANYWHERE in the file?".

## What was silently lost

- `LoggingConfiguration.TargetObjectKeyFormat` — a bucket asking for the partitioned server-access-log key format got the FLAT default.
- The four `ReplicationConfiguration.Rules.Destination` blocks, now built by a new `buildReplicationDestination` helper:
  - `AccessControlTranslation` — cross-account replication kept SOURCE ownership.
  - `EncryptionConfiguration.ReplicaKmsKeyID` — replicas were not encrypted with the declared key. CDK's `aws-s3` L2 emits this for a `replicationRules` entry with a KMS key.
  - `ReplicationTime` + `Metrics` — a template asking for S3 Replication Time Control (a billed SLA) got plain asynchronous replication.
- Rule-level `SourceSelectionCriteria` — replica-modification / SSE-KMS-object selection, so the rule replicated a different set of objects than declared.
- `LifecycleConfiguration.TransitionDefaultMinimumObjectSize` — it sits on the PutBucketLifecycleConfiguration **REQUEST**, not inside `LifecycleConfiguration`, which is exactly why the rules-only mapper never reached it.
- `BucketEncryption…BlockedEncryptionTypes` — a bucket kept accepting the SSE-C it declared it would refuse.

Every block is assembled member by member rather than forwarded, so a future CFn-only member cannot ride through unnoticed.

## Reads, not just writes

The matching `readCurrentState` reads land in the same change: a field cdkd SENDS but does not READ becomes permanent phantom drift that `cdkd drift --revert` then re-applies forever. Every member is emitted **only when AWS actually returned it** — `state.json`'s JSON round-trip drops undefined keys and the comparator compares arrays wholesale, so echoing an omitted member would turn the whole `Rules` array into permanent drift.

**Corrected in review:** the first cut filtered `TransitionDefaultMinimumObjectSize` down to "the NON-default value", with the polarity **inverted** — AWS defaults buckets created after September 2024 to `all_storage_classes_128K`, so the filter suppressed the one value a template meaningfully declares. It also made the live assertion vacuous (it asserted the value a new bucket returns anyway, so it passed with the write removed). The read is now unconditional and the fixture declares the NON-default value.

## Measured

S3's write-evidence residual **98 -> 81**, never-written count **20 -> 0**. The pinned test is INVERTED from "these 20 are missing" into a fence asserting the set is EMPTY.

**Measured limitation, found by review and recorded rather than papered over:** that empty-set assertion is NOT on its own a re-drop fence. `REVERSE_MAP_FUNCTION_PREFIXES` is prefix-only, so the CFn-spelled writes this PR adds inside `readEncryption` / `readLifecycle` / `readLogging` / `readReplication` land in the evidence set and vouch for the forward mapper. A variant with the whole write half reverted measures `no-write-evidence` 98 with the never-written set still EMPTY. What actually fails on that variant is the **count pin** (98 vs 81) — so that is the fence, it just fails as a number rather than by member name. Both tests now say so; the suffix-match mechanism fix belongs to #1520, where it is recorded with the measurement. The script header's reason (C) is re-measured accordingly.

Only 13 of the 20 cleared the bucket. The other 7 are written and still unresolvable at the audited CHAIN for the same structural reasons as the surviving 81 — a renamed CFn segment (`LoggingConfiguration` is the SDK's `BucketLoggingStatus.LoggingEnabled`), an SDK-only wrapper segment (`ServerSideEncryptionConfiguration` -> `...Rules`), and the request-hoisted member. That is `segmentRenames` work, split out as #1520; S3 stays out of the write pass for a purely structural reason now, with **no known silent drop behind it**.

## Test plan

- New `tests/unit/provisioning/s3-bucket-provider-nested-write-drops.test.ts` — 14 tests covering each block's write, the omit-when-undeclared case, the empty-`SimplePrefix`-must-survive case (its PRESENCE is the signal), the send-only-what-was-declared case (RTC without Metrics), and each read-back.
- Full suite: 532 files / 9583 tests pass.
- **Real AWS, `s3-replication-and-filter` integ (2026-08-11)**, extended for this PR: five of the twenty members are read back off the live bucket — `TargetObjectKeyFormat.PartitionedPrefix`, `Destination.ReplicationTime` + `Metrics` (RTC really on), `SourceSelectionCriteria.ReplicaModifications`, `LifecycleConfiguration.TransitionDefaultMinimumObjectSize`. All passed. Destroy: 6 deleted, 0 errors; all three buckets and the state file gone; no orphans.

## Won't-do here (recorded, not deferred silently)

Two members stay unit-only, with the reason in the fixture header: `Destination.EncryptionConfiguration` / `SourceSelectionCriteria.SseKmsEncryptedObjects` need a customer-managed KMS key whose 7-day minimum deletion window would leave a pending-deletion orphan behind **every** run, and `AccessControlTranslation` is only meaningful across two accounts. Options are written up in #1523.

Closes #1495.

## Review fix-back (6 reviewers, 3 axes x 2 PRs)

- **Blocker:** `TransitionDefaultMinimumObjectSize` polarity inverted (above), which also made its live assertion vacuous. Fixed; integ re-run confirms `varies_by_storage_class` really comes back off the bucket.
- Read-side members are now emitted only when AWS returned them.
- The four `Destination` blocks gained empty-block guards, so a template's `AccessControlTranslation: {}` no longer ships `{Owner: undefined}` (AWS rejects that; the pre-fix code sent nothing).
- Every new gate is `!= null` rather than truthiness, per the post-#1493 rule.
- The empty-block guards are INLINE, not routed through a shared `attach()` helper: a helper taking a string-literal key hides the write from the evidence pass — measured as the residual going 93 instead of 81.


go-to-k:
## pr-review marker set manually, with maintainer go-ahead

Recording the reason the `pr-review` gate is being satisfied by hand rather than by another `/review-pr` cycle.

**This PR was reviewed at full 3-axis tier**, plus a verification pass:

- spec / code / test reviewers ran against `93fbf15e`; the code reviewer found a BLOCKER (the `TransitionDefaultMinimumObjectSize` polarity was inverted, which also made the live assertion vacuous) and the test reviewer proved the inverted pin was over-crediting by reverting the write half and measuring.
- a re-review ran against `dfe69df6` and confirmed the blocker CLOSED, leaving 2 minors + 3 nits — all of which are fixed in `62b2ad2a` except the two recorded as follow-ups (#1523, #1530).

**Why the sha binding no longer matches:** `main` moved twice mid-session (four PRs merged, including #1518 which restructured the very coverage table this PR pins). The rebase onto `72fea4ec` resolved conflicts in three docs and re-applied the S3 row of the opt-in table (`98 -> 81`, total `260 -> 81`) on top of #1475's new `+spread` column. **No provider logic changed in the rebase** — it was conflict resolution in `docs/changelog-cdkd.md`, `.claude/rules/code-layout.md` and the `gen-nested-key-coverage.ts` header table.

**Re-verified after the rebase, not assumed:**

- full suite green (533 files / 9657 tests), including the opt-in-table pin, which is what proves the table merge was resolved correctly rather than merely plausibly;
- the real-AWS integ re-run clean — all five #1495 read-backs pass, destroy 6 deleted / 0 errors, all three buckets and the state file gone, no orphans;
- `integ-destroy` re-set off that run (it had legitimately staled, because the rebase pulled main's provider changes into the gate's scope).

Re-running six reviewer agents over a doc-conflict rebase was judged poor value against that evidence. Filed under the gate's documented escape hatch: a comment explaining why, plus the maintainer's explicit go-ahead.

Issue #1543, opened 2026-08-11
integ lint: flag a fixture resource gated on a CDKD_TEST_UPDATE token that a LATER deploy's mode list drops


go-to-k:
## Summary

Add a `check-integ-*` lint for the mode-gating trap documented in `.claude/rules/testing.md` ("A mode-gated fixture resource DISAPPEARS in every later step that omits the token").

Multi-step fixtures drive phases with `CDKD_TEST_UPDATE=<comma,separated,modes>` and the stack reads `updateMode.includes('x')`. Gating a NEW resource on a NEW token is the natural spelling and is wrong whenever a later deploy in the same `verify.sh` uses a mode list without that token: the resource leaves the template and cdkd correctly DELETEs it.

Found during #1512 by hand, before the run reached it. `OnDemandReplicaTable` gated its `eu-west-1` replica on `cross-region-ondemand*` (steps 12g/h/i); the fixture then deploys six more times with mode lists that omit the token, so step 12f would have removed the replica **from a still-live table** — the operation that arms DynamoDB's 24h source-region delete lock (#1442 wedged a table in `UPDATING` for 90+ minutes that way). The fixture comments, the commit message and the issue comment had all already claimed the design "only ever ADDs". Nothing would have caught the contradiction.

## Why this is a checker, not a judgment call

Unlike the order-insensitivity rule (which needs a human to decide whether a list is order-significant), both halves here are statically extractable:

- the ORDERED list of `CDKD_TEST_UPDATE=` mode lists per deploy invocation in `verify.sh` (the `integ-cli-flags` classifier already parses these invocation shapes, including the inline-env-prefix form);
- the set of tokens each fixture stack tests via `updateMode.includes('<literal>')`, and which construct/property declaration each one gates (TS Compiler API, as the other classifiers do).

The verdict is then: for a token `T` gating a declaration, is there a deploy AFTER the last deploy carrying `T` whose mode list omits `T`? If yes, that declaration is dropped mid-run — flag it.

## Suggested shape

- `scripts/check-integ-mode-gated-resources.ts` + `tests/unit/scripts/integ-mode-gated-resources.test.ts`, matching the seven existing `check-integ-*` siblings.
- Escape hatch comment for the deliberate case (a fixture that WANTS to exercise the removal — e.g. the pre-existing step 12d replica removal on the main table, and every `CDKD_TEST_REMOVAL` fixture, which are removal tests by design). Something like `// allow-mode-gated-drop: <reason>`.
[...]

Issue #1442, opened 2026-08-09
cleanup: DynamoDB probe table cdkd-1434-probe is stuck UPDATING and cannot be deleted


go-to-k:
Residual from the #1434 live-probe session (2026-08-10). The probe global table
`cdkd-1434-probe` (us-east-1 source + us-west-2 replica, PAY_PER_REQUEST, empty)
could not be torn down and is still present.

## Why it did not delete

Two AWS-side constraints stacked:

1. `DeleteTable` on the source region is refused for 24h after it has sourced a
   replica:

   ```
   ValidationException: Replica cannot be deleted because it has acted as a
   source region for new replica(s) being added to the table in the last 24 hours.
   ```

2. The documented workaround (remove the replica first, then delete the table) is
   itself blocked, because the table has been stuck `UPDATING` for ~1h with every
   sub-resource reporting ACTIVE (GSI `gsi1` ACTIVE, replica `us-west-2` ACTIVE):

   ```
   ResourceInUseException: The resource which you are attempting to change is in use.
   ```

The likely cause is an in-flight `update-table` from the probe run that was
interrupted mid-sequence (the replica-GSI `OnDemandThroughputOverride` probes).

## Cost

Negligible but non-zero: PAY_PER_REQUEST with no items and no traffic, so storage
and request charges are ~0. It is an orphan on hygiene grounds, not a billing risk.

## Teardown steps

```bash
T=cdkd-1434-probe
# 1. wait for the table to leave UPDATING
aws dynamodb describe-table --region us-east-1 --table-name $T --query Table.TableStatus --output text
# 2. drop the replica (needed: the 24h rule blocks deleting the source directly)
aws dynamodb update-table --region us-east-1 --table-name $T \\
  --replica-updates '[{"Delete":{"RegionName":"us-west-2"}}]'
# 3. delete the table
aws dynamodb delete-table --region us-east-1 --table-name $T
# 4. verify BOTH regions report not-found
for r in us-east-1 us-west-2; do aws dynamodb describe-table --region $r --table-name $T; done
```

If step 2 still returns `ResourceInUseException`, the in-flight op has not
settled; if step 3 still returns the 24h `ValidationException`, wait out the
window (it expires ~2026-08-10T15:00Z) and retry.

## Lesson worth keeping

A live probe that mutates a GLOBAL table is not cheaply reversible: adding a
replica arms a 24h deletion lock on the source region. Future probes of
replica-level behavior should either use a throwaway table per probe step, or
[...]

Issue #1436, opened 2026-08-09
AWS::DynamoDB::GlobalTable: replica-level ReadOnDemandThroughputSettings is never wired (Billing.onDemand maxReadRequestUnits silently dropped)


go-to-k:
Found while implementing the table-level half of #1434.

`AWS::DynamoDB::GlobalTable` carries the table's on-demand READ ceiling on the
LOCAL REPLICA (`Replicas[?Region==<deploy region>].ReadOnDemandThroughputSettings.MaxReadRequestUnits`),
the mirror of how per-GSI read capacity lives on the replica's index entry
(the shape #1387 established). `DynamoDBGlobalTableProvider` never reads it.

This is NOT the absent-field-reset class of #1423 / #1434 — the value is
dropped on the way IN, so it never reaches AWS at all.

## Evidence

`create()` wires only the write half:

```ts
// OnDemand throughput (GlobalTable: WriteOnDemandThroughputSettings).
const wodts = properties['WriteOnDemandThroughputSettings'] as ...;
if (wodts?.['MaxWriteRequestUnits'] !== undefined) {
  createParams.OnDemandThroughput = { MaxWriteRequestUnits: Number(...) };
}
```

`update()`'s flat step-3 block does the same. Every `MaxReadRequestUnits` write
in the file is at GSI level (`toSdkGlobalSecondaryIndexes`, and the #1423 reset
in the per-GSI Update action); nothing ever populates the TABLE-level
`OnDemandThroughput.MaxReadRequestUnits`.

`readCurrentState` matches the same gap — it reverse-maps
`WriteOnDemandThroughputSettings` from `Table.OnDemandThroughput.MaxWriteRequestUnits`
and has no read counterpart, so drift cannot see it either.

## Reachability: canonical CDK, not a hand-authored edge case

This is the plain `TableV2` spelling:

```ts
new ddb.TableV2(stack, 'OnDemand', {
  partitionKey: { name: 'pk', type: STRING },
  billing: Billing.onDemand({ maxReadRequestUnits: 100, maxWriteRequestUnits: 200 }),
});
```

The synthesized template puts `maxWriteRequestUnits` at
`WriteOnDemandThroughputSettings` (handled) and `maxReadRequestUnits` at
`Replicas[local].ReadOnDemandThroughputSettings` (dropped).

The repo's own test fixture already contains it and nobody noticed: the
verbatim-`cdk synth` bag `ON_DEMAND_TABLE_PROPS` in
[...]

Issue #1435, opened 2026-08-09
AWS::DynamoDB::GlobalTable: initial capacity derives from SeedCapacity where CloudFormation uses MinCapacity (silent over-provision on create)


go-to-k:
Found by the live CloudFormation probe run for #1427 (see [that comment](https://github.com/go-to-k/cdkd/issues/1427#issuecomment-5232192847) for the full probe record).

## Summary

`deriveReadCapacityUnits` / `deriveWriteCapacityUnits` in `src/provisioning/providers/dynamodb-globaltable-provider.ts` both end with:

```ts
return toFiniteNumber(autoScaling['SeedCapacity']) ?? toFiniteNumber(autoScaling['MinCapacity']);
```

`SeedCapacity` first. CloudFormation uses `MinCapacity` on a fresh `CREATE`, so cdkd over-provisions any autoscaled PROVISIONED GlobalTable by the seed-to-min ratio. It is a billing symptom, not an error: the deploy succeeds and the table is simply more expensive than the same template under CloudFormation.

## Evidence (real AWS, us-east-1)

Stack `CdkdIssue1427Control`, a `AWS::DynamoDB::GlobalTable` with `BillingMode: PROVISIONED` and, at table level and on the GSI:

```yaml
WriteProvisionedThroughputSettings:
  WriteCapacityAutoScalingSettings:
    MinCapacity: 1
    MaxCapacity: 30
    SeedCapacity: 20
    TargetTrackingScalingPolicyConfiguration:
      TargetValue: 70
```

`CREATE_COMPLETE`, then `dynamodb describe-table`:

```json
{"Table": {"ReadCapacityUnits": 5, "WriteCapacityUnits": 1, "NumberOfDecreasesToday": 0},
 "GSI": [{"Name": "gsi1", "PT": {"ReadCapacityUnits": 5, "WriteCapacityUnits": 1, "NumberOfDecreasesToday": 0}}]}
```

CloudFormation created at **1** (`MinCapacity`), not 20 (`SeedCapacity`). cdkd's helpers return 20 for the same input.

`NumberOfDecreasesToday: 0` rules out an auto-scaling scale-down between create and read-back, and `application-autoscaling describe-scalable-targets` confirms `Min: 1 / Max: 30` registered on both `dynamodb:table:WriteCapacityUnits` and `dynamodb:index:WriteCapacityUnits`.

## Why CloudFormation is right here

[...]

PR #1550, opened 2026-08-11
test(integ): cover both replica throughput-override shapes in the multi-region GlobalTable block

### LINKED ISSUE #1512: test(integ): assert the cross-region replica table-level throughput overrides in the multi-region gated block
Follow-up from PR (#1503) (the throughput cluster), recorded there as a won't-do; re-classified as deferred test coverage on session review — "it lives behind the multi-region gate" explains why it was not in the default flow, not why the gated block should never assert it.

## The gap

PR (#1503) wired NEW cross-region replica behavior that no real-AWS run has ever asserted:

- `addReplica` / the replica-modify action now send the replica's TABLE-level `{Provisioned,OnDemand}ThroughputOverride` (from `Replicas[].Read{Provisioned,OnDemand}ThroughputSettings`) — previously dropped entirely.
- A DROPPED non-local replica override now WARNS that the old value is still in effect (no reset payload exists — the (#1436) probes).

The `dynamodb-globaltable` fixture's `CDKD_INTEG_MULTI_REGION=1` block (steps 12b-e) predates this and asserts none of it. Unit coverage exists (mocked SDK), but the mocked-SDK-agrees-with-wrong-wire-assumptions trap is exactly what this repo's rules budget real-AWS runs for on new call shapes — `OnDemandThroughputOverride` on `CreateReplicationGroupMemberAction` / `UpdateReplicationGroupMemberAction` is a new call shape.

## Scope

Extend the `CDKD_INTEG_MULTI_REGION=1` block of `tests/integration/dynamodb-globaltable/verify.sh`:

- The added `eu-west-1` replica declares `ReadOnDemandThroughputSettings`; after it flips ACTIVE, assert the replica-region `DescribeTable` reads the override back.
- A follow-up deploy changes the override value; assert the `Update` action applied it.
- A deploy that DROPS the override asserts the live value is UNCHANGED and the "STILL IN EFFECT" warning fired (grep the deploy log).

Wall-clock is the known 15-25 min per round-trip, which is why this stays behind the gate rather than in the default flow.

go-to-k:
## Summary

PR (#1503) wired new cross-region replica behavior that no real-AWS run had ever asserted: `addReplica` and the replica-modify action now send the replica's TABLE-level `{Provisioned,OnDemand}ThroughputOverride`, and a DROPPED override warns that the old value is still in effect. Unit coverage exists but is mocked-SDK only — and a mocked SDK agrees with whatever wire assumption the code makes, which is exactly the trap this repo budgets real-AWS runs for on a new call shape.

## One scoping correction to the issue

Issue (#1512) asks the added `eu-west-1` replica to declare `ReadOnDemandThroughputSettings`. But the existing `CDKD_INTEG_MULTI_REGION=1` cross-region step deploys under `CDKD_TEST_UPDATE=deletion-protection,autoscaling,cross-region`, i.e. **PROVISIONED** billing, and `toSdkReplicaThroughputOverrides` branches on the table's billing mode — so that step structurally cannot reach the `OnDemandThroughputOverride` arm. Covering the shape the issue names requires a PAY_PER_REQUEST table.

This PR therefore covers **both** override shapes rather than only the one that fits the existing step.

## Changes

**PROVISIONED half** (existing step, no new round-trip): the `eu-west-1` replica now declares its own autoscaled `readCapacity` (min 7 — deliberately different from the source table's 5, so a replica that merely *inherited* the source capacity cannot pass). New step 12c-2 asserts `Replicas[eu-west-1].ProvisionedThroughputOverride.ReadCapacityUnits == 7`; cdkd derives that single number from `MinCapacity` per the (#1435) `CapacitySource` rule.

**ON-DEMAND half** (new `OnDemandReplicaTable`, PAY_PER_REQUEST): the table exists from the first deploy and gains its replica only in the multi-region block. That ordering is load-bearing — declaring the replica at CREATE time would exercise `create()`, whereas the code under test is `addReplica`, the UpdateTable `ReplicaUpdates[].Create` path taken when a new region appears on an EXISTING table. Presence is carried forward monotonically once added (see defect 1 below). Three rounds, exactly as the issue asks:

- **12e1** add the replica with `maxReadRequestUnits: 20`; assert the override reads back as 20 from BOTH the source table's `Replicas[]` entry and the replica region's own `DescribeTable`.
- **12e2** CHANGE it to 40; assert the Update action applied it.
- **12e3** DROP it; assert the live value is **unchanged at 40** and that cdkd emitted the `STILL IN EFFECT` warning. AWS offers no way to clear a replica-level override — a `-1` sentinel stores literally and an empty block wedges the table in UPDATING, both live-probed on (#1436) — so leaving the value and saying so is the correct behavior, and this pins it.

## Safety of the teardown shape

The new rounds only ever **ADD** a replica, and step 12e4 waits for the table and its replicas to reach ACTIVE before the run moves on. Removing one from a still-live table is what arms DynamoDB's 24h source-region delete lock — a probe wedged a table in UPDATING for 90+ minutes that way (#1442). Teardown is left to `cdkd destroy`, which drops the GlobalTable as ONE resource with all replicas together and never issues a standalone replica-delete. New step 16a3 asserts the `eu-west-1` copy is actually gone afterwards, so a cross-region orphan cannot hide from a same-region-only sweep.

(Step 12d's replica removal on the main table is pre-existing and unchanged.)

## Probe hygiene

The probe helpers are tail-less captures that branch on **exit status**, not on a sentinel string. A first draft used `2>/dev/null || echo MISSING`, which the (#1120) gone-probe lint correctly rejected: a throttle would have been laundered into a plain assertion failure blaming the fix — the false-negative shape `.claude/rules/testing.md` calls out. The timeout path reports the last-seen values rather than re-probing, since a re-probe can itself fail and mask what was actually observed.

## Test plan

- Synth-validated across all four modes: the expected `ReadProvisionedThroughputSettings` (min 7) and `ReadOnDemandThroughputSettings` (20 / 40) blocks appear exactly where intended, and the drop round emits neither.
- Full unit suite green, including the (#1120) gone-probe lint and the (#1097) CLI-flag / signal-trap lints.
- Real-AWS `CDKD_INTEG_MULTI_REGION=1` run: result below.

## What the runs actually found

Three defects in this fixture, none visible to unit tests or `cdk synth` — each surfaced only by running against real AWS:

1. **The replica was deleted mid-run.** Presence was keyed on the `CDKD_TEST_UPDATE` token, but `verify.sh` keeps deploying afterwards and none of those mode lists carry it — so step 12f would have removed the replica from a still-live table, the operation that arms the 24h source-region delete lock (#1442). Fixed with `OD_MODE_SUFFIX`, which carries the token forward once the replica exists, making presence monotonic. The suffix stays empty when `CDKD_INTEG_MULTI_REGION` is unset, so the default flow is byte-for-byte unchanged.
2. **The first fix for (1) silently changed what was tested.** Keying presence on the env var meant the replica was declared from step 1, so it was created *with* the table and the round became a replica-**modify**, not `addReplica`. The assertion passed anyway — it only checks the resulting value. Caught by reading the log (`Creating DynamoDB GlobalTable` at step 1).
3. **Destroy failed with `ResourceInUseException`**, leaving the table and its `eu-west-1` replica as orphans (cleaned up manually). The drop round was still settling. Added step 12e4, which polls until the table **and** its replicas are ACTIVE — a global table reports ACTIVE at the table level while a replica is still UPDATING.

Also de-raced the **pre-existing** `assert_scalable_target`, unrelated to this issue: it read Application Auto Scaling once with no retry and failed with `Min=None` while cdkd's own log showed the policy upserted a second earlier. The retry cannot hide a real non-registration — every poll would read `None` and still fail. `assert_target_gone` (the absence side) is untouched.

Steps renumbered `12e1`-`12e4`; they had collided with the pre-existing `12f`/`12g`.

## Real-AWS result

`CDKD_INTEG_MULTI_REGION=1`, **1019s, RC=0, 6 resources deleted, 0 errors, 0 orphans** in either region.

```
step 12c-2 ok: eu-west-1 ProvisionedThroughputOverride = 7
step 12e1  ok: addReplica sent OnDemandThroughputOverride = 20
step 12e2  ok: the Update action applied OnDemandThroughputOverride = 40
step 12e3  ok: dropped override left at 40 and cdkd warned STILL IN EFFECT
step 12e4  ok: on-demand table + replica ACTIVE
step 16a3  ok: no cross-region orphan left behind
```

The `addReplica` claim is verified from the LOG rather than from the assertion, because defect (2) is exactly the case where the assertion passes while the wrong path runs: the step-1 CREATE region carries **zero** `eu-west-1` mentions for `OnDemandReplicaTable` in this run, against **26** in the broken one.

## Convention added

The mode-gating trap behind (1) and (2) is documented in `.claude/rules/testing.md` and `docs/testing.md` so it survives a machine change, with the mechanical lint filed as (#1543) — both halves it needs (the ordered mode lists and the token-gated declarations) are statically extractable, so it is a checker rather than a judgment call.

Closes #1512

### DIFF
--- a/.claude/rules/testing.md
+++ b/.claude/rules/testing.md
@@ -330,31 +330,49 @@ delete lock (a probe wedged a table in `UPDATING` for 90+ minutes that way,
 #1442), and the fixture's own comments, the commit message and the issue comment
 had all already claimed the design "only ever ADDs".
 
-**Separate PRESENCE from CONFIGURATION.** Presence goes on something constant
-for the whole run; only the configuration varies per step:
+**Make the token MONOTONIC — carry it forward in a shell suffix.** Set a
+variable once the resource exists and append it to the mode list of every later
+deploy:
 
-```ts
-// PRESENCE: constant for the whole verify.sh run.
-const wantsReplica = process.env.CDKD_INTEG_MULTI_REGION === '1';
-// CONFIGURATION: per-step.
-const declaresCeiling = mode === 'initial' || mode === 'changed';
+```bash
+OD_MODE_SUFFIX=""                              # seeded with the other run vars
+...
+OD_MODE_SUFFIX=",cross-region-ondemand-dropped" # set right after the rounds
+...
+CDKD_TEST_UPDATE=ttl,tags${OD_MODE_SUFFIX} ${CLI} deploy ...
 ```
 
-Then no step can drop the resource and only `cdkd destroy` removes it — which
+The suffix stays empty when the scenario is not gated on, so the default flow is
+byte-for-byte unchanged; only `cdkd destroy` then removes the resource — which
 for a GlobalTable deletes every replica as ONE resource and never issues a
 standalone replica-delete.
 
+**Do NOT instead key presence on a run-scoped env var**, which is the tempting
+one-line fix. It stops the deletion but silently changes what the test tests:
+the resource is then declared from step 1, so it is created WITH its parent and
+the step you meant to exercise becomes an UPDATE. That is how the #1512 fixture
+briefly stopped covering `addReplica` while its assertion still passed — the
+assertion only checked the resulting value. Use the env var only for presence
+that no step needs to transition.
+
 **Before adding a mode-gated resource to a multi-step fixture:**
 
-- Enumerate every deploy (`grep -n 'CLI} deploy' verify.sh`) and read the mode
-  list of each one that runs AFTER your steps. Any that omits your token deletes
-  your resource there.
+- Enumerate every deploy and read the mode list of each one that runs AFTER your
+  steps. Any that omits your token deletes your resource there. Grep for the
+  invocation, **not** for a single-line pattern: fixtures wrap long mode lists
+  with a trailing `\` so the env prefix and `${CLI} deploy` sit on DIFFERENT
+  lines, and a one-line `grep 'CDKD_TEST_UPDATE=.*deploy'` silently misses those.
+  That is exactly how #1512 shipped a run that still performed the live
+  replica-delete while reporting PASS.
 - Ask what that deletion COSTS. For a plain queue, nothing. For anything whose
   removal is slow, locked, or destructive (DynamoDB replicas, RDS, stateful
   storage) it is the whole point.
 - Verify by synthesizing the LATER modes, not your own: the check that catches
   this is `cdk synth` under `ttl,tags` showing the resource still intact — not
   `cdk synth` under your own mode showing it created.
+- Make sure some assertion would NOTICE the drop. A post-destroy "it is gone"
+  check passes vacuously when the resource was deleted mid-run, so it is not a
+  guard; assert presence at the point the resource must still exist.
 
 NOT yet mechanically enforced — issue
 [#1543](https://github.com/go-to-k/cdkd/issues/1543) tracks the lint (the
--- a/.claude/rules/testing.md
+++ b/.claude/rules/testing.md
@@ -309,6 +309,59 @@ judgment call a lint cannot make, so this one stays a read-it-and-follow-it
 rule by design. User-facing writeup in
 [docs/testing.md](../../docs/testing.md).
 
+### A mode-gated fixture resource DISAPPEARS in every later step that omits the token (mandatory)
+
+Multi-step fixtures drive each phase with
+`CDKD_TEST_UPDATE=<comma,separated,modes>` and the stack reads
+`updateMode.includes('x')`. The natural way to add a scenario is to gate the new
+resource on a new token — and that is **wrong whenever a later deploy in the
+same `verify.sh` uses a mode list without it**. The resource vanishes from the
+template and cdkd correctly issues a DELETE for it. A per-step conditional in a
+long fixture is really a *step function over the whole run*, not a flag for your
+step.
+
+Caught during issue #1512, before the run reached it. The new
+`OnDemandReplicaTable` gated its `eu-west-1` replica on `cross-region-ondemand*`
+(steps 12g/h/i). The fixture then keeps deploying — `deletion-protection,
+autoscaling,ttl,tags` and five more `ttl,tags,...` rounds — none of which carry
+the token, so step 12f would have removed the replica **from a still-live
+table**. That is exactly the operation that arms DynamoDB's 24h source-region
+delete lock (a probe wedged a table in `UPDATING` for 90+ minutes that way,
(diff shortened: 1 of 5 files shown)


go-to-k:
## Review round 1 — blocker found, and it had already fired

`/review-pr` (1-reviewer tier) found a **blocker that invalidated the run I had reported as passing**, so recording it here rather than quietly fixing it.

**Step 13f (`unresolvable-capacity`) is a MULTI-LINE invocation** — the `CDKD_TEST_UPDATE=` prefix and `${CLI} deploy` sit on different lines — so the one-line grep I used to place `OD_MODE_SUFFIX` never saw it. That deploy therefore dropped the token, the replica left the template, and cdkd issued the live replica-delete this whole design exists to prevent. The prior run's log shows it plainly: 13f raised a 2-property UPDATE on `OnDemandReplicaTable` that took ~3 minutes (02:41:57 -> 02:44:59).

Worse, **step 16a3 passed vacuously**: it asserts the `eu-west-1` copy is gone *after destroy*, and it was already gone mid-run. A green multi-region run was not evidence the gap was absent — which is exactly what the reviewer flagged.

**Fixed and re-verified.** 13f now carries the suffix. On the re-run, 13f reports `NO_CHANGE` for that table (previously UPDATE/3min), so the replica survives to destroy and 16a3 is meaningful. `addReplica` remains genuine — zero `eu-west-1` mentions in the step-1 CREATE region.

**Minors, all applied:**

- The settle-wait's `ACTIVE:` arm accepted an EMPTY replica list as "settled" — the one shape that would have masked this deletion. Now `ACTIVE:ACTIVE` only.
- Step 12c-2 single-shot probed the same eventually-consistent value its on-demand twin polls for; it would have flaked with a FAIL blaming (#1503). Now polled.
- Drop round checks both source and replica sides; `OD_DROP_LOG` no longer leaks on failure paths; `OD_LAST_SRC`/`OD_LAST_REP` refresh independently; stale 12g/12h/12i references corrected.
- **Rule correction**: the convention I added had prescribed keying presence on the env var — the remedy this implementation deliberately rejects, since it makes the resource be created with its parent and turns the step under test into an UPDATE. Both copies now teach the suffix pattern, explain why the env-var form is wrong here, warn that a one-line grep misses wrapped invocations, and note that a post-destroy "it is gone" check is vacuous as a guard.

**Re-run on post-(#1542) main:** PASS, 928s, 6 resources deleted, 0 errors, 0 orphans in either region. Unit suite 542 files / 9807 tests green.

Issue #1609, opened 2026-08-11
elbv2: follow-ups from the #609 LB+TargetGroup batch — attribute-reset A/B for LB/Listener, flag-omission A/B, test polish, Targets drift readback


go-to-k:
Follow-ups from the #609 ELBv2 LoadBalancer + TargetGroup batch (PR: fix/609-elbv2-lb-targetgroup-props). Bundled per the residual-review-nit sweep.

## Live-verification gaps (the important ones)

1. **A/B the `Value: ''` removal arm of ModifyLoadBalancerAttributes / ModifyListenerAttributes against real AWS.** The batch's live run proved ModifyTargetGroupAttributes REJECTS the empty-string reset ("A target group attribute value must be specified") that the LB / Listener attribute diff arms have shipped since their introduction — and those two arms have never been exercised by an integ removal phase. If either API also rejects `''`, the same defaults-table treatment (`TARGET_GROUP_ATTRIBUTE_DEFAULTS` pattern in `elbv2-provider.ts`) is needed there. Cheapest probe: extend the `alb` fixture's removal phase to also drop the Listener's `ListenerAttributes` (a live listener is already in the fixture); the LB attributes arm needs a fixture that templates `LoadBalancerAttributes` and then drops one.
2. **A/B the omission semantics of `EnablePrefixForIpv6SourceNat` / `EnforceSecurityGroupInboundRulesOnPrivateLinkTraffic` for the combined removal-plus-change deploy.** When only the flag is removed, cdkd retains (no call — CFn-retains precedent). But when the flag is removed in the SAME deploy that changes Subnets / SecurityGroups, the Set* call is issued WITHOUT the flag member, and AWS's retain-vs-default behavior on omission is unverified (recorded in-code in `updateLoadBalancer`). Needs an NLB fixture.

## Test polish (reviewer nits, deliberately deferred)

3. Pin the register-before-deregister CALL ORDER in the TG targets update test (`elbv2-lb-targetgroup-props.test.ts` — payloads are asserted, order is only a comment).
4. Cover `readCurrentState`'s best-effort catch arms (NotFound -> undefined vs permission-error -> key absent) for the new DescribeCapacityReservation / DescribeTargetGroupAttributes reads.
5. Comment in `tests/integration/alb/lib/alb-stack.ts` that the removal-phase deploy intentionally flips the registered target back to 10.0.0.100 (scalar churn a future reader may otherwise chase).

## Feature gap

[...]

Issue #1602, opened 2026-08-11
ApiGatewayV2::Integration: TlsConfig on a public integration is permanent drift, and ResponseParameters does not round-trip for the flat SDK spelling


go-to-k:
Split out of the `AWS::ApiGatewayV2::Integration` silent-drop backfill (issue
#609). Both items are pre-existing shapes that the backfill makes REACHABLE for
the first time — neither is a regression, and neither was in that PR's scope.

**1. `TlsConfig` on a public integration is permanent drift.**

Direct A/B against the API (2026-08-11, us-east-1): AWS SILENTLY IGNORES
`TlsConfig` on an integration whose `ConnectionType` is `INTERNET` — it is
absent from both the `CreateIntegration` echo and the `GetIntegration`
read-back, with or without an explicit `ConnectionType`. The field is only
meaningful for a PRIVATE (VPC-Link) integration.

So for a template that declares `TlsConfig` on a public integration:

- cdkd records `TlsConfig` in state (it is in the desired bag),
- `readCurrentState` cannot return it (AWS has nothing),
- and `drift-calculator` descends into every key present in state,

which makes `cdkd drift` report the key forever, and `drift --revert` re-send a
value AWS discards on every run. This is the `#1552` warn-and-continue junk-state
class one property over, except the discarding happens AWS-side rather than in
cdkd.

The template is arguably wrong to begin with (AWS accepts and ignores it rather
than rejecting), so the options are the usual three: declare the path
drift-unknown via `getDriftUnknownPaths` when the integration is not private,
refuse the combination at pre-flight, or leave it and document. Worth measuring
which before picking — the fixture deliberately does NOT set the field, so this
is not currently exercised.

**2. `ResponseParameters` does not round-trip when a template uses the flat
SDK spelling.**

`toSdkResponseParameters` accepts an ALREADY-FLAT block
(`{"<status>": {"<Destination>": "<Source>"}}`) and passes it through, which is
what a hand-written template using the SDK spelling produces. That branch makes
the mapping non-injective: `toCfnResponseParameters` always rebuilds the CFn
list-of-pairs shape, so the AWS-current side comes back in the CFn shape while
the state baseline holds the flat one, and the two never compare equal.

Templates written from the CloudFormation schema (every CDK-synthesized one,
and the integ fixture) are unaffected — they carry the list shape on both sides
and round-trip cleanly. The exposure is limited to a hand-written L1 that
borrowed the SDK spelling.

[...]

Issue #1498, opened 2026-08-10
`cdkd drift` phantom-detects sibling-resource-populated keys in the observed baseline; `--revert` then detaches ECS capacity providers and deletes lifecycle hooks


go-to-k:
## Summary

`cdkd drift` uses `observedProperties` (the deploy-time AWS snapshot, captured per-resource right after that resource's create/update settles) as the drift baseline. That snapshot is taken BEFORE sibling resources in the same stack run, so any parent-resource key that a **sibling cdkd-managed resource** (or an AWS service itself) populates later is permanently reported as drift on a freshly deployed, untouched stack. `cdkd drift --revert` then "fixes" the phantom by **stripping the sibling-managed configuration from AWS** — while reporting `✓ reverted`.

CloudFormation drift detection only compares property values **explicitly set in the stack template** (documented behavior), so none of this fires there. This is a cdkd-specific false positive with a destructive revert path.

## Live repro (2026-08-10, us-east-1, cdkd @ main 944d4ce3)

Standard CDK EC2-backed ECS capacity provider pattern (`cluster.addAsgCapacityProvider(...)` — the canonical L2 shape):

```ts
const cluster = new ecs.Cluster(this, 'Cluster', { vpc, clusterName: 'CdkdBughuntCluster' });
const asg = new autoscaling.AutoScalingGroup(this, 'Asg', { vpc, instanceType: t3.micro,
  machineImage: ecs.EcsOptimizedImage.amazonLinux2023(), minCapacity: 0, maxCapacity: 2, desiredCapacity: 0 });
const cp = new ecs.AsgCapacityProvider(this, 'Cp', { autoScalingGroup: asg,
  capacityProviderName: 'CdkdBughuntCp', enableManagedTerminationProtection: false });
cluster.addAsgCapacityProvider(cp);
```

1. `cdkd deploy` — clean, all 29 resources created, post-deploy no-op `cdkd diff --fail` exits 0.
2. `cdkd drift` — **exit 1, drift on 2 resources** with NO out-of-band change ever made:

```
~ AsgASGD1D7B4E2 (AWS::AutoScaling::AutoScalingGroup)
  - Tags: [{"Key":"Name",...}]
  + Tags: [{"Key":"AmazonECSManaged","Value":""},{"Key":"Name",...}]
  - LifecycleHookSpecificationList: []
  + LifecycleHookSpecificationList: [ecs-managed-draining-termination-hook,
[...]

PR #1635, opened 2026-08-11
fix(drift): unordered comparison for object arrays, and make TargetGroup.Targets visible to cdkd drift (#1609 item 6)

### LINKED ISSUE #1620: drift: unordered comparison for object arrays, so TargetGroup.Targets can leave getDriftUnknownPaths (#1609 item 6)
Split out of issue 1609 item 6 (#1609), which the 1609 follow-up lane closed the other
items of. This one is a drift-layer FEATURE with a blast radius well past
ELBv2, so it does not belong in an ELBv2 follow-up PR.

## The gap

`AWS::ElasticLoadBalancingV2::TargetGroup.Targets` is declared in
`ELBv2Provider.getDriftUnknownPaths`, so `cdkd drift` cannot see a
console-side target registration change at all. That declaration is a
workaround, not a decision: the provider CAN read the targets back
(`DescribeTargetHealth`), but the drift comparator compares arrays
POSITIONALLY (`deepEqual` is `JSON.stringify`), and AWS does not guarantee
readback order for a target list. Reading them back without an unordered
compare would trade a blind spot for permanent phantom drift — the 1096 /
1498 class (see https://github.com/go-to-k/cdkd/issues/1498).

## What exists today

`src/analyzer/drift-normalize.ts` canonicalizes two shapes on BOTH comparison
sides:

- `canonicalizeTagListsDeep` — any `{Key,...}[]` tag list, sorted by `Key`.
- `canonicalizeIdArraysDeep` — arrays whose every element is an AWS resource
  id or ARN.

plus `canonicalizeUnorderedArraysAtPaths`, an explicit per-provider opt-in
(`ResourceProvider.getDriftUnorderedPaths`) — but that one handles PLAIN
STRING arrays only. There is no equivalent for an array of OBJECTS.

## Proposal

Extend the provider-declared unordered-path mechanism to object arrays:
sort by canonical JSON at the declared paths, applied to both sides, same as
the existing passes. Then `TargetGroup.Targets` can move out of
`getDriftUnknownPaths` and into `getDriftUnorderedPaths`, and `cdkd drift`
starts detecting console-side target changes.

The same mechanism would serve other providers carrying unordered object
lists; the point of doing it in `drift-normalize.ts` rather than inside a
provider's `readCurrentState` reverse-mapper is the one that file's header
already records — the normalizer runs on BOTH sides, so it stays correct for
the `properties`-fallback baseline (a resource deployed before
observed-capture, whose baseline is the user's template order). Sorting only
the read side would MANUFACTURE drift there.

## Constraints worth carrying over

- Same judgment call as the existing opt-in: a list that IS order-significant
  must stay undeclared. Declaring an order-significant list silently HIDES
  real drift, which is worse than the visible false positive of leaving it
  out. (`getDriftUnorderedPaths`'s existing header records why
  `SelfManagedActiveDirectoryConfiguration.DnsIps` and ElastiCache
  `PreferredAvailabilityZones` are deliberately NOT declared.)
- Nested arrays are not descended into today; decide explicitly whether
  array-of-arrays is in scope rather than inheriting the behavior by
  accident.
- Needs a real-AWS drift round-trip (`drift-revert` or a TargetGroup-specific
  fixture) before the `getDriftUnknownPaths` entry is removed — a
  reorder-only readback must report NO drift across two consecutive runs.
  Two back-to-back runs disagreeing with no code change in between is the
  signature this class produces (see https://github.com/go-to-k/cdkd/issues/1515).

go-to-k:
## Summary

Splits `#1609` item 6 out as its own drift-layer feature, and makes
`ElasticLoadBalancingV2::TargetGroup.Targets` visible to `cdkd drift` for the
first time.

`canonicalizeUnorderedArraysAtPaths` — the `getDriftUnorderedPaths` opt-in
seam — handled **plain-string arrays only**, so a type whose unordered list is
an array of OBJECTS had no way to declare it. `Targets` was therefore parked in
`getDriftUnknownPaths`, meaning a console-side target registration was
invisible, even though the provider can read it back via
`DescribeTargetHealth`.

### 1. The normalizer learns object arrays

Sorted at a declared path only, keyed on a private key-order-independent
`canonicalJson`. **Key order deliberately does not participate**: AWS's
readback order for an object's own keys is no more guaranteed than its order
for the list, so a raw `JSON.stringify` sort key would reintroduce the very
phantom drift the pass exists to remove.

Each shape is sorted only when the array is HOMOGENEOUS in it. A mixed
strings-and-objects list, a `null` element, and a nested array inside a
declared path are all left untouched, so a mis-declared path can never reorder
a heterogeneous list. Array-of-arrays stays out by an explicit decision rather
than an inherited accident (the issue asked for that to be settled): no CFn
shape in tree is one, and an inner list's order-significance is a separate
question the single declared path cannot express.

### 2. Ordering was only half of why `Targets` was parked

The in-code note parking it named **two** blockers, and the issue's title only
covers the first. Deregistration is asynchronous: `DescribeTargetHealth` keeps
reporting a just-removed target as `draining` for minutes, so including one
would freeze it into the deploy-time `observedProperties` snapshot and produce
*permanent* phantom drift once it finally disappeared.

The readback excludes exactly `draining` and INCLUDES every other state —
`initial` / `unused` / `unhealthy` / `unavailable` are all registered targets,
and health is not registration. It also drops the `AvailabilityZone: 'all'`
AWS substitutes for an unscoped `ip` target, and leaves the key **absent**
(never empty) when the health read fails, so a missing
`elasticloadbalancing:DescribeTargetHealth` permission cannot read as "every
target was removed".

### 3. A third hazard, found by self-review before merge (second commit)

Retiring the type-level entry made `Targets` compared for *every* target
group — including one fronting an **ECS service or an ASG**, which declares no
`Targets` because the SIBLING resource registers them and re-registers on every
scale event. That would report drift on an untouched stack, and `--revert`
would deregister the tasks the service had just placed. This is the
[#1498](https://github.com/go-to-k/cdkd/issues/1498) class, and
`undeclaredEmptyObservedKeys` only covers it when the capture happened to be
EMPTY — on an UPDATE deploy of a running service, it is not.

So the unknown-path entry did not disappear, it became **per resource** via the
[#1602](https://github.com/go-to-k/cdkd/issues/1602) seam: compared only when
the template declares `Targets`. An explicit `Targets: []` is a real
declaration and stays compared; an absent properties bag falls back to
comparing, per the method's contract.

## Test plan

**Unit** — `canonicalJson` key-order independence (including a nested-key
case), homogeneity refusals (mixed, `null` element, array-of-arrays),
declared-vs-undeclared paths; the `draining` exclusion paired with an
explicit **include-every-other-state** assertion so the polarity is pinned
both ways; the `AvailabilityZone: 'all'` drop; absent-not-empty on a failed
read; and the five per-resource scoping cases. 574 files / 10773 tests pass.

**Live (real AWS)** — `tests/integration/drift-revert-arrays` gains a
standalone `targetType: 'ip'` TargetGroup (no load balancer, so it is free,
and its `unused` targets exercise the include-non-draining decision live) with
three deliberately unsorted IPs:

- a **second back-to-back** `cdkd drift` must also be clean — two consecutive
  runs disagreeing is this class's signature and one run cannot detect it;
- an out-of-band `RegisterTargets` of a fourth untemplated IP is asserted
  **live before** the revert, so the post-revert assertion cannot pass
  vacuously, and asserted deregistered after with the other three RETAINED.

Run end-to-end three times (the first two caught real defects: an AWS
constraint the CFn schema does not express — `HealthCheckEnabled: false` is
rejected for `targetType: 'ip'` — and then the per-resource scoping fix).
Final run: deploy -> 2x clean drift -> reorder -> detect -> revert -> clean ->
destroy, **0 errors / 0 orphans**.

Closes #1620


go-to-k:
## 3-axis review round (spec / code / test) — all blockers addressed

All three reviewers independently converged on the same **destructive** defect, which is now fixed in `fefee0aa`.

### Blocker: `--revert` deregistered a live target over an omitted `Port`

`updateTargetGroup` keys its register/deregister diff on the whole `(Id, Port, AvailabilityZone)` tuple. AWS substitutes the target group's own port for a target registered without one, so a template-shaped `Targets: [{Id}]` and an AWS-readback-shaped `[{Id, Port: 80}]` describe the **same live target under two different keys** — the diff registered the "new" one (a no-op) and **deregistered the real one**. `cdkd drift --revert` hits exactly that, since it hands `update()` a template-shaped desired side and the AWS snapshot as the previous side.

Closed from **both** directions, because either side can be the one that omits it:

- `convertTargets` defaults an absent `Port` to the group's port at both call sites;
- the readback drops `Port` when it merely echoes the group's own.

Each call site is pinned by its own test — reverting either turns exactly one of them red (verified by real-code probe, not by inspection).

### Also fixed

| Finding | Resolution |
|---|---|
| `AvailabilityZone` echoed on non-`ip` groups | Dropped entirely there — CFn only accepts it for `ip` targets and `DeregisterTargets` rejects it on an instance group, so `--revert` would have failed on the resources it repairs |
| `canonicalJson` collided `undefined` with `null` | Skips `undefined`-valued keys, mirroring `JSON.stringify` — one side arrives raw from the SDK, the other JSON round-tripped from state |
| **My nested-key test was vacuous** | Its elements also differed in a leading `Id`, which decided the order before `Meta` was reached, so it passed with a non-recursive sort key. Replaced with a discriminating fixture, verified to fail when `canonicalJson` is reverted |
| No calculator-level coverage for object arrays | Added both polarities end-to-end through `calculateResourceDrift`, plus the ignore-vs-unordered interaction that models the ECS/ASG-managed group |
| **Integ steps passed with the feature reverted** | Step 4 now greps the drift report for a `Targets` path. Exit 1 alone was satisfied by the S3 / IAM / SG drifts |
| `getDriftUnorderedPaths` contract said "plain-string array" | Corrected in `resource.ts` + `drift-calculator.ts`, including a stale example citing FSx `DnsIps` — a path deliberately **not** declared |

### Deliberately not changed

The empty-readback window (`DescribeTargetHealth` returning `[]` right after `RegisterTargets`) is recorded as a known bound in the readback's JSDoc rather than guarded: `RegisterTargets` is synchronous, the integ's post-deploy drift sees all three fixture targets, and treating an empty readback as unreadable would blind a legitimately emptied target group.

### Verification

Real AWS (`drift-revert-arrays`) re-run end-to-end on the rebased, review-fixed build — deploy → 2× clean drift → reorder → detect → revert → clean → destroy, **0 errors / 0 orphans**. 575 files / 10820 unit tests pass; CI green.

Issue #1634, opened 2026-08-11
A CFn-invalid multi-destination Route is only caught on CREATE — add a pre-flight mutually-exclusive-property check


go-to-k:
Raised by the spec + code reviews on the PR for issue #1591, and partly
addressed there — this issue tracks the remainder.

## The gap

`EC2Provider.createRoute` REFUSES a template declaring more than one of
`DestinationCidrBlock` / `DestinationIpv6CidrBlock` / `DestinationPrefixListId`
(#1566), because CloudFormation and EC2 both reject the combination. But that
refusal only fires on a template-borne CREATE. #1591 then made the deploy diff
normalize both comparison sides, so for a stack that already has the resource:

- `cdkd deploy` classifies NO_CHANGE, never calls the provider, and the
  create-path refusal is never reached.
- `cdkd diff` reports no change either.

#1591 added a diff-side WARNING so this is not silent — it fires whenever
canonicalization actually changes the desired bag. That covers `deploy` and
`diff`. What is still missing is a check that catches the invalid template
BEFORE it is ever deployed, and one that does not depend on a provider
happening to implement `canonicalizeDesiredProperties`.

## Why a pre-flight validator is the right layer

cdkd already refuses whole TYPES (`unsupported-types`) and individual
PROPERTIES (`property-coverage`) at pre-flight, with a per-resource error and a
one-click issue link. A mutually-exclusive-property rule is the same shape and
the same layer, and it would apply on the very first `cdkd synth` / `cdkd diff`
rather than only once the resource exists.

Note the constraint recorded in `.claude/rules/providers.md`: pre-flight runs
BEFORE intrinsic resolution, so a legitimate `Fn::If`-valued destination is an
object whose branch is not yet known. A naive presence check would reject valid
templates. The rule has to either run after resolution or treat an unresolved
intrinsic as "unknown, do not refuse".

## Scope

1. Decide the layer: a pre-flight mutually-exclusive-property check (general,
   reusable — `AWS::EC2::Route` is unlikely to be the only type with one) vs a
   type-specific validator.
2. Handle the unresolved-intrinsic case without false positives.
3. Unit-pin: an invalid template is refused before any AWS call; a template
   whose destinations are behind an `Fn::If` is NOT refused.

Refs #1566, #1591.

go-to-k:
[...]

Issue #1633, opened 2026-08-11
fix(ec2): SecurityGroupIngress IpProtocol substitution produces the same permanent phantom drift as #1591


go-to-k:
Found by the code review on the PR for issue #1591, which fixed the identical
shape one method over.

`EC2Provider.createSecurityGroupIngress` warn-substitutes a default for a
malformed `IpProtocol` and SENDS that default, while
`updateSecurityGroupIngress` returns no `effectiveProperties` — so cdkd state
records the malformed value the template declared, and
`readSecurityGroupIngressCurrentState` returns what AWS actually holds
(`'-1'`). The difference is permanent phantom drift: reported by every
`cdkd drift`, and `drift --revert` "repairs" it by calling `update()` again,
which revokes and re-authorizes the rule and re-emits the same warning.

That is exactly the #1591 shape (#1552 junk-state class), and #1591 shipped the
machinery it needs:

- `EffectivePropertiesResult.effectiveProperties` on the create/update result —
  the provider hands back the bag it actually SENT and the engine records that.
- `ResourceProvider.canonicalizeDesiredProperties` — the diff-side half, so the
  narrowing does not read back as a user-made change on the next deploy.

**Both halves are required.** #1591 learned the hard way that shipping only the
first breaks the next deploy, and that normalizing only ONE comparison side
breaks the population whose state was written before the narrowing existed. See
`.claude/rules/providers.md` for the full contract and
`src/provisioning/canonicalize-properties.ts` for the shared builder.

Scope for a fix:

1. Extract the substitution rule into one helper shared by the provisioning
   path and the canonicalizer (the Route fix uses `narrowRouteDestinations`) —
   re-deriving it lets state and template normalize differently.
2. Return `effectiveProperties` from the create arm and forward it from
   `updateSecurityGroupIngress`.
3. Implement `canonicalizeDesiredProperties` for
   `AWS::EC2::SecurityGroupIngress`.
4. Check whether `IpProtocol` is create-only for this type — that is what
   decides whether a one-sided fix would produce a REPLACEMENT (and therefore
   whether the deploy fails) or an in-place update.

Not urgent: reaching it needs a template or state record that already carries a
malformed `IpProtocol`.

go-to-k:
[...]

Issue #1626, opened 2026-08-11
drift --revert hands providers a FULL observed bag as previousProperties, so 'absent from desired' means two different things


go-to-k:
Found by the code review on PR (#1624). The destructive half is already
closed there; this issue is the residual visibility gap.

## The shape

`cdkd drift --revert` calls, in `src/cli/commands/drift.ts`:

```ts
provider.update(
  outcome.logicalId,
  stateResource.physicalId,
  outcome.resourceType,
  newProperties,             // buildRevertNewProperties(...) over observedProperties ?? properties
  outcome.awsProperties      // the FULL readCurrentState snapshot
);
```

So `previousProperties` is **not** template-shaped there. For a provider that
KEY-DIFFS a collection — ELBv2's `LoadBalancerAttributes`, where
`readCurrentState` deliberately emits every attribute AWS reports — the
previous side carries ~20 keys while the desired side may carry one or two.
Every key AWS reports and the template never set therefore lands on the
provider's REMOVAL path.

For ELBv2 this is reachable whenever `observedProperties` is absent from the
state record: a pre-v3 record, or one where the best-effort
`DescribeLoadBalancerAttributes` read inside `readLoadBalancer` failed at
deploy time (it is deliberately swallowed so a permission gap does not blank
the snapshot).

## What is already fixed

PR (#1624) added the removal-value skip: a key whose CURRENT value already
equals its documented default is not submitted at all. An untemplated
attribute is by definition sitting at its default, so the bulk case writes
nothing. That closes the destructive path for ELBv2.

## What is left

1. **The plan does not warn.** `findRevertDroppedAwsKeys` /
   `collectMissingPaths` return early for anything that is not a plain record
   (`drift.ts`), and `LoadBalancerAttributes` is an array of `{Key, Value}`
   objects — so the (#1478) "these AWS-side keys would be dropped" pre-flight
   is structurally blind to exactly the shape that has the most keys to drop.
   Extending `collectMissingPaths` to `{Key, Value}` arrays would make the
   revert plan honest about what it is about to touch.
2. **The residual write.** A key the template never set that AWS reports at a
   NON-default value (someone changed it in the console) still takes the
   removal path and is reset. That is arguably drift the user asked to revert,
   but they did not ask for it via the template and the plan does not name it —
   see item 1.
3. **The general question.** Handing a provider a full observed bag as
   `previousProperties` makes "absent from desired" mean two different things
   depending on the caller (a template removal vs never-templated). Every
[...]

Issue #1591, opened 2026-08-11
EC2::Route: the multi-destination warn arm records a bag AWS cannot represent, producing permanent phantom drift (#1552 class)


go-to-k:
Found by the code review on PR #1590 (issue #1566).

#1590 makes `EC2Provider.createRoute` REFUSE a template declaring more than one of `DestinationCidrBlock` / `DestinationIpv6CidrBlock` / `DestinationPrefixListId`, and downgrades that refusal to a warning on the state-borne paths — including `updateRoute`, which passes the callback unconditionally because it deletes the route before re-creating it and a throw there would strand it.

That downgrade leaves a residue, and it is the `warn-and-continue update path becomes a producer of junk state` class (#1552) one provider over:

- The warn arm narrows to the precedence winner and the update SUCCEEDS, so the deploy engine records the DESIRED bag — carrying **all** the declared destination keys — into cdkd state.
- `readRouteCurrentState` (`src/provisioning/providers/ec2-provider.ts`, the `readRoute*` reverse map) returns exactly ONE destination key, because that is all AWS has.
- So the extra key is permanent phantom drift: every subsequent `cdkd diff` / `cdkd drift` reports it, and `drift --revert` acts on it by calling `update()` again — which delete-and-recreates the route and re-emits the same warning on every run.

Scope / reachability: this needs a template (or state record) that already carries the CFn-invalid multi-destination shape. On the CREATE path #1590 now refuses it outright, so the only way in is (a) a state record written by a pre-#1590 binary, which silently narrowed and recorded the full bag, or (b) the update path, which cannot refuse. Both are exactly the populations that would hit the phantom drift.

Options, roughly in increasing order of intrusiveness:

1. **Record the narrowed bag.** On the warn arm, strip the losing destination keys from the properties the engine persists, so state matches what AWS holds. This is the `#1552` remedy shape (seed the recorded value from what was actually delivered) and kills the drift at the source. Needs care: the provider does not own the state write, so this likely means returning the effective properties rather than mutating the input bag.
2. **Declare the losing keys drift-unknown** via `getDriftUnknownPaths` for `AWS::EC2::Route`. Cheap, but wrong in the general case — it would also hide a genuine destination change.
3. **Leave as-is and document.** The shape is CFn-invalid to begin with and the warning is emitted on every affected update, so the drift is at least not silent.

[...]

PR #1647, opened 2026-08-12
fix(drift): honour a provider's effectiveProperties on drift --revert and the rollback revert arms

### LINKED ISSUE #1644: fix(drift): drift --revert discards update()'s effectiveProperties, so the narrowing loop persists on that path
Found by the test review on the PR for issue #1633 (PR #1641).

## The gap

#1633 / #1591 both rest on the provider returning `effectiveProperties` and the ENGINE recording it in place of the desired bag (`DeployEngine.propertiesToRecord`, `src/deployment/deploy-engine.ts`).

`cdkd drift --revert` also calls `provider.update(...)`, but it **discards the return value** — see the call in `src/cli/commands/drift.ts`:

```ts
provider.update(
  outcome.logicalId,
  stateResource.physicalId,
  outcome.resourceType,
  newProperties,
  outcome.awsProperties
);
```

There is no `propertiesToRecord`-equivalent on that path, so a narrowing the provider reports on a revert is dropped: the record keeps the un-narrowed value, the next `cdkd drift` reports the same difference again, and `--revert` re-issues the same call. For `AWS::EC2::SecurityGroupIngress` that means a revoke + re-authorize + the same warning on every invocation — the exact loop `effectiveProperties` exists to break, still live on this one command.

The deploy path IS fixed, so a user who runs `cdkd deploy` once converges. The loop only persists for someone using `drift --revert` as their repair tool.

## Scope

- Thread the `update()` result through `runRevert` and persist `effectiveProperties` the way the deploy engine does (`result.effectiveProperties ?? desiredProperties`).
- Check the same question for the ROLLBACK path (`rollback-executor.ts`'s revert arm calls `update()` too) — it may have the identical gap.
- Unit-pin both: a provider returning `effectiveProperties` from `update()` must leave the saved record carrying the narrowed bag.

Applies to every `effectiveProperties` producer, not just this type — today `AWS::EC2::Route` (#1591) and `AWS::EC2::SecurityGroupIngress` (#1633).

go-to-k:
## Summary

`provider.update()` may answer with `effectiveProperties` — the bag it ACTUALLY
sent, which is what AWS holds. It is the loop breaker behind #1591
(`AWS::EC2::Route`) and #1633 (`AWS::EC2::SecurityGroupIngress`), and
`DeployEngine.propertiesToRecord` honours it.

The three OTHER `provider.update()` call sites discarded the return value:
`cdkd drift --revert` and the rollback executor's `revert` /
`revert-failed-update` arms. So a narrowing announced there never reached state,
the record kept describing a value AWS does not hold, the next `cdkd drift`
reported the identical difference, and `--revert` re-issued the identical call —
forever, for anyone using it as their repair tool.

## What changed

The two shapes differ because the bag each caller HANDS the provider differs.

- **Rollback arms** send `previousState.properties` verbatim, so
  `effectiveProperties` is its complete replacement:
  `{...restored, properties: {...effectiveProperties}}` (copied, not aliased).
  Physical id, attributes, dependencies and policies survive untouched.
- **`drift --revert`** sends `buildRevertNewProperties`'s output — AWS-CURRENT
  values for every non-drifted key merged with the state baseline for the
  drifted ones. Writing that back wholesale would turn `--revert` into
  `--accept`, so it persists a per-key DELTA (`collectNarrowedTopLevelKeys`)
  into the field the comparator uses as its baseline (`observedProperties` when
  present, else `properties`) — `properties` keeps the user's template intent,
  because a narrowing is an AWS-side fact, not a template edit.

Four guards keep the delta from leaking AWS values into state (three of them
found by the review):

1. presence is `key in effective`, not a compare against `undefined` — a `null`
   on both sides is not a phantom drop, a real drop of one still is;
2. the comparison is key-order-INDEPENDENT deep equality, not `JSON.stringify`
   — a provider rebuilding a nested object would otherwise register as changed,
   and the value written back for such a key is the AWS-current one that was
   sent;
3. only a key the BASELINE already declares may move — the sent bag starts as
   the AWS-current snapshot and carries keys state never tracked;
4. with NO `observedProperties` only a DROP is recorded, never a value — that
   baseline is the raw template and `buildRevertNewProperties` ran in
   `preserveUntemplated` mode (#1626), so the sent value deliberately carries
   untemplated AWS paths that would disable the #1160 removal derivation.

The write happens once per stack, after the concurrent tasks settle and still
under the stack lock, with the same `expectedEtag` / `migrateLegacy` handling as
`--accept`. It is **best-effort**: AWS has already been reverted, so a failed
write warns and the run carries on — a throw would skip every later stack's
revert under `--all`. The capture also sits after the per-resource success
accounting in its own `try`, so a non-comparable bag cannot re-report a
succeeded revert as `AWS update failed`.

## Test plan

- 11 drift command cases + 10 `collectNarrowedTopLevelKeys` unit cases + 5
  rollback cases. Mutation-probed: neutering the drift wiring fails 4; the
  rollback recorder 2; the best-effort `catch` 1; the baseline-restriction
  guard 1; the template-only-baseline guard 2; reverting to `JSON.stringify` 2.
- Full suite: 585 files / 11017 tests green; typecheck, test-project typecheck,
  lint, format, build clean; every generated matrix regenerates clean.
- **Live-tested against real AWS** (us-east-1, scratch stack
  `CdkdIssue1644Verify`, VPC + IGW + route table + `AWS::EC2::Route`), re-run
  against the final build. The legacy record shape was seeded by hand-patching
  `state.json` with a second destination key, since #1591 + #1634 mean no
  template deployable today produces it:
  - `cdkd drift` reports the difference (exit 1);
  - `cdkd drift --revert --yes` warns about the extra destination, reverts, and
    prints `State updated ... recorded the value the provider actually applied`;
  - `cdkd drift` again is **clean (exit 0)**; `observedProperties` lost only the
    narrowed key and `properties` (template intent) is untouched;
  - **negative control**: with the fix neutered and rebuilt, the second `drift`
    reports the identical difference again (exit 1) — the loop this PR closes.
  - Torn down: 5 resources deleted, 0 errors; state, event history and every EC2
    resource swept to zero.

Reviewed by `pr-code-reviewer` + `pr-test-reviewer`; every finding is either
fixed here or recorded below.

**Won't-do (recorded):** the rollback arms leave `restored.observedProperties`
describing the un-narrowed value, so a `cdkd drift` right after a rollback can
still report it. Refreshing it would mean claiming a read-back that never
happened — `observedProperties` is a deploy-time capture by contract, and the
deploy engine's own `propertiesToRecord` writes `properties` only. Out of scope
for this PR.

No CLI flag, dependency, or state-schema change.

Closes #1644


go-to-k:
Post-merge note — residual reviewer nits, for the record.

Two coverage gaps the test reviewer flagged were consciously NOT closed, both low-severity and neither reachable as a defect today:

- **Multi-stack `--all` best-effort is not covered end-to-end.** `drift.test.ts` proves the catch-vs-throw behavior on a single stack (mutation-probed), but not the downstream claim that a failed state write on stack A still lets stack B revert. Closing it needs a two-report fixture for one `continue`-vs-`throw` edge that the single-stack test already fences at the same statement.
- **`if (!existing) continue` in the state-write loop is untested.** It is defensive against a resource vanishing from `report.state.resources` between the drift read and the write, which cannot happen under the stack lock this block holds.

Everything else from the code + test reviews was fixed in `a2339901`: the `key in effective` presence test, the key-order-independent comparison, the baseline-declares-it restriction, the DROP-only rule on a template-only baseline, the capture moved after the success accounting into its own `try`, the copied (not aliased) rollback bag, and the rollback drop-key / identity / `revert-failed-update` pins.

The one deliberate non-fix is already in the PR body: the rollback arms leave `restored.observedProperties` describing the un-narrowed value, because refreshing it would claim a read-back that never happened.

Issue #1618, opened 2026-08-11
test(lint): require mockReset() in suites using mockResolvedValueOnce — clearAllMocks leaves the queue and shifts later tests


go-to-k:
Found while fixing issue (#1588): a unit suite using `mockResolvedValueOnce`
queues with only `vi.clearAllMocks()` in `beforeEach` can silently feed the
WRONG AWS response to later tests, and the affected test still passes.

## The mechanism

`vi.clearAllMocks()` clears call RECORDS. It does not drain the queue seeded by
`mockResolvedValueOnce`. Confirmed in the installed spy —
`node_modules/.pnpm/@vitest+spy@4.1.10/…/dist/index.js`: `mockReset` sets
`config.onceMockImplementations = []`; `mockClear` does not.

So a test that primes more responses than its code path consumes leaves the
remainder queued, and the NEXT test's first `send()` receives the previous
test's leftover — every later call shifted by the same offset.

## Why it is dangerous rather than merely annoying

The shifted test does not error. It reads a `DescribeTable` describing a
different table, takes a different branch, and then satisfies its own
assertions — because the assertions on that branch are usually ABSENCE
assertions (`toBeUndefined()`, `toHaveLength(0)`,
`not.toHaveBeenCalled()`), and an absence assertion is satisfied both by
"the guard correctly declined" and by "the code never got there".

In (#1588) the last test in a new suite primed a live PAY_PER_REQUEST table with
one GSI and the provider instead saw a leftover response with no indexes, so the
branch under test never ran. It still passed its `toBeUndefined()` assertion; the
only symptom was one `logger.warn` that was mysteriously never called, and
locating it took a full instrumentation pass.

## Scope

`clearAllMocks`-without-`mockReset` alongside `mockResolvedValueOnce` is the
prevailing idiom across `tests/unit/provisioning/` (40+ files by the reviewer's
count). Most are correct TODAY only because their priming happens to match
their consumption exactly — which is not a property anyone maintains
deliberately, and which silently breaks the moment a provider gains or loses one
AWS call (adding a wait poll, an extra `DescribeTable`, a second `UpdateTable`).
That is a routine change; issue (#1588) made exactly that kind of change.

## Proposal

A lint in the codegen'd-critic family: **a test file that calls
`mockResolvedValueOnce` must also call `mockReset()` (or `vi.resetAllMocks()`)
in its `beforeEach`.**

- statically decidable, so it belongs with the other `scripts/check-*.ts`
[...]

PR #1684, opened 2026-08-12
test: fix the three grandfathered *Once-leak suites and drop them from the allow-list

### LINKED ISSUE #1655: test: three suites silently consume another test's primed mock response (found by the 1618 detector)
Found by the runtime `*Once`-leak detector added for issue (#1618), on its first
full-suite run.

## What was measured

Three files contain tests that consume a mock response primed by an EARLIER
test in the same file — 27 such consumptions in total, measured 2026-08-12 on
`main` plus the (#1618) branch:

- `tests/unit/provisioning/elbv2-provider-roundtrip.test.ts`
- `tests/unit/provisioning/iam-role-provider.test.ts`
- `tests/unit/provisioning/s3-tables-provider-roundtrip.test.ts`

These are the real thing, not a proxy: the detector flags only a value actually
shifted off the queue during a different test than the one that primed it. An
over-priming that nothing consumes, and a suite that drains with `mockReset()`
in `beforeEach`, are both correctly ignored — the first pass of the detector
used the "queue non-empty at end of test" proxy instead and reported 15 files,
12 of which were not defects.

## Why this needs its own pass

Per issue (#1588)'s write-up, a test in this state usually still PASSES: it reads
a response describing a different resource, takes a different branch, and then
satisfies that branch's assertions — which are typically absence assertions
(`toBeUndefined()`, `not.toHaveBeenCalled()`), and an absence assertion is
satisfied both by "the guard correctly declined" and by "the code never got
there".

So each of the 27 needs to be read individually to establish what the test was
supposed to assert and whether the branch it means to cover has ever actually
run. That is investigative work that may surface real provider bugs, which is
why it did not ride the PR that added the detector — that PR is scoped to the
enforcement mechanism.

## Scope

1. For each of the three files, fix the priming so each test consumes exactly
   what it primes (or drain with `mockReset()` in `beforeEach`).
2. For every test whose branch turns out never to have executed, verify the
   assertion still holds once it does — and file separately if it does not.
3. Drop each file from `tests/once-leak-allowlist.json` as it is fixed
   (`vp run gen:once-leak-allowlist` regenerates). An empty list is the goal
   state; the allow-list test caps the size but deliberately has no floor.

## Reproduce

Remove the file from `tests/once-leak-allowlist.json`, then:

```bash
vp run test:once-leak
```

go-to-k:
## Summary

The runtime `*Once`-leak detector from issue (#1618) measured **27 cross-test mock-response consumptions** across three test files and grandfathered them in `tests/once-leak-allowlist.json`, so each was exempt from the check — including for any test added to it later.

All 27 are fixed by removing the over-priming, and all three files are dropped from the allow-list, which now holds nothing but the deliberately-leaking canary: the documented goal state.

The fix is **priming-side rather than a `mockReset()` drain in `beforeEach`**. A drained suite can never be flagged again, which would re-create exactly the grandfathering the ratchet exists to remove.

## What the 27 turned out to be

In all three files the surplus described a call the code path **never makes**, and in all three the priming's own COMMENT asserted that call:

| File | Leaks | The priming claimed | What actually happens |
|---|---:|---|---|
| `s3-tables-provider-roundtrip` | 1 | `GetTable` + `ListTagsForResource` | `readTableCurrentState` gates the tags read on `resp.tableARN`, a field the mocked `GetTable` response omitted — so it never fired |
| `iam-role-provider` | 8 | `UpdateRole` + 3 policy/tag no-ops + `GetRole` | with no policies/tags in the props those helpers issue **zero** sends — as a sibling test's own helper in the same file already documented |
| `elbv2-provider-roundtrip` | 18 | `ModifyTargetGroup` → `DescribeTargetGroups` → `DescribeTags` | `update()` derives the tag diff from the desired-vs-previous property bags and never reads AWS tags (`readCurrentState` does, which is why the spelling looked right) |

So the measured consumption contradicted the annotation every time. Counts were obtained by draining the queue with `mockReset()` and logging the per-test call count; **measuring while the queue still leaks is circular**, because a consumer's count includes the responses it inherited.

## Was any assertion hiding a defect?

**No** — checked per the issue's scope item 2 rather than assumed. Two shifted tests had genuinely never executed their own branch:

- `s3-tables`' Format-less `GetTable` case was reading the previous test's leftover tags response.
- The `iam-role` clear-on-removal trio was taking a leaked `{}` as its `GetRole` response.

Both still pass once the branch actually runs, because each asserts on the **command input** rather than on the response. Nothing to file separately.

## Two guards, two different regressions

A probe settled this rather than intuition, and the first draft of the docs got it wrong:

- **Dropping the files from the allow-list** re-arms the detector for the **surplus** case. Verified against real code: re-introducing one primer and running with `CDKD_ONCE_LEAK_DETECT=1` fails, naming the priming test; the same edit under a plain `vp test run` passes.
- **`expect(mockSend).toHaveBeenCalledTimes(N)`**, added beside the existing assertions at every site whose priming was wrong, catches the complementary case — the code path changing how many calls it makes, which is what silently invalidates a priming in the first place.

The count pin does **not** catch a surplus primer: an unconsumed response leaves the count unchanged. That was measured, and the corrected statement is now in `.claude/rules/testing.md` and `docs/testing.md`.

## Also

- Fixes a mis-titled test — `s3-tables`' Format-less case announced `AWS::S3Tables::TableBucket` while exercising `::Table`.
- Updates the generator's `$comment` (it still said fixing these files was *tracked by* this issue) and regenerates the allow-list so the two cannot drift.
- Adds the measured lessons to `.claude/rules/testing.md` / `docs/testing.md` and a changelog entry.

## Review

Two read-only reviewers (code-quality + test-adequacy) ran against this diff. **No blockers.** Both independently re-derived the claims rather than taking them on trust — one confirmed the call sequences against provider source, the other cloned `origin/main` and re-measured **exactly 27** cross-test consumptions in snapshot mode, so green-on-branch is a real delta and not a dead instrument. Two findings are worth recording because they strengthen the change:

- `tableARN` is a **required** member of `GetTableResponse`, so the old mock modelled a response AWS cannot return and silently exercised the defensive `: []` arm. The new fixture is strictly more faithful.
- A mutation probe (neutering the table tags read to `result['Tags'] = []`) fails **only** the two new `toHaveBeenCalledTimes(2)` assertions — the `toEqual({... Tags: []})` assertions still pass. So the pin is the sole guard for that regression, not decoration.

All four raised items are fixed here:

1. **should-fix** — the `import` describe still drained with `mockSend.mockReset()` under a now-false comment. That is exactly the remedy this PR argues against: it would make any future leak crossing into that block invisible to the detector, re-grandfathering the tail of a file just un-grandfathered. Hook deleted, with the reasoning left in place of it.
2. **nit** — `once-leak-allowlist.test.ts` still called an EMPTY list the goal state, contradicting the new allow-list `$comment` and the assertion three lines above that the canary must always be listed. Reworded to canary-only.
3. **nit** — the `TABLE_ARN` comment claimed AWS returns the table NAME in the ARN; real ones end in an opaque table id. Constant and comment corrected.
4. **nit** — the aggregate cross-type test pinned its TG leg but not its listener leg. Now pinned at 3, which also closes the detector's structural blind spot at a file's TAIL: a surplus primed by the last test is never consumed, so the detector alone cannot see it.

Re-verified after the fixes: full suite 594 files / 11179 tests, and `test:once-leak` exit 0 with zero leaks.

## Test plan

- `vp run test` — 594 files, 11179 tests, all pass.
- `CDKD_ONCE_LEAK_DETECT=1 vp run test:once-leak` — 594 files, **11179 tests, zero leaks** (exit 0).
- `vp run test:once-leak-canary` — still **exits 1** carrying the detector's own wording `primed by an EARLIER test`, so the detector is proven to still reject rather than having gone silently dead.
- Real-code fail probe: re-introduced one over-priming into `elbv2-provider-roundtrip.test.ts` and confirmed the detector fails naming `Class 2 — TargetGroup TCP no-drift round-trip...` as the primer; reverted.
- `vp run gen:all-matrices` + `vp run audit:coverage:check` — no generated artifact staled.

No `src/**` change, so the `integ-destroy` / `integ-broad` / `integ-local` / `integ-schema-migration` gates are all out of scope for this diff, and no AWS resources were created.

Closes #1655


go-to-k:
Post-merge note — one reviewer observation deliberately NOT acted on, recorded here so the decision is visible.

The test-adequacy review flagged, as pre-existing info rather than a finding: no test in `iam-role-provider.test.ts` asserts `update()`'s RETURNED `attributes`, so the `GetRole` readback is exercised only as "a call happened".

Decided not to close that here. The fields in question (`Role?.Arn`, `Role?.RoleId`, read at `iam-role-provider.ts:172-174`) ARE asserted — through the provider's own `getAttribute` tests in the same file, which prime a `GetRole` response and check both values. So a regression in the field reads is caught; what is untested is only the plumbing that copies them onto the `update()` return.

It is also outside this issue's scope: the gap predates the leak and is not one of the shifted branches the leak was hiding (all of those were verified to still pass, per the PR body). Fixing it here would have meant editing the file for a reason unrelated to the priming, on a PR whose whole point is that the priming and the assertions were out of step.

Not filing an issue for it — the exposure is one assignment with the underlying reads already covered. Raising this later is cheap if it ever bites.

Issue #1671, opened 2026-08-12
fix(s3): the empty-collection placeholder skip in applySubConfigDiffs records an empty declaration for a Put that never ran


go-to-k:
Found by the code review on PR (#1660) (issue (#1612)).

## The gap

`S3BucketProvider.applySubConfigDiffs` has two `onPut` arms that skip the Put for
an EMPTY-collection desired value ("Class 2 placeholder"):

```ts
// Lifecycle
async (cfg) => {
  if (!cfg.Rules || !Array.isArray(cfg.Rules) || cfg.Rules.length === 0) return;
  ...
}
// CORS
async (cfg) => {
  if (!cfg.CorsRules || !Array.isArray(cfg.CorsRules) || cfg.CorsRules.length === 0) return;
  ...
}
```

`diffSubConfig` reaches `onPut` only when the value CHANGED, so the live
configuration still holds the previous rules while the engine records
`{Rules: []}` / `{CorsRules: []}`. That is the same shape (#1612) just fixed for
the malformed-value skips: a Put that never ran, recorded as though it had.

Unlike those, this arm is reachable from an ORDINARY template path, not only
from a state replay — a condition-pruned or intrinsic-collapsed template can
synthesize an empty rules array.

## Why it was not fixed in the (#1612) PR

(#1612)'s remedy for a skipped whole-Put is `retainPrevious(key)`, and applying
it here is one line — but the placeholder arm exists to absorb the
`readCurrentState` round-trip (`cdkd drift --revert` feeds an always-emitted
empty-rules block back through `update()`), so changing what it RECORDS
interacts with the drift-revert path in a way that needs its own live test. That
is a different verification story from the malformed-value skips, so it was
split out rather than bundled.

## Scope

- Decide whether the empty-collection placeholder should record the PREVIOUS
  value (making state describe AWS) or keep recording the empty declaration
  (making the next deploy derive a removal). These are opposite answers and the
  right one depends on whether an empty array is a REMOVAL intent — check what
  CloudFormation does with `LifecycleConfiguration: { Rules: [] }` by live A/B,
  the way (#1466) settled the sibling removal semantics.
- Whichever is chosen, unit-pin it and note the decision in
  `.claude/rules/providers.md` next to the (#1612) carve-out.
- Live-verify against the `s3-lifecycle` / `drift-revert` fixtures.

go-to-k:
[...]

Issue #1466, opened 2026-08-10
fix(s3): removing VersioningConfiguration / OwnershipControls / BucketEncryption is a silent no-op (diverges from CloudFormation)


go-to-k:
## Summary

Removing `VersioningConfiguration`, `OwnershipControls`, or `BucketEncryption` from an
`AWS::S3::Bucket` in the template is a **silent no-op** in cdkd. CloudFormation resets each
of these to its type default on removal; cdkd leaves the previously-applied value in place
on the real bucket, reports `updated ✓`, and then reports **no drift**.

The user gets no signal from any command that the bucket still carries the removed config.

Found by `/hunt-bugs` (deploy-first sweep, 2026-08-10). Verified against real AWS with a
live CloudFormation A/B on the same template pair.

## Root cause

`S3BucketProvider.applySubConfigDiffs` documents the split at
[`s3-bucket-provider.ts:1436-1443`](https://github.com/go-to-k/cdkd/blob/main/src/provisioning/providers/s3-bucket-provider.ts#L1436-L1443):

> `Versioning / PublicAccessBlock / Tags / OwnershipControls / BucketEncryption` stay on
> `applyConfiguration` (the unconditional always-PUT path) because their AWS APIs don't have
> a clean "delete" counterpart and they round-trip safely as no-ops when state == AWS-current.

The premise "no clean delete counterpart" does not hold for all five, and "round-trip safely
as no-ops" is only true while the property is still PRESENT. Each is guarded by a
presence check, e.g. `OwnershipControls` at
[`s3-bucket-provider.ts:1325`](https://github.com/go-to-k/cdkd/blob/main/src/provisioning/providers/s3-bucket-provider.ts#L1325):

```ts
const ownershipControls = properties['OwnershipControls'] as ... | undefined;
if (ownershipControls?.Rules) {          // <-- absent on removal => nothing runs
  await this.s3Client.send(new PutBucketOwnershipControlsCommand({ ... }));
}
```

So when the property disappears from the template, the branch simply never runs and the old
AWS-side value survives. `applySubConfigDiffs`' `diffSubConfig` helper (which DOES handle
removal via an `onRemove` callback) is never consulted for these five.

## Live CloudFormation A/B

Same two templates (`BucketName` differing only so both can exist at once), deployed base then
stripped, through CloudFormation and through cdkd. Complete enumeration of the always-PUT set:

| Property removed | CloudFormation | cdkd | Verdict |
| --- | --- | --- | --- |
[...]

PR #1731, opened 2026-08-12
fix(s3): an empty BucketEncryption / OwnershipControls collection must not DELETE the live config

### LINKED ISSUE #1713: fix(s3): an empty BucketEncryption / OwnershipControls collection DELETES the live configuration, where CloudFormation rejects the template (1671 sibling)
Found while fixing (#1671) (the lifecycle / CORS empty-collection skip). Filed
rather than folded in because it is a different pair of properties, the fix is a
behavior change on a data-safety-relevant path, and it needs its own tests and
live verification.

## The gap

`S3BucketProvider.applySubConfigDiffs` treats an empty collection in TWO
incompatible ways:

- **Lifecycle / CORS** — the `onPut` arm SKIPS the Put, leaving the live
  configuration alone. This is the (#1671) arm, now also recording the previous
  value.
- **OwnershipControls / BucketEncryption** — both sides are normalized through
  `emptyListConfigToUndefined(...)` BEFORE `diffSubConfig`, so an empty
  collection becomes `undefined` and a non-empty previous against an empty
  desired takes the **onDelete** arm — issuing
  `DeleteBucketOwnershipControls` / `DeleteBucketEncryption`.

Measured against live CloudFormation (us-east-1, 2026-08-12), the second
behavior DIVERGES. Updating a deployed bucket from

    "BucketEncryption": { "ServerSideEncryptionConfiguration": [ { "ServerSideEncryptionByDefault": { "SSEAlgorithm": "aws:kms" } } ] },
    "OwnershipControls": { "Rules": [ { "ObjectOwnership": "BucketOwnerPreferred" } ] }

to

    "BucketEncryption": { "ServerSideEncryptionConfiguration": [] },
    "OwnershipControls": { "Rules": [] }

drives the stack to `UPDATE_ROLLBACK_COMPLETE` with

    Resource handler returned message: "The XML you provided was not well-formed
    or did not validate against our published schema (Service: S3, Status Code: 400 ...)"

and BOTH live configurations survive unchanged — the KMS default encryption and
the ownership rule are still in place after the rollback. So CloudFormation
treats an empty collection as an INVALID template, not as a removal, exactly as
it does for `LifecycleConfiguration` / `CorsConfiguration` (the A/B that settled
(#1671), same date, same account).

cdkd instead DELETES. For `BucketEncryption` that silently drops a declared
`aws:kms` default down to the SSE-S3 / AES256 default on a template whose only
fault is a collapsed array — a condition-pruned or intrinsic-collapsed template
synthesizes exactly that, which is why (#1671) called the arm reachable from an
ordinary template path rather than only a state replay.

## Why the normalization exists (do not just delete it)

`emptyListConfigToUndefined` is load-bearing for the round-trip:
`readCurrentState` ALWAYS emits `BucketEncryption: { ServerSideEncryptionConfiguration: [] }`
and `OwnershipControls: { Rules: [] }` for a bucket with no explicit setting, so
`cdkd drift --revert` feeds that placeholder back through `update()`. Because
the helper normalizes BOTH sides, empty-vs-empty compares equal and no call is
made — that part is correct and must keep working. The defect is only the
asymmetric case: previous NON-empty, desired EMPTY, which currently derives a
removal.

## Scope

- Split the two cases. Keep normalizing for the empty-vs-empty round-trip, but
  make previous-non-empty-vs-desired-empty take the (#1671) treatment: SKIP the
  call, record the PREVIOUS value, and warn — rather than deleting. The shared
  `emptyCollectionSkip` helper added by the (#1671) PR is the precedent.
- Removing the property ENTIRELY must keep deleting: that is the (#1466)
  behavior (CFn suspends versioning / deletes ownership controls / resets
  encryption on property REMOVAL), which was measured on an ABSENT property, not
  on an empty collection. The two inputs are being conflated today and they are
  not the same question.
- Unit-pin both polarities per property, with a control asserting the
  empty-vs-empty round-trip still issues no call, and a control asserting a
  genuine property removal still Deletes.
- Live-verify with the `s3-subconfig-removal` fixture, which already exercises
  the removal path for these properties.

## Related

- (#1671) — the lifecycle / CORS empty-collection skip, and the CFn A/B that
  established empty-is-not-a-removal.
- (#1466) — the A/B that established the PROPERTY-REMOVAL semantics these two
  sites correctly implement, and which the empty-collection case is being folded
  into by mistake.
- (#1612) — the warn-and-SKIP recording class.

go-to-k:
## Summary

`applySubConfigDiffs` treated an empty collection in **two incompatible ways**.
Lifecycle / CORS SKIP the Put and leave the live configuration alone (the
(#1671) arm), while `OwnershipControls` / `BucketEncryption` normalize BOTH
sides through `emptyListConfigToUndefined` *before* `diffSubConfig` — so a
non-empty previous against a declared-but-empty desired became `undefined`, took
the **onDelete** arm, and issued `DeleteBucketOwnershipControls` /
`DeleteBucketEncryption`.

For `BucketEncryption` that silently drops a declared `aws:kms` default to the
SSE-S3 / AES256 default, on a template whose only fault is a condition-pruned or
intrinsic-collapsed array — reachable from an **ordinary template path**, not
just a state replay.

## What CloudFormation actually does

Measured live (us-east-1, same account and date as the (#1671) A/B): updating a
deployed bucket from an `aws:kms` default plus a `BucketOwnerPreferred` rule to
`ServerSideEncryptionConfiguration: []` / `Rules: []` drives the stack to
`UPDATE_ROLLBACK_COMPLETE` with *"The XML you provided was not well-formed or
did not validate against our published schema"*, and **both live configurations
survive unchanged**. So the empty collection is an INVALID TEMPLATE, not a
removal — the identical answer (#1671) measured for lifecycle / CORS, which is
what makes the two arms' disagreement a defect rather than a design.

## The fix

Both arms now route through the same `emptyCollectionSkip` helper the lifecycle
/ CORS arms use: the call is skipped, the live configuration is untouched, the
PREVIOUS value is recorded (so a later template that genuinely REMOVES the block
still derives the removal), and the skip is ANNOUNCED rather than silent.

**The fold is deliberately not what changed.** `readCurrentState` ALWAYS emits
`BucketEncryption: { ServerSideEncryptionConfiguration: [] }` /
`OwnershipControls: { Rules: [] }` for a bucket with no explicit setting, so
`cdkd drift --revert` feeds that placeholder straight back through `update()`
and both sides must keep normalizing for empty-vs-empty to compare EQUAL and
issue no call at all. What was missing is that the fold **collapses** two
different desired sides — declared-but-empty and ABSENT — into one `undefined`,
and only the second is a removal. `declaresEmptyCollection` splits exactly that
arm: **declared-but-empty SKIPS, absent DELETES**.

## This reverses a contract (#1466) pinned — please confirm

That issue's row asserted `empty placeholder desired + real previous REMOVES
(empty == not declared)`, reasoning it is *"a genuine `declared -> not declared`
transition … matching CloudFormation"*. **The CFn-parity half was asserted, not
measured**, and the A/B above measures the opposite. The row is inverted with
both the old reasoning and the new measurement recorded in place, plus a new
sibling row pinning the transition (#1466) actually meant (ABSENT desired still
REMOVES) — which is the discrimination the whole fix rests on, since before it
the two desired sides were literally the same `undefined` after the fold and
could not disagree.

**Scope is exactly what was measured, and two failing (#1466) rows enforced
that during the work.** `emptyListConfigToUndefined` also folds a bare `{}` and
a `{listKey: null}`-only block to `undefined`, and (#1466) pinned BOTH of those
as removals. This issue's A/B exercised neither — so `declaresEmptyCollection`
tests the list key present-and-empty **directly** rather than delegating to the
fold. Delegating would have reversed two further contracts on evidence that does
not cover them.

## Test plan

- **5 new unit cases** in `s3-bucket-provider-shape-and-empty-skip.test.ts`: the
  two regression rows (no Delete, no Put, previous value recorded) plus **three
  controls that are what make them non-vacuous** — removing the property
  entirely still Deletes BOTH configurations, the `readCurrentState` placeholder
  round-trip still issues NO call in either direction, and a malformed block
  still reaches the applier and throws.
- **2 reversed / added rows** in `s3-bucket-provider-roundtrip.test.ts` for the
  (#1466) contract change described above.
- **Mutation-probed against the real provider**: neutralizing the two skip
  guards fails exactly the 2 regression rows with all 3 controls surviving.
- **Full suite**: 605 files / 11621 tests pass.
- **Live-tested end to end against real AWS** (us-east-1), both directions:

  | phase | template | live encryption | live ownership |
  |---|---|---|---|
  | 1 — baseline | `aws:kms` + `BucketOwnerPreferred` | `aws:kms` | `BucketOwnerPreferred` |
  | 2 — collapsed | both collections `[]` | **`aws:kms` (survived)** | **`BucketOwnerPreferred` (survived)** |
  | 3 — removed | properties dropped entirely | `AES256` (deleted) | `OwnershipControlsNotFoundError` (deleted) |

  Phase 2 printed both skip warnings, and state recorded the **previous** value
  rather than `[]` — so no phantom drift. Phase 3 is the control proving the
  template-side remedy the warning prescribes still works. Destroyed clean (2
  deleted, 0 errors); bucket gone, state gone, KMS key `PendingDeletion`.
- **`/run-integ`** against a bucket-touching fixture — required by the
  `integ-destroy` gate; run before merge.

Closes #1713


go-to-k:
## Parking this in draft — review found it not shippable

A `pr-code-reviewer` pass found a **blocker**, and I verified it before acting on it: `readOwnershipControls` / `readBucketEncryption` return the empty placeholder (`{Rules: []}`) for an UNSET feature, so `cdkd drift --revert` hands `update()` a desired bag that is byte-identical to the collapsed-template shape this PR teaches the provider to SKIP.

The two callers mean opposite things by it:

- **template** — a condition-pruned array; CFn rejects it and keeps the live value, so SKIP is right (this PR's measurement).
- **`drift --revert`** — "restore the state where this feature was unset", so DELETE is right.

`update()` has no context to tell them apart. As written, this PR regresses revert for the two properties where it previously WORKED, and worse: the skip records the previous (= AWS-current) value via `retainPrevious` into `effectiveProperties`, which `drift.ts` persists into `observedProperties` — so an out-of-band change is silently **laundered into the baseline** and the next `cdkd drift` reports clean.

Filed as (#1732) with four candidate directions; it needs a design call rather than a patch.

**Why nothing caught it:** the unit tests, the mutation probe and a real-AWS three-phase live test all pass, because every one of them exercises the TEMPLATE caller. That is now recorded as a rule in `.claude/rules/providers.md` — before adding a skip on an update path, enumerate every caller that can produce the skipped shape.

Two documentation errors the same review found are fixed in `adbda371` (the `{Rules: null}` claim was wrong, and the rules file prescribed the opposite of the code). The rest of the branch stands and the live A/B in the description is unaffected — it is only the revert caller that is unresolved.

go-to-k:
## Blocker closed — back out of draft

(#1732) is implemented on this branch: `update()` gains an optional `UpdateContext`, the sibling of `CreateContext`, and `cdkd drift --revert` is the one caller that sets `desiredFromAwsReadback: true`. Optional, so none of the 77 providers implementing `update()` changed.

**The field name is the part worth reviewing.** It says `desiredFromAwsReadback`, not `stateBorne` — the rollback executor's revert arms are state-borne too, but their desired bag is `previousState.properties` (a TEMPLATE recorded earlier), so `{Rules: []}` there means what the template meant and SKIP is correct. Those arms deliberately pass no context; the wider spelling would have swept them in and deleted a live configuration during a rollback.

**Mutation-probed in both directions**, which is what proves the two arms cannot be collapsed into one:

| mutation | fails |
|---|---|
| ignore the flag (skip always wins) | exactly the revert row |
| invert it (delete always wins) | exactly the 2 template rows |
| neutralize the skip guards | exactly the 2 template rows |

**Live-verified end to end on one binary**, both directions, on a real bucket:

- out-of-band `OwnershipControls` added → `cdkd drift` reports it → `--revert` **removes it** (I ran the pre-fix binary first to confirm that is what it did, so the regression is genuinely closed rather than assumed);
- collapsed-array redeploy → both skips announced → declared `aws:kms` default **survives**.

Destroyed clean, zero orphans. The (#1466) contract reversal described in the original description is unchanged and still wants your eye.

Issue #1806, opened 2026-08-13
fix(ecs): DeploymentConfiguration's Linear/Canary/LifecycleHook nested blocks have no required list, so a kept-but-partial block IS reachable and is unmeasured


go-to-k:
## The residual

The #1225 item-1 classification of `AWS::ECS::Service.DeploymentConfiguration` (PR #1805) settled the two TOP-LEVEL shapes as CloudFormation parity, and settled the depth-2 case for the two nested definitions that carry a `required` list — `DeploymentCircuitBreaker` [Enable, Rollback] and `DeploymentAlarms` [AlarmNames, Rollback, Enable]. For those, a kept-but-partial nested block is refused by CloudFormation, so the only exposure is the permissive-direction divergence filed as #1802.

**Three sibling nested blocks on the same property carry NO such refusal**, so a kept-but-partial block there is reachable from a template CloudFormation ACCEPTS (live registry schema, us-east-1, 2026-08-13):

```
{"k":"LinearConfiguration","required":null,"props":["StepBakeTimeInMinutes","StepPercent"]}
{"k":"CanaryConfiguration","required":null,"props":["CanaryBakeTimeInMinutes","CanaryPercent"]}
{"k":"DeploymentLifecycleHook","required":["LifecycleStages"],
 "props":["HookDetails","HookTargetArn","LifecycleStages","RoleArn","TargetType","TimeoutConfiguration"]}
```

So `LinearConfiguration` and `CanaryConfiguration` are fully optional, and `DeploymentLifecycleHook` leaves five of its six members droppable.

## Why that matters

The #1225 probe measured that `UpdateService` REPLACES a nested struct rather than merging it — a kept `DeploymentCircuitBreaker` missing `Rollback` had its live `rollback` flipped true -> false. If that replace behavior generalizes to these three blocks (likely, but NOT measured), then a template that keeps `LinearConfiguration` while dropping `StepBakeTimeInMinutes` silently resets it — the exact #1225 silent-drop class, on a shape CloudFormation permits, which is strictly worse than the #1802 case because there is no CFn-side refusal to point at.

This was NOT covered by the #1225 measurement: the probe used a ROLLING/REPLICA Fargate service, and `Strategy: BLUE_GREEN` is what makes `LinearConfiguration` / `CanaryConfiguration` / `LifecycleHooks` reachable at all. PR #1805's in-code comment was corrected to scope its claim to the top-level shapes rather than "every shape" after review caught the overreach.

## What to measure

[...]

Issue #1802, opened 2026-08-13
cdkd enforces no nested-definition `required` list, so it deploys templates CloudFormation refuses (measured: a partial DeploymentCircuitBreaker silently flips live rollback)


go-to-k:
## The divergence

CloudFormation validates a nested definition's `required` list before it touches AWS. cdkd validates nothing below the top level, so a template CloudFormation REFUSES is deployed by cdkd — and where the underlying API replaces the nested struct wholesale, the effect is a silent change to a live setting.

Measured us-east-1, 2026-08-13, on `AWS::ECS::Service.DeploymentConfiguration.DeploymentCircuitBreaker` (live baseline `{Enable: true, Rollback: true}`), template keeping the block but dropping `Rollback`:

| Engine | Result |
|---|---|
| CloudFormation | **refuses** — `Model validation failed (#/DeploymentConfiguration/DeploymentCircuitBreaker: required key [Rollback] not found)`, stack reaches `UPDATE_ROLLBACK_COMPLETE`, live config intact |
| cdkd | **deploys** — `UpdateService` accepts the partial, REPLACES the nested struct, and the live `rollback` goes true -> false |

The registry schema is explicit:

```
$ aws cloudformation describe-type --type RESOURCE --type-name AWS::ECS::Service \
    --query Schema --output text | jq -c '.definitions | to_entries[]
      | select(.key|test("DeploymentCircuitBreaker|DeploymentAlarms"))
      | {k:.key, required:.value.required}'
{"k":"DeploymentCircuitBreaker","required":["Enable","Rollback"]}
{"k":"DeploymentAlarms","required":["AlarmNames","Rollback","Enable"]}
```

## Why nothing catches it today

cdkd's pre-flight has two property-level checks and neither reads a definition's `required` list:

- `src/provisioning/property-coverage.ts` — TOP-LEVEL properties only ("each unhandled top-level key"), i.e. the silent-drop class, not required-ness.
- `src/provisioning/mutually-exclusive-properties.ts` — property COMBINATIONS AWS itself rejects; no `required` handling at all.

So the gap is structural rather than an oversight in one provider.

## Scope — how wide is this?

`AWS::ECS::Service` is where it was measured, but the shape is generic: any type with a nested definition carrying `required`, whose API replaces that struct wholesale rather than merging it. Worth enumerating from the registry schemas before choosing a fix. Note that `DeploymentAlarms` (the `Alarms` property) has the same exposure on the same type and was not separately probed.

[...]

Issue #1464, opened 2026-08-10
gen-nested-key-coverage: same-named members inside ONE top-level still vouch for each other — needs per-PATH fixture capture


go-to-k:
## Summary

`gen-nested-key-coverage`'s write-evidence pass now audits **paths** and scopes
evidence per object literal (#1448). That closed the cross-top-level case —
`BuildBatchConfig.ServiceRole` no longer gets vouched for by the top-level
`serviceRole:` write. But it did **not** close the same bug class *inside* one
top-level property, and it structurally cannot at the critic layer.

## Why the critic cannot close it

The fixture's `nestedProperties` capture is a **flattened transitive closure per
top-level property** — every nested name reachable under `Environment`, with the
chain discarded. So `Environment.Type` and
`Environment.EnvironmentVariables[].Type` are literally the **same audited
path**. The CFn side has already collapsed them before the critic runs; making
the write-side scope index deeper cannot re-separate them.

Measured on the real tree during #1448 (both are genuine silent drops that stay
silent):

```
# delete `type: ((environment?.['Type'] …))` from codebuild-provider.ts
$ node scripts/gen-nested-key-coverage.ts --check --providers-dir=<scratch>
nested-key-coverage: OK — 587 nested key paths ... 0 divergences    EXIT=0
   (covered by the surviving environmentVariables[].type write)

# delete `type: ((source['Type'] …))` from codebuild-provider.ts
   EXIT=0    (covered by the surviving auth.type write)
```

A second, related bound has the same root: `record()` keys the scope index by
NAME and unions across every write site of that name, so two unrelated
`environment: {…}` writes merge into one scope, and a purely nested `cache:`
write creates a `Cache` scope that vouches for the whole `Cache.*` family.

Both are documented as measured known-bounds in the script header's
`WHAT PATH-SCOPING DOES NOT CLOSE` section, in `.claude/rules/providers.md`,
and pinned by regression tests — so the bound is a recorded fact, not a
surprise. This issue is about actually closing it.

## What closing it requires

Per-PATH fixture capture:

1. Extend `scripts/refresh-cfn-schemas.mjs` to emit a `nestedPropertyPaths`
   capture alongside the existing flattened `nestedProperties` — full
   `Top.A.B` chains, `$ref`-resolved and cycle-guarded, the same walk the
   current capture does but without collapsing.
2. Re-capture every fixture under `tests/fixtures/cfn-schemas/` (needs AWS
[...]

Issue #1225, opened 2026-07-25
Audit: sub-field removal inside kept config objects (the #1160 bug class one level down)


go-to-k:
Surfaced by the PR #1224 reviews (test + spec reviewers), applies to every `clearOnUpdateRemoval` batch shipped so far (#1157 Lambda, #1164 ECS, #1222 RDS, #1224 ASG).

`clearOnUpdateRemoval` handles TOP-LEVEL property removal. When a config object is KEPT but one of its sub-fields is dropped (e.g. ASG `InstanceMaintenancePolicy` present in both old and new templates but `MaxHealthyPercentage` removed; `CapacityReservationSpecification.CapacityReservationTarget` dropped while the preference is kept), the provider passes the new object verbatim — and for APIs whose sub-shapes ALSO merge server-side (the ASG `-1`-to-clear sub-field sentinels imply exactly that), the dropped sub-field silently keeps its live value. This is the #1160 silent-drop bug class one level down, called out in docs/provider-development.md §2a ('Sub-structures can carry the same hazard one level down') but not yet audited per provider.

Suggested scope: for each shipped batch, classify every object-typed reset field's sub-shape as server-side merge vs whole-object replace (SDK doc or live probe), and where merge semantics are confirmed, normalize kept-but-partial objects to the explicit clear shape (e.g. fill `-1` for dropped InstanceMaintenancePolicy sub-fields). Unit-test the kept-object-with-dropped-sub-field shape per §2a's 'per-key removal does NOT cover whole-block removal — test both' rule (and vice versa).

go-to-k:
[...]

PR #1832, opened 2026-08-13
test(schemas): capture each CFn definition's required list so a required-ness-dependent parity verdict can be fenced

### LINKED ISSUE #1800: refresh-cfn-schemas does not capture each definition's `required` list, so a required-ness-dependent parity verdict cannot be fenced
## The gap

`tests/fixtures/cfn-schemas/*.json` (written by `scripts/refresh-cfn-schemas.mjs`) captures `properties`, `readOnlyProperties`, `createOnlyProperties`, `primaryIdentifier`, `nestedProperties`, `nestedPropertyPaths` and `definitionShapes` — but **not** each definition's `required` list, which the live registry schema does carry:

```
$ aws cloudformation describe-type --type RESOURCE --type-name AWS::ECS::Service \
    --query Schema --output text | jq -c '.definitions.DeploymentCircuitBreaker | {required, props: (.properties|keys)}'
{"required":["Enable","Rollback"],"props":["Enable","ResetOnHealthyTask","Rollback","ThresholdConfiguration"]}
```

## Why it matters (the case that surfaced it)

The #1225 item-1 classification of `AWS::ECS::Service.DeploymentConfiguration` concluded that cdkd's verbatim pass-through is CloudFormation parity at every depth. The depth-2 arm of that verdict rests **entirely on required-ness**: a kept `DeploymentCircuitBreaker` missing `Rollback` has its `rollback` REPLACED true -> false by `UpdateService`, which would be a real silent drop — except that the shape is unreachable from a valid template, because CloudFormation refuses it:

```
Model validation failed (#/DeploymentConfiguration/DeploymentCircuitBreaker: required key [Rollback] not found)
```

(measured us-east-1, 2026-08-13, `UPDATE_ROLLBACK_COMPLETE` with the live config intact).

So if AWS ever relaxes that `required` list, the depth-2 REPLACE becomes reachable and the parity verdict silently becomes wrong — a real drop, with no test failing. The PR closing #1225 item 1 tried to fence exactly this and **could not**: with no `required` in the fixture, the best available assertion is that the member paths are still modelled, which stays true under a relaxed `required` list. That test now states its own limitation in-code and points here rather than over-claiming.

The same reasoning generalizes past ECS. `.claude/rules/providers.md` already tells provider authors to "check the nested definition's `required` list before treating a depth-2 replace as reachable at all" — advice no checker can currently verify, because the data is not captured.

## Suggested direction

1. Capture `required` per definition in `extractDefinitionShapes` / alongside `definitionShapes` in `scripts/refresh-cfn-schemas.mjs` (additive; a re-capture rewrites `generatedAt` and adds the section, leaving `properties` / `readOnlyProperties` / `createOnlyProperties` untouched — the same additive shape #1699 documents for `nestedPropertyPaths`).
2. Re-run `vp run gen:all-matrices` afterwards, per the #1694 note that `primaryIdentifier` and friends feed `gen-enrichment-coverage`.
3. Upgrade the ECS test at `tests/unit/provisioning/ecs-deployment-configuration-subfield.test.ts` (the "still modelled" test, which carries a `guard the guard` assertion that FAILS the day the fixture starts carrying `required`, so this issue cannot be silently forgotten) to assert the actual required lists.
4. Consider whether the nested-key critic should surface a required-ness-dependent verdict at all, since a "parity by loud reject" classification is only as durable as the schema it rests on.

## Scope note

Deliberately NOT done in the #1225 PR: that lane was a no-behavior-change classification, and re-capturing every fixture plus touching the shared schema generator is a different blast radius with its own review needs.

go-to-k:
Closes #1800

## The gap

`tests/fixtures/cfn-schemas/*.json` captured `properties` / `readOnlyProperties`
/ `createOnlyProperties` / `primaryIdentifier` / `nestedProperties` /
`nestedPropertyPaths` / `definitionShapes` — but **not** each definition's
`required` list, which the live registry schema does carry.

That matters because a whole class of parity verdicts rests entirely on
required-ness. The [#1225](https://github.com/go-to-k/cdkd/issues/1225) classification of
`AWS::ECS::Service.DeploymentConfiguration` concluded that cdkd's verbatim
pass-through is CloudFormation parity at depth 2 **only** because CFn refuses a
kept-but-partial `DeploymentCircuitBreaker`:

```
Model validation failed (#/DeploymentConfiguration/DeploymentCircuitBreaker:
  required key [Rollback] not found)
```

`UpdateService` REPLACES the nested struct rather than merging it (measured
live — a kept block missing `Rollback` flipped a live `rollback` true -> false),
so the ONLY thing making that shape unreachable is CFn's own model validation.
If AWS ever relaxes that list, the depth-2 replace becomes reachable and the
parity verdict silently becomes wrong, **with no test failing**.

The PR closing [#1225](https://github.com/go-to-k/cdkd/issues/1225) item 1 tried to fence exactly this and could not: with no
`required` in the fixture, the strongest available assertion was that the member
paths are still MODELLED, which stays true under a relaxed list. It stated that
limit in-code and carried a `guard the guard` assertion designed to RED the day
the fixture started carrying `required`.

## What shipped

**The capture.** `extractDefinitionRequired` in
`scripts/refresh-cfn-schemas.mjs` emits a `definitionRequired` section: per
definition, its `required` list (sorted), with the schema's top-level list under
the same reserved `#top` key `definitionShapes` uses.

A definition with no `required` list, or an empty one, gets **no entry** — the
absence IS the fact ("nothing is required here"), and emitting `[]` for the ~90%
of definitions in that state would bloat every fixture for no signal. Consumers
read a missing key as "nothing required". (The section's ABSENCE is NOT a
not-captured marker — see the review round below; `generatedAt` is the
discriminator.) Same omit-when-empty rule as the two sibling sections, so a type
requiring nothing anywhere keeps its prior byte shape.

All 134 fixtures were re-captured live (us-east-1, 2026-08-13), and
`vp run gen:all-matrices` re-run afterwards — 0 gaps across every critic.

**The fence it unblocks.** The `guard the guard` assertion fired as designed.
`tests/unit/provisioning/ecs-deployment-configuration-subfield.test.ts` now
asserts the ACTUAL lists — `DeploymentCircuitBreaker` [Enable, Rollback],
`DeploymentAlarms` [AlarmNames, Enable, Rollback], `DeploymentLifecycleHook`
[LifecycleStages] — with exact `toEqual`, because a member DROPPED from a
required list is precisely the relaxation being fenced and `toContain` cannot
see a removal. It also asserts the ABSENCE of a list on `LinearConfiguration` /
`CanaryConfiguration`, which SCOPES the parity verdict rather than over-claiming
it and keeps [#1806](https://github.com/go-to-k/cdkd/issues/1806) visible — those two are fully optional, so a kept-but-partial
block there IS reachable from a template CFn accepts. A non-vacuity guard pins
that the section exists and is populated, so a capture regression fails there
rather than silently downgrading the test to the shape it replaced.

## Two things found while doing it

**1. The re-capture defused a fence, exactly as that fence predicted.** The
`definitionShapes` loud-failure probe picked its stand-in DYNAMICALLY from
fixtures predating the shape capture; once every fixture had one it had nothing
to select and failed with its own "no pre-shape-capture fixture left to probe
with" message. Its comment prescribed adding a fixture-dir seam — one already
existed (`loadReport`'s fourth parameter, added by [#1464](https://github.com/go-to-k/cdkd/issues/1464) for the sibling
`nestedPropertyPaths` fence), so the probe now builds its own partially captured
tree and is immune to the next re-capture. That is the property the dynamic pick
was reaching for and could not actually hold.

**2. The route-flip side effect was MEASURED, not assumed.** A fixture stale
since 2026-05-16 gains whatever AWS has added since, and a new top-level
property with no provider wiring becomes a `silentDrop` entry — which does not
reject the deploy but AUTO-ROUTES the resource through Cloud Control (the [#614](https://github.com/go-to-k/cdkd/issues/614)
rule). Nine appeared:

| type | property |
|---|---|
| `AWS::CloudWatch::Alarm` | `EvaluationWindow` |
| `AWS::Cognito::UserPool` | `IssuerConfiguration`, `KeyConfiguration` |
| `AWS::EC2::Route` | `OdbNetworkArn` |
| `AWS::EC2::VPC` | `VpcEncryptionControl` |
| `AWS::ElasticLoadBalancingV2::Listener` | `Tags` |
| `AWS::Neptune::DBCluster` | `GlobalClusterIdentifier`, `NetworkType` |
| `AWS::WAFv2::WebACL` | `MonetizationConfig` |

`Listener.Tags` looked like a live route flip for the `alb` / `alb-advanced`
fixtures, both of which apply stack-level `cdk.Tags.of(this)`. It is not: each
of the nine was checked individually against its `aws-cdk-lib` 2.244.0 L1 props
interface and **all nine are absent**, so they are CFn-schema-only properties
ahead of CDK support and no CDK app can emit one — the same disposition the
pre-existing hand-written `EvaluationInterval` entry records. A hand-authored L1
using `addPropertyOverride` can still reach them, and the CC auto-route is the
correct answer there.

**3. The re-capture surfaced three things the coverage critics are built to
catch, and the full suite is what found them** (the targeted runs were green).

- **Nine unaccounted properties** — the table above. Each now has an entry in
  `tests/fixtures/cfn-schemas/_todo-backfill.json`, the ledger that exists for
  exactly this ("allowed only for incremental backfill — must be migrated to
  (a) or (b) eventually"), regenerated with the documented
  `CDKD_GENERATE_BACKFILL=true`. The umbrella backfill campaign is [#609](https://github.com/go-to-k/cdkd/issues/609).
- **That regeneration also CLEANED stale entries** — `AWS::ApiGateway::Stage`
  entirely, plus `LaunchConfigurationName`, `NotificationConfiguration`,
  `AvailabilityZone`, `ElasticGpuSpecifications`, `ElasticInferenceAccelerators`,
  `EvaluationCriteria`, `EvaluationInterval` and others. These are NOT schema
  removals (checked — every one is still in the fixture): they are properties
  the providers already account for via `handledProperties` /
  `unhandledByDesign`, which the ledger kept listing because retiring an entry
  is a manual step its own comment describes and people forgot. Spot-verified in
  `cloudwatch-alarm-provider.ts`, `apigateway-provider.ts` and `asg-provider.ts`.
- **Two REAL latent bugs** — AWS added an ARN read-only attribute to
  `AWS::RDS::DBSubnetGroup` (`DBSubnetGroupArn`) and `AWS::SSM::Parameter`
  (`Arn`), and neither provider caches it, so an output `Fn::GetAtt` on either
  hard-fails the resolver's `*Arn` shape guard (the [#1179](https://github.com/go-to-k/cdkd/issues/1179) class). Both are
  recorded as `KNOWN GAP` entries in `SDK_ATTR_ALLOW_LIST` pointing at the
  filed issue 1824 — the same disposition
  `AWS::Lambda::EventSourceMapping.EventSourceMappingArn` had under [#1190](https://github.com/go-to-k/cdkd/issues/1190) — so
  the critic is green while the gaps stay VISIBLE and deleting each entry is
  what verifies the eventual fix. They are not fixed here because that means
  editing `src/provisioning/providers/**`, which activates the `integ-destroy`
  gate and drags a real-AWS integ run behind a no-behavior-change capture PR.

## Review round (second commit)

A 3-axis review found no blockers, and independently re-verified the load-bearing
claims (a live re-extraction of the ECS schema matched the committed
`definitionRequired` byte-for-byte; all 59 backfill removals confirmed still
present in the re-captured `properties`; the two ARN rationales checked against
the real SDK response shapes). What it did find were **two claims this PR made
that were false**, plus a fence that then caught a third thing:

- **"a stale entry fails the critic in both modes, which is what verifies it"**
  — false. `classifyType` tests `cachedKeys` BEFORE the allow-list, so the
  moment issue 1824 lands the entry silently classifies `cached` and goes
  inert; there is no `findStaleAllowListEntries` here the way there is in
  `gen-nested-key-coverage.ts`. A test now asserts every allow-listed attribute
  still classifies `allow-listed`, which is what makes the "DELETE this entry
  when it is fixed" note enforceable rather than aspirational.
- **That fence immediately retired `AWS::SNS::Subscription`.** Its entry is now
  provably inert — and this PR is what made it so: the fixture predated the
  [#1694](https://github.com/go-to-k/cdkd/issues/1694) `primaryIdentifier` capture, so `Arn` used to reach the allow-list, and
  once the re-capture gave it `primaryIdentifier: ['Arn']` the classifier
  filters the attribute out first. That is exactly the "auto-classify those as
  not-a-gap instead of requiring a hand-written entry per type" behavior
  `extractPrimaryIdentifier` exists for, so the entry is deleted rather than
  kept.
- **"the section's presence is what marks the fixture as captured"** — false.
  Seven types legitimately require nothing anywhere and carry no
  `definitionRequired` at all; `generatedAt` is the discriminator.
- **The contract was too strong in a way that mattered.** Only a definition's
  OWN top-level `required` array is captured. Required-ness expressed through a
  `oneOf` combinator (`AWS::S3::Bucket`'s `TargetObjectKeyFormat`) or on an
  inline nested object (`AWS::WAFv2::WebACL`'s `FieldToMatch.SingleHeader`) is
  absent — both measured live — so for those shapes a missing entry means
  UNKNOWN, not "CloudFormation requires nothing". A consumer reading it the
  strong way would derive precisely the false "a kept-but-partial block is
  reachable" verdict this capture exists to prevent. Narrowed in the JSDoc, in
  `docs/provider-development.md` and in `.claude/rules/code-layout.md`, and
  pinned by a test so the bound cannot be silently widened.

Also: the summary reported `gap: 0` for two types whose `Fn::GetAtt` hard-fails,
because a NOT-A-BUG entry and a tracked real gap shared one bucket —
`AllowListEntry` gained a `knownGap` flag and the summary counts those
separately. The two [#1806](https://github.com/go-to-k/cdkd/issues/1806) scope absences now also pin that the definitions still
EXIST (a `toBeUndefined()` passes just as well if AWS renames them). The
non-vacuity rationale on the ECS guard was inverted and is corrected. A
fixture-level pin was added for the omit-when-empty contract.

## Testing

- 3 new unit tests for the extractor (sorted lists, `#top`, the omit-when-empty
  branch, non-string members, a non-array `required`).
- The ECS depth-2 fence upgraded from "still modelled" to the actual lists.
- The `definitionShapes` probe repointed at the seam, with a non-vacuity guard
  that the section it deletes was actually there.
- Full suite + `vp run check` + `vp run typecheck:test` + build green.

## Scope

No behavior change: no runtime code reads the new `definitionRequired` section.
No CLI flag, dependency, or state-schema change. The follow-on work the capture
enables — a required-ness-aware pre-flight (#1802) and the `LinearConfiguration`
/ `CanaryConfiguration` measurement (#1806) — is deliberately NOT in this PR.


go-to-k:
## `pr-review` marker re-bound after a content-neutral rebase

The 3-axis review for this PR completed against sha `0605e03d`. Main then
advanced (five PRs merged during this session), so the branch was rebased and
HEAD moved to `018d66f8`, which invalidated the sha-bound `pr-review` marker.

Rather than assume the rebase was harmless, the two patches were compared
directly:

```
git diff 0605e03d^ 0605e03d  > reviewed.diff     # my patch, as reviewed
git diff 018d66f8^ 018d66f8  > now.diff          # my patch, at current HEAD

# added/removed lines only (excluding +++/--- headers):
18304 reviewed_content.txt
18304 now_content.txt
diff  -> IDENTICAL
```

Every one of the 18,304 added/removed lines is byte-identical. The only
differences between the two patch files are blob index hashes, hunk line
offsets, and context lines belonging to main's own new content — i.e. exactly
the artifacts a rebase produces and nothing else. The reviewed content and the
merging content are the same content.

The rebase itself was verified separately rather than trusted: the only
conflict was `docs/changelog-cdkd.md` (the usual top-of-file append collision,
resolved keeping both sides), and `vp run gen:all-matrices` +
`vp run audit:coverage:check` were re-run afterwards against main's new
provider ARN changes and came back clean.

Marker re-bound to `018d66f8` with the maintainer's explicit go-ahead, per the
escape hatch in `.claude/hooks/pr-review-gate.sh`. Recorded here so the
override is visible and challengeable rather than silent.

Review coverage that applies unchanged:
- `pr-spec-reviewer` vs issue 1800's four prescribed items — no blockers; the
  additivity claim was independently measured across all 134 fixtures.
- `pr-code-reviewer` — 2 should-fix (a false in-code claim about allow-list
  staleness detection; the "section presence marks a captured fixture" claim),
  both fixed.
- `pr-test-reviewer` — 2 should-fix (combinator/inline-object `required` shapes
  silently dropped; an inverted non-vacuity rationale), both fixed.

CI on `018d66f8`: all 6 checks pass.

PR #1868, opened 2026-08-13
fix(provisioning): preserve Route 53 NameServers as a list

minorun365:
## Summary

- preserve `AWS::Route53::HostedZone.NameServers` as a list across create, update, and on-demand attribute lookup
- normalize the legacy comma-delimited state shape at the intrinsic-resolution boundary so existing stacks do not require recreation or manual state edits
- cover `Fn::Join` with unit tests and a live Route 53 integration assertion against the actual delegation set

## Backward compatibility

Earlier cdkd versions stored `NameServers` as a comma-delimited string. The resolver converts only the legacy `AWS::Route53::HostedZone.NameServers` value back to a list when it is read. This avoids a state schema bump and leaves unrelated string-valued attributes unchanged.

## Verification

- post-rebase `vp check`, test typecheck, and build passed
- post-rebase targeted tests: 292 passed
- full suite before the final upstream rebase: 645 test files / 12,796 tests passed
- Route 53 live integration: 8 created, 8 deleted, 0 errors, 0 orphans
- broad `microservices` integration: 19 created, 19 deleted, 0 errors, 0 orphans

---

## Maintainer notes (added at merge time)

Reviewed with a code-quality and a test-adequacy agent; no blockers. All four
shipped source lines were mutation-probed and each is fenced by a failing test.
Re-verified on the maintainer account against this branch's HEAD: `route53`
verify.sh PASS (8 deleted, 0 errors, 0 orphans, 349s) with the joined Output
matching the live `GetHostedZone` delegation set, and the broad `microservices`
integ PASS (19 created / 19 deleted, 0 errors, 0 orphans) for the shared
`intrinsic-function-resolver.ts` change.

Merging as-is. The review's minor findings are filed and will be fixed in
separate maintainer PRs rather than sent back here:

- go-to-k/cdkd#1873 — test-strength gaps in this change: the legacy-split
  assertion is tautological (joining with the same `,` it splits on), neither
  guard on the normalization block has a negative control, the empty-string
  edge is claimed in the changelog but untested, and two `verify.sh`
  robustness nits.
- go-to-k/cdkd#1874 — `Fn::Split` refuses a list-valued `Fn::GetAtt`, so the
  workaround someone hit by this bug would have written breaks after upgrading.
- go-to-k/cdkd#1875 — pre-existing and not caused by this PR: an IMPORTED
  hosted zone still cannot resolve `NameServers`, because `importHostedZone`
  records no attributes and `constructAttribute` has no case for the type.
- go-to-k/cdkd#1876 — `CLAUDE.md`'s `StackState` snippet says
  `outputs: Record<string, string>` where the real type is
  `Record<string, unknown>`.


go-to-k:
@minorun365 Thank you so much!

PR #1890, opened 2026-08-13
fix(provisioning): skip ZIP-only Lambda reads for image functions

minorun365:
## Summary

- skip `GetRuntimeManagementConfig` and `GetFunctionCodeSigningConfig` when `GetFunction` reports `PackageType: Image`
- keep the existing readback behavior unchanged for ZIP functions
- add regression coverage verifying both unsupported API calls are omitted while image code metadata is preserved

## Why

AWS Lambda container-image functions manage their runtime in the image and do not support Lambda code signing. Calling these ZIP-only getter APIs during drift readback produces warnings and can make missing values look like removals.

## Verification

- post-rebase `vp run check` passed with 0 errors and 6 pre-existing warnings
- post-rebase test typecheck and build passed
- targeted Lambda property-backfill suite: 32 passed
- full suite before the final upstream rebase: 659 test files / 13,274 tests passed
- generated coverage matrices and provider coverage audit passed
- live `docker-image-asset` integration was attempted but could not reach Lambda creation because the existing sandbox CDK image-publishing role lacks `ecr:GetDownloadUrlForLayer`; cleanup confirmed 0 state, IAM, Lambda, and ECR orphans

---

## Maintainer verification (2026-08-14)

The live run the contributor could not complete was run here, and it settles the premise this change rests on. Against a real `PackageType: Image` function deployed by the `docker-image-asset` fixture (us-east-1):

```
GetRuntimeManagementConfig   -> InvalidParameterValueException:
  "Lambda couldn't get a runtime management configuration because <fn> is a container image function."
GetFunctionCodeSigningConfig -> InvalidParameterValueException:
  "Code signing is not supported for functions created with container images."
```

Both rejections are non-`ResourceNotFoundException`, so the pre-PR code took the `logger.warn` arm and emitted NEITHER key for an image function. That matters beyond the warning noise: a change that stops `readCurrentState` emitting a key normally manufactures a one-sided REMOVAL drift against every `observedProperties` baseline the old binary wrote, and would need a `getDriftUnknownPaths` companion. Because AWS always rejected here, no baseline can carry either key, and this skip is invisible to drift — no companion needed.

- `/run-integ docker-image-asset`: PASS, 73s — deploy + ECR push + invoke + destroy (2 deleted, 0 errors), 0 orphans, state gone.
- Deploy-time `readCurrentState` capture logged zero `GetRuntimeManagementConfig failed` / `GetFunctionCodeSigningConfig failed` warnings on the new binary (the branch under test ran).
- Full suite 659 files / 13,274 tests; `handled-property-wiring` audit still reports read evidence for both properties (1136/1138, 0 gaps) — the conditional wrap did not sever the evidence walk.
- Code review: 1-reviewer tier (`src/provisioning/providers/**` up-bias). Mutation probe confirms the new test is RED without the guard.

Minor follow-ups are tracked in #1894 (write-side asymmetry, comment rationale, two vacuous assertions, missing integ coverage of the branch, changelog entry) rather than being pushed onto this branch.

### DIFF
--- a/src/provisioning/providers/lambda-function-provider.ts
+++ b/src/provisioning/providers/lambda-function-provider.ts
@@ -1829,54 +1829,62 @@ export class LambdaFunctionProvider implements ResourceProvider {
         }
       }
 
-      // RuntimeManagementConfig: a SEPARATE control-plane read
-      // (GetRuntimeManagementConfig), emit-when-present. AWS returns the
-      // default `UpdateRuntimeOn: 'Auto'` for functions that never set it;
-      // the comparator's state-keys-only walk ignores the field unless state
-      // carries it, so the always-emit shape produces no phantom drift.
-      try {
-        const rmResp = await this.lambdaClient.send(
-          new GetRuntimeManagementConfigCommand({ FunctionName: physicalId })
-        );
-        if (rmResp.UpdateRuntimeOn !== undefined) {
-          const rm: Record<string, unknown> = { UpdateRuntimeOn: rmResp.UpdateRuntimeOn };
-          // Only meaningful under Manual; AWS omits it otherwise.
-          if (rmResp.RuntimeVersionArn !== undefined) {
-            rm['RuntimeVersionArn'] = rmResp.RuntimeVersionArn;
+      // Runtime-management controls and code signing apply to managed-runtime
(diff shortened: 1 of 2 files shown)


go-to-k:
Thanks!

PR #1970, opened 2026-08-18
chore(hooks): let a foreign repo's own gate decide its own commands

### LINKED ISSUE #1961: hooks: a cdkd session judges another gate-equipped repo's commands with cdkd's policy
A session whose primary working directory is cdkd loads cdkd's `.claude/settings.json` hooks, and those hooks then fire on **every** Bash call in the session — including commands whose target is a different repository. When that other repository is also gate-equipped, cdkd's copy of a gate adjudicates it using **cdkd's** policy rather than the target repo's.

Observed twice today while landing the markgate `hash: diff` adoption across three repos:

- A `gh pr create` for go-to-k/cdk-real-drift#1757 was refused by `verify-pr-gate`. Feeding the same payload to both copies settled it: cdk-real-drift's own hook answered `verify-pr-gate: PR diff touches no src/** (docs/tooling-only) — exempt` and exited 0, while **cdkd's** copy exited 2. cdkd's hook has no docs/tooling exemption (`grep -c "docs/tooling-only" .claude/hooks/verify-pr-gate.sh` -> 0).
- The same thing recurred on go-to-k/cdk-real-drift#1759, whose diff is documentation only.

Both times the agent correctly refused to route around the block and instead completed cdk-real-drift's own `/verify-pr` checklist by hand, which is the right behavior but is pure waste: the target repo's policy had already exempted the change.

## Why this is not #1259

#1259 scopes the main-tree / branch hooks to repos that opt in by carrying a `.markgate.yml`, so they stop firing in unrelated personal repos. That fix does not apply here: **cdk-real-drift carries `.markgate.yml`**, so it passes the opt-in check and the hook proceeds — just with the wrong repo's rules. The two are complementary. #1259 asks "should this hook run at all?"; this asks "when it runs against another repo, whose policy applies?".

## Shape of the problem

The gates already resolve the TARGET working tree correctly — that is what #559 fixed, and it is why `markgate verify` consults the right per-worktree marker store. What does not follow the target is the hook SCRIPT: it is whichever copy the session loaded at startup. So the marker lookup is target-correct while the policy around it is session-correct, and the two disagree exactly when the repos' gate scripts have diverged — which they have, since each repo's gates evolved separately.

## Options, none obviously right

1. **Early-exit when the resolved target repo is not the session's own repo.** Simple and removes the wrong-policy case entirely, but it also removes the gate: a cdkd session doing genuine cdkd work via `gh -C` would go unguarded. Probably needs pairing with a loud stderr note so the absence of a gate is visible.
2. **Delegate to the target repo's copy of the same-named hook** when one exists, exec'ing it with the original payload. Correct in principle; means a hook must be safe to invoke from a foreign repo's session, and silently does nothing when the target has no counterpart.
3. **Leave it and document** that cross-repo work from a cdkd session inherits cdkd's policy, so agents stop treating a foreign block as a bug in their own repo. Cheapest, but it keeps paying the friction on every multi-repo session, and today it cost two hand-run verification cycles.

Worth deciding before the next multi-repo campaign rather than during one.

Session-fit: next (not this session) — it is a hooks design decision affecting all 35 gates, needs its own verification cycle across at least two repos, and the sibling adoption work it surfaced during has already landed / Effort: ~1-3 h (decision + one hook changed as a pilot + its smoke suite, run under both bash 5 and /bin/bash 3.2)

go-to-k:
## Summary

A session's hooks come from one repo's `.claude/settings.json` and fire on **every** Bash call — including commands whose target is a different repository. Issue 559 made the gates resolve the target working tree, so the marker lookup is target-correct; the hook SCRIPT is still ours, so the policy around it is session-correct. Those disagree exactly when the repos' gates have diverged, and they have: 35 gates here, 17 in cdk-local, 10 in cdk-real-drift, each evolved separately.

`lib/foreign-repo.sh` adds `hook_delegate_if_foreign`, called by all 24 gates that resolve a target dir, right after resolution completes. A foreign target that ships its own copy of the gate gets the payload piped to it, and its exit status becomes ours.

Closes #1961.

## What went wrong twice

While landing the markgate `hash: diff` adoption across three repos on 2026-08-18, a `gh pr create` for cdk-real-drift was refused by **this repo's** copy of `verify-pr-gate`. Feeding one payload to both copies settled it:

| copy | verdict |
|---|---|
| cdk-real-drift's own | `PR diff touches no src/** (docs/tooling-only) — exempt`, exit 0 |
| ours (no such exemption) | exit 2 |

Both times the agent correctly refused to route around the block and hand-ran the target repo's checklist instead. Right behavior, pure waste: the target's own policy had already exempted the change.

## Why delegation and not an early exit

"Exit 0 when the target is foreign" is the obvious fix and it is wrong, for a reason that is easy to miss and that I got wrong in the issue body before checking: **the foreign repo's hooks are not registered in this session**. Silencing our copy leaves the command with no gate at all — on both incidents that would have skipped the `verify-pr` requirement rather than applying the wrong one. Strictly weaker than the status quo.

Delegation is the only shape that is never weaker on any path: the target's policy when a counterpart exists, today's behaviour plus a stderr note when it does not.

## Identity is the git common dir, not the toplevel

Two worktrees of one repo have different toplevels and one common dir, and gate work legitimately runs from `.claude/worktrees/<branch>/` against the main tree. A toplevel comparison would have called that foreign and delegated to the same repo's own hook for nothing. Pinned by a test using a real `git worktree`.

## Failure directions all fall back, never pass

| situation | result |
|---|---|
| target repo unknown (not a repo, missing dir) | caller's own policy |
| target has no counterpart for this gate | caller's own policy + stderr note |
| counterpart not executable | caller's own policy |
| counterpart unrunnable (exit 126/127 — "could not run", not a verdict) | caller's own policy |
| library missing or unloadable | exactly the pre-change behaviour |
| counterpart also carries the helper | `CDKD_HOOK_DELEGATED=1` stops a second hop |

The last two matter most: a gate must never open because the helper broke, and a delegation loop must not be reachable.

## Test plan

`.claude/hooks/lib/foreign-repo.test.sh` — 13 cases against real throwaway repos and a real `git worktree`. Git is not mocked: the whole decision is what `git rev-parse --git-common-dir` reports, so a mock would only be testing the mock. Covers the linked-worktree case, the payload arriving intact, a blocking verdict propagating rather than being swallowed, and each unrunnable-counterpart shape.

Full hook suite green under **both** bashes — `bash 5.3.9` and `/bin/bash 3.2.57` (the macOS system bash the gates actually run under on a bare machine), 33 suites including the new one. `vp run check` / `typecheck:test` / `build` clean; `vp run test` 685 files, 13931 tests, no type errors.

## Review note

24 safety gates change at once, so the diff is mechanical but the blast radius is not: a mistake here can make a gate silently not fire, which is the failure mode the whole gate family exists to prevent. The insertion is identical in every gate and sits immediately after target-dir resolution; the interesting file is `lib/foreign-repo.sh`.

### DIFF
--- a/.claude/hooks/branch-gate.sh
+++ b/.claude/hooks/branch-gate.sh
@@ -109,6 +109,21 @@ if [[ "$cmd" =~ git[[:space:]]+-C[[:space:]]+([^[:space:]]+) ]]; then
   target_dir="$c_target"
 fi
 
+# Cross-repo delegation (issue 1961). The hooks a session runs come from
+# ONE repo's settings, but they fire on every Bash call -- including
+# commands whose target is a different repository. When that target ships
+# its own copy of this gate, that copy decides; ours must not impose a
+# policy the target never adopted. Falls back to this repo's policy (with
+# a stderr note) when the target has no counterpart, and degrades to
+# exactly today's behaviour if the library is missing -- never weaker.
+__fr_dir="${BASH_SOURCE[0]%/*}"
+[ "$__fr_dir" = "${BASH_SOURCE[0]}" ] && __fr_dir="."
+# shellcheck source=lib/foreign-repo.sh
+if . "$__fr_dir/lib/foreign-repo.sh" 2>/dev/null \
+   && declare -F hook_delegate_if_foreign >/dev/null; then
+  hook_delegate_if_foreign "$target_dir" "${BASH_SOURCE[0]##*/}" "$input" || true
+fi
+
 # Repo opt-in scope (issue #1259): this gate protects repos that follow
 # the feature-branch + PR + markgate convention. A session rooted in
 # such a repo can still run git against OTHER repos (a personal blog, a
--- a/.claude/hooks/bughunt-clean-gate.sh
+++ b/.claude/hooks/bughunt-clean-gate.sh
@@ -91,6 +91,21 @@ if [[ -n "$cd_target" ]]; then
   target_dir="$cd_target"
 fi
 
+# Cross-repo delegation (issue 1961). The hooks a session runs come from
+# ONE repo's settings, but they fire on every Bash call -- including
+# commands whose target is a different repository. When that target ships
+# its own copy of this gate, that copy decides; ours must not impose a
+# policy the target never adopted. Falls back to this repo's policy (with
+# a stderr note) when the target has no counterpart, and degrades to
+# exactly today's behaviour if the library is missing -- never weaker.
+__fr_dir="${BASH_SOURCE[0]%/*}"
+[ "$__fr_dir" = "${BASH_SOURCE[0]}" ] && __fr_dir="."
+# shellcheck source=lib/foreign-repo.sh
+if . "$__fr_dir/lib/foreign-repo.sh" 2>/dev/null \
+   && declare -F hook_delegate_if_foreign >/dev/null; then
+  hook_delegate_if_foreign "$target_dir" "${BASH_SOURCE[0]##*/}" "$input" || true
+fi
+
 # Not a git repo (or git unavailable) → nothing to gate.
 git -C "$target_dir" rev-parse --git-dir >/dev/null 2>&1 || exit 0
 
--- a/.claude/hooks/check-gate.sh
+++ b/.claude/hooks/check-gate.sh
@@ -95,6 +95,21 @@ if [[ "$cmd" =~ git[[:space:]]+-C[[:space:]]+([^[:space:]]+) ]]; then
   target_dir="$c_target"
 fi
 
+# Cross-repo delegation (issue 1961). The hooks a session runs come from
+# ONE repo's settings, but they fire on every Bash call -- including
+# commands whose target is a different repository. When that target ships
+# its own copy of this gate, that copy decides; ours must not impose a
+# policy the target never adopted. Falls back to this repo's policy (with
+# a stderr note) when the target has no counterpart, and degrades to
+# exactly today's behaviour if the library is missing -- never weaker.
+__fr_dir="${BASH_SOURCE[0]%/*}"
+[ "$__fr_dir" = "${BASH_SOURCE[0]}" ] && __fr_dir="."
+# shellcheck source=lib/foreign-repo.sh
+if . "$__fr_dir/lib/foreign-repo.sh" 2>/dev/null \
+   && declare -F hook_delegate_if_foreign >/dev/null; then
+  hook_delegate_if_foreign "$target_dir" "${BASH_SOURCE[0]##*/}" "$input" || true
+fi
+
 # If the resolved target dir is not a git repo, silently pass — we
 # can't audit what we can't see (mirrors branch-gate.sh).
 if ! git -C "$target_dir" rev-parse --git-dir >/dev/null 2>&1; then
--- a/.claude/hooks/ci-green-gate.sh
+++ b/.claude/hooks/ci-green-gate.sh
@@ -99,6 +99,21 @@ if [[ "$cmd" =~ gh[[:space:]]+-C[[:space:]]+([^[:space:]]+) ]]; then
   target_dir="$c_target"
 fi
 
+# Cross-repo delegation (issue 1961). The hooks a session runs come from
+# ONE repo's settings, but they fire on every Bash call -- including
+# commands whose target is a different repository. When that target ships
+# its own copy of this gate, that copy decides; ours must not impose a
+# policy the target never adopted. Falls back to this repo's policy (with
+# a stderr note) when the target has no counterpart, and degrades to
+# exactly today's behaviour if the library is missing -- never weaker.
+__fr_dir="${BASH_SOURCE[0]%/*}"
+[ "$__fr_dir" = "${BASH_SOURCE[0]}" ] && __fr_dir="."
+# shellcheck source=lib/foreign-repo.sh
+if . "$__fr_dir/lib/foreign-repo.sh" 2>/dev/null \
+   && declare -F hook_delegate_if_foreign >/dev/null; then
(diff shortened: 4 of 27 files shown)


go-to-k:
## Converted to draft — security blocker, independently reproduced

A security review found that this PR introduces **arbitrary code execution from an untrusted checkout**, and I reproduced it before acting on the report.

`hook_delegate_if_foreign` execs `<target-repo>/.claude/hooks/<gate-name>` with no trust check on that repo. The target directory is named by the command itself — a `cd`, a `git -C` / `gh -C` flag, or the payload's `cwd` — so any directory the agent can be induced to touch, that happens to contain an executable file at that path, gets it run with the session's full environment.

Reproduced: a throwaway git repo with a planted `.claude/hooks/main-tree-branch-gate.sh`, and a payload no more exotic than `{"cwd":"<that repo>","tool_input":{"command":"git checkout feature-x"}}`. The planted script ran, counted 3 inherited `AWS_*` / `GH_TOKEN`-shaped environment variables, captured the full payload from stdin, and the gate exited 0.

The trigger set is every common git write verb — `checkout` / `switch`, `commit`, `push`, `stash`, `reset --hard`, `clean -f`, `restore`, `merge`, `gh pr create` / `merge` — and the gate filenames are public. "Clone this repro repo and check out my branch" therefore becomes code execution with session credentials, which is precisely the class this repo's rules forbid, on a machine whose maintainer holds AWS credentials.

Two things compound it:

- Delegation is placed **before** the repo opt-in check that issue 1259 added, in `branch-gate.sh` and `main-tree-branch-gate.sh`. Repos explicitly declared out of scope there — "a personal blog, a scratch clone" — now get their hooks executed anyway.
- The exit-status handling treats only 126/127 as "could not run". Under the hook protocol only 0 and 2 are verdicts, so a counterpart that dies on an unbound variable or a missing `jq` exits 1, which propagates as a non-blocking error and the command **proceeds** — a block turned into a pass, where our own policy would have applied.

## Why this is not a patch away

The premise of the chosen design was "never weaker than the status quo on any path". That holds for the paths I enumerated, and misses the one that matters: the mechanism itself introduces execution of code from a directory the command names.

Any trust signal taken **from the target repo** is forgeable, because the attacker controls that repo — a planted `.markgate.yml`, a plausible `origin` remote, a committed rather than working-tree file, all equally plantable. Trust has to come from outside the target, which means either a maintainer-maintained allow-list or dropping delegation. That is a decision for the maintainer rather than a fix I should pick unilaterally, so this stays a draft until it is made.

Not merging. The other two reviews (code, tests) are still running and will be appended.

go-to-k:
## Closing — delegation abandoned in favour of documentation

Maintainer decision: drop cross-repo delegation, take the original option 3 from issue 1961 (document that a session applies its own repo's policy to sibling repos). Replacement PR follows and will close the issue.

Recording the four reviews here so the work is not lost, because two of them found things worth keeping even though the mechanism is going away.

### Security — blocker (independently reproduced before acting)

`hook_delegate_if_foreign` execs `<target-repo>/.claude/hooks/<gate-name>` with no trust check. The target is named by the command itself, so any directory the agent can be induced to touch that contains an executable file at that path gets it run with the session's environment. Reproduced with a throwaway repo, a planted `main-tree-branch-gate.sh`, and a payload no more exotic than `git checkout feature-x`: the planted script ran, counted 3 inherited `AWS_*` / `GH_TOKEN`-shaped variables, captured the payload, and the gate exited 0.

Not patchable from inside the design: every trust signal available from the target repo is forgeable, since the attacker controls that repo.

Two compounding issues, both real regardless: delegation ran **before** the repo opt-in check that issue 1259 added in `branch-gate.sh` / `main-tree-branch-gate.sh`, so repos explicitly declared out of scope got their hooks executed anyway; and only 126/127 were treated as "could not run", so a counterpart erroring with exit 1 propagated as a non-blocking error and the command proceeded.

### Code — no blocker, three minors worth recording

- **Exit ≥128 propagates as a verdict.** A counterpart killed by a signal (timeout, OOM) exits 128+N, which Claude Code treats as a non-blocking error, so the tool call proceeds — a counterpart that never finished converts a block into a pass. Same class as 126/127; the guard should have been `>= 126`.
- **No guard that `target_dir` is a non-empty existing directory.** `git -C ""` is a silent no-op that resolves the hook process's cwd instead, so an empty value silently names an unrelated repo as the target. Latent today (all callers seed `${hook_cwd:-$PWD}`), but it was becoming a 24-caller dependency.
- **`bughunt-clean-gate` is the one gate delegation genuinely weakened.** Every other gate's subject is target-repo state, so handing it over is neutral; that one's subject is *session* state (live AWS resources), and its pr-create/merge paths deliberately aggregate across owners. Delegating it meant a foreign-repo merge stopped consulting cdkd's live hunt. This falsified the PR body's "never weaker on any path" claim.

### Tests — the finding I most needed

Of the mutations tried, **five were not caught**, including the two that matter most: removing the common-dir equality check entirely, and swapping `--git-common-dir` for `--show-toplevel` — the exact design error the source file's own comment warns against. Cause: the fixture's linked worktree had no `.claude/hooks/` at all (`mk_repo` commits before the caller script is written, so `git worktree add` produced a tree without it), so the mutant fell into the *foreign, no-counterpart* arm and still produced the expected exit code. Rebuilt realistically — hooks committed before `worktree add`, as they are in this repo — the mutant delegates and executes the worktree's own copy, and the exit code is *still* identical because the counterpart is the same script. Exit status alone can never discriminate that case; it needs a negative assertion (counterpart-ran marker absent, stderr free of "delegating").

So the suite looked thorough and was partly vacuous, which is the same failure this session flagged in an upstream review two hours earlier. Worth carrying forward as a lesson about fixtures rather than about this mechanism.

### What survives

Nothing from the diff. The friction that motivated it is real but bounded — two hand-run checklists during a three-repo campaign — and the replacement makes it legible rather than mysterious, which was the cheaper half of the value all along.

Issue #2017, opened 2026-08-19
chore(work-issues): section 8 tells agents to pass a vp flag that does not exist, and its error rc reads as the failure the paragraph is hunting


go-to-k:
`/work-issues` section 8, in the no-`src/**` verification tier, tells the next agent:

> **Run it more than once, BEFORE and AFTER — and pass `--no-cache`.**
> `run.cache.tasks` is true in `vite.config.ts` and the `check` task does not opt out of it, so repeats replay instead of re-running

The diagnosis is right and the remedy does not exist.

## Measured 2026-08-19

```
$ vp run check --no-cache
error: Invalid vite task command: vp with args ["check", "--no-cache"]
* error: unexpected argument '--no-cache' found
  tip: to pass '--no-cache' as a value, use '-- --no-cache'
Usage: vp check [OPTIONS] [PATHS]...
```

Exit code **1**. And `vp run --help` lists no cache option at all — the full set is `-r/--recursive`, `-t/--transitive`, `-w/--workspace-root`, `-F/--filter`, `--ignore-depends-on`, `-v/--verbose`, `--last-details`, `-h/--help`. There is no `--no-cache`, `--force` or equivalent on `vp run`.

## Why this is worse than a dead flag

The rc is **1**, and the surrounding instruction is specifically about watching for *an rc that differs across identical runs*. So an agent that follows the sentence literally gets exactly the signal the sentence told it to treat as a real failure, from a command that never ran the check at all. In this session that misread lasted until the log was opened. The next agent may not open it — the whole point of the paragraph is that a cached replay hides a differing rc, so a differing rc is what it has been primed to find.

The cached-vs-uncached gap is real, so the paragraph should not simply be deleted:

```
$ vp run check     # first
vp run: 0/12 cache hit (0%)
$ vp run check     # again
vp run: cache hit, 2.55s saved.
```

## What needs deciding

Find the invocation that actually re-runs the task and put THAT in section 8, having run it. Candidates, none yet verified:

- `vp check` directly rather than through `vp run` (note section 8 elsewhere states these two are NOT equivalent — `vp check --fix` has passed 0-errors while `vp run check` failed, so substituting one for the other changes what is being attested);
- clearing whatever cache directory `run.cache.tasks` writes to, before the second run;
- a per-task `cache: false` in `vite.config.ts` for `check` — note `typecheck:test` already prints `cache disabled`, so the mechanism exists and `check` simply does not use it.

[...]

Issue #2006, opened 2026-08-19
test(scripts): security-surface-list-sync compares sorted SETS, so a deleted list entry that prose also names still passes


go-to-k:
`tests/unit/scripts/security-surface-list-sync.test.ts` fences four copies of the security-surface path list against drift. Two of its four extractors compare as a **sorted set**, and one of those also matches paths in PROSE, so an entry deleted from a list can still be "found" and the test passes.

This is the optional third item of (#1993), verified against this repo rather than carried over: `cdk-local` hit the same shape and its write-up predicted it here.

## The defect

`agentPaths()` (around line 84):

```js
return [
  ...new Set((section.match(/`(src\/[a-z0-9\-/]+\.ts)`/g) ?? []).map((m) => m.slice(1, -1))),
].sort();
```

Two properties combine badly:

1. **The pattern matches any code span in the section**, not just list items — `/`(src\/[a-z0-9\-/]+\.ts)`/g` has no list-item anchor. `.claude/agents/pr-security-reviewer.md` section 4 is prose plus a list, and the prose names paths (that is what makes the section readable).
2. **`new Set` then dedups**, so one path contributed by BOTH the list and a prose sentence collapses to a single element.

Net effect: delete a path from the list while a prose sentence in the same section still mentions it, and the extracted set is unchanged. The fence reports green on exactly the drift it exists to catch.

`skillPaths()` (around line 58) is NOT vulnerable to the prose half — it anchors on the list-item shape, `/^\s+- `(src\/[^`]+)`$/gm` — but it still dedups with `new Set`, so a list carrying the same path twice is indistinguishable from one carrying it once. That matters less, but it is the same weakening.

`hookPaths()` (around line 55) and `claudeMdPaths()` (around line 113) sort without a Set, so they are order-insensitive but duplicate-preserving.

## Why comparing as a set is the wrong contract here

These four lists are hand-maintained copies of one list. What the fence must prove is that the copies are the SAME LIST, and a list is an ordered sequence with multiplicity — not a set. Comparing sets discards precisely the two properties whose drift is undetectable by reading: a dropped duplicate, and an entry present only because something else in the file happens to name it.

## Fix direction

- Extract in **document order with duplicates preserved** and compare sequences, or compare multisets if the four files genuinely disagree on order (check first — if they do agree, order is free signal and should be asserted).
- Scope `agentPaths()` to list items the way `skillPaths()` already does, so prose cannot refill the set.
[...]

Issue #1993, opened 2026-08-19
chore(work-issues): mirror two flow lessons from cdk-local (English-only for filed issue bodies; $0-relative harnesses must run beside their subject)


go-to-k:
## What

Two flow lessons from a `/work-issues` run in the sibling repo
(go-to-k/cdk-local#506 / PR go-to-k/cdk-local#513, 2026-08-19) belong in this
repo. Filing rather than PR-ing per `.claude/skills/work-issues/SKILL.md:694`
("otherwise file one issue per repo carrying the `Session-fit` line") — the
mirror fallback working as written.

Every claim below was checked against THIS repo's files first, per the
verify-the-copy rule in §10-c. Where cdk-local's wording would be false here, the
correct statement for cdkd is given instead.

## Lesson 1 — nothing covers the English-only rule for issues the flow FILES

This one wants a **hook**, not prose, and the matcher already exists in-repo.

### The gap

The flow files issues in two places — `.claude/skills/hunt-bugs/SKILL.md:113`
(`gh issue create`) and `work-issues/SKILL.md:694`, `:843-845` (the deferral
issue carrying `Session-fit`) — and neither prose nor hook covers their language:

- `CLAUDE.md:323` enumerates "source code, shell scripts, hook messages, config
  files..., docs, comments, commit messages, PR titles/bodies" and then closes
  with "this rule applies only to files that land in the repository". An issue
  body is not a committed file, so it is outside the rule by that clause, not
  merely unmentioned.
- `.claude/hooks/non-english-text-gate.sh:71-74` matches
  `gh pr (create|edit|merge)` only, and its subject is the **PR diff**
  (`gh pr diff <N> --name-only`, `:150`) — so it cannot see a PR body, let alone
  an issue body.
- `work-issues/SKILL.md:820-821` covers "issue **comments** on this repo";
  `:352` covers the **claim comment**. Neither covers a body you file.
- `CLAUDE.md:244` pins the `Session-fit` KEY to one spelling and explicitly
  contemplates "an English or a Japanese report alike", while `CLAUDE.md:249`
  mandates a readable gloss on every value. So the repo requires a gloss,
  permits a Japanese report, and has nothing saying which language the gloss
  takes when the destination is a public issue.

### It already bit

In cdk-local — same hook, same PR-only scope — a `/work-issues` run filed its
follow-up with both halves of the `Session-fit` line glossed in the session's
chat language rather than English, and had to patch the body after creation.
Nothing caught it.

[...]

PR #2019, opened 2026-08-19
chore(hooks): gate the English-only rule on the issue / PR text this flow publishes

go-to-k:
## Summary

Closes lesson 1 of (#1993): nothing enforced the English-only rule on the issue / PR text this flow PUBLISHES.

`CLAUDE.md`'s rule enumerated committed artifacts and then narrowed itself with "this rule applies only to files that land in the repository" — so an issue body sat OUTSIDE the rule by its own terms, while `/work-issues` (the `Session-fit` deferral issue) and `/hunt-bugs` both FILE issues as a normal step. `non-english-text-gate.sh` cannot cover them: its subject is the PR **diff**, and a body is never a file in the repo. It bit in `cdk-local`, where a run filed its follow-up with the `Session-fit` gloss in the session's chat language and had to patch the body after creation.

New sibling gate `.claude/hooks/gh-body-english-gate.sh` scans the TEXT handed to `gh`. `CLAUDE.md`'s rule is widened in the same commit — the line is now drawn at whether the text becomes PUBLIC, not at whether it lands in a file — since the hook would otherwise enforce something the canonical rule disclaims.

## Review round

Two read-only reviewers were dispatched despite the size heuristic giving `inline` at the time (agent-instruction files carry no down-bias, and a gate that silently fails is the worst case). They converged independently on the same defects, all now fixed:

| defect | why it mattered |
|---|---|
| multi-line inline bodies never scanned (`perl -ne` is line-wise) | the NORMAL inline shape — the headline case was wide open |
| `gh pr comment` / `pr review` / `release create` outside the verb set | `CLAUDE.md` had just been widened to claim review comments were covered |
| `gh api -f title=` and `--raw-field` passed | `-f title=` is this repo's own blessed title-edit form |
| relative `--body-file` silently passed | hook resolved against its own cwd, never read the payload `cwd` |
| escaped quotes truncated the scan; unquoted values unmatched | |
| hook did not use `lib/command-match.sh` | prose QUOTING these commands fired the gate — reproduced on a `git commit -m` and an `echo` |

## Verification

No `src/**` change, so this is section 8's **command arm**: the artifact is a command, so the verification is running it.

- suite 15 → **42 cases**, green under bash 5.2 **and** macOS system bash 3.2
- **binding proof**: the old hook was swapped back in and the suite re-run — **10 cases fail against it, 42/42 pass against the fix**. Each regression case was verified to fail before, not merely asserted.
- failure direction driven twice more: a missing `perl -CSD` (8/7) and a deliberate gate-disabling mutation (8/7), then restored from a scratch copy and re-run green
- each of the five Unicode ranges now covered in **isolation** — without that, deleting any single range from the character class failed nothing
- adds the repo-mandated quoted-body false-positive cases, two pass cases that can only pass for the right reason, and a registration check (the suite invokes the hook directly, so unregistering it from `settings.json` was previously invisible)
- full `vp run test:hooks` aggregate green; 690 test files / 14054 tests green
- every count asserted in the docs was measured (42 cases, 33 suites, 37 hooks); the suite-count sentence was already stale at 31/35 before this change

## Scope — this does NOT close 1993

Lesson 2 (the `$0` / `${BASH_SOURCE[0]}`-relative harness trap) lands in `.claude/skills/work-issues/SKILL.md` sections 5 and 8, which a live lane is holding with uncommitted work, so section 2's file-disjointness gate refuses a second lane on it. (#1993) stays open for that half.

## Filed from this work

- (#2017) — section 8 tells agents to pass `vp run check --no-cache`; that flag does not exist on `vp run`, and its argument-parse rc=1 reads as exactly the "rc differs across runs" failure the paragraph primes you to hunt. Contested file, bundled with the SKILL.md items.
- (#2006) — `security-surface-list-sync.test.ts` compares sorted SETS and one extractor also matches prose, so a deleted list entry that prose names still passes. This was the optional third item of 1993, verified against this repo.

Refs #1993


go-to-k:
## Not ready to merge — recording round 6, and a design conclusion

Six review rounds have now run against this PR, finding 48+ defects. Rounds 1-5 are fixed and each fix is pinned by a regression case verified to fail against the pre-fix hook. **Round 6 found a blocker that is a regression introduced by the round-5 fix**, and that pattern — a fix round introducing the next round's blocker — has now repeated three times (rounds 2, 4 and 5 each introduced a defect the next round caught).

### Round 6 blocker

`@inq` is built over the RAW command with no heredoc / comment neutralisation. An unterminated single quote leaves the quote state open to end-of-string, so every later offset is marked in-quote and the real invocation is discarded as prose by the `next if $inq[$start]` guard.

```
cat > /tmp/pr-body-en.md <<'EOF'
It's fixed now.
EOF
gh pr create --title "<non-English>" --body-file /tmp/pr-body-en.md
```

Exit **0**. With `It is fixed now.` in the body: exit 2. It is apostrophe-PARITY dependent.

This matters more than a normal miss: heredoc → file → `gh --body-file` in one Bash call is this repo's own blessed shape, and an apostrophe in English prose is near-certain. The same input exits 2 against the previous commit, so round 5 caused it.

### Also found

- **Perf past the hook's own timeout.** The character-class test spawns one `perl -CSD` per extracted value: 200 values 1.5 s, 1000 values 7.3 s, 3703 values **38.1 s** — the hook's timeout is 15 s. The scanning pass itself is fine (0.10 s on 100 KB). Fix is to do the character-class test inside the same perl program.
- **Heredoc-to-stdin body unscanned**: `gh issue create -F - <<EOF` exits 0. Not in the documented known limits.
- **Adjacent quoted chunks**: `-b 'It'\''s <non-English>'` exits 0 — only the first chunk is extracted.
- **Quoted body-file path with a space** is not matched.
- **Doc count wrong**: `.claude/rules/hooks.md` says 43 regression cases contributed 10/12/7/9/5; the round-labelled sections actually hold 17/13/8/9/5 = 52. Also a duplicated clause typo ("without which a `\`-continued a `\`-continued").

### Conclusion — this needs a design change, not a seventh patch round

The recurring failure is always the same shape: this hook parses arbitrary shell command strings with a hand-rolled quote/separator machine, because it needs RAW offsets to bound the segment and therefore cannot use `lib/command-match.sh`, which already solves heredoc and quoted-span neutralisation for the other 16 gates. Every round has been a rediscovery of a case that library already handles.

Two directions worth weighing before more patching:

1. **Extend `lib/command-match.sh` to expose an offset-preserving neutralisation** — same-length filler instead of a single placeholder — so this hook (and any future one needing spans) can reuse the audited implementation rather than reimplementing it. The library's own comment already states the ambiguity rule this hook got wrong: resolve toward KEEPING, i.e. an unterminated span should fail toward scanning, not toward skipping.
2. **Drop segment-bounding entirely** and accept the short-flag false positives by removing `-b` / `-t` / `-n` from the inline arms, keeping only the unambiguous long flags plus the file forms. Strictly smaller coverage, but no shell parsing at all — and the file forms (`--body-file` / `-F <file>`), which the repo actually uses most, never needed the segment logic.

Direction 1 is better if the offset-preserving variant is genuinely reusable; direction 2 ships a smaller but honest gate now.

Marking this draft rather than merging. A gate that silently skips the repo's most common publishing shape is worse than no gate, because its presence is taken as coverage.

The CLAUDE.md rule widening and the `-CSD` / `-0777` findings in this branch are independently correct and worth keeping whichever direction is taken.

Issue #2065, opened 2026-08-19
fix(deploy): cdkd deploy --region US-EAST-1 fails at the state-bucket preflight, while gc and bootstrap --destroy fold the same spelling and work


go-to-k:
## Summary

`cdkd deploy --region US-EAST-1` fails outright, at the state-bucket preflight, before it does anything. `cdkd gc --region US-EAST-1` and `cdkd bootstrap --destroy --region US-EAST-1` both work, because those two commands canonicalize and `deploy` does not. The two sibling commands disagree about what a region spelling means.

Found while building the live-test arm for #2021, by measurement against real AWS rather than by reading.

## Measured

**SigV4 rejects a non-canonical region at every service cdkd calls** — probed with this repo's own vendored SDK:

```text
FAIL STS GetCallerIdentity     region=US-WEST-2  SignatureDoesNotMatch: Credential should be scoped to a valid region.
FAIL S3  ListObjectsV2         region=US-EAST-1  AuthorizationHeaderMalformed: the region 'US-EAST-1' is wrong; expecting 'us-east-1'
FAIL Lambda ListFunctions      region=US-WEST-2  InvalidSignatureException
FAIL ECR DescribeRepositories  region=US-WEST-2  InvalidSignatureException
```

**So cdkd dies at its own preflight:**

```text
$ AWS_REGION=US-EAST-1 node dist/cli.js state info --state-bucket cdkd-state-<acct> --json
StateError: Failed to verify state bucket '...': S3 error during HeadBucket on '...' (HTTP 400).
```

`--region US-EAST-1` is strictly worse than the env var, because `deploy.ts:218-220` copies the flag into `AWS_REGION` / `AWS_DEFAULT_REGION` and the state-bucket resolution needs STS first.

## The asymmetry

| site | folds? |
| --- | --- |
| `src/cli/commands/gc.ts:943`, `:967` | yes |
| `src/cli/commands/bootstrap-destroy.ts:338`, `:358`, `:375` | yes |
| `src/cli/commands/deploy.ts:529` (`baseRegion`) | **no** |
| `src/cli/commands/deploy.ts:558` (`assetRegion`), `:567`, `:583` | **no** |
| `src/cli/commands/publish-assets.ts:147`, `:149`, `:178`, `:184`, `:238`, `:258` | **no** |
| `src/utils/aws-clients.ts` constructor | **no** — it stores the config verbatim |

[...]

Issue #2029, opened 2026-08-19
fix(gc): a literal us-east-1 default is pinned over the profile's region, so bare `cdkd gc` can collect against a region the user never named


go-to-k:
Surfaced while reviewing PR #2014. **Pre-existing on `main`** — not introduced by that PR, which is why it is filed rather than folded in.

## The defect

`src/cli/commands/gc.ts:965` resolves the region with a literal fallback and then passes it to the client factory unconditionally:

```ts
const rawRegion = options.region || process.env['AWS_REGION'] || 'us-east-1';
const region = canonicalizeRegion(rawRegion);

const awsClients = new AwsClients({
  region,
  ...(options.profile && { profile: options.profile }),
});
```

So a user who names no region at all — no `--region`, no `AWS_REGION`, but a configured `~/.aws/config` region — gets `us-east-1` **pinned over their profile's region**. The literal default wins because it is materialised into a value before the client ever gets a chance to resolve the profile.

## Why it matters more here than in a read-only command

`cdkd gc` DELETES. The failure is silent in both directions:

- against `us-east-1`, it evaluates a region the user did not ask about and may delete assets there;
- it does nothing about the region they actually work in, while reporting success.

There is no error, because `us-east-1` is a perfectly valid region — the command simply operates on the wrong one.

## The contrast that surfaced it

PR #2014 fixed the sibling command, `cdkd bootstrap --destroy`, and deliberately did NOT copy this shape. It folds the region on both named paths (`--region` and `AWS_REGION`) but lets an absent region stay absent, so `AwsClients` resolves the profile — matching `loadBootstrapContainerRepo`'s established shape. The two sibling commands now disagree about what "no region given" means.

## Direction

Fold on the named paths and let absent stay absent, as `bootstrap-destroy.ts` now does. Check every other command that materialises a `|| 'us-east-1'` literal before client construction — the defect class is "a default written as a value rather than left to the resolver", and it rarely occurs once. Report that audit in both directions: sites with the same shape, and sites where the literal default is genuinely wanted.

## Verification

Unit: a case with no `--region` and no `AWS_REGION` asserting the client factory receives NO `region` property (the `expected { region: 'us-east-1' } to not have property "region"` shape PR #2014 uses to pin its own half), plus the two named paths still folding. Mutation-probe each — the absent-region arm is a negative assertion and those are easy to write green.

[...]

Issue #1820, opened 2026-08-13
fix(bootstrap): the region variable falls back to a hardcoded us-east-1 while the clients resolve the profile region, so asset storage is named and keyed for the wrong region


go-to-k:
Found by the 3-axis PR review of the issue 1794 lane (both the spec and the code reviewer landed on the same root cause independently). The 1794 PR fixes the CONSEQUENCE for the bucket-policy partition; this issue is the underlying region divergence, which has a second consequence that PR deliberately does not touch.

## The divergence

`src/cli/commands/bootstrap.ts`:

```ts
const awsClients = new AwsClients({
  ...(options.region && { region: options.region }),   // omitted when --region absent
  ...(options.profile && { profile: options.profile }),
});
const region = options.region || process.env['AWS_REGION'] || 'us-east-1';
```

The two lines disagree about what "the region" is whenever `--region` is absent:

- the CLIENTS get no `region`, so the AWS SDK resolves it through its own chain — which reads `~/.aws/config`'s profile `region =` line and `AWS_DEFAULT_REGION`;
- the `region` VARIABLE consults neither, and falls back to a hardcoded `us-east-1`.

So for a user whose profile says `region = ap-northeast-1`, with no `AWS_REGION` and no `--region`, a plain `cdkd bootstrap` talks to ap-northeast-1 while every use of `region` says `us-east-1`.

## Consequence still open after the 1794 PR

`region` is threaded into `ensureAssetStorage` (`src/assets/asset-storage.ts`), where it NAMES the per-region asset storage:

- `getCdkdAssetBucketName(accountId, region)` -> `cdkd-assets-{account}-us-east-1`, but `CreateBucket` runs on the ap-northeast-1 client, so the bucket is **named after a region it is not in**;
- `getCdkdContainerRepoName(accountId, region)` has the same problem for the ECR repo;
- `getBootstrapMarkerKey(region)` writes the marker at `cdkd-bootstrap/us-east-1.json`, so the DEPLOY-side `AssetModeResolver` looks the region up under the wrong key. A deploy that then genuinely targets ap-northeast-1 finds no marker and silently stays in legacy (CDK bootstrap) asset mode, while a deploy targeting us-east-1 finds a marker pointing at storage that does not exist in us-east-1 and hard-errors.

The last one is the interesting failure: the marker is the single source of truth for asset mode, and it is keyed by the wrong region.

## Already fixed (do not re-report)

[...]

PR #2071, opened 2026-08-19
fix(assets): canonicalize the region at the AssetModeResolver boundary and share the marker read

### LINKED ISSUE #2021: fix(assets): an upper-cased deploy region misses the bootstrap marker, silently downgrading assets to legacy (cdk-gc-collectable) storage
Third instance of the region-case defect fixed for `cdkd gc` in (#1995) and for `cdkd bootstrap --destroy` in the same PR. Found by the "audit for a third" sweep that pair prompted: every reader of `getBootstrapMarkerKey` was traced back to where its region comes from.

## The defect

`AssetModeResolver.doResolve` (`src/assets/asset-storage.ts:672`) keys the bootstrap marker off the region it is handed:

```ts
const markerKey = getBootstrapMarkerKey(region);
const body = await this.stateBackend.getRawObject(markerKey);
```

Both deploy-time callers hand it an UN-canonicalized region:

- `src/cli/commands/deploy.ts:523` — `const baseRegion = options.region || process.env['AWS_REGION'] || 'us-east-1';` then `:552` `const assetRegion = stack.region || baseRegion;`
- `src/cli/commands/publish-assets.ts:147` / `:238` — the same two lines.

So `cdkd deploy --region US-EAST-1` (or `AWS_REGION=US-EAST-1`) on an env-agnostic stack looks for `cdkd-bootstrap/US-EAST-1.json`, misses the marker that `cdkd bootstrap` wrote at `cdkd-bootstrap/us-east-1.json`, and takes the `body === null` arm.

A stack whose `env.region` is pinned in CDK is unaffected — `stack.region` comes from the Cloud Assembly and is canonical. The exposure is exactly the env-agnostic case, which is the CDK default.

## Why it matters, stated honestly

Milder than the two already fixed, and in a different way from each:

- gc's version collected nothing (deleted nothing).
- `bootstrap --destroy`'s version reported success while leaving billable resources alive.
- **This one silently downgrades to LEGACY asset mode**: assets are published to the CDK bootstrap bucket/repo instead of the cdkd-owned storage the user opted into. Nothing breaks at deploy time, and the notice even fires — but the whole point of `cdkd bootstrap` (#1002) is that `cdk gc` cannot reach cdkd's asset storage, and the legacy destination is precisely what `cdk gc` DOES collect. So the failure surfaces later, as assets vanishing from under deployed stacks.

With `autoCreate` on, the miss instead triggers a redundant `tryAutoCreate` for a region that is already bootstrapped, whose marker write would then land at the raw key — which is how a raw-key marker gets created in the first place (the population `loadBootstrapContainerRepo` and now gc / `bootstrap --destroy` all carry a second probe for).

## Suggested direction

Canonicalize inside `AssetModeResolver.doResolve` rather than at each caller: the resolver also CACHES by region (`this.cache.get(region)`), so folding at the boundary fixes a second, quieter bug where `us-east-1` and `US-EAST-1` occupy two cache slots and re-probe S3. Give it the canonical-then-raw two-probe shape the other three readers now use.

**A shared helper is the right answer at this point, and this is the issue that justifies it.** There are now three hand-written copies of "read the bootstrap marker, canonical key first, raw spelling second" — `local-state-loader.ts:368`, `gc.ts:994`, `bootstrap-destroy.ts:331` — and a fourth is needed here. They already differ in ways that are easy to get wrong: `bootstrap-destroy` must also DELETE the key it read from (a bug caught in review — deleting the canonical key when the marker came from the raw one orphans the marker), and `local-state-loader` is best-effort where the other two hard-error. Suggested shape, in `src/assets/asset-storage.ts` beside `getBootstrapMarkerKey`:

```ts
readBootstrapMarkerBody(stateBackend, rawRegion): Promise<{ body: string | null; resolvedKey: string }>
```

returning the key actually read so every caller's error message, plan line and delete target can follow it. Each caller keeps its own policy on top (throw vs. fall back vs. destroy).

Deliberately NOT in scope, so it is not "fixed" by mistake:

- `ensureAssetStorage` (`src/assets/asset-storage.ts:351`) reads the marker with a raw region too, but its WRITE uses the same raw region, so read and write are self-consistent. Folding the read alone would break that pairing. Aligning the write side is issue (#1820)'s lane, and this issue should land after or with it.

## Why not folded into the (#1847) / (#1995) PR

Different subsystem and a different verification cycle: this is the deploy/publish asset hot path, so it needs its own integ (a deploy under an upper-cased region asserting the asset lands in the cdkd-owned bucket rather than the CDK bootstrap one), whereas the two fixed instances rode the `gc-custom-asset-names` fixture that PR already runs. `deploy.ts` was also being edited by another lane at the time.

Session-fit: next (not this session) — the deploy asset path needs its own integ fixture and a decision on the shared helper, and the `ensureAssetStorage` half is sequenced behind (#1820). / Effort: ~3 h+ (the shared helper + four call sites, plus a new deploy-under-upper-cased-region integ arm and a review round)

go-to-k:
## What this changes

`AssetModeResolver.resolve` now folds region case at its own boundary rather than trusting each caller, and the canonical-then-raw bootstrap-marker read becomes ONE shared helper instead of three hand-written copies.

- `readBootstrapMarkerBody(stateBackend, rawRegion) -> { body, resolvedKey }` in `src/assets/asset-storage.ts`, returning the key ACTUALLY read so every caller's message, plan line and delete target can follow it.
- Folding at the resolver boundary also collapses the two cache slots `us-east-1` / `US-EAST-1` used to occupy, each re-probing S3.
- The three existing copies (`gc.ts`, `bootstrap-destroy.ts`, `local-state-loader.ts`) are folded onto it, and the helper deliberately does NOT catch, so each caller keeps its own policy unchanged.

## The part worth reading: the issue's premise did not survive measurement

Issue (#2021) describes an env-agnostic stack under an upper-cased region silently downgrading to legacy (`cdk gc`-collectable) asset storage, with assets later vanishing from under deployed stacks. **That narrative is not reachable against real AWS, and this PR deliberately does not repeat it.** Measured while building the live-test arm:

- SigV4 rejects a non-canonical region at every service cdkd calls — `AuthorizationHeaderMalformed` from S3, `InvalidSignatureException` from Lambda / ECR, `SignatureDoesNotMatch` from STS.
- `deploy.ts:529` does NOT canonicalize where `gc.ts:943` and `bootstrap-destroy.ts:338` do, so `cdkd deploy --region US-EAST-1` dies at the state-bucket preflight and never reaches the marker read at all.
- The `body === null` arm the issue describes needs a CANONICAL client region paired with a RAW key spelling. On the deploy path those are the same string, so the pairing cannot occur. Measured GetObject matrix, real state bucket:

  | client region | key | result |
  | --- | --- | --- |
  | `us-east-1` | `cdkd-bootstrap/us-east-1.json` | marker body |
  | `US-EAST-1` | `cdkd-bootstrap/us-east-1.json` | `AuthorizationHeaderMalformed` (400) |
  | `US-EAST-1` | `cdkd-bootstrap/US-EAST-1.json` | `AuthorizationHeaderMalformed` (400) |
  | `us-east-1` | `cdkd-bootstrap/US-EAST-1.json` | `NoSuchKey` (404) — the only `null` |

- The downgrade could not publish anything either: the legacy destination `cdk-hnb659fds-assets-<acct>-US-EAST-1` is not a legal S3 bucket name (`HeadBucket` -> `NotFound`).
- Pinning `env: { region: 'US-EAST-1' }` to decouple the two does not work — CDK refuses the assembly outright, aborting synth for the whole app.

That deploy-boundary fold is filed as (#2065) with the full measurement. **Until it lands, the reachable value of this PR is the two DESTRUCTIVE callers** — `cdkd gc` and `cdkd bootstrap --destroy`, which DO fold and DO reach the helper under an upper-cased region — plus the shared-helper consolidation and the cache-slot fix.

## Deliberately out of scope

- `ensureAssetStorage`'s read is NOT folded: it is paired with its marker WRITE and both use the same raw region, so folding the read alone would break that pairing — a custom asset-bucket name recorded by an earlier run would stop being reused and a conflicting one would stop being refused. Aligning the write side is (#1820)'s lane; the reason is recorded at the call site.
- (#2029)'s `|| 'us-east-1'` literal default in `gc.ts` is untouched (`git diff -U0 src/cli/commands/gc.ts | grep -c "us-east-1'"` -> 0).

## The second probe's reachability, documented from measurement

The helper's JSDoc no longer asserts that a plain `AWS_REGION=US-EAST-1 cdkd bootstrap` "really wrote" the raw key — it cannot, because the marker is written LAST and both resources must be created first from that same raw region, where the conventional bucket name is illegal and `CreateBucket` also carries an invalid `LocationConstraint`. The raw key needs BOTH `--asset-bucket` and `--container-repo` as valid lowercase names AND a pre-existing bucket, or a marker written by hand / by a cdkd predating those guards. Probe 2 is kept anyway, because that state is real and losing it silently re-points a bootstrapped region at `cdk gc`-collectable storage; both conditions are pinned by tests so the claim cannot rot.

## Tests

14 new cases in `tests/unit/assets/asset-storage.test.ts`, every one mutation-probed. Selected results:

| probe | result |
| --- | --- |
| de-canonicalize the helper | 22 RED across 4 suites — proves all three legacy callers really route through it |
| report canonical as `resolvedKey` | 8 RED, including `bootstrap-destroy`'s DELETE-target case |
| de-fold `resolve()` | 4 RED, including the cache probe-count |
| fold region in `ensureAssetStorage` | 2 RED (both reachability tests) |
| ignore `logPrefix` / delete the raw-hit debug line | 1 / 2 RED |

The cache-slot case asserts the PROBE COUNT, not the returned mode — a returned-mode assertion stays green with the cache bug intact.

**One correction made during review, worth flagging.** An earlier round removed a test with an in-code note claiming its property "is TRUE, but nothing can make it false". The test reviewer measured that and it was wrong: the test DID go red under the unfolded-cache mutation, and only stayed green when `legacyNoticeShownRegions` itself was deleted. The note now says what is actually true. A wrong "cannot be fenced" note suppresses the retest, which is why this was not left as-is.

## Review

3-axis + security reviewer (tier raised above the size heuristic's `1-reviewer`, because the diff includes the DELETE-target key of a destructive command).

- spec: clean — both spec parts implemented, no out-of-scope drift, per-caller policy preservation proved individually rather than accepted.
- security: no blocker, no minor — traced every destructive caller's region usage end to end; bucket / repo names come from the marker BODY and are never re-derived from a region, `ExpectedBucketOwner` and the owned-elsewhere refusal are untouched, and no failed-read path leads to a destructive default.
- code: 4 minor / nit, all addressed (the raw-probe premise, the cache order-dependency note, `legacyNoticeShownRegions`, and the changelog reading wider than the code).
- test: 3 claims verified by probe, 1 in-code note found factually wrong (above), 2 gaps closed.

## Live verification

`/run-integ asset-bootstrap` (eu-west-1, 166s, destroy 2 deleted / 0 errors / 0 orphans) — it walks legacy mode with no marker, `cdkd bootstrap`, cdkd-assets detection WITH the marker, the deleted-resource hard error, and a clean destroy. That is `AssetModeResolver` plus the shared helper through `bootstrap` and `bootstrap --destroy`, i.e. the callers that actually reach the changed code today.

`gc-custom-asset-names` was the first choice and was abandoned after two environment blockers, both recorded in the integ ledger as blocked-before-assertions rather than as a defect in this change: the docker daemon was down, and the retry then hung in `docker push` for 4h17m with no container running. It was killed and the bucket / ECR repo / marker it had created were removed by hand. `asset-bootstrap` needs no docker and exercises the resolver more directly, so it is the better fixture for this change regardless.

Region choice is not arbitrary and is worth recording, because it cost three attempts: the fixture needs a region that is CDK-bootstrapped (its Phase 1 publishes to the CDK bootstrap bucket) AND has no cdkd marker (its own guard refuses otherwise). On this account only `eu-west-1` satisfies both — `us-east-1` / `ap-northeast-1` / `us-west-2` carry cdkd markers, and `ca-central-1` is not CDK-bootstrapped.

Closes #2021


go-to-k:
Status: ready to merge except for one gate that this machine cannot currently satisfy.

`integ-local` fires on this PR, correctly: it changes `src/cli/commands/local-state-loader.ts`, which is the `--from-state` path that `cdkd local run-task` uses to recognise images published to a custom-named cdkd container-asset repo. That gate needs a real Docker run, and Docker's registry connectivity is broken on this machine right now:

- `docker pull hello-world:latest` was killed at 120s, three separate times, including after a full Docker Desktop restart.
- `docker version` answers (server 20.10.22) and `com.docker.backend` is running, so the daemon is up — it is registry networking specifically that hangs.
- The first attempt at the intended fixture, `local-run-task-cdkd-assets`, hung in `step 1c: pre-pulling sidecar + asset base image` and was killed by a watchdog at 901s. Post-kill sweep: 0 orphan containers, 0 orphan networks.

Everything else on this PR is complete:

- `check` / `docs` / `verify-pr` / `pr-review` markers recorded; `integ-destroy` fresh.
- All CI checks green (`check`, `check-build-test`, `once-leak-detect`, `runtime-compat` 20/22/24).
- Live-tested against real AWS with `/run-integ asset-bootstrap` (eu-west-1, 166s, destroy 2 deleted / 0 errors / 0 orphans), which walks `AssetModeResolver`'s legacy -> cdkd-assets detection and the shared marker read through `bootstrap` and `bootstrap --destroy`.
- 4 reviewers (spec / code / test / security); spec and security both came back clean, and every code and test finding is addressed.

Deliberately NOT done: narrowing `.markgate.yml`'s `integ-local` scope to make this mergeable. The gate's own message anticipates that move and rejects it, and it is right to — the changed file really is on the local-execution path, so narrowing the scope would be abandoning the verification rather than satisfying it.

To finish this PR once Docker's registry access works:

```
/run-integ local-run-task-cdkd-assets
gh pr merge 2071 --squash --delete-branch
```

PR #2165, opened 2026-08-21
fix(state): fail fast when a stack lock is held instead of running under contention and releasing it

### LINKED ISSUE #2161: fix(state): destroy/import/orphan/drift/state ignore acquireLock()'s boolean return, so they run against a stack another process holds the lock on — and then release that process's lock
## Summary

Six command paths that take the stack lock call `LockManager.acquireLock(...)` but **discard its boolean return value**, then proceed unconditionally. `acquireLock` returns `false` (it does **not** throw) on contention — a live, non-expired lock already held by another owner (`src/state/lock-manager.ts:266`) — so the surrounding `try/catch` never fires. The result: `destroy` / `import` / `orphan` / `drift --accept` / `drift --revert` / `state refresh-observed` run against a stack that another process (e.g. an in-flight `cdkd deploy`) currently holds the lock on — and on completion they release that other process's lock via the normal owner-blind release path (`releaseLock` does an unconditional `DeleteObject` with no owner/ETag check, `src/state/lock-manager.ts:342-355`).

`cdkd deploy` is unaffected **by this discarded-return bug** because it uses `acquireLockWithRetry(...)`, which checks the boolean and **throws** on contention (`src/state/lock-manager.ts:435-464`). (The owner-blind `releaseLock` is a separate, system-wide weakness — a long-running owner whose fixed-TTL lock expires and is replaced can later delete the replacement owner's lock — but that is out of scope here.)

## Affected call sites (all discard the `boolean`)

- `src/cli/commands/destroy-runner.ts:603` (`destroy` / `state destroy`) — then unconditionally sets `lockHeld = true` at :610
- `src/cli/commands/import.ts:605` and `src/cli/commands/import.ts:1915` (top-level + nested child; note `--dry-run` still calls the broken acquire)
- `src/cli/commands/orphan.ts:188` (non-dry-run only; `--dry-run` skips acquisition)
- `src/cli/commands/state.ts:2074` (`state refresh-observed`)
- `src/cli/commands/drift.ts:2238` (`drift --accept`) and `src/cli/commands/drift.ts:3208` (`drift --revert`)

The correct pattern is already used elsewhere: `deploy-engine.ts:1488`, `rollback.ts:292`, `export.ts:6013` (nested export), `scrub.ts:2704` all call `acquireLockWithRetry(...)`; and top-level `export.ts:2615` uses the raw `acquireLock` but **does** check the boolean and reject `false` at `export.ts:2621`. The six paths above are the ones that ignore it.

## Root cause

`acquireLock` returns `Promise<boolean>` and, on a `PreconditionFailed` conditional-PUT (`IfNoneMatch: '*'`) against a live non-expired lock, returns `false` rather than throwing:

```ts
// src/state/lock-manager.ts (PreconditionFailed branch)
// ... expired-lock cleanup path ...
return false;   // contention: a live foreign lock is held (or another process won the post-cleanup race)
```

The affected call sites treat it like a throwing acquire:

```ts
// src/cli/commands/destroy-runner.ts:601-610
try {
  await ctx.lockManager.acquireLock(stackName, regionForState, undefined, 'destroy');
} catch (error) {
  process.removeListener('SIGINT', sigintHandler);
  throw error;
}
lockHeld = true;   // reached even when acquireLock returned false
```

Because a held-lock result is a `false` return (not an exception), the `catch` is skipped and the command continues.

## Reproduction (real AWS, us-east-1)

1. `cdkd bootstrap`
2. Deploy a trivial stack (one `AWS::SSM::Parameter`) with `cdkd deploy` — writes `state.json` and the live resource.
3. Simulate a concurrent process holding the lock: `PutObject` `cdkd/<stack>/us-east-1/lock.json` with a **different owner** and a **future** `expiresAt` (now + 30 min), `operation: "deploy"`.
4. Control — `cdkd deploy` (uses `acquireLockWithRetry`):
   ```
   Stack '<stack>' (us-east-1) is locked by process-A@otherhost:9999 (operation: deploy). Retrying ... (attempt 1/3, 2/3, 3/3)
   Failed to acquire lock ... after 4 attempts. Use --force-unlock ...
   ```
   Correct: it respects the lock and makes no changes.
5. Bug — `cdkd destroy <stack>` with the same lock present:
   ```
   Acquiring lock for stack <stack>...
     ✓ Param... (AWS::SSM::Parameter) deleted
   ✓ Stack <stack> destroyed (1 deleted, 0 errors)
   ```
   It proceeds despite the held lock.

Post-conditions verified against AWS:
- process-A's `lock.json` → **404** (destroy released another process's lock)
- the SSM parameter → **ParameterNotFound** (a resource the lock holder believed it owned was deleted)
- `state.json` → **404**

## Impact

Mutual exclusion is defeated on the six paths above, and the impact differs by command:

- **`destroy` / `state destroy`** — deletes live AWS resources out from under a concurrent `deploy`/`import`/etc. on the same stack.
- **`drift --revert`** — issues `provider.update` against live AWS resources concurrently with another writer.
- **`import` / `orphan` / `drift --accept` / `state refresh-observed`** — primarily mutate cdkd state (no AWS resource delete), but still corrupt the concurrency invariant and release the other owner's lock.

In every case the completing command releases the **other** process's lock (owner-blind `DeleteObject`), so contention is silently lost rather than surfaced.

## Not a duplicate of the existing lock issues

The prior lock work — go-to-k/cdkd#1348, go-to-k/cdkd#2118, go-to-k/cdkd#1342 — addresses the **stranded-lock** failure mode (a lock left behind after SIGINT/SIGTERM). This issue is the inverse: a *held* lock being **ignored**, and another owner's lock being released. `#1348` introduced the `lockHeld` gate and does discuss foreign-lock deletion, so it is adjacent prior work, but none of those changes check the discarded boolean at the six call sites above.

## Proposed fix

Switch the six call sites to `acquireLockWithRetry(...)` (which throws on contention and already carries the expired-lock cleanup + retry UX), or — where a non-retrying acquire is intended — check the boolean and abort with the same "locked by <owner>, use --force-unlock" error the retry helper raises (this is exactly what top-level `export.ts:2621` already does). A regression test should assert each of the six commands surfaces a lock error (not success) when a live foreign lock is present.

nix-tkobayashi:
## Summary

Fixes #2161. Six command paths that take the stack lock called
`LockManager.acquireLock(...)` and **discarded the boolean it returns**.
`acquireLock` returns `false` (it does **not** throw) when a live,
non-expired lock is already held by another owner, so these commands ran
against a stack another process (e.g. an in-flight `cdkd deploy`) currently
held the lock on — and on completion released that other process's lock via
the normal owner-blind release path (`releaseLock` does an unconditional
`DeleteObject` with no owner / ETag check).

Affected sites, all now fixed:

- `src/cli/commands/destroy-runner.ts` (`destroy` / `state destroy`)
- `src/cli/commands/import.ts` (top-level **and** nested child)
- `src/cli/commands/orphan.ts`
- `src/cli/commands/state.ts` (`state refresh-observed`)
- `src/cli/commands/drift.ts` (`drift --accept` and `drift --revert`)

`cdkd deploy` was unaffected (it uses `acquireLockWithRetry`, which throws on
contention). `cdkd export` was already correct (it checks the boolean); this
PR aligns its two `force-unlock` hints to also carry `--stack-region` so the
fail-fast message class is spelled consistently across every site.

## Fix

Each site now checks the boolean and throws on `!acquired`, mirroring the
fail-fast pattern `cdkd export` already uses (`export.ts` checks the boolean
and aborts with a "locked by another process, use `cdkd force-unlock`"
error). The throw happens **before** any lock-release / state-write /
provider call, so the foreign lock is left untouched and no resource is
deleted / mutated under contention. Failing fast (rather than switching to
`acquireLockWithRetry`) keeps the behaviour consistent with `export` and
honours the existing "fail fast if another process is mid-deploy" intent
stated in `import.ts`.

### destroy-runner cross-region cleanup hardening

`destroy` switches the process-global `AWS_REGION` / AWS clients for a
cross-region stack **before** acquiring the lock, and restores them only in
the main `try/finally`. Because the fix now throws before that `finally` is
entered, the switch would leak on ordinary contention. This PR extracts one
**idempotent** `restoreBaseRegionAndClients()` helper and calls it at every
exit that can happen before / around the main finally:

- the cross-region **setup** failure,
- the **lock-acquisition** failure,
- the **strong-ref refusal** path, and
- the guaranteed inner `finally` on the success path (so a rejecting
  `releaseLock` cannot skip it).

The helper claims its `regionSwitched` flag first (idempotent, no
double-destroy), restores the globals **before** `destroy()`, and swallows a
throwing `destroy()` so a teardown failure can neither mask the real error
nor skip listener/region cleanup. `renderer.start()` was also moved inside
the protected `try` (it writes to stdout and can `EPIPE`, which would
otherwise strand the lock).

## Tests

- A dedicated `destroy-runner-lock-contention.test.ts` fences the destroy
  path: contention (`acquireLock` resolves `false`) → aborts, no provider
  delete, no `releaseLock`, no `deleteState`; plus cross-region cleanup
  fences for the setup-failure / acquire / strong-ref / release-reject /
  throwing-destroy / `renderer.start()`-throw exits.
- Per-command contention fences added for `orphan`, `state refresh-observed`,
  `drift --accept` / `--revert`, top-level `import`, and the **nested-child**
  import acquire.
- Every guarded behaviour was verified with a mutation probe (reverting the
  guard turns the fence red).
- Existing tests that mocked `acquireLock` were updated to return `true`
  (the boolean is now checked).

## Verification

- typecheck / lint / build green; the full unit suite passes for every file
  this PR touches.
- Reviewed independently for correctness, cross-region-cleanup safety
  (idempotency / double-free / ordering), and test faithfulness.
- Live-tested against real AWS: with a foreign lock planted on a deployed
  stack, `cdkd destroy` now refuses ("Could not acquire lock ...") and leaves
  the resource, the foreign lock, and `state.json` untouched; with no
  contention it destroys normally.

Note: this touches deletion logic (`destroy-runner.ts`), so a real-AWS
destroy integration run is required before merge per the repo's
`integ-destroy` / `integ-broad` gates.

## Review method

This change was reviewed adversarially by an independent reviewer over 8
rounds until zero findings remained. Each round's finding was fixed and
re-reviewed; the progression walked from the core fail-fast fix to a
sequence of pre-existing latent cross-region cleanup leaks in `destroy-runner`
(setup failure, strong-ref refusal, rejecting `releaseLock`, throwing
`destroy()`, throwing `renderer.start()`), each of which is now restored and
fenced with a mutation-probed test.

## Retrospective

Lesson recorded for future work: when a fix makes a function throw earlier
than before, any process-global state switched before the main `try/finally`
leaks on the new exit — and the same leak usually already exists at every
sibling exit before that `try/finally`. Extract one idempotent cleanup helper
and call it at all of them, restoring globals before any `.destroy()` and
swallowing a teardown throw.

## Follow-ups from maintainer review

- Every fail-fast `force-unlock` hint now carries `--stack-region <region>`,
  so `cdkd force-unlock` targets the right region even for a first-time
  `import` or a nested-stack child that has no state record yet.
- `destroy-runner.ts`: the `Acquiring lock...` log line moved **inside** the
  protected `try` (it reaches `process.stdout.write`, which can `EPIPE`, and
  sits after the SIGINT registration + region switch — same leak class the
  rest of this PR closes).
- Contention unit fences hardened: each now asserts the surfaced error is the
  **lock** error (`/Could not acquire lock/`), `drift` asserts the provider's
  live-AWS `update` never fired, the nested-child `import` fence adds positive
  markers that the child acquire was attempted, and `orphan` stubs state so a
  reverted guard would actually reach `saveState` (otherwise the assertion is
  vacuous). All re-verified with a mutation probe.
- `docs/troubleshooting.md` documents the fail-fast message + `--stack-region`
  recovery, and counts the seven fail-fast commands (noting `cdkd export`'s
  nested-stack children retry briefly first).

## Behaviour changes (correct, but called out)

- `cdkd import --dry-run` now **aborts on contention**. Unlike
  `cdkd orphan --dry-run` (which skips lock acquisition entirely), import
  deliberately takes the lock even in `--dry-run` so its plan cannot lie about
  the current AWS state, so a held lock now fails the dry-run fast.
- `cdkd destroy --all` now **aborts the whole run at the first contended
  stack** rather than continuing to the remaining stacks — the fail-fast throw
  propagates out of the per-stack runner. (A contended stack was never safe to
  skip past silently.)

## Maintainer verification (added on merge)

Reviewed at head `fa563361` with the 3-axis tier (26 files) plus the security
add-on: `pr-spec-reviewer` / `pr-code-reviewer` / `pr-test-reviewer` /
`pr-security-reviewer` in parallel. **Zero blockers.** The spec reviewer
confirmed all seven review items implemented with file:line evidence; the
security reviewer traced every one of the nine acquire sites and confirmed the
throw precedes every committing statement (`lockHeld`, the releasing `try`,
every state write, every provider call), and enumerated every path to
`releaseLock` / `deleteLock` / `forceReleaseLock` as post-acquire.

Independently verified here, not just re-read:

- Every one of the 11 guards this PR adds was mutation-probed — reverting each
  one individually turns its own fence red, so none of the new tests are inert.
- Full unit suite green on this head: 15512 passed, 1 skipped, 0 failed.
- **Live-tested against real AWS.** With a foreign lock planted on a deployed
  stack (`owner=process-A@otherhost:9999`, `operation=deploy`, expiring in 30
  min), `cdkd state destroy` exited 1 with
  `Could not acquire lock for stack 'Cdkd2165LiveTest' (us-east-1) — ... run
  'cdkd force-unlock Cdkd2165LiveTest --stack-region us-east-1' ...`, and the
  foreign `lock.json` and `state.json` were both intact afterwards. With the
  lock removed, the same command destroyed normally (1 deleted, 0 errors) —
  so the refusal is the guard firing, not a fixture that cannot proceed.
- Real-AWS integration runs on this head, both clean:
  `drift-revert` (13 deleted / 0 errors / 0 orphans) and `lambda`
  (9 deleted / 0 errors / 0 orphans). `drift-revert` is the arm that matters
  most here — `drift --revert` is the path whose `provider.update` the new
  check keeps from firing under a foreign lock. Ledger rows are in
  `docs/_generated/integ-last-run.tsv` on this branch.

Post-run scans were empty for every stack involved: no `state.json`, no
`lock.json`, and no IAM / Lambda / DynamoDB / SQS / SNS / KMS / ECS / Glue
leftovers.

## Follow-ups (not blocking this PR)

The reviewers' minor / nit findings are filed rather than dropped:

- go-to-k/cdkd#2170 — the contention message prints no owner / operation /
  expiry, so it asks the user to judge whether another process is live while
  withholding the evidence; plus the hint propagates `--stack-region` but not
  `--profile` / `--state-bucket` / `--state-prefix`, and the moved
  `logger.info` has no fence.
- go-to-k/cdkd#2171 — three pre-existing siblings of this bug family found by
  the security pass: `destroy` deletes a 0-resource stack's state *before*
  acquiring, `state rm` force-releases a live foreign lock with no acquire,
  and `deploy-engine.ts` still has the EPIPE lock-strand this PR closed in
  `destroy-runner.ts`.
- go-to-k/cdkd#2168 — the lock TTL has no renewal, so any operation outliving
  it loses mutual exclusion outright. This is the residual the issue correctly
  scoped out.

Thanks @nix-tkobayashi — good catch, and a thorough fix.


go-to-k (changes requested):
Thank you for this — it is a genuinely good catch, and a nasty one. A `false` return that everyone reads as "it threw on failure" is exactly the kind of bug that survives review for a long time, and the blast radius you documented (a `destroy` deleting live resources under a concurrent `deploy`, then deleting that process's lock on the way out) is about as bad as it gets for this project.

The fix is complete and I could not fault the core of it. I confirmed all 7 raw `acquireLock` sites are covered (the 8th, `export.ts:2615`, was already correct), mutation-probed all 11 new guards — every fence goes red when its guard is reverted, so none of them are inert — and ran the full unit suite green on your head. The `restoreBaseRegionAndClients()` extraction is a real improvement over the block it replaces: it now covers the partial-switch case and restores the globals before `destroy()`, neither of which the old `if (destroyAwsClients)` did.

**Before the list: the real-AWS integration run this PR needs is mine to do, not yours.** You noted in the PR body that the `integ-destroy` / `integ-broad` gates require one — I will run it here once the changes below are pushed. Please do not run one yourself; there is nothing you need to do about AWS for this PR.

Seven changes below, all small. Please make them on this branch rather than in a follow-up: `destroy-runner.ts` is in the `integ-broad` and `integ-destroy` gate scopes, so a second PR touching it would cost me a second real-AWS integration run on top of this one.

### 1. `destroy-runner.ts:646` — one statement is still outside the hardening

Move `logger.info(`\nAcquiring lock for stack ${stackName}...`)` inside the `try` that starts on the next line:

```ts
  process.on('SIGINT', sigintHandler);

  try {
    logger.info(`\nAcquiring lock for stack ${stackName}...`);
    const acquired = await ctx.lockManager.acquireLock(...);
```

It currently sits after `regionSwitched = true` and after the SIGINT registration but outside the `try`, and it reaches `process.stdout.write`. An EPIPE there (`cdkd destroy | head`) leaks the SIGINT listener and the cross-region globals — the same two leaks this PR closes everywhere else. It is the same reasoning you used to move `renderer.start()`.

### 2. All 7 sites — the `force-unlock` hint needs `--stack-region`

Plain `cdkd force-unlock <stack>` resolves the region from `listStacks()` (`force-unlock.ts:88-94`). On a first-time import there is no state record yet — and for a nested child `<parent>~<Child>` there never is one at that point — so it falls back to `namedCliRegion(...) ?? 'us-east-1'`, reports "No lock found", and leaves the user stuck for the full 30-minute TTL while the lock sits at `cdkd/<stack>/<targetRegion>/lock.json`.

Every site already has the region in the same scope. Please append ` --stack-region ${<region>}` to each hint, matching the form `state.ts:1045` already uses:

| Site | stack expression | region expression |
| --- | --- | --- |
| `destroy-runner.ts:669` | `stackName` | `regionForState` |
| `import.ts:620` | `stackInfo.stackName` | `targetRegion` |
| `import.ts:1946` | `childStackName` | `childRegion` |
| `orphan.ts:202` | `stackInfo.stackName` | `targetRegion` |
| `state.ts:2088` | `stackName` | `region` |
| `drift.ts:2253` | `report.stackName` | `report.region` |
| `drift.ts:3240` | `report.stackName` | `report.region` |

Only the import pair is actually broken today, but leaving five siblings on the weaker form is how the next person copies the wrong one. (The pre-existing hint at `destroy-runner.ts:619` is a different message — leave it alone.)

### 3. `import.test.ts:745` — the nested-child fence needs a positive marker

Its three assertions (untyped `rejects.toThrow()`, `childSave === undefined`, no `releaseLock('P~Child')`) are all satisfied by any root-side failure in that hand-built `getCfnResourceTree` fixture. It does discriminate today — I verified by probe — but it is one fixture change from going silently inert. Please add:

```ts
expect(mockAcquireLock).toHaveBeenCalledWith('P~Child', expect.any(String), expect.any(String), 'import');
expect(errorSpy.mock.calls[0]?.[0]).toMatch(/Could not acquire lock for nested stack/);
```

### 4. `orphan.test.ts:336` — `expect(mockSaveState).not.toHaveBeenCalled()` is vacuous

That test never stubs `mockGetState` (the `beforeEach` only resets it), so with the guard reverted the `getState` block at `orphan.ts:209` throws "No state found" before `saveState` is reachable. Only the `releaseLock` assertion carries the fence. Please stub state the way the test at `orphan.test.ts:240` does.

### 5. All five contention fences — assert the message, not just that something threw

`process.exit` is stubbed to throw, so `expect(error).toBeDefined()` only proves *some* exit happened. Each file already has an `errorSpy`. Please add a `toMatch(/Could not acquire lock/)` at `import.test.ts:312`, `import.test.ts:745`, `orphan.test.ts:336`, `state-refresh-observed.test.ts:198`, and `drift.test.ts:1290`.

In the `drift.test.ts:1290` fence, also assert the provider `update` was not called. It is currently an inline `update: vi.fn()` at `drift.test.ts:1304`, so hoist it into a local the way the tests at `drift.test.ts:1467` and `:1545` do:

```ts
const updateMock = vi.fn();
mockRegistryGetProvider.mockReturnValue({
  readCurrentState: async () => ({ VersioningConfiguration: { Status: 'Suspended' } }),
  update: updateMock,
});
...
expect(updateMock).not.toHaveBeenCalled();
```

That fence asserts `saveState` and `releaseLock` were not called, but for `--revert` the live-AWS mutation is the load-bearing part of the path — it is the one thing that must not happen under a foreign lock.

### 6. `docs/troubleshooting.md:18-32` — document the new message

That section (`## Lock Issues`, lines 18-32) documents only the retry helper's final error, `Failed to acquire lock for stack 'MyStack' after 3 attempts. / Locked by: ...`. Six commands now emit a different, non-retrying `Could not acquire lock for stack '<name>' (<region>)` instead. A couple of lines saying which commands fail fast, and that `cdkd force-unlock <stack> --stack-region <region>` is the recovery, is enough.

### 7. PR body — call out the two behaviour changes

`cdkd import --dry-run` now aborts on contention (it deliberately takes the lock, unlike `orphan --dry-run` which skips acquisition), and `destroy --all` now aborts the whole run at the first contended stack instead of continuing to the rest. Both are correct and I am not asking you to change them — but a reader should not have to infer them from the diff.

---

Once these are pushed I will take it from there: approve the CI workflow run, run the real-AWS integration test, and push the ledger row to this branch. As above, please do not run an integ yourself — besides the cost landing on you, it would not satisfy the gate anyway: the markgate markers are local to the worktree I merge from, and they digest this branch's delta against `origin/main`, so the run has to happen here and on the final head.

I filed https://github.com/go-to-k/cdkd/issues/2168 for the TTL residual you flagged, since it deserves more than a footnote: the TTL defaults to 30 minutes (`lock-manager.ts:53`) and nothing renews it, so any operation running past it loses mutual exclusion outright — and several providers wait well beyond that by design (`fsx-filesystem-provider.ts:57` and `emr-cluster-provider.ts:56` allow 60 minutes, `custom-resource-provider.ts:218` an hour, matching CDK's default `totalTimeout`). Correctly out of scope here; thanks for drawing the line in the issue rather than expanding the PR.

nix-tkobayashi:
Thanks for the review — all seven points are addressed in the latest push (`4d52e2c4`):

1. `destroy-runner.ts`: moved the `Acquiring lock...` log **inside** the protected `try` (it writes to stdout and can `EPIPE`, and it sits after the SIGINT registration and the cross-region switch, so a throw outside the `try` would leak both).
2–3. Every fail-fast `force-unlock` hint now carries `--stack-region <region>`, so `cdkd force-unlock` targets the right region even for a first-time `import` or a nested-stack child that has no state record yet.
4–5. Hardened the contention fences: each now asserts the surfaced error is the lock error (`/Could not acquire lock/`); `drift` asserts the provider's live `update` never fired; the nested-child `import` fence adds positive markers that the child acquire was actually attempted; and `orphan` stubs state so a reverted guard would actually reach `saveState` (otherwise the assertion was vacuous). Re-verified each with a mutation probe.
6. `docs/troubleshooting.md` documents the fail-fast message and the `--stack-region` recovery.
7. PR body updated with the two behaviour-change callouts (`import --dry-run` aborts on contention; `destroy --all` aborts the whole run at the first contended stack).

Two consistency items I also folded in while there (same concern, no behaviour-break risk):

- `cdkd export` already fails fast with the identical message, but its two `force-unlock` hints still lacked `--stack-region`; aligned them so the message class is spelled consistently across every site.
- The troubleshooting note now counts seven fail-fast commands (including `cdkd export`) and notes that `export`'s nested-stack children retry briefly before failing.

typecheck / lint / build are green and the full unit suite passes (15512/15512).

go-to-k:
Merged — thank you for the contribution, and for the quality of it.

The diagnosis was the hard part here: a `false` return that every call site read as "it threw on failure" is the kind of bug that survives review for a long time, and you found it, reproduced it against real AWS with the post-conditions spelled out, and drew the scope line in the right place (leaving the owner-blind release and the TTL residual as separate concerns rather than growing the PR). The fix then went further than the report needed it to — extracting one idempotent cleanup helper and fencing every pre-existing cross-region leak it exposed, each with a mutation probe.

Verified before merge: all 11 guards mutation-probed here as well, the full unit suite green, `drift-revert` and `lambda` run against real AWS (0 errors / 0 orphans each), and a live test with a foreign lock planted on a deployed stack — `destroy` refused and left the other owner's lock and `state.json` untouched. A 3-axis plus security review returned zero blockers.

Shipped in v0.284.38. The reviewers' minor findings are filed as #2170 and #2171, and the TTL residual you flagged is #2168; I am picking all three up now, so nothing further is needed from you.

PR #2186, opened 2026-08-23
fix(local): keep ECS run-task secrets and credentials off the docker argv

nix-tkobayashi:
## Summary

Fixes #2183. `cdkd local run-task` wrote AWS credentials and resolved Secrets Manager / SSM secret **values** into the `docker run` argument vector as `-e KEY=value`, so they were visible in `/proc/<pid>/cmdline` to any local process (co-tenant users, monitoring / telemetry agents, crash collectors). The Lambda path (`docker-runner.ts`'s `runDetached`) already avoided this with docker's value-less `-e KEY` form (the value is read from the spawn-time process env), so this was an asymmetry between two sibling modules rather than a missing capability.

Leaked material: static or session AWS credentials, plus every resolved Secrets Manager / SSM secret referenced by the task definition.

## Fix

Sensitive keys — the AWS credential set (`SENSITIVE_ENV_KEYS`) plus each **resolved secret name** — are emitted value-less as `-e KEY`, and the value is handed to docker through the spawn-time `env` instead of the argv:

- `src/local/ecs-task-runner.ts`: `buildDockerRunArgs` now returns `{ args, sensitiveEnv, collisions }`; the orchestrator threads `sensitiveEnv` into the real `execFileAsync('docker', args, { env })` spawn.
- `src/local/ecs-network.ts`: the metadata sidecar's AWS credentials get the same treatment.

Non-sensitive template env / `--env-vars` overrides stay as `-e KEY=value` (unchanged).

### Why not `--env-file`

`docker run --env-file` was rejected: it is line-based and **truncates a multi-line value at the first newline** (confirmed empirically — a PEM-style value came back cut off), which would silently corrupt multi-line ECS secrets. The env channel preserves them.

### Client-env hijack protection

[...]

Issue #2183, opened 2026-08-23
fix(local): ECS local run-task and metadata sidecar leak AWS credentials and resolved secrets via docker argv (/proc/<pid>/cmdline)


nix-tkobayashi:
## Summary

The `cdkd local run-task` ECS path writes AWS credentials and resolved Secrets Manager / SSM secret values into the `docker run` argument vector as `-e KEY=value`, so they are visible to any local process that can read `/proc/<pid>/cmdline` (co-tenant users, monitoring/telemetry agents, crash collectors). The Lambda path already avoids this via docker's value-from-process-env form (`-e KEY`, no `=value`), so this is an asymmetry between two sibling modules rather than a missing capability.

## Affected sites (current `main`)

1. **Metadata sidecar — AWS credentials** — `src/local/ecs-network.ts`
   - Lines 170-173 fold `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` / `AWS_SESSION_TOKEN` into `sidecarEnv`.
   - Line 178 pushes every entry as `-e ${k}=${v}` into `sidecarArgs`, run via `execFileAsync(getDockerCmd(), sidecarArgs)` at line 184.

2. **Task containers — resolved secrets + all env** — `src/local/ecs-task-runner.ts`
   - Line 1047 folds resolved Secrets Manager / SSM secret VALUES into `finalEnv` (`for (const s of secrets) finalEnv[s.name] = s.value;`).
   - Lines 1056-1058 push every `finalEnv` entry as `-e ${k}=${v}`; the resulting `args` are run via `execFileAsync(getDockerCmd(), args)` at line 453.

## Safe reference (Lambda path)

`src/local/docker-runner.ts`'s `runDetached` already does the right thing: for keys in `SENSITIVE_ENV_KEYS` (the built-in AWS credential set, extensible per call via `sensitiveEnvKeys`) it pushes `-e KEY` with NO `=value` and supplies the value through the spawn-time process environment, so the value never appears on argv. Non-credential env stays as `-e KEY=value`. See `docker-runner.ts:220` (`runDetached`), `:262-275` (the `-e KEY` vs `-e KEY=value` split), `:441` (`SENSITIVE_ENV_KEYS`), `:267-268` (`sensitiveEnvKeys` extension).

## Proposed fix

Bring both ECS `execFileAsync(getDockerCmd(), args)` sites onto that pattern:

[...]

PR #2187, opened 2026-08-23
fix(local): close the runDetached docker-client hijack (local invoke) — stacked on #2186

### LINKED ISSUE #2184: fix(local): runDetached (local invoke) lets a SecureString name hijack the docker client env — same class as #2183, deferred there
## Summary

`runDetached` in `src/local/docker-runner.ts` (the Lambda / AgentCore `cdkd local invoke` path) forwards sensitive env vars value-less as `-e KEY` and spawns docker with:

```ts
env: { ...process.env, ...passthroughEnv }
```

`passthroughEnv` OVERRIDES `process.env`, and its keys can be user/template-controlled: the AgentCore / `--from-cfn-stack` path extends `sensitiveEnvKeys` with SecureString SSM **names** (same trust level as an ECS task's secret names). So a SecureString whose name collides with a var the docker CLIENT reads — e.g. `DOCKER_HOST` — overrides `env.DOCKER_HOST` and redirects the docker client to a different daemon (or `PATH` breaks locating docker).

This is the exact docker-client hijack that go-to-k/cdkd#2183 closed for the ECS `run-task` path. That PR deliberately scoped itself to ECS and left `runDetached` as a documented deferral (see the `docker-cmd.ts` note in `.claude/rules/code-layout.md`).

## Fix

Route `runDetached`'s spawn env through the shared helper introduced in go-to-k/cdkd#2183:

```ts
env: dockerSpawnEnvWithSensitive(passthroughEnv)
```

`dockerSpawnEnvWithSensitive` (`src/utils/docker-cmd.ts`) starts from `process.env` and adds only the sensitive keys that do NOT name a docker-client var (`DOCKER_CLIENT_ENV_KEYS`, matched case-insensitively), so a colliding SecureString name cannot hijack the client. Add the same call-site collision `logger.warn` (via `dockerClientEnvCollisions`) so a dropped key is not silent.

## Verification plan

- Unit: a `sensitiveEnvKeys`-extended `runDetached` call with a `DOCKER_HOST` passthrough key must keep the client's own `DOCKER_HOST` authoritative in the spawn `env`.
- Integ: `local-invoke` (Docker) still delivers a normal SecureString value to the container.

## Notes

- Only the argv-vs-client-env HIJACK surface is at issue here; `runDetached` already keeps secret VALUES off the argv (that is the pattern go-to-k/cdkd#2183 mirrored).
- Surfaced by the security review of the go-to-k/cdkd#2183 PR.

nix-tkobayashi:
> **Stacked on go-to-k/cdkd#2186.** This reuses the `partitionSensitiveEnv` helper that go-to-k/cdkd#2186 adds. Until it merges, the diff below includes its changes; it resolves to just the `src/local/docker-runner.ts` + test delta once go-to-k/cdkd#2186 lands. **Review the `docker-runner.ts` change; merge after go-to-k/cdkd#2186.**

## Summary

Fixes #2184. `runDetached` in `src/local/docker-runner.ts` (the Lambda / AgentCore `cdkd local invoke` path) already keeps secret **values** off the docker argv (value-less `-e KEY`, value from the spawn env). But a template-controlled **env var name** (the template's own variable names, `audit.sensitiveKeys`, via `--from-cfn-stack` — not the SSM SecureString parameter names) colliding with a variable the docker CLIENT reads — e.g. `DOCKER_HOST` — still emitted a value-less `-e KEY`, which docker then resolved against the client's **own** env, handing the container the HOST's value (the daemon socket, or a proxy credential). This is the exact host-value-injection blocker go-to-k/cdkd#2186 fixed for the ECS `run-task` path.

## Fix

`runDetached` now partitions its env through the shared `partitionSensitiveEnv` (added in go-to-k/cdkd#2186):

- a non-sensitive key stays `-e KEY=value` (unchanged);
- a sensitive key becomes value-less `-e KEY`, its value carried via the spawn `env` (`dockerSpawnEnvWithSensitive`);
- a sensitive key that **names a docker-client var** gets **no flag at all** — dropped entirely so the container never receives the host's value — and is reported in `collisions` for a `logger.warn`.

## Tests

- A `sensitiveEnvKeys`-extended `runDetached` call with a `DOCKER_HOST` secret asserts: no `-e DOCKER_HOST` flag on the argv, the client's own `DOCKER_HOST` stays authoritative in the spawn env, a non-colliding secret still reaches the container value-less, and the drop is **warned** (via a logger spy). Mutation-probed: re-emitting `-e KEY` for a collision reds it.

## Verification

- typecheck / lint / build green; the full unit suite passes.
- Behaviour is unchanged for every non-colliding key; only a pathological docker-client-named sensitive key is now dropped ENTIRELY (no `-e` flag — the container receives neither the template value nor the host's value) and warned, instead of injecting the host's value.


nix-tkobayashi:
### Verification

- **Adversarial review** looped to zero findings on the `docker-runner.ts` delta (the shared `dockerSpawnEnvWithSensitive` helper it reuses was already reviewed in go-to-k/cdkd#2186, where a security + code review both returned no blockers).
- **Mutation-probed**: reverting the spawn env back to the naive `{ ...process.env, ...passthroughEnv }` merge turns the new `DOCKER_HOST`-hijack fence red.
- typecheck / lint / build green; the full unit suite passes (15,726 tests).

### Merge order

This is **stacked on go-to-k/cdkd#2186** — it reuses the helper that PR adds. The diff shows both PRs' changes until go-to-k/cdkd#2186 merges, then resolves to just `src/local/docker-runner.ts` + its test. **Please merge go-to-k/cdkd#2186 first**; this one rebases to a ~15-line delta.

go-to-k:
Reviewed alongside #2186 (four passes: spec / code / test / security), citations re-verified against `f6e3209`.

The delta itself is right and minimal, and the stacking is clean — #2186 added a deferral note to `.claude/rules/code-layout.md` saying `runDetached` had not adopted the helper, and this PR correctly retires that same sentence rather than leaving two contradicting statements in the file. Nice.

**I verified the premise of #2184**, since the whole severity argument rests on it: `sensitiveEnvKeys` really is template-controlled on the AgentCore path — `local-invoke-agentcore.ts` folds `audit.sensitiveKeys` from `substituteEnvVarsFromStateAsync` into the set, which flows to `runDetached`'s `opts.sensitiveEnvKeys` and is matched against `Object.keys(opts.env)`. Reachable only through that path (`resolveTemplateSsmParameters` has a single caller), so plain `local invoke` / `start-api` cannot produce one — worth noting, but it does not weaken the fix.

Three things.

---

## 1. The blocker from #2186 applies here verbatim

`runDetached`'s argv partition has the same shape as `buildDockerRunArgs`:

https://github.com/go-to-k/cdkd/blob/f6e32097ac40881b9f5598f61bb72b1eeda8ad51/src/local/docker-runner.ts#L275-L283

A colliding key still gets `-e KEY` pushed while `dockerSpawnEnvWithSensitive` skips it, so docker resolves the flag against the client's own environment and the container receives the **host's** value rather than nothing.

I measured this on real docker 20.10.22, driving the exact argv/env pair the code produces (`--entrypoint /usr/bin/env`): a key named `PATH` gave the container the full macOS host `PATH` in place of the image's, and a key named `HTTPS_PROXY` gave it the host's `http://alice:hunter2@corp-proxy:3128`. A colliding key the host has *not* set is correctly absent — which is exactly the outcome the design wants, and shows the flag is what makes the set case unsafe. Full table in my #2186 comment.

The warning here says the keys "were not forwarded with their value":

https://github.com/go-to-k/cdkd/blob/f6e32097ac40881b9f5598f61bb72b1eeda8ad51/src/local/docker-runner.ts#L319-L324

which is only half true — nothing was forwarded *with the secret value*, but the host's value was. Same fix as #2186: consult `isDockerClientEnvKey(k)` in the partition and skip the `-e` push on collision. Since #2186 suggests lifting the partition into a shared `partitionSensitiveEnv()` helper, doing that there makes this PR's delta shrink to swapping one loop for one call, and the fix lands in both paths at once.

---

## 2. The new warning block is untested

Nothing in `docker-runner.test.ts` references the message, so `docker-runner.ts:319-324` can be deleted with the suite green. The block is the only signal a user gets that a `-e` key did not carry its intended value, so its absence is a silent-data-loss path rather than a cosmetic one.

The env-hijack test itself is good — `tests/unit/local/docker-runner.test.ts:420-450` pins `opts.env['DOCKER_HOST']` to the client's own socket, asserts a non-colliding secret still arrives, and checks neither value reaches the argv. That is the correctly-paired positive/negative shape. Note though that its fixture runs straight through the blocker above without asserting on it: with `DOCKER_HOST` in `sensitiveEnvKeys` the args still carry a value-less `-e DOCKER_HOST`, so in a real run that container would come up with the host's `unix:///var/run/docker.sock` in its environment. Adding an assertion on that would make the test pin the intended behaviour rather than the current one.

---

## 3. PR body wording

- "the AgentCore / `--from-cfn-stack` path extends `sensitiveEnvKeys` with SecureString **names**" — the keys added are the *environment variable* names from the template (`audit.sensitiveKeys`), not the SSM parameter names. The premise holds either way (both are template-controlled), but the sentence as written points at the wrong identifier.
- "A colliding key is dropped and **warned**" — per item 1 it is not dropped from the argv, and the host value takes its place.

Worth fixing so the body does not end up read as the spec.

---

Everything else checked out: no secret value reaches state / journal / events / exports, the warning interpolates key names only, `redactAwsCredentialsInArgs` at `:313` has nothing left to redact for these keys (values are off the argv), and no other spawn site in the file carries sensitive args. `fix(local):` prefix and English-only both match the repo's conventions.

Since this is stacked, the natural sequence is: fix the argv partition once in #2186 (ideally in a shared helper), then this PR rebases onto it.

go-to-k:
Follow-up to my review above — I built the fix locally and verified it, so here is the `runDetached` half as a patch rather than a description. It assumes #2186 lands the shared `partitionSensitiveEnv` helper (patch posted there); with that in place this PR's delta becomes one loop replaced by one call.

Also please read the correction I posted on #2186: the table-driven test I recommended there was self-referential and could not detect a deletion. It does not affect this PR's code, but it does affect the test I suggested.

---

## One thing you need to know before applying this

Once `partitionSensitiveEnv` excludes a colliding key from `passthroughEnv`, the `dockerSpawnEnvWithSensitive` call at the spawn site becomes **defence in depth rather than the load-bearing guard** — and that makes **this PR's existing test stop discriminating it**.

Measured: with the patch applied, reverting the spawn line back to `env: { ...process.env, ...passthroughEnv }` leaves `docker-runner.test.ts` **29/29 green**, because the colliding key is no longer in `passthroughEnv` for the naive merge to pick up. On this PR as it stands today the same mutation turns your test red, since the un-partitioned loop does put `DOCKER_HOST` into `passthroughEnv`.

That is not an argument against the change — moving the guard earlier is the right direction, and the helper itself stays directly fenced by its own unit tests in `docker-cmd.test.ts`. But it is worth knowing that after this refactor the spawn-site call is redundant by construction, so nobody should read a green suite as proof that it is still doing work. I noted it in the code comment in the patch.

---

## Verified

Applied on top of #2186's patch, in a worktree, with the repo's toolchain:

- `vp run test` — **15757 passed, 1 skipped, 0 failed**, no type errors (full suite)
- `npx tsc --noEmit` clean, `vp run lint` 0 errors, `vp run format` applied

Mutation probe on the new fence (anchor-count receipt confirmed each mutation applied, then reverted from a scratch copy):

| probe | result |
|---|---|
| colliding key gets its `-e KEY` flag back | **red** — 1 test |
| spawn env reverted to `{ ...process.env, ...passthroughEnv }` | green — see the caveat above |

<details>
<summary><b>Patch (src + test)</b></summary>

```diff
diff --git a/src/local/docker-runner.ts b/src/local/docker-runner.ts
index 5b12b28b..ab1a9f81 100644
--- a/src/local/docker-runner.ts
+++ b/src/local/docker-runner.ts
@@ -1,7 +1,13 @@
 import { execFile, spawn } from 'node:child_process';
 import { createServer } from 'node:net';
 import { promisify } from 'node:util';
-import { getDockerCmd, runDockerForeground, runDockerStreaming } from '../utils/docker-cmd.js';
+import {
+  dockerSpawnEnvWithSensitive,
+  getDockerCmd,
+  partitionSensitiveEnv,
+  runDockerForeground,
+  runDockerStreaming,
+} from '../utils/docker-cmd.js';
 import { getLogger } from '../utils/logger.js';
 
 const execFileAsync = promisify(execFile);
@@ -266,15 +272,12 @@ export async function runDetached(opts: DockerRunOptions): Promise<string> {
     opts.sensitiveEnvKeys && opts.sensitiveEnvKeys.size > 0
       ? new Set<string>([...SENSITIVE_ENV_KEYS, ...opts.sensitiveEnvKeys])
       : SENSITIVE_ENV_KEYS;
-  const passthroughEnv: Record<string, string> = {};
-  for (const [k, v] of Object.entries(opts.env)) {
-    if (sensitiveKeys.has(k)) {
-      args.push('-e', k);
-      passthroughEnv[k] = v;
-    } else {
-      args.push('-e', `${k}=${v}`);
-    }
-  }
+  const {
+    flags: envFlags,
+    sensitiveEnv: passthroughEnv,
+    collisions: clientCollisions,
+  } = partitionSensitiveEnv(opts.env, sensitiveKeys);
+  args.push(...envFlags);
 
   // Issue #440 — Lambda EphemeralStorage.Size: emit `--tmpfs
   // <target>:rw,size=<N>m` so the container's `/tmp` is capped at the
@@ -306,14 +309,26 @@ export async function runDetached(opts: DockerRunOptions): Promise<string> {
   const logger = getLogger().child('docker');
   logger.debug(`${getDockerCmd()} ${redactAwsCredentialsInArgs(args).join(' ')}`);
 
+  // A sensitive key whose NAME collides with a var the docker CLIENT itself
+  // reads reaches neither the argv nor the spawn env: forwarding the value
+  // would hijack the client, and emitting a value-less `-e KEY` would make
+  // docker resolve it against the CLIENT's env, handing the container the
+  // HOST's value instead. Warn so the drop is not silent (issue #2184).
+  if (clientCollisions.length > 0) {
+    logger.warn(
+      `Env var(s) ${clientCollisions.join(', ')} share a name with a docker-client environment variable and were NOT passed to the container at all (they would hijack the docker client).`
+    );
+  }
+
   try {
     const { stdout } = await execFileAsync(getDockerCmd(), args, {
       maxBuffer: 10 * 1024 * 1024,
-      // Forward the sensitive-key passthrough values via process env so
-      // docker picks them up for the `-e KEY` flags (the values never
-      // touched the argv). Inherit the rest of the parent's env so the
-      // user's local docker config (DOCKER_HOST etc.) keeps working.
-      env: { ...process.env, ...passthroughEnv },
+      // Forward the sensitive-key passthrough values via process env so docker
+      // picks them up for the value-less `-e KEY` flags (the values never touch
+      // the argv). `dockerSpawnEnvWithSensitive` keeps the docker client's own
+      // critical vars (DOCKER_HOST / PATH / ...) authoritative — defence in
+      // depth, since `partitionSensitiveEnv` already excluded them (issue #2184).
+      env: dockerSpawnEnvWithSensitive(passthroughEnv),
     });
     return stdout.trim();
   } catch (error) {
diff --git a/tests/unit/local/docker-runner.test.ts b/tests/unit/local/docker-runner.test.ts
index 78116e03..6ff84b65 100644
--- a/tests/unit/local/docker-runner.test.ts
+++ b/tests/unit/local/docker-runner.test.ts
@@ -452,6 +452,44 @@ describe('runDetached', () => {
     expect(pIdx).toBeGreaterThanOrEqual(0);
     expect(args[pIdx + 1]).toBe('127.0.0.1:56789:8000');
   });
+  it('emits no -e flag for the colliding key and keeps the client env authoritative', async () => {
+    // The AgentCore / --from-cfn-stack path extends sensitiveEnvKeys with
+    // template-controlled env var NAMES. A value-less `-e DOCKER_HOST` would
+    // be resolved by docker against the CLIENT's own environment, so the
+    // container would receive the HOST's docker socket rather than nothing.
+    const savedHost = process.env['DOCKER_HOST'];
+    process.env['DOCKER_HOST'] = 'unix:///var/run/docker.sock';
+    try {
+      await runDetached({
+        image: 'my-image:latest',
+        mounts: [],
+        env: { DOCKER_HOST: 'tcp://attacker.example:2375', DB_PASSWORD: 'real-secret' },
+        sensitiveEnvKeys: new Set(['DOCKER_HOST', 'DB_PASSWORD']),
+        cmd: [],
+        hostPort: 9000,
+      });
+      const joined = lastArgs().join(' ');
+      // Neither `-e DOCKER_HOST` nor `-e DOCKER_HOST=<value>` on the argv.
+      expect(lastArgs()).not.toContain('DOCKER_HOST');
+      expect(joined).not.toContain('DOCKER_HOST');
+      expect(joined).not.toContain('tcp://attacker.example:2375');
+
+      const calls = childProcessMock.execFile.mock.calls;
+      const opts = calls[calls.length - 1]![2] as { env?: Record<string, string> };
+      // The client's own value is untouched...
+      expect(opts.env?.['DOCKER_HOST']).toBe('unix:///var/run/docker.sock');
+      // ...and the secret value was not smuggled in either.
+      expect(opts.env?.['DOCKER_HOST']).not.toBe('tcp://attacker.example:2375');
+      // Positive control: an ordinary sensitive key is unaffected, so this
+      // test cannot pass by the whole passthrough being broken.
+      expect(lastArgs()).toContain('DB_PASSWORD');
+      expect(joined).not.toContain('real-secret');
+      expect(opts.env?.['DB_PASSWORD']).toBe('real-secret');
+    } finally {
+      if (savedHost === undefined) delete process.env['DOCKER_HOST'];
+      else process.env['DOCKER_HOST'] = savedHost;
+    }
+  });
 });
 
 // `pullImage` exercises `runDockerStreaming` (default compact log level
```

</details>

---

The new test is placed inside the existing `describe('runDetached', ...)` block so it can use that scope's `lastArgs()` helper. It keeps a positive control (`DB_PASSWORD` still arrives with its value, and still never on the argv) so it cannot pass by the whole passthrough being broken.

Take, adapt, or ignore.

go-to-k:
Same summary as I just posted on #2186, for this PR's delta. Nothing new is reported here.

| # | Item | Blocking? | In my patch? |
|---|---|---|---|
| 1 | Same collision blocker in `runDetached` — a colliding key keeps its `-e KEY`, so the container gets the **host's** value | **Yes** | Yes |
| 2 | The `dockerClientEnvCollisions` warn block is untested (deleting it keeps the suite green) | No | **No** — the patch fences `collisions` as a return value, not the `logger.warn` call |
| 3 | PR body: "extends `sensitiveEnvKeys` with SecureString **names**" — they are the template's **env var** names (`audit.sensitiveKeys`); and "a colliding key is dropped" is inaccurate per item 1 | No, but worth fixing before merge | **No** |
| 4 | Rebase onto #2186's `partitionSensitiveEnv` once it lands | — | Assumed by the patch |

**If you apply the patch as-is, what remains is 2 and 3.**

Item 1 is the only one I would hold the merge on.

**Merge order:** #2186 first (it carries the shared helper), then rebase this one — at which point this PR's diff is the `docker-runner.ts` change plus its test. And please do read the caveat in my previous comment: after the refactor the spawn-site `dockerSpawnEnvWithSensitive` call is redundant by construction, so this PR's existing env-hijack test no longer discriminates it. The helper stays fenced by its own unit tests; just do not read a green suite as proof that the spawn-site call is load-bearing.

go-to-k:
Same as the disposition I just posted on #2186, for this PR's two non-blocking items.

**Do in this PR:**

- **2** (the `dockerClientEnvCollisions` warn block is untested). I suggested this might be a follow-up; on checking, that was too cautious. **508 test files in this repo already mock the logger**, so the pattern is cheap and well-worn — `tests/unit/local/cfn-local-state-provider.test.ts:6-14` is the shape to copy. `docker-runner.ts` takes `getLogger().child('docker')`, so the mocked `getLogger` needs to return an object whose `child()` returns the spied logger. The warn is the only signal a user gets that a key did not reach the container, so it is worth a spy rather than an issue.
- **3** (PR body wording — "SecureString names" should be the template's env var names, and "dropped" is inaccurate). The body becomes the squash commit message, so it wants to be right before merge.

**Nothing here belongs in a follow-up issue.** The one item on #2186 I suggested filing (the podman / containerd connection vars) sits in `docker-cmd.ts`, so it is #2186's to track, not this PR's.

**1 is the only item I would hold the merge on.**

nix-tkobayashi:
Updated to follow the go-to-k/cdkd#2186 review. Rebased onto the reviewed go-to-k/cdkd#2186 head and reworked `runDetached` to route through the shared `partitionSensitiveEnv` — so the **same blocker fix applies here**: a docker-client-name collision (e.g. a `--from-cfn-stack` SecureString named `DOCKER_HOST`) is now dropped **entirely** (no `-e` flag) rather than emitting a value-less `-e KEY` that docker would resolve against the client's own env and hand the container the host's value. The collision is reported and warned.

Verified: adversarial pass to zero findings on the `docker-runner.ts` delta, full unit suite green, and the drop-entirely + warn fence mutation-probed (re-emitting `-e KEY` for a collision reds it). Still stacked on go-to-k/cdkd#2186 — merge that first.

nix-tkobayashi:
Aligned with your review of this PR (thank you — I reworked before seeing your patch, and it converged on the same shape):

- **item 1 (blocker)** — `runDetached` now routes through the shared `partitionSensitiveEnv`, so a docker-client-name collision is dropped entirely rather than emitting a value-less `-e KEY` that resolves to the host's value.
- **item 2 (warn untested)** — added a logger-spy test asserting the collision `warn` fires (the key name + "were NOT passed to the container"); it reds on deleting the warn block.
- **item 3 (PR body)** — corrected: the body now says the colliding name is the template's **env var name** (`audit.sensitiveKeys`), not the SSM SecureString parameter name, and that a colliding key is dropped entirely (the container receives neither the template value nor the host's value).
- **your caveat** — noted in the spawn comment that `dockerSpawnEnvWithSensitive(passthroughEnv)` is now defence in depth, redundant by construction after the partition, so a green suite is not proof that call is load-bearing.

The env-hijack test now asserts the drop (no `-e DOCKER_HOST` flag) rather than running through the old blocker. Mutation-probed: re-emitting `-e KEY` for a collision reds it. Full suite green. Still stacked on go-to-k/cdkd#2186.

go-to-k:
Reviewed your `2eca6692` delta over #2186's head. The `runDetached` rework is right: it routes through the shared `partitionSensitiveEnv`, so a collision is dropped entirely, and the spawn-site comment records the defence-in-depth caveat I raised — that is exactly what I was asking for. The env-hijack test now asserts the drop rather than running through the old blocker.

Nothing blocking on this PR's own delta.

Two things carry over from the #2186 round-2 review I just posted, both landing in files this PR inherits rather than changes:

- **`DOCKER_CLIENT_ENV_KEYS` is missing the loader / TLS-trust class** (`LD_PRELOAD`, `LD_LIBRARY_PATH`, `LD_AUDIT`, `DYLD_INSERT_LIBRARIES`, `DYLD_LIBRARY_PATH`, `SSL_CERT_FILE`, `SSL_CERT_DIR`). Worth noting for this PR specifically: the reachable-today instance of that class is **`runDetached` on `origin/main`**, via `audit.sensitiveKeys` from `--from-cfn-stack` — so once this PR lands, the denylist becomes the guard for the path where the problem actually exists. Fix it in #2186 and this one inherits it; no separate change needed here.
- **`"Mirrors runDetached"`** in `ecs-task-runner.ts` becomes true again once this merges, so nothing to do — flagging it only so the note in the #2186 thread does not read as an outstanding item here.

Sequence from my side: #2186 fixes land → I re-run the review tier there (the marker is sha-bound) → `local-*` integ → merge #2186 → you rebase this → same checks → merge. This PR's delta needs no further changes from you right now.

nix-tkobayashi:
Rebased onto the updated #2186 HEAD (final denylist); the runDetached diff is unchanged, only the `code-layout.md` merge resolution differs. Force-push changed the sha, so any sha-bound review is invalidated — happy to revert if you'd rather handle the rebase at merge time.

go-to-k:
Heads-up rather than a request — nothing to change here.

Round 4 on #2186 turned up a blocking finding that this PR inherits: a secret NAME containing `=` bypasses `isDockerClientEnvKey` entirely, because the denylist matches the whole key string while the OS parses the variable name as everything before the first `=`. Measured at #2186's head: a key named `PATH=/tmp/evil:` gets `predicate false`, `collisions []` (so no warning), a `-e` flag, AND an entry in the spawn env — and since `dockerSpawnEnvWithSensitive` writes after `{ ...process.env }`, the poisoned duplicate wins:

```
sh -c 'printf "%s" "$PATH"'   ->  /tmp/evil:=SECRETVALUE
sh -c 'command -v ls'         ->  fails
```

The docker CLI execs `docker-credential-*` off `PATH`, so that is code execution as the operator.

Why it matters here specifically: `runDetached` routes through both `partitionSensitiveEnv` and `dockerSpawnEnvWithSensitive`, and its key names are template-controlled via `audit.sensitiveKeys` on the `--from-cfn-stack` path — the same reachability argument as #2184's original `DOCKER_HOST` case. So this PR's path is affected identically, and the fix landing in #2186 (drop-and-warn on a key containing `=`, plus the belt-and-braces guard inside `dockerSpawnEnvWithSensitive`) is what covers it.

No action needed on this branch. Merge order is unchanged: #2186 fixes → tier re-run → `local-*` integ → merge #2186 → you rebase → same checks here.

go-to-k:
go-to-k/cdkd#2186 is merged (squashed as `b7e0f697`). **Please rebase this branch onto the new `main` before it goes any further — as it stands it would revert the last three review rounds.**

This is not a conflict warning, it is a content one. `4b0a1fe6` carries the round-2-era `src/utils/docker-cmd.ts`, and `main` now carries the round-6 one. Measured just now:

| | this branch | `main` |
|---|---|---|
| `DOCKER_CLIENT_ENV_KEYS` size | **34** | **47** |
| `isMalformedEnvKey` | absent | present |
| `AWS_ENDPOINT_URL_` prefix | absent | present |
| `SSH_SK_PROVIDER` / `GCONV_PATH` / `BASH_ENV` | absent | present |

So merging as-is would take back, among others:

```
REVERTS: export function isMalformedEnvKey(key: string)     <- the round-4 code-execution fix
REVERTS: export const DOCKER_CLIENT_ENV_PREFIXES = ['LD_', 'DYLD_', 'AWS_ENDPOINT_URL_']
REVERTS: 'AWS_ENDPOINT_URL', 'AWS_CA_BUNDLE', 'AWS_PROFILE', 'AWS_ROLE_ARN'
REVERTS: 'GCONV_PATH', 'BASH_ENV', 'SSH_SK_PROVIDER'
```

The first line is the one that matters most: without `isMalformedEnvKey`, a secret named `PATH=/tmp/evil:` is back to poisoning the docker client's `PATH` and giving code execution as the operator. That is the finding this PR's own `runDetached` path is most exposed to, since its key names are template-controlled via `audit.sensitiveKeys`.

GitHub still reports `mergeable=UNKNOWN` while it recomputes against the moved base, so this may well surface as a plain conflict too — but I did not want the resolution decided by whichever side git picks.

After the rebase the delta should shrink to what we actually reviewed here: `src/local/docker-runner.ts` plus its test, and whatever `code-layout.md` / changelog resolution the rebase needs. If it comes out larger than that, something from the old stack survived.

Once it is up I will re-check the delta, re-bind the review marker to the new head, and merge — no real-AWS re-run needed if the change stays inside `src/local/docker-runner.ts` (the `integ-local` marker is fresh and `docker-runner.ts` is in its scope, so I will confirm rather than assume).

nix-tkobayashi:
Rebased onto the new `main` (the merged go-to-k/cdkd#2186 at `b7e0f697`). Replayed only the runDetached commit onto main and dropped the three old go-to-k/cdkd#2186 stack commits, so nothing from the round-2 era survives:

- `src/utils/docker-cmd.ts` is **byte-identical to `main`** — `isMalformedEnvKey`, the 47-key set, the `AWS_ENDPOINT_URL_` prefix and the rest of the round-6 shape are all intact (nothing reverted).
- The delta is exactly what you said it should be: `src/local/docker-runner.ts` + its test, plus the one-line `code-layout.md` / changelog resolutions. The changelog keeps `main`'s round-6 go-to-k/cdkd#2186 entry and adds the go-to-k/cdkd#2184 entry; `code-layout.md`'s docker-cmd note now says the helper IS adopted by `runDetached` (it said "NOT yet adopted... tracked as #2184").

**One correctness fix the rebase forced, flagged rather than smuggled:** post-go-to-k/cdkd#2186, the shared `partitionSensitiveEnv`'s `collisions` array now also contains **malformed** keys (empty / `=` / NUL), not just docker-client-var name collisions. `runDetached`'s warn text read "share a name with a docker-client environment variable" — which would mislabel a malformed-key drop. I updated it to mirror the ECS-side warn go-to-k/cdkd#2186 round-5 already shipped ("...or have a malformed name (empty, or containing '=' / NUL)..."), and added a malformed-key test (`env: { 'PATH=/tmp/evil:': ... }` → no flag, no spawn-env entry, warning names the malformed cause). Mutation-probed: reverting the warn text to the docker-client-only wording reds it. This is the only behaviourally-meaningful change beyond the mechanical rebase, and it lands in `docker-runner.ts` (inside the delta you expected).

An independent Codex pass over the delta returned no findings (confirmed the four-file scope, the byte-identical `docker-cmd.ts`, the helper call correctness, and the warn accuracy).

Verification: typecheck / lint / CI-parity check / build green; the docker-runner suite is green locally (30 tests, incl. the new malformed case); full suite run with the same WSL2 flake caveat (every failure a `Test timed out` in the heavyweight checker files, each green in isolation), CI is the parity run. Over to you for the delta re-check, the marker re-bind, and merge — `integ-local` is fresh and `docker-runner.ts` is in its scope, so no real-AWS re-run should be needed, as you said.

go-to-k:
Merged! Thank you — both of these are in now.

The rebase was exactly right. I checked the thing I was worried about rather than taking it on trust: `src/utils/docker-cmd.ts` at your head is byte-identical to `main`, so nothing from go-to-k/cdkd#2186's six rounds came back. And you flagged the one behavioural change the rebase forced — the warn text having to name the malformed cause too, now that the shared `collisions` array carries both — instead of folding it in quietly. That is the part that makes a stacked PR reviewable.

Two independent reviewers came back with no blockers. Verified here before merging: full suite 15804 passing, CI 6/6, and `local-invoke` green against real Docker with a clean sweep — that is the fixture that actually drives `runDetached`, so it exercises this change directly.

One thing I hit while gating this, unrelated to your work: `local-invoke-agentcore` no longer synthesizes. `aws-cdk-lib@2.244.0` dropped the L2 constructs that fixture imports from `aws-bedrockagentcore` (only `CfnRuntime` / `CfnRuntimeEndpoint` are exported now), so it fails at synth before reaching a single assertion. Upstream fixture staleness, nothing to do with go-to-k/cdkd#2184 — I am filing it and recording the FAIL row myself.

Also mine to clean up, not yours: the `docker-cmd.ts` comment and the go-to-k/cdkd#2183 changelog entry both still say `runDetached` "still spreads its passthrough env directly", which this merge just made false. Both live in files your PR deliberately kept byte-identical, so they belong in the follow-up I already have open.

Thanks for the whole run of this — seven pushes across the two PRs, and every round you took the finding rather than arguing it. The denylist that came out the other end is a lot better than the one either of us started with.

Issue #1919, opened 2026-08-15
fix(state): an Export.Name colliding with an output NAME makes the outputs redaction take the wrong position source


go-to-k:
## Context

Found by the security reviewer on PR (#1918) (the (#1910) / (#1916) redaction
lane). It is a residual of the (#1910) work in `deploy-engine.ts`, not of the
skeleton matcher, and it is a distinct defect from either issue, so it is filed
rather than folded in.

## The bound

`deploy-engine.ts` aliases an output's value under its `Export.Name` so a
cross-stack consumer can find it:

```ts
outputs[exportName] = value;
```

A separate post-loop pass then populates the POSITION source for the redaction:

```ts
outputsTemplateSource[outputKey] = output.Value;   // for every output key
```

Those two loops key the same bag differently. When a template's `Export.Name`
COLLIDES with the NAME of another output, the alias writes output B's resolved
value under key A, while the source pass writes output A's unresolved value
under key A. `redactByPath`'s plain-string arm then returns the SOURCE leaf
verbatim, so A's `{{resolve:...}}` expression is persisted as B's value into the
exports index.

## Why it matters

The pre-(#1910) value scan was weaker but not wrong here: it returned an
expression that at least resolved to the value it replaced. The position pass
turns a name collision into a stored reference that names a DIFFERENT secret,
which is the wrong-reference class (#1910) exists to remove -- one layer up, in
the exports index rather than in `state.resources`.

## Likelihood

Low. It needs a stack where some output's `Export.Name` is spelled exactly like
another output's logical NAME, and at least one of the two to carry a secret
dynamic reference. CDK does not generate that shape; a hand-written template
can.

## Direction

Guard the alias write against a key that is also an output name -- either skip
the alias on a collision (and warn, since the export is then unreachable by
name anyway), or key the source pass by the same alias rule so the two bags
cannot disagree. Whichever is chosen, the two loops must agree on the key
space; the defect is that they do not.

## Verification

A unit case building an output set whose `Export.Name` collides with a sibling
output name, asserting the exports index keeps each value's OWN expression,
mutation-probed. No integ arm needed -- the fixture set has no such collision
and the bug is expressible entirely in the deploy engine's output handling.

Session-fit: next (not this session) -- a different subsystem
(`deploy-engine.ts` output aliasing) from the redaction matcher the current
lane changed, with no file overlap in the affected function, and it needs its
own key-space decision plus a review round / Effort: ~1-3 h (the guard, a unit
[...]

PR #2194, opened 2026-08-25
fix(state): scope Fn::ImportValue to a producer's declared exports so a same-named plain Output can't shadow a real export (#2193)

### LINKED ISSUE #2193: fix(state): the exports index and the Fn::ImportValue state scan treat EVERY output key as importable, so a same-named plain Output in another stack silently shadows a valid export
## Summary

`Fn::ImportValue` can bind to the WRONG stack's value, silently. cdkd's exports index and its `state.json` fallback scan both treat **every key of `state.outputs`** as an importable export name — but that bag holds plain Output names AND `Export.Name` aliases side by side. So a stack that merely declares a plain `CfnOutput` named `X` (no `Export`) is indexed as the producer of export `X`, and whichever stack wrote the index last wins.

Realistic shape: two network stacks in the same account + region (say staging and prod) both declare a plain `CfnOutput('VpcId')`; only prod exports it (`exportName: 'VpcId'`). An app stack with `Fn::ImportValue: VpcId` resolves to **whichever of the two deployed most recently** — with no warning, and with a value CloudFormation would never have handed out (CFn keeps export names in a separate namespace and refuses a second producer of the same export name).

Two consequences, both silent:

1. A **valid** import can be shadowed: a same-named plain Output in an unrelated stack replaces the real exporter in the index (last writer wins) or is found first by the scan.
2. An **invalid** template deploys: `Fn::ImportValue: DatabaseUrl` with no such export anywhere resolves as long as some stack has a plain Output called `DatabaseUrl`. CloudFormation rejects it.

## Evidence (current `main`)

Three readers, one shared assumption:

- `src/deployment/deploy-engine.ts` (`updateForStack(stackName, this.stackRegion, newState.outputs)`, in the post-save block) hands the exports index the FULL mixed bag.
- `src/state/export-index-store.ts` `applyStackUpdate`: `for (const [name, value] of Object.entries(outputs)) next.set(name, ...)` — every key becomes an index entry, and `next.set` overwrites an entry owned by a DIFFERENT stack with no check and no log. `rebuildFromStateBackend` does the same over every stack's `state.outputs`.
- `src/deployment/intrinsic-function-resolver.ts` fallback scan: `if (state.outputs && exportName in state.outputs)` — first stack found wins. This is also the arm `cdkd scrub` takes (it deliberately passes no `exportIndex`).

The intra-stack version of this collision (an export alias landing on a published output name of the SAME stack) is already refused and warned by `isExportAliasCollision` (issue #1919). The cross-stack version has no guard at all.

The root cause is that `state.json` does not record WHICH keys of `outputs` are export aliases. No reader can tell a plain Output from an export after the fact, so filtering in the deploy engine alone would leave the rebuild path, the fallback scan, `cdkd scrub`, and every existing state file wrong.

## Proposed fix (schema v9)

- `StackState.exportNames?: string[]` (v9). v9 writers ALWAYS emit the field, `[]` included — `undefined` is the discriminator for a pre-v9 record, which keeps reading as before (every key importable) until its next deploy populates the field. Transparent auto-migration, no user action.
- One predicate in `src/types/state.ts` — `importableOutputKeys(state)` = `state.exportNames ?? Object.keys(state.outputs)` — used by all three readers: the deploy engine's `updateForStack` call, `rebuildFromStateBackend`, and the resolver's fallback scan.
- `applyStackUpdate` / rebuild warn when a stack's export overwrites an entry another stack owns (a genuine duplicate export, which CloudFormation refuses outright; the index write is best-effort and post-save, so this is a warning rather than a refusal).
- `cdkd import` carries `exportNames` forward from the existing record (today it carries only `outputs`).
- Docs: `.claude/rules/state-schema.md`, `CLAUDE.md`, `docs/state-management.md`, `docs/cross-stack-references.md`, changelog.

Why a version bump rather than a bare optional field: an older binary that rewrites a record would silently DROP the field, regressing the stack to "every key importable" with no signal. The version literal makes that binary fail loudly instead, which is the contract every prior bump relied on.

## Verification plan

- Unit: `tests/unit/state/schema-v8-to-v9-migration.test.ts` (same shape as the v7→v8 file); index ingestion + collision warning; resolver scan; deploy-engine persistence. One mutation probe per guard.
- New integ `tests/integration/schema-v8-to-v9-migration/`: Producer (exports `X`), Decoy (plain Output named `X`, different value), Consumer (`Fn::ImportValue X`). Phase 1 deploys all three under the latest v8 binary (`@go-to-k/cdkd@0.284.45`) and asserts the Consumer received the DECOY's value — the bug reproduced on real AWS. Phase 2 re-deploys under the local v9 binary and asserts `version: 9`, `exportNames` present on the producer (and `[]` on the decoy), and the Consumer now holding the PRODUCER's value. Destroy clean.
- Existing broad integ `multi-stack-deps` (uses `exportName`) for the cross-cutting deploy-engine / resolver change.

Working on this — branch `fix/export-index-plain-output-shadowing` (worktree), files: `src/types/state.ts`, `src/deployment/deploy-engine.ts`, `src/state/export-index-store.ts`, `src/deployment/intrinsic-function-resolver.ts`, `src/cli/commands/import.ts`, the tests and docs above.

nix-tkobayashi:
Closes #2193

## What was wrong

`state.outputs` holds plain Output names AND `Export.Name` aliases in one bag, and nothing in the record said which was which. Four readers derived "what does this stack export" by walking the bag wholesale:

- the exports index on `updateForStack` (fed the whole bag by the deploy engine) and on `rebuildFromStateBackend`;
- the resolver's `state.json` fallback scan (`exportName in state.outputs`);
- the `cdkd local` commands' `--from-state` fallback scan in `local-state-loader.ts` (same spelling).

So a plain `CfnOutput('VpcId')` in an unrelated stack was indexed as the producer of export `VpcId`, `applyStackUpdate` overwrote the real producer's entry with no check and no log (last writer wins), and a consumer's `Fn::ImportValue: VpcId` bound to whichever stack deployed most recently — a deterministic silent wrong value, and one CloudFormation would never hand out (its export namespace is separate; it refuses a second producer of one name). The intra-stack version of this collision was already refused by #1919's `isExportAliasCollision`; the cross-stack one had no guard, and no reader could have one, because the record did not carry the information.

## What this PR does

**Schema v9: `StackState.exportNames`** — exactly the alias keys `resolveOutputs` wrote (deduplicated) — and ONE predicate, `importableOutputKeys(state)` in `src/types/state.ts`, that all four readers go through.

- The discriminator is the FIELD, not the version. `undefined` (a pre-v9 record, or a v9 partial save carrying a pre-v9 bag forward) keeps the legacy every-key rule, so no existing cross-stack reference breaks on upgrade. `[]` means the stack is known to export nothing — so unlike `imports` / `outputReads`, an empty array is WRITTEN, never omitted.
- A save that re-resolves outputs (success path, no-change refresh) always writes the set. The five failure-path saves that carry `outputs: currentState.outputs` forward spread `exportNamesCarriedFrom(currentState)` beside it; the fresh in-memory placeholder for a stack with no record yet carries `[]` (a first deploy that fails before its outputs resolve must not persist "not known"). The existing source-scan fence (`deploy-engine-cross-stack-read-writers.test.ts`) derives both rules from the file rather than trusting an enumeration.
- The no-change deploy path BACKFILLS a record that lacks the field and re-feeds the index with the exports only — the part that evicts the plain-name entries a pre-v9 deploy published, without which a producer whose template never changes would pollute the index forever.
- Two stacks both exporting one name keep the index's latest-writer policy but now WARN, on update and on rebuild. The warning names the transitional case too (a pre-v9 record's stale plain-output entry, cleared by that stack's next deploy) — it could not exist before v9, since it would have fired on every plain-output collision.
- `cdkd import` over an existing record carries the set with the outputs it already carried; the destroy snapshot writes `exportNames: []` beside its emptied bag.
- Why a version bump rather than a bare optional field: a v8 binary rewriting a v9 record would silently DROP the field and regress the stack to the every-key rule; the bump makes it fail loudly instead — the contract every prior bump relied on. Transparent auto-migration, no user action.

## Verification

- **Unit**: `tests/unit/state/schema-v8-to-v9-migration.test.ts` (in-memory contract); index store ingestion + collision warning; resolver scan and local loader scan (decoy-first / plain-name refused / legacy still served); deploy engine persisted set, no-change backfill, failure-path carry (absent stays absent, never `[]`; fresh placeholder writes `[]`), export-set dedupe; destroy snapshot; six existing collision tests re-pinned to "a refused alias publishes NOTHING to the index". **15 guards mutation-probed**, each red when its guard is reverted.
- **Integ (new, real AWS)**: `tests/integration/schema-v8-to-v9-migration/` — Producer (exports `X`), Decoy (plain output named `X`, different value), Consumer (`Fn::ImportValue X`). Deploys all three under `@go-to-k/cdkd@0.284.45` (the last v8 binary) and asserts the Consumer received the DECOY's value — the bug on real AWS — then re-deploys under this branch and asserts `version: 9`, the producer's `[X]`, the decoy's `[]`, the transitional warning, `imports[0]` naming the producer, and the Consumer rebound to the PRODUCER's value. Destroy clean.
- **Integ (broad)**: `drift-revert-vpc` (21-resource VPC + Lambda hyperplane-ENI + IAM + CustomResource) PASS, destroy 0 errors / 0 orphans — the cross-cutting backstop for the deploy-engine / resolver change, since a narrow feature integ never exercises multi-resource VPC/Lambda paths. (`export` was attempted first but its `cdkd export` → CloudFormation IMPORT step needs a `cdk bootstrap` toolkit stack this account lacks; its deploy+destroy of 22 resources still ran clean, exercising the same resolveOutputs path.)
- Codex review looped to zero findings (3 rounds; round 1 surfaced the `local-state-loader.ts` reader this PR had missed).
- Three parallel repo reviewers (spec / code / test): spec clean (no drift), test adequate, code nothing above minor. The three code nits were fixed (commit `e5d038f5`): `importableOutputs` now rebuilds through `Object.create(null)` (the resolver's `__proto__` own-key defense), a dead backfill ternary was inlined, and the duplicate-producer warning strips control bytes from the resolved export name.

## Docs

`CLAUDE.md`, `.claude/rules/state-schema.md`, `docs/state-management.md` (new `version: 9` section), `docs/cross-stack-references.md`, `docs/changelog-cdkd.md`; coverage matrices regenerated for the new fixture.

## Out of scope (recorded)

- Three low-severity test-coverage gaps the test reviewer flagged are **won't-do** (benign): `cdkd import`'s first-time `{ exportNames: [] }` else-branch is nil-impact (a first import's bag is `{}`, so `importableOutputKeys` returns `[]` whether the field is `[]` or absent); the "every key is already an export → no backfill" sub-arm is a benign suppressed no-op write whose predicate direction is already fenced; and the no-change `resolutionFailed` carry arm needs a simultaneous resolution-failure + observed-refresh on the no-change path (the carry helper itself is unit-tested). None can regress the shadowing bug.
- Backfilling `exportNames` on a legacy record from `cdkd scrub` (it already reconstructs the alias set from the template) would let a stack that never redeploys stop polluting the index without a deploy. Deferred: it widens scrub's contract, and the no-change backfill already covers every stack that deploys at all.

## Review round 2 (blocker fix)

- **No-change path now persists an effective-export-set change** (`exportSetChanged`: set comparison of `importableOutputKeys(currentState)` vs the freshly-resolved set, subsuming the pre-v9 backfill): a self-named export (`Export.Name` equal to the output's own key) toggled on a v9 record leaves the outputs bag byte-equal, and previously neither saved nor re-indexed — adding one never became importable, removing one left a phantom export. Kept out of `outputsChanged` so the "Outputs-only change" log and bag selection are unaffected; the index re-feed covers a simultaneous observed-refresh save. Unit tests pin both directions plus an over-fire guard.
- **Rebuild keeps the higher-`lastModified` producer** on a name collision, making the duplicate-producer warning's "deployed last" wording true on the rebuild arm too (review's non-blocking nit, fixed here).
- Maintainer-side verification at this head: `schema-v8-to-v9-migration` integ PASS (v8 shadowing reproduced under published 0.284.45, transparent v9 auto-migration, consumer rebound) and `export` broad integ PASS (deploy-engine / resolver / destroy-runner are cross-cutting scope), both 0 errors / 0 orphans.
- Remaining non-blocking residuals (rebuild tie-break edge tests; set-change summary warn under `resolutionFailed`) are tracked in (#2232).


go-to-k:
Thanks for the thorough PR — the three-party integ fixture (real published v8 binary reproducing the shadowing on real AWS before proving the fix) is exactly the right shape, and the spec/test review both came back clean. One blocker surfaced in code review, verified by hand against the branch.

## Blocker: an `exportNames`-only change on a v9 record is never persisted or indexed

`src/deployment/deploy-engine.ts:1773-1777` — the no-change arm saves only on `observedRefresh || outputsChanged || exportNamesBackfill`, and `exportNamesBackfill` requires `currentState.exportNames === undefined`, so it only ever fires for pre-v9 records.

But a **self-named export** changes the export set without changing the outputs bag. `isExportAliasCollision` (`src/deployment/outputs-export-alias.ts:217`) deliberately returns false when `exportName === outputKey`, and the alias write then lands on the same key with the same value:

1. A v9 record has a plain output `Foo` (`exportNames: []`). The user adds `Export: { Name: "Foo" }` to that output — a completely ordinary operation.
2. Redeploy: no resource diff → no-change path. `resolvedOutputs` is byte-equal to `persistedOutputs` (`outputs[exportName] = value` rewrote the same key), so `outputsChanged` is false; `exportNames` is `[]`, not `undefined`, so the backfill term is false. Nothing saves, nothing is indexed.
3. The deploy reports "up to date", state keeps `exportNames: []`, and the consumer's `Fn::ImportValue "Foo"` hard-fails: the index has no entry, and the state-scan fallback now correctly refuses a KNOWN-empty set — the pre-v9 every-key fallback that would have served it is gone.

The reverse direction is a phantom export: remove `Export.Name` from a self-named export and `exportNames: ["Foo"]` is carried forever — the export keeps being served with nothing in the template backing it.

**Suggested fix**: replace the `=== undefined` term with an effective-set comparison — save (and re-index) when `importableOutputKeys(currentState)` differs as a set from `this.resolvedExportNames`. That subsumes the pre-v9 backfill (`undefined` → every-key differs whenever there is anything to suppress) and catches both self-named directions. Two things to avoid while fixing:

- Don't fold the set difference into `outputsChanged` — that would flip the "Persisted Outputs-only change" log and the `outputs:` bag selection (`resolvedOutputs` vs `persistedOutputs`) for a change that is not an outputs-value change.
- The pure-`observedRefresh` save (gate at `deploy-engine.ts:1800`, index update gated at `:1842`) writes the corrected set into state but skips the index update — after the fix, an observed-refresh save that also changed the effective set should update the index too, or the index stays stale while state says otherwise.

A unit test for both directions (add `Export.Name` == key on an unchanged bag → state + index updated; remove it → entry evicted) would pin this — the existing backfill tests only cover the `undefined` arm.

## Non-blocking

- `src/state/export-index-store.ts:476` (rebuild): the warning says the index "keeps the latest writer / binds to whichever deployed last", but on the rebuild path the survivor is the last stack in `listStacks()` iteration order, not deploy order. Either compare `lastModified` when rebuilding or soften the message on that arm.
- Disclosed already in the PR body, no action needed: `drift-revert-vpc` instead of the planned `multi-stack-deps` as the broad integ is fine — both are broad-set.

Everything else checked out: all four wholesale-bag readers go through `importableOutputKeys`, the untouched state writers are safe (`rollback.ts` spreads the prior record; `deploy.ts` / `diff-recursive.ts` placeholders are in-memory only; scrub/orphan-rewriter rewrite values, not keys), the v8→v9 transparent-migration contract is pinned by both the unit blob test and the integ, and the unit assertions were traced to be red on pre-fix code.

go-to-k:
Maintainer-side integ run: `schema-v8-to-v9-migration` **PASS** on our AWS account (143 s, verify.sh flow) — v8 shadowing reproduced under the published 0.284.45 binary, v9 read of v8 state clean with no rewrite, redeploy auto-migrated all three stacks to `version: 9` with `exportNames` written (`[\"CdkdSchemaV8ToV9SharedName\"]` / `[]`), consumer rebound to the producer's value, destroy 0 errors 0 orphans. Ledger row recorded on our side; `integ-schema-migration` + `integ-destroy` markers set for this head (they will go stale on your next push and we re-run then as needed). The blocker in the previous comment still stands before merge.

go-to-k:
To make the path to merge unambiguous:

**Required before merge (blocker):**
- [ ] Fix the no-change arm so an effective-set change saves AND re-indexes (the set-comparison approach above, or equivalent)
- [ ] Cover the `observedRefresh`-only save's index gap as part of it
- [ ] Unit tests for both self-named directions (add `Export.Name` == key on an unchanged bag → state + index updated; remove it → entry evicted)

**Optional — your call, either is fine:**
- [ ] The rebuild-path warning wording (`export-index-store.ts:476`): fix it in this PR, or leave it and we'll file a follow-up issue — say which in the PR body.

After your push we re-run `/review-pr` and the `schema-v8-to-v9-migration` integ on our side (no action needed from you there); the integ fixture itself already passed maintainer-side as-is.

nix-tkobayashi:
Fixed the blocker, and pushed after rebasing onto the latest `main`.

**The blocker (an `exportNames`-only change never persisted/indexed):** replaced the `=== undefined` backfill term with the effective-set comparison you suggested — save (and re-index) when `importableOutputKeys(currentState)` differs as a set from the freshly-resolved set. It subsumes the pre-v9 backfill and catches both self-named directions. Both caveats handled:
- Kept OUT of `outputsChanged`, so the "Persisted Outputs-only change" log and the `resolvedOutputs`-vs-`persistedOutputs` bag selection are untouched — a set-only change writes `persistedOutputs` (byte-equal bag) with the corrected `exportNames`.
- The index re-feed gate is `outputsChanged || exportSetChanged`, so an observed-refresh save that also changed the effective set updates the index too (no stale-index-vs-state divergence).

New unit tests, both directions on an unchanged bag: adding `Export.Name` == the output key → state gets `exportNames: [Foo]` and the index is fed `{Foo: ...}`; removing it → `exportNames: []` and the index is re-fed `{}` (phantom evicted). Plus an over-fire guard (unchanged v9 set → no save / no index). Four new mutation-probed guards.

**Non-blocking (rebuild "deployed last" wording):** took the `lastModified` route — rebuild now keeps the more-recently-modified producer on a name collision, so the warning's claim holds on that arm too (a test proves the newer-`lastModified` stack wins even when iterated first).

**Docs:** `state-schema.md` / `state-management.md` / changelog now describe the no-change save as an effective-export-set change (pre-v9 migration OR a self-named toggle) rather than a pre-v9-only backfill.

Re-ran `schema-v8-to-v9-migration` on the rebased tree (v8 shadowing reproduced under 0.284.45, v9 rebind, 0 errors / 0 orphans — note an SSM DeleteParameter lagged ~21 min on our side but completed clean). Codex looped to zero over the delta. Thanks for catching this — the self-named case was exactly the gap my reviewers and I missed.

go-to-k:
Verified the new head f9b1fcff is content-identical to 488dc073 (rebase offsets + ledger normalize only), so all review verdicts and the maintainer-side integ runs (schema-v8-to-v9-migration PASS, export broad PASS, both 0 err / 0 orphans) carry over. Finalizing the merge now — please hold further pushes; I'll add one ledger commit recording our runs and squash-merge.

Issue #2277, opened 2026-08-26
chore(local): delete the orphaned cdk-local fork under src/local now that the reachability critic has it annotated


go-to-k:
## Summary

go-to-k/cdkd#2228 asked for one of three outcomes: delete the orphaned cdk-local fork, mechanize the drift check, or annotate the orphans. The PR for it did the second and third together -- `scripts/check-local-reachability.ts` now fails when an exported value symbol under `src/local/**` has no live caller and no annotation, AND when an annotation is stale -- and deliberately did NOT do the first. This issue carries the deletion, with the audit already done so the next session does not repeat it.

The audit is the reason to file this rather than leave the annotations as the end state. An annotation stops a fix landing in dead code by accident; it does not stop the code being carried, compiled, typechecked, linted and unit-tested on every CI run, and it does not stop a reader having to work out which of two copies is real.

## What is orphaned, measured 2026-08-26 at b8f97714

29 exported symbols across 8 files carry `@no-live-caller` / `@test-only-export`. Four of the files are orphaned WHOLE:

| File | Lines | State | cdk-local equivalent |
| --- | --- | --- | --- |
| `src/local/vtl-engine.ts` | 1229 | loaded-only (all 5 exports dead) | `evaluateVtl`, `buildDefaultUtil`, `buildVtlInput`, `applyJsonPath`, `buildVtlRequestContext` all present in cdk-local 0.147.7's bundle |
| `src/local/httpv2-service-integration.ts` | 717 | unreferenced (all 5 exports dead) | `dispatchServiceIntegration`, `isSupportedSubtype`, `SUPPORTED_SUBTYPES`, `applyResponseParameters` all present |
| `src/local/cfn-local-state-provider.ts` | 436 | unreferenced (all 5 exports dead) | `CfnLocalStateProvider` is exported from the `cdk-local` package ROOT and re-exported by `src/cli/commands/local-state-source.ts:31`, which is what `--from-cfn-stack` actually uses |
| `src/local/reload-orchestrator.ts` | 343 | loaded-only (its 1 export dead) | none by that name; superseded IN-REPO by the non-exported `reloadAllServers` in `src/cli/commands/local-start-api.ts:2865`, which drives `--watch` for the N-server topology. Only the `NextStateMaterial` TYPE is still imported from it (`local-start-api.ts:108`) |

Two files are orphaned in PART and must be edited rather than deleted:

[...]

Issue #2276, opened 2026-08-26
chore(scripts): widen check-local-reachability's default scope from src/local to src


go-to-k:
## Summary

`scripts/check-local-reachability.ts` (issue go-to-k/cdkd#2228) requires every exported value symbol with a cdkd-authored body under `src/local/**` to be transitively reachable from a shipped entry point, or to carry `@no-live-caller` / `@test-only-export` at its declaration. Its default scope is `src/local` because that is where the cdk-local fork boundary runs. The rule is not specific to that directory, though, and the same run over the whole tree reports work the narrow scope does not.

## What a repo-wide run reports

Measured 2026-08-26 with `node --experimental-strip-types scripts/check-local-reachability.ts --scope=src` against `b8f97714`: roughly 30 unreachable exported symbols outside `src/local/**`, across ~25 files. The distribution is materially different from `src/local`'s, and that is why the default was NOT widened in the go-to-k/cdkd#2228 PR:

- The large majority are TEST-RESET SEAMS: `resetLiveRenderer`, `clearWriteOnlyPropertiesCache`, `clearCreateOnlyPropertiesCache`, `resetIdempotencyTokensForTests`, `clearOaiCanonicalUserIdCacheForTest`, `disarmInterruptWatchForTests`, `resetAccountInfoCache`, `resetEcrLoginCache`, `clearResolvedResourceTimeouts`. Each needs `@test-only-export` plus one line of reason. Mechanical, but ~20 declarations.
[...]

Issue #2203, opened 2026-08-25
fix(local): two more JSON.parse error-message echoes with the go-to-k/cdkd#2189 shape but a non-secret population (rie-client, vtl-engine)


go-to-k:
## Summary

Issue go-to-k/cdkd#2189 removed a `JSON.parse` error-message echo from `src/local/ecs-secrets-resolver.ts`, because V8 embeds a prefix of the PARSED INPUT in `SyntaxError.message` and the parsed input was a resolved secret. The sweep for that fix found two more sites with the **identical shape** but a **different population**, and they were deliberately left alone. This issue records that call so it is visible and challengeable rather than silently dropped.

## The two sites

- `src/local/rie-client.ts:466` -- parses a Lambda streaming-response prelude emitted by the function under test.
- `src/local/vtl-engine.ts:200` -- parses the argument of a VTL `$util.parseJson(...)`, which under `cdkd local start-api` can be the incoming HTTP request body.

Both interpolate the parser's message on failure, so a prefix of what they parsed reaches the terminal.

## Why they were NOT fixed in go-to-k/cdkd#2189

The criterion in go-to-k/cdkd#2189 is explicit: an interpolation whose thrown-from operation **consumed secret material cdkd resolved**. Neither of these does. They consume application payload that the developer supplied on their own machine, in their own terminal -- cdkd did not fetch it from a secret store, and the developer already holds it. Truncating the parse diagnostic there removes real debugging value from a local-emulation path with no cdkd-managed secret at stake.

`vtl-engine.ts` is the arguable one: a developer testing a login endpoint through `cdkd local start-api` has a password in the request body, and a malformed body would put a prefix of it on their terminal.

## What this issue is asking for

A decision, not necessarily a fix. Three admissible outcomes:

1. **Won't-do** -- close it, and put the reasoning in a comment at both sites so the next sweep does not re-raise it.
2. **Narrow fix** -- redact in `vtl-engine.ts` only (the request-body path), leave `rie-client.ts` alone.
3. **Both** -- apply the go-to-k/cdkd#2189 treatment (`err.name` plus a fixed discriminator) at both.

Whichever is chosen, the sites should end up carrying a comment saying so, since the shape is identical to one that WAS fixed and a future reader will otherwise assume an oversight.

## Evidence from the go-to-k/cdkd#2189 sweep (so it is not re-derived)

[...]

PR #2278, opened 2026-08-26
chore(local): fence cdkd's orphaned cdk-local forks with a symbol-level reachability critic

### LINKED ISSUE #2228: chore(local): cdkd forks the REST v1 dispatch chain but cdk-local's copy is what runs, so a fix can land in the fork and ship as a no-op
## Summary

cdkd carries a full fork of the REST v1 local-integration dispatch chain that **no longer executes**. `cdkd local start-api` runs cdk-local's implementation instead. The fork is still maintained -- it has tests, doc comments and active edits -- so a fix landed in it looks complete, passes CI, and changes nothing a user experiences.

This is not hypothetical. It happened on the go-to-k/cdkd#2203 lane in this run and was caught only by a security reviewer who booted the server and probed it live; unit tests, a code review and the author's own mutation probes had all passed.

## The evidence

- `src/local/http-server.ts` is a shim: it re-exports `startApiServer` from `cdk-local/internal`, and its own doc comment says cdkd "consumes it verbatim instead of carrying a byte-identical copy".
- `src/cli/commands/local-start-api.ts` calls that re-exported server with no dispatcher hook.
- These five symbols have **zero callers anywhere in `src/`** -- only in `tests/`:
  `dispatchMockIntegration`, `dispatchHttpIntegration`, `dispatchAwsLambdaIntegration` (`src/local/rest-v1-integrations.ts`), `invokeRieStreaming` (`src/local/rie-client.ts`), `evaluateVtl` (`src/local/vtl-engine.ts`).
- The single live import out of `rest-v1-integrations.ts` is `warnSsrfRiskyUri` (`local-start-api.ts:68`).
- `src/local/integration-response-selector.ts` is reachable only from `rest-v1-integrations.ts`, so it is orphaned by the same chain.

Measured against the installed `cdk-local@0.147.0` bundle: it carries its own `dispatchRestV1Integration`, its own `buildDefaultUtil().parseJson`, its own `vtlFailure` and its own `invokeRieStreaming` -- and those are what a running `cdkd local start-api` executes.

## Correction (2026-08-26): the scope is the start-api / HTTP chain ONLY

An earlier revision of this body implied `src/local/**` broadly. That is wrong and the correction matters, because it changes which fixes are at risk of shipping as no-ops.

`src/local/ecs-task-runner.ts` is a REAL implementation on cdkd's shipped path -- `src/cli/commands/local-run-task.ts:45` imports it directly -- and it in turn imports `ecs-secrets-resolver.ts` at `:29`. So `cdkl`-style task running and the ECS secret resolution under it are LIVE in cdkd, and go-to-k/cdkd#2189's fix genuinely reaches users.

What is shimmed is the **start-api / HTTP dispatch chain** specifically: `http-server.ts` re-exports `startApiServer` from `cdk-local/internal`, and the REST v1 dispatchers, `evaluateVtl` and `invokeRieStreaming` have no live `src/` caller behind it. The audit this issue asks for should therefore START from "which modules have no live importer" rather than assuming a directory-wide answer -- the boundary runs through the middle of `src/local/`, not around it.

## Why this is worse than ordinary dead code

Ordinary dead code is inert. This fork is a **decoy**: it is a plausible-looking, test-covered, comment-documented copy sitting at the path an author would search. Every signal a normal verification pass produces -- the grep finds it, the unit tests exercise it, a mutation probe reddens -- is satisfied by editing the copy that does not run. Only a live end-to-end probe can tell the difference, and nothing in the flow demands one for a change of this shape.

The same class already bit this repo once: go-to-k/cdkd#1972, where `src/local/lambda-authorizer.ts` outlived its move to cdk-local and kept a dead entry in four separate lists.

## What needs deciding

1. **Delete the fork** and keep only what is genuinely cdkd-owned (`warnSsrfRiskyUri`, and whatever else has a live importer). Cleanest, and it makes the next author's grep land on cdk-local where the code actually is. Cost: the tests covering the fork go too, and any behaviour cdkd intends to diverge on has to be re-expressed as a cdk-local option.
2. **Keep it and make the drift mechanical**: a test that fails when a cdkd file has no live `src/` importer, or one that diffs the fork against the cdk-local source it mirrors. Cheaper to land, but it preserves a copy that will keep attracting fixes.
3. **Keep it and mark it**: a header on each orphaned file naming cdk-local as the live owner. Weakest -- it is a comment asking the next reader to remember, which is what the two prior instances of this class already had.

Whichever is chosen, the audit should be repo-wide rather than limited to these four files: enumerate every `src/local/**` module and check each has a live importer, since the shim boundary has moved more than once.

Dup-check: searched open issues for `cdk-local shim`, `no runtime caller`, `dead code`, `rest-v1-integrations`, `vtl-engine` in titles and bodies. Hits were go-to-k/cdkd#1831 and go-to-k/cdkd#1821 (a cdk-local partition table, a different root cause) and go-to-k/cdkd#2203 (the lane that surfaced this, scoped to the error-message echo itself). No open issue names this root cause.
Session-fit: next (not this session) -- it is a decision between deleting a subsystem, writing a repo-wide reachability fence, or annotating; and the audit it implies spans every module under `src/local/**`, which is a different subsystem from the error-message echo this run was fixing.
Severity: medium -- no user-visible defect today, since the fork is inert. What it costs is that a future fix can land in it and ship as a no-op while every gate reports green, which is exactly what happened here and cost a full review round to catch.
Effort: medium (M) -- option 2 rides the existing unit suite and needs no new fixture; option 1 needs a re-review round because it deletes a subsystem and its tests.
Estimate: ~2-4 h -- dominated by the repo-wide reachability audit and by deciding, per orphaned module, whether cdkd intends to diverge from cdk-local or not; the mechanical fence itself is under an hour.

go-to-k:
## Summary

cdkd carries orphaned copies of code that cdk-local actually runs. The copies are test-covered, comment-documented, and sit exactly where a grep lands — so a fix can be written into one, pass every gate, and ship as a no-op. That is not hypothetical: it happened to go-to-k/cdkd#2203 earlier today and was caught only by a security reviewer who booted the server.

Closes #2228.

## Module-level reachability would NOT have caught it

The obvious fence — "every `src/local/**` module must have a live importer" — is green on the exact defect. `vtl-engine.ts` HAS a live importer: `rest-v1-integrations.ts` imports it, and `local-start-api.ts` imports `warnSsrfRiskyUri` from that. ESM loads the module; not one of its five exports is ever reached.

So the critic is **symbol-level**, computed over the TS AST from the build entries (`src/index.ts`, `src/cli/index.ts`), handling dynamic `import()`, namespace imports, `extends` clauses and inline type imports under `verbatimModuleSyntax`.

## The audit — 57 modules

| class | n | notable |
|---|---|---|
| live (all exports reached) | 12 | incl. `ecs-task-runner`, `ecs-secrets-resolver` — fully live, correcting an earlier assumption |
| live-partial | 4 | `rie-client` 2/6 — `invokeRie` and `waitForRieReady` ARE live; only the streaming half is dead |
| loaded-only (zero exports called) | 2 | `vtl-engine` (a fork), `reload-orchestrator` (an orphan — no cdk-local equivalent) |
| unreferenced | 2 | `httpv2-service-integration`, `cfn-local-state-provider` — both forks |
| shim-consumed | 28 | `http-server`, `sigv4-verify`, … — legitimate re-exports |
| shim-unreferenced | 8 | re-exports with no body, so they cannot act as decoys |
| types-only | 1 | |

All 29 "dead" verdicts were cross-checked by grep; zero false positives.

## Why both an annotation AND a fence

An annotation alone is a comment asking the next reader to remember — which is what already failed here twice. A fence alone does not reach the author: the tree already contains orphans, so any check that is green today must tolerate them, and a tolerated orphan is invisible at the moment someone edits it.

So it runs in **both directions**: an unannotated unreachable symbol FAILS, and an annotation on a symbol that has become reachable FAILS as stale. 29 `@no-live-caller` / `@test-only-export` annotations were added — comments only, no logic changed in `src/`.

Shims are excluded **structurally**, not by allowlist: a file declaring no value symbols never enters the population. `http-server.ts` gets no special-casing.

## Probes, both directions

21 mutations of the checker: **20 RED**, plus one deliberate semantics-preserving control that stayed GREEN (proving the suite is not failing for an unrelated reason).

Worth stating plainly: **6 of those were GREEN on the first run.** One was passing for the wrong reason, so the fixture was rebuilt as a discriminating type-only-vs-value pair; four covered shapes absent from the current tree, so a self-probe corpus was added for them; one was unobservable, so a `CDKD_SELF_PROBE_FORCE_FAIL=1` seam now proves it in one spawn.

Must-not-fire direction: 28 consumed shims, 8 unconsumed shims, types-only, all 55 live symbols and every annotated orphan produce zero findings.

## Two byproducts the calibration surfaced

- `ecr-puller.ts`'s doc claimed `isImageInLocalCache` is reused by `docker-image-builder`'s `--no-build` path. That file is a shim; the reuse does not exist.
- `tests/integration/local-start-api/verify.sh:263` claimed cdkd's `invokeRieStreaming` parses the prelude. cdk-local's copy is what runs.

## Deliberately NOT in this PR

Deletion of the orphaned forks (~4.4k src + ~3.7k test lines) is go-to-k/cdkd#2277, and widening the critic's scope beyond `src/local` is go-to-k/cdkd#2276. Deleting a subsystem is a different review question ("was it really unreachable?") from adding a critic, and bundling them puts two unrelated risk judgements in one diff.

## Verification

`vp run test` **806 files / 16,608 tests**, all pass. `vp run check` 0 errors, `typecheck:test` 0, `build` 0. The CI step lands inside `check-build-test` (verified by parsing the YAML).

Note for merging: this touches `src/local/**` — comments only, but the `integ-local` gate is path-based, not content-based — so it needs a Docker `local-*` integ before it can merge.


go-to-k:
Closing and re-opening as a fresh PR from the same branch.

GitHub fired no workflow at all for this PR after `666fd613` — zero check-runs on the head sha, commit status `pending`, and not even `PR Title Check`, which runs on every push. `ci.yml` has no path filters (`pull_request: branches: [main]`), so an empty commit is not the explanation, and other branches ran normally at 16:11Z, so it is not the earlier repo-wide outage either.

Tried and did not re-trigger: close + reopen, and two pushes of new shas. The close/reopen is the likeliest cause of the stuck state, so a fresh PR should clear it. Same branch, same commits — nothing is lost.

Issue #2274, opened 2026-08-26
fix(state): a custom resource's Data attributes are persisted verbatim, so a handler-generated secret sits in plaintext in state.json


go-to-k:
Dup-check: searched open issues for `state.json attributes secret plaintext`, and ran a body-window scan for `ResourceState.attributes` / `cfnResponse.Data`. No hits -- none covers this root cause. go-to-k/cdkd#2250 is the nearest neighbour and covers the `--verbose` LOG channel only; this is the durable one, and that issue's body explicitly (and wrongly) said this channel does not exist.

## Summary

A custom resource returns generated values in its response `Data` -- the documented place for a handler to hand back something it created, including a secret. cdkd persists that object verbatim into the state record:

```ts
// src/provisioning/providers/custom-resource-provider.ts:1128 and :1208
const attributes: Record<string, unknown> = cfnResponse.Data || {};
```

which `src/deployment/deploy-engine.ts` writes to `ResourceState.attributes`. So a secret returned by a custom-resource handler sits in plaintext in `s3://<state-bucket>/cdkd/<stack>/<region>/state.json`.

## Why this is filed separately from go-to-k/cdkd#2250

go-to-k/cdkd#2250 closed the `--verbose` log line. Its body claimed the value was "not otherwise exposed", relaying a flow trace that was correct about `cdkd gc`'s sweep and wrong about this. A security review of the fix re-derived the readers independently and found the persistence path; that issue now carries the correction.

The two are different problems with different remedies. A log line is transient and gated behind `--verbose`; the state record is durable, is read back on every subsequent operation, and lives in a bucket whose whole point is to outlive the run.

## Why it is not simply "the same as any other attribute"

Most `attributes` are AWS-assigned identifiers -- an ARN, a DNS name -- which are not secret. `Data` is different in kind: it is the one attribute source whose contents are chosen by USER-AUTHORED handler code, and the CloudFormation docs point handlers at it precisely for values the stack could not know in advance. That is the same shape as GHSA-p5qg-v9gv-hc7w, where a dynamic reference's resolved value reached the state file.

## Direction (to settle, not assume)

1. **Does the existing masking machinery already cover this and simply not run here?** This repo applies secret-masking helpers to several persistence paths. Check whether `attributes` is inside or outside that population before writing anything new -- if it is outside, the fix may be one call site rather than a mechanism.
[...]

Issue #2170, opened 2026-08-22
chore(state): follow-ups from the PR 2165 lock fail-fast review (contention message evidence, flag propagation, unfenced guard)


go-to-k:
Follow-ups from the 3-axis + security review of go-to-k/cdkd#2165 (the `acquireLock` discarded-return fix). None of these blocked that merge; they are collected here so the review output is not lost.

## 1. The contention message gives no evidence for its own hedge

All nine fail-fast sites print `... another cdkd process holds it. Wait for it to finish, or run 'cdkd force-unlock <stack> --stack-region <region>' if you are certain no other process is active.` — with no owner, no operation, and no expiry. `cdkd deploy` already prints `Locked by: user@hostname:12345, operation: deploy` from `acquireLockWithRetry` (`src/state/lock-manager.ts:464-472`), and `docs/troubleshooting.md` now contrasts the two messages side by side, so the gap is user-visible.

Why this matters more than a wording nit: `acquireLock` auto-reaps an EXPIRED foreign lock (`src/state/lock-manager.ts:223-230`), so **every message a user actually sees is a live lock younger than the TTL** — in practice a running `cdkd deploy`. The message asks the user to decide whether another process is active while withholding the only evidence that would answer it, and the likeliest response (`force-unlock`, then re-run) reproduces go-to-k/cdkd#2161's exact harm by hand.

Fix: one best-effort `getLockInfo` (a single GetObject) at each site, interpolating owner / operation / time-to-expiry; ignore a read failure and fall back to today's text.

Sites: `src/cli/commands/destroy-runner.ts:672`, `src/cli/commands/import.ts:617` and `:1944`, `src/cli/commands/orphan.ts:200`, `src/cli/commands/state.ts:2086`, `src/cli/commands/drift.ts:2251` and `:3238`, `src/cli/commands/export.ts:2623` and `:6024`.

Session-fit: next (not this session) — src change across nine sites, so it drags a broad integ re-run; belongs with the lock-contract work in go-to-k/cdkd#2168 rather than on its own.
Severity: medium — the recovery a user is steered toward can delete a live deploy's lock, but only when they act on the hedge without checking.
Effort: medium (M) — one re-review round plus a broad integ this session was not otherwise going to run.
Estimate: ~1-2 h — nine call sites plus their fences; the broad integ (`drift-revert` or `lambda`, ~2-4 min each) is the small part.

## 2. The hint propagates `--stack-region` but not the other resolution flags

[...]

PR #2281, opened 2026-08-26
fix(providers): log the custom-resource response ENVELOPE, not a 200-byte prefix of the body

### LINKED ISSUE #2250: fix(providers): the custom-resource poll debug-logs the first 200 bytes of the response body, so a generated secret in Data reaches --verbose output
Dup-check: searched the open issues for `custom-resource debug log Data`, for `verbose log secret redact`, and with a body-window scan for `substring(0, 200)` / `Got S3 response for` / `custom-resource-provider.ts:22`. No hits -- no open issue covers this root cause.

## Summary

`CustomResourceProvider` debug-logs the first 200 bytes of the custom-resource RESPONSE BODY on every poll. That body is the CloudFormation custom-resource response document, whose `Data` field is the documented place a custom resource returns a generated value -- including a generated secret.

`src/provisioning/providers/custom-resource-provider.ts:2278`:

```ts
const body = await response.Body?.transformToString();
if (body && body.length > 0) {
  this.logger.debug(`Got S3 response for ${logicalId}: ${body.substring(0, 200)}`);
```

The body has the shape `{"Status":...,"PhysicalResourceId":...,"Data":{...}}`, so with a short `PhysicalResourceId` the `Data` values land inside the 200-character window and reach the terminal -- and in CI, the retained build log.

## Why it matters

- It is unconditional for every custom-resource response, not a rare path. The only gate is `--verbose` (`logger.debug` is gated by `ConsoleLogger.shouldLog`).
- It is the same shape as the exposure half of GHSA-p5qg-v9gv-hc7w: a sensitive value reaching a channel that outlives the process.
- The value is not otherwise exposed. This was confirmed by a full write-to-every-reader trace during the review of go-to-k/cdkd#2249: `listRawObjects` returns metadata only, gc never issues `GetObject`, `scanReferencedAssets` is scoped to keys ending `/state.json` so a placeholder leaf cannot match, and the body is never written to `state.json`, the rollback journal, deployment events or the exports index. This log line is the ONLY reader that emits the payload.

## How it was found

Surfaced by the security add-on reviewer while reviewing go-to-k/cdkd#2249 (a follow-up hardening `cdkd gc`'s sweep of abandoned custom-resource response placeholders), and confirmed by reading the source directly. `git diff d44cd29a..HEAD -- src/provisioning/providers/custom-resource-provider.ts` on that branch is EMPTY: go-to-k/cdkd#2249 neither introduces nor worsens it, which is why it is filed separately rather than folded in.

## Suggested fix

Log the non-sensitive envelope instead of a raw prefix -- `Status`, `PhysicalResourceId`, and `Object.keys(Data)` rather than the values. That keeps the diagnostic usefulness (which resource answered, whether it succeeded, which keys came back) and drops the payload.

A test should assert the RENDERED line, not that a mocked logger was called with something: a `vi.fn()` logger records any argument, so it cannot see the payload being interpolated.

## Classification

Session-fit: next (not this session) -- this session is driving two unrelated follow-up PRs to merge, and this is an independent file with no overlap with either lane. Neither `now` criterion fires: it is not in a file this session has open, and its evidence is a source read that survives in this body rather than a live repro that dies with the session.
Severity: medium -- a generated secret can reach the terminal and CI logs, but only under `--verbose`, and no persisted artifact carries it.
Effort: small (S) -- one log statement plus a unit test asserting the rendered line; it rides verification this repo already pays for and needs no integ fixture.
Estimate: ~30 min -- the edit is a few lines; most of the time is writing a test that asserts the rendered string rather than a mock call, and checking the sibling providers for the same shape.

go-to-k:
Closes https://github.com/go-to-k/cdkd/issues/2250

## The defect

Each poll of the pre-signed S3 response key logged `body.substring(0, 200)` of the
CloudFormation custom-resource response document. `Data` is the documented place a handler
returns a GENERATED VALUE — including a generated secret — so behind a short
`PhysicalResourceId` those values landed inside the 200-character window and reached the
terminal and, in CI, the retained build log. It fired for EVERY custom-resource response,
gated only by `--verbose`.

## The fix

The line renders the non-sensitive ENVELOPE: `Status`, `PhysicalResourceId` and
`Object.keys(Data)` — never a `Data` value, and never `Reason`, which is free-form handler
text that can quote them.

The body stays UNTRUSTED input. The parse is hoisted so one result feeds both the summary
and the terminal-status check, but the log still fires for a body that does NOT parse,
reporting its LENGTH only — a malformed response is exactly when the diagnostic is worth
most, and the obvious "show a prefix to diagnose it" fallback would reintroduce the same
leak through another door, since a truncated body is a prefix of the same document.

## Two regressions the fix itself introduced, both found by security review

**Control characters.** The line this replaced printed raw WIRE json, where the encoder had
already escaped control characters as a `` sequence. Parsing before logging UNDOES
that: an ESC and a newline reach the terminal as real bytes. Measured — a
`PhysicalResourceId` carrying a clear-screen sequence plus a newline cleared the screen and
printed a forged `ERROR [cdkd]` line into the transcript. Anyone holding the pre-signed
response URL can author that body. Every handler-controlled field now goes through
`displaySafe`, which is this repo's one answer to that (issue
https://github.com/go-to-k/cdkd/issues/2170) and which `state.ts` already applies to this
very value.

**Bound.** `substring(0, 200)` also capped the old line by construction; dropping it removed
the cap. Measured: a 5000-character id with 300 `Data` keys rendered a **19,714-character**
line, re-emitted on every poll of a resource that can run for an hour. Each field is capped
at 200 characters and the key list at 20 names, with a `+N more` count rather than silent
truncation.

## A premise this issue carried was wrong

The issue said the payload was "not otherwise exposed", relaying an earlier flow trace that
was correct about `cdkd gc`'s sweep and wrong about state. `cfnResponse.Data` becomes
`attributes` (`custom-resource-provider.ts:1128` / `:1208`) and is persisted into
`ResourceState.attributes`, so a generated secret ALSO sits in plaintext in `state.json` —
a durable channel strictly worse than the transient one this closes. Filed as
https://github.com/go-to-k/cdkd/issues/2274; the issue body carries the correction. **What
this PR closes is the `--verbose` channel, not the exposure as a whole.**

## Verification

`vp check --fix` / `vp run check` / `vp run typecheck:test` / `vp run typecheck` /
`vp run build` all rc 0. Full suite **806 files / 16584 tests**, 0 failures, no type errors.

Cases assert the **RENDERED line**, captured at the `console.debug` boundary of a REAL
`ConsoleLogger` at debug level — not a `vi.fn()`, which records any argument handed to it
and so cannot see a payload being interpolated.

| mutation | measured |
|---|---|
| restore `body.substring(0, 200)` | 4 RED, each failing on the leak assertion itself |
| bypass `displaySafe` | exactly 1 RED (the control-character case) |
| lift the per-field cap | exactly 1 RED (the bound case) |

Each probed alone, restored from a snapshot, restore verified.

**Live test / integ:** `custom-resource-provider` against real AWS from this branch —
rc 0, clean destroy, and an independent post-run orphan scan (state, Lambda, Step
Functions, S3, IAM, SSM) clean. `integ-destroy` marker set, ledger row recorded. A first
attempt was killed by a 10-minute tool timeout mid-Phase-6; its deploy child was allowed to
finish rather than killed, and the only leftover — a state machine already in `DELETING`,
i.e. the interrupted run's own delete in flight — was polled to gone before the re-run.

## Round 2 — two residuals the security review measured

**Delimiter forgery.** `displaySafe` removes control characters and nothing else, so a
handler-chosen value carrying `]`, `=` or `,` forges the rest of the record while every
character in it stays printable. Measured: a `Data` KEY of
`x] Status=SUCCESS PhysicalResourceId=forged-id Data keys [y` rendered a line on which
`grep 'Status=SUCCESS'` matches a response whose real `Status` is `FAILED`. Weaker than the
newline forgery round 1 closed — it cannot start a new log RECORD — but the same intent, and
unreachable by sanitising characters, because the characters are legitimate. Each field is
now `JSON.stringify`'d, which quotes, escapes the delimiters, and well-forms a lone
surrogate the per-field slice can create by cutting a pair in half.

**The WARN channel.** `truncateReason` feeds `logger.warn` and is therefore NOT gated by
`--verbose`, while its inputs are the handler-authored `Reason` and the Lambda `logTail` —
the same untrusted document. Hardening the quiet channel and leaving the loud one raw would
have moved the problem rather than closed it. Also an outer whole-line clamp: the per-field
caps bound each PART, and 20 keys at the field cap still rendered ~4.6 KB.

**The warn fence took two tries and BOTH first attempts passed while proving nothing**,
which is worth recording because a vacuous test is worse than none. The first drove
`pollS3Response`, which never reaches `truncateReason` — the premise was out of scope. The
second, re-aimed at the authz-retry path, was vacuous again: the forged bytes were appended
*after* a ~300-character `AUTHZ_REASON`, and `truncateReason` caps at 200, so the payload sat
past the cut and the case passed whether or not the sanitiser ran. Moving them to the front
makes it discriminate. Both dead ends are recorded in the case's own comment.

| round-2 mutation | measured |
|---|---|
| drop the per-field quoting | 5 RED (incl. the forgery case) |
| un-sanitise `truncateReason` | exactly 1 RED (the warn case) |

## Sibling sites found and NOT fixed here

Both are outside this diff's file scope, same shape (a raw prefix of untrusted input
interpolated into a log line):

- `src/local/websocket-server.ts:356` — raw client frame prefix at debug. Local-emulation
  traffic, but frames routinely carry auth tokens / session payloads.
- `src/local/rest-v1-integrations.ts:726` — a 200-char prefix of rendered VTL output on a
  MOCK statusCode fallback; that output derives from the request body. Weaker, same shape.


go-to-k:
Reopening to re-dispatch CI: the GitHub Actions outage today dropped the `synchronize` event for head 0d0a3a5b, so the CI workflow never ran on it (only PR Title Check and the label-inherit job did). No content change.

Issue #2275, opened 2026-08-26
fix(cli): nine confirmation prompts still hang forever on EOF stdin — the class go-to-k/cdkd#2259 fixed for destroy only


go-to-k:
Dup-check: searched open issues for `prompt non-interactive guard rl.question`, and ran a body-window scan for `rl.question` / `NON_INTERACTIVE_CONFIRM`. The only hit is go-to-k/cdkd#2259, which is the DESTROY prompt and is being fixed separately -- this covers the sites that fix deliberately leaves open, so it is one root cause rather than a duplicate.

## Summary

`rl.question` never settles once stdin is at EOF, and EOF delivers no signal, so a command that prompts without a non-TTY guard parks FOREVER in CI rather than failing -- burning the job's whole timeout budget. go-to-k/cdkd#2259 closed that for the per-stack destroy prompt. It is not the only one.

Still a bare `await rl.question(...)` with neither an `isTTY` guard nor an abort signal:

- [ ] `src/cli/commands/rollback.ts` -- reached whenever `--force` is absent
- [ ] `src/cli/commands/state.ts` -- the `state rm` prompt
- [ ] `src/cli/commands/state.ts` -- the second prompt further down the same file
- [ ] `src/cli/commands/orphan.ts`
- [ ] `src/cli/commands/import.ts`
- [ ] `src/cli/commands/export.ts`
- [ ] `src/cli/commands/drift.ts`
- [ ] `src/cli/commands/retire-cfn-stack.ts`
- [ ] `src/cli/commands/state-migrate.ts`
- [ ] `src/cli/commands/confirm-prompt.ts` -- the SHARED helper, so whatever calls it inherits the gap

Found while reviewing the go-to-k/cdkd#2259 fix: its changelog entry originally claimed the destroy prompt was "the ONE prompt in the repo that had neither", which a reviewer measured to be false. That entry now names this issue instead of making the claim.

## Why one issue rather than ten

One root cause, one sentence describing the fix at every site: guard on `process.stdin.isTTY` before creating the interface and refuse. Splitting it into ten would pay the full triage / claim / review / merge cost ten times for the same edit, and would let sites 2..10 drift away from the fix at site 1.

## Direction

`confirm-prompt.ts` first -- it is the shared helper, so fixing it may cover several of the others for free. Check which of the listed sites route through it before writing anything per-site.

Two decisions that go-to-k/cdkd#2259 already settled and should be inherited rather than re-litigated, unless a site has a reason to differ:

[...]

PR #2284, opened 2026-08-26
fix(destroy): refuse the per-stack confirmation prompt on a non-TTY stdin instead of hanging forever

### LINKED ISSUE #2259: fix(destroy): the per-stack confirmation prompt hangs forever on EOF stdin, so a CI destroy without --yes parks until its own timeout
Dup-check: searched the open issues for `prompt EOF hang non-interactive destroy`, and ran a body-window scan for `rl.question` / `isTTY` / `NON_INTERACTIVE_CONFIRM`. No hits. go-to-k/cdkd#1342 (the nearest neighbour, about a cancelled CI job stranding a stack lock) is CLOSED and is about the SIGNAL case, not EOF.

## Summary

`destroy-runner.ts`'s per-stack confirmation prompt is a bare `await rl.question(prompt)` with neither a non-TTY guard nor an abort signal. `rl.question` never settles when stdin is already at EOF, so

```
cdkd destroy <stack>        # without --yes / --force, stdin not a terminal
cdkd destroy --all
cdkd state destroy <stack>
```

park forever in CI on nothing more than an absent stdin. No signal is delivered, so nothing wakes them.

Measured on Node 24.19 against real `node:readline/promises`:

| stdin | `rl.question` |
| --- | --- |
| `echo y \|` | resolves `"y"` |
| `printf 'y' \|` (no trailing newline) | **hangs** |
| `< /dev/null` | **hangs** |

## Why the sibling fix did not cover it

go-to-k/cdkd#2247 fixed the same shape for `state.ts`'s `--all` BATCH prompt by refusing a non-TTY run up front (`process.stdin.isTTY !== true` -> `CdkdError(..., 'NON_INTERACTIVE_CONFIRM')`). Its changelog left the per-stack prompt alone with the reason "it sits in the watch's PRE-REGISTRATION window where the watch already force-quits, so it is an exit rather than a hang".

That reason is correct for a SIGNAL and does not cover EOF, which delivers no signal at all. The entry has been corrected to say so and to point here.

## Why destroy is the outlier

`deploy.ts` already handles it the other way: it auto-confirms when `!process.stdin.isTTY`. Five prompts (`gc.ts`, `bootstrap-destroy.ts`, `recreate-confirm-prompt.ts`, `prefix-migration-check.ts`, `migrate-command.ts`) guard on `isTTY` before creating the interface. `destroy-runner.ts` is the one prompt in the repo that does neither.

## Direction

Apply the same non-TTY refusal the batch prompt now uses. Two decisions to settle rather than assume:

1. **Refuse vs auto-confirm.** `deploy.ts` auto-confirms; the five guarded prompts refuse. For a destroy, refusing is the safer default and matches what go-to-k/cdkd#2247 chose for the batch prompt one layer up — but it is a behaviour break for anyone piping `y` today, so it needs an upgrade note like the one `cdkd gc` carries.
2. **Error shape.** Only `gc.ts` and `bootstrap-destroy.ts` throw `CdkdError` with `NON_INTERACTIVE_CONFIRM`; the other three throw a bare `Error` / `LocalMigrateError`. Match the two carrying the code, so CI can branch on it.

Fence it the way go-to-k/cdkd#2247 fenced its twin: a probe that removes the guard must reproduce the production symptom as a TIMEOUT, not as an assertion about a mock.

## Classification

Session-fit: next (not this session) -- found at review round FOUR of go-to-k/cdkd#2247, which had already withdrawn one attempt at this exact class. Adding a new refusal to a second command that late is how a fifth round happens; the defect is pre-existing and the PR's claims have been corrected to stop implying it is covered.
Severity: medium -- a CI job hangs until its own timeout rather than failing, and it is reachable by an ordinary `cdkd destroy <stack>` with no `--yes`. Not silent data loss, and `--yes` avoids it entirely.
Effort: small (S) -- one guard plus a timeout-shaped unit probe, riding verification this repo already pays for. It does touch `destroy-runner.ts`, which is in the `integ-destroy` gate scope, so the PR carrying it needs a real-AWS integ run.
Estimate: ~1 h -- the edit is minutes; most of the time is the integ run the gate scope forces plus deciding refuse-vs-auto-confirm against the `deploy.ts` precedent.

go-to-k:
Closes https://github.com/go-to-k/cdkd/issues/2259

## The defect

`destroy-runner.ts`'s per-stack confirmation prompt was a bare `await rl.question(prompt)`
with neither a non-TTY guard nor an abort signal. `rl.question` never settles once stdin is
at EOF, and **EOF delivers no signal at all** — so the abort arm that
https://github.com/go-to-k/cdkd/pull/2247 added for the BATCH prompt could not help here:
there was nothing to abort ON.

`cdkd destroy <stack>`, `cdkd destroy --all` and `cdkd state destroy <stack>` without
`--yes` / `--force` therefore parked FOREVER in CI on nothing more than an absent stdin,
burning the job's whole timeout budget instead of failing. Measured on Node **24.15.0** (the version `.node-version` pins) against real
`node:readline/promises`:

| stdin | `rl.question` |
|---|---|
| `echo y \|` | resolves `"y"` |
| `printf 'y' \|` (answer, no trailing newline) | **hangs** |
| `< /dev/null` | **hangs** |

This is exactly the gap https://github.com/go-to-k/cdkd/pull/2247 left open and pointed
here: its narrow reason — the per-stack prompt sits in the watch's PRE-REGISTRATION window
where the watch already force-quits, so a SIGNAL there is an exit rather than a hang — is
correct for a signal and says nothing about EOF.

## The fix

The same non-interactive REFUSAL the batch prompt one layer up now uses:
`process.stdin.isTTY !== true` -> `CdkdError(..., 'NON_INTERACTIVE_CONFIRM')`, thrown
BEFORE `createInterface` so there is no window a never-settling question can be awaited in.

**Two decisions settled rather than assumed.**

*Refuse, not auto-confirm.* `deploy.ts` takes the other branch (`!isTTY` -> proceed), and
five prompts refuse. A destroy belongs with the refusers: a deploy that assumes "yes" is
recoverable, while silently answering "yes" for an absent operator here deletes every
resource in the stack. It also makes the two layers of the same command agree.

*Error shape.* Matched to the only TWO of those five that carry a code — `gc.ts` and
`bootstrap-destroy.ts` throw `CdkdError` with `NON_INTERACTIVE_CONFIRM`, while
`recreate-confirm-prompt.ts` / `prefix-migration-check.ts` throw a bare `Error` and
`migrate-command.ts` a `LocalMigrateError`. (An earlier revision of the sibling entry
claimed all five carried the code; that was measured false and is not repeated.)

## Four claims a review measured FALSE, corrected here

- `docs/cli-reference.md` said "Nothing is read, locked or deleted" of the refusal. True of
  the BATCH prompt, copied down and false here: the per-stack guard sits after the
  strong-reference SCAN, so the refusal is preceded by a cross-stack READ. (An earlier
  revision of this bullet also said the guard sits after the empty-state branch "which
  takes the lock and calls `deleteState`". That was wrong in a way worth naming: the two
  paths are MUTUALLY EXCLUSIVE — the empty-state branch RETURNS at
  `destroy-runner.ts:582`, so a zero-resource stack never reaches the refusal at all and a
  non-interactive run of one still succeeds. Measured both ways: zero resources on a
  non-TTY gives `acquireCalls 1, deleteStateCalls 1` and no throw, while one resource with
  outputs gives `listStacksCalls 1, acquireCalls 0, deleteStateCalls 0` and the refusal.)
- The changelog called this "the ONE prompt in the repo that had neither". **NINE others
  still are**: `rollback.ts`, `state.ts` TWICE (the `state rm` prompt and
  `confirmRefresh`), `orphan.ts`, `import.ts`, `export.ts`, `drift.ts`,
  `retire-cfn-stack.ts` and `state-migrate.ts`. Telling readers the class is closed
  repo-wide when it is not is worse than saying nothing; the entry now names them and
  points at https://github.com/go-to-k/cdkd/issues/2275, filed for the rest. (The count
  was published as TEN, listing the shared `confirm-prompt.ts`. That entry was WRONG — it
  is a bare prompt, but its only caller short-circuits on a non-TTY stdin before reaching
  it, so it cannot hang on EOF; the same criterion had already excluded the `events.ts`
  helper, applied inconsistently. It mattered beyond the count: https://github.com/go-to-k/cdkd/issues/2275's Direction
  said to fix that helper FIRST because doing so "may cover several of the others for
  free", and it covers none. The issue carries the correction.)
- The docs said "exits non-zero" without naming the code. `CdkdError` carries no
  `exitCode`, so `handleError` maps it to **1** — and a destroy also emits 2 for a partial
  failure, so "non-zero" alone does not let CI tell the two apart.

## BREAKING

`printf 'y\n' | cdkd destroy MyStack` succeeded before (piped stdin does settle
`rl.question` when the input ends in a newline) and now exits 1. `--yes` / `-y` (or `-f` /
`--force` on `cdkd destroy`) is the supported non-interactive path and never consults
stdin. Documented as an upgrade note in `docs/cli-reference.md` and the changelog — the
same trade `cdkd gc` and `cdkd state destroy --all` each made.

## Verification

`vp check --fix` / `vp run check` / `vp run typecheck:test` / `vp run typecheck` /
`vp run build` all rc 0. Full suite **805 files / 16584 tests**, 0 failures, no type errors.

**The probe reproduces the PRODUCTION SYMPTOM, not a proxy for it**, the way the batch
prompt's twin was fenced:

| mutation | measured |
|---|---|
| remove the guard | 1 case fails as a **5005 ms TIMEOUT** (+2 ordinary assertion fails) |
| hoist it above the `--yes` short-circuit | **8 RED** — the `--yes` case plus all 7 pre-existing SIGINT cases, so the POSITION is pinned too |
| invert `!== true` to `=== true` | 5 RED, incl. both TTY negative controls |
| delete the file-level readline mock | 7 pass — the mock is genuinely INERT for the pre-existing block |

Negative controls that stay green: a TTY run still prompts and still destroys, a TTY user
answering `n` still cancels and deletes nothing, and a `--yes` / `--force` run never
consults stdin even on a non-TTY.

**Caller enumeration** (the false-refusal direction, which is where a guard like this does
damage): only 3 importers of `runDestroyForStack`. `nested-stack-provider.ts` passes
`skipConfirmation: true`, so nested-stack children are unaffected — the parent already
confirmed the cascade; `state.ts` passes `yes || all`; `destroy.ts` passes `yes || force`.
Every integ `verify.sh` destroy call carries `--force` / `--yes`. Not exported from
`src/index.ts`, so there is no programmatic blast radius.

## The doc-correction round these bullets came out of

The three bullets above were themselves published wrong once, which is why each now
carries what it used to say. A review of the SHIPPED docs found four claims false: the
branch relationship (asserted as sequential, measured as mutually exclusive), the site
census (ten with a wrong entry, actually nine), an "exits non-zero" left standing in the
same section where a sibling paragraph had already been corrected to name the code, and a
Node version cited as 24.19 in four places while `.node-version` pins 24.15.0 — so the
published evidence came from a binary the project does not run. All four are corrected in
the diff and re-measured on the pinned version.

The branch-relationship claim had **no test behind it**: every fixture in the confirm
block used an empty outputs map, which skips the strong-reference scan entirely, so the
only evidence for that paragraph was out of tree. A case with non-empty outputs now pins
it — `listStacks` called, `acquireLock` and `deleteState` not — and disabling the scan
reds exactly that case.

**Live test / integ:** `lambda` (broad set) against real AWS from this branch —
`destroy-runner.ts` is in BOTH the `integ-destroy` and `integ-broad` scopes. rc 0, 9
deleted / 0 errors, independent post-run orphan scan (state, IAM, Lambda, DynamoDB, SQS)
clean. Both markers set, ledger row recorded.


go-to-k:
Reopening to re-dispatch CI: the GitHub Actions outage today dropped the `synchronize` event for head 6078a20d, so the CI workflow never ran on it (only PR Title Check and the label-inherit job did). No content change.

PR #2362, opened 2026-08-28
chore(work-issues): split the 231 KB SKILL.md into a thin orchestrator plus per-stage reference files

go-to-k:
## Summary

`/work-issues`'s SKILL.md had grown to 231,172 B (~83k tokens) and is loaded whole on every invocation — more than half a context window spent before a run's first action. The growth is structural: section 10 folds every run's lessons back into the same always-loaded file.

This PR converts the skill to progressive disclosure:

- **SKILL.md becomes a ~6.7 KB orchestrator** — stage table, hard invariants, and a mandatory "read `references/<stage>.md` at stage entry" rule. It also directs the read-heavy stages (triage, retro) to run in read-only subagents so raw backlog listings and issue bodies stay out of the parent context.
- **Sections move VERBATIM** into `references/{triage,claim,implement,gates-and-pr,verify,ship,retro,gotchas}.md` (sections 0-3 / 4 / 5 / 6-7 / 8 / 9 / 10 / appendix). Verified mechanically: re-concatenating the reference files (minus their 2-line headers) and diffing against the original body shows only the three deliberate retro.md edits below.
- **retro.md section 10-b now targets the stage file where a lesson fires**, never the orchestrator, and its already-landed grep command targets the skill directory instead of the single file. This redirects the growth loop away from the always-loaded file.
- **New fence `tests/unit/scripts/skill-file-payload.test.ts`** — the skills-side sibling of `rule-file-payload.test.ts`: 12 KB cap on a split skill's SKILL.md, 36 KB on any other SKILL.md, 64 KB per reference file, a vacuous-scan floor, and a deletion floor so "reducing payload" by dropping stage files fails rather than passing.
- **Consumers re-pointed**: `cross-cutting-list-sync.test.ts` reads the section-2 contested-file list from `references/triage.md`; `work-issues-skill-refs.test.ts` scans every `.md` in the skill directory (directory-derived list with a count floor, per-file proportional stripper checks, corpus-level qualified-ref floor); hook messages (`gated-command-preamble-gate.sh`, `issue-dup-check-gate.sh`, `main-tree-git-cwd-detector.sh(.test.sh)`) and JSDoc pointers (`check-provider-secret-mask.ts`, `secret-mask-baseline-2269.ts`, `.claude/rules/provider-masking.md`) name the stage file that now holds their section.

The same treatment is being mirrored to cdk-local and cdk-real-drift in parallel branches.

## Test plan

- `vp check` / `vp run check` / `vp run typecheck:test` / `vp run build`: pass
- `vp run test`: 17,303 tests pass (includes the new payload fence and the two updated extractor tests)
- `vp run gen:all-matrices` + `vp run audit:coverage:check`: clean, no dirty generated files
- Verbatim-move proof: reassembled reference files diffed against the pre-split SKILL.md body — only the three deliberate retro.md edits differ
- Every `references/*.md` named by the orchestrator's stage table exists (checked mechanically)


go-to-k:
Post-merge record of the two review nits accepted as-is (won't-do), so the decision is visible:

1. skill-file-payload.test.ts: when the split-skill floor fails because `references/` is missing entirely, `isSplit` also flips false and SKILL.md silently falls to the loose 36 KB cap. Theoretical — the floor itself fails first and names the real problem; adding a second signal would duplicate it.
2. work-issues-skill-refs.test.ts: SKILL.md (the orchestrator) carries 0 qualified refs, so its proportional stripper check is vacuous for that one file. Accepted: the backtick-parity guard covers the known blanking mechanism per file, and the corpus-level floor (40 vs 124 measured) bounds wholesale loss.

Both were graded nit/acceptable by the round-1 test reviewer and re-confirmed in round 2.

PR #2398, opened 2026-08-31
feat(utils): route AWS SDK calls through HTTPS_PROXY / HTTP_PROXY, honouring NO_PROXY

### LINKED ISSUE #2388: feat(cli): no AWS call is routed through HTTPS_PROXY, so cdkd cannot run behind a corporate proxy
cdkd does not route its AWS calls through an HTTP proxy, so on a machine whose only
egress is a corporate proxy nothing works. The very first command fails:

```
$ cdkd bootstrap --profile my-sso-profile
Starting cdkd bootstrap...
No --state-bucket specified, resolving default bucket name...
CredentialsProviderError: Error: self-signed certificate in certificate chain
```

The certificate wording is a red herring: cdkd is connecting directly, and on this
network it is the direct route that gets intercepted. Routed through the proxy it
sees Amazon's own certificate and needs no extra CA.

The AWS SDK for JavaScript v3 does not read the proxy variables on its own —
[the guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/node-configuring-proxies.html)
says a proxy has to be supplied "through a third-party HTTP agent" by whoever
constructs the client. In the same shell the AWS CLI, `gh` and npm all work; cdkd is
the odd one out. The AWS CDK CLI covers it with `--proxy`.

It also cannot be worked around from outside the tool. Node 24's
`NODE_USE_ENV_PROXY=1` only rewires the *global* agent, and the SDK builds its own
`https.Agent` per client, so SDK traffic bypasses it.

Could cdkd honor `HTTPS_PROXY` / `HTTP_PROXY` / `NO_PROXY`, or take a `--proxy` flag?

## Environment

- cdkd 0.284.85, Node.js 24.16.0, Ubuntu 24.04 on WSL2
- SSO profile, corporate forward proxy, no direct egress

## Repro without a proxy

Point cdkd at a proxy that cannot work and it succeeds anyway, which shows the
variable is never consulted:

```sh
HTTPS_PROXY=http://127.0.0.1:1 cdkd state list --profile <any working profile>
```

---

I dug into this a fair way and have a branch that fixes the `AwsClients` path with
`vp run typecheck` / `vp run test` green. Two things there might be worth knowing
before anyone starts: the credential providers need the handler passed separately
from the service clients, and around 91 clients are constructed outside
`AwsClients`. Happy to write that up properly or open a PR — just say which is more
useful.

masucamp:
Built to the design in the issue thread. The AWS SDK for JavaScript v3 does not
read `HTTPS_PROXY` / `HTTP_PROXY`, so on a machine whose only egress is a
corporate proxy every cdkd command failed at credential resolution -- before any
service call -- as `CredentialsProviderError: self-signed certificate in
certificate chain`.

`awsClientDefaults({ profile })` returns `{}` unless a proxy variable is set, so
the unproxied path is byte-identical and this is a no-op for existing users.

## Evidence

Same command, same machine, same corporate proxy, untouched environment:

| build | result |
| --- | --- |
| `main` | `Error: self-signed certificate in certificate chain` |
| this branch | `InvalidClientTokenId: The security token included in the request is invalid.` |

The second is an AWS API response: the request reached STS through the proxy and
the only thing left wrong was the deliberately fake key used for the comparison.
With a real profile the command exits 0 and lists the account's stack.

Inverted, with every spelling pointed at a proxy that cannot work: `main`
ignores the variable and goes direct, this branch reports
`connect ECONNREFUSED 127.0.0.1:1`.

The tunnel measurement asked for in the thread is posted there rather than
repeated here.

## Scope

Migrated: `aws-clients.ts` plus the bootstrap and state-write paths
(`bootstrap.ts`, `expected-bucket-owner.ts`, `aws-region-resolver.ts`,
`bucket-region-client.ts`, `asset-storage.ts`, `config-loader.ts`) -- the set
that makes the reported repro pass. The remaining 61 files are allow-listed and
are the follow-up PR, which empties the list.

Out of scope as agreed: a `--proxy` flag, PAC files, the Docker daemon's own
egress.

## Two checks, because neither covers both halves

`tests/unit/cli/proxy-env-routing.test.ts` is the repro inverted: a non-zero exit
plus evidence of a connection attempt to the named proxy, with a no-proxy
control so the assertion is discriminating rather than trivially true. No error
CLASS is asserted -- the SDK retries and which call fails first moves as the
follow-up migrates the rest. Neutering `awsClientDefaults()` and rebuilding makes
it fail, naming the missing evidence.

It cannot show the sweep is COMPLETE: it passes as soon as any one call fails,
which is how the `config-loader.ts` gap survived the original bootstrap repro. So
`scripts/check-aws-client-defaults.ts` covers the population instead. It binds to
identifiers imported from an `@aws-sdk/client-*` package plus the `Client`
suffix (the import alone counts commands and reports 1726 sites where there are
134), resolves the shared-bag and `this.clientOptions` indirections, and checks
the spread ORDER rather than mere presence. Its allow-list may only shrink -- a
stale entry (a file now clean) and a dead one (a file with no client site) both
fail.

A shared config BAG is its own verdict. Two clients built from one
`const opts = { ...awsClientDefaults(), ... }` share the `requestHandler` it
carries, and therefore one routing agent -- the per-client rule broken through
the back door, and invisible at both call sites because each reads as
`defaults-first`. The correct form spreads the shared bag inside each client's
own literal, which calls the helper per client.

The checks found four defects in my own work: two spread-order violations in
`aws-region-resolver.ts` and an import-alias hole in the checker itself, both
from the fence; and the shared-bag hazard above in `verifyAssetStorageExists`,
from reading the diff. That last one is why `shared-defaults` exists at all --
the instance was a two-line fix, and the class was what needed closing. Its
regression probe restores the exact shape this branch shipped with for an hour.

## Three things worth knowing before reviewing

**The rules-corpus fence.** Documenting the new modules crossed the ceiling in
`tests/unit/scripts/rule-file-payload.test.ts`. I split them into a
narrow-`paths:` `.claude/rules/proxy-support.md` the way the failure message
prescribes, which moved `CORPUS_FILE_COUNT` 35 -> 36 (after the rebase onto
39c2e91) and needed a `REACH_FLOORS` row (wildcard-free, so asserted exactly)
plus a `PAYLOAD_BUDGETS` row naming a path only this satellite matches.

Measured on `origin/main` at 39c2e91: 35 files, 933,620 B. This branch:
36 files, 942,951 B, a delta of 9,331 B.

`CORPUS_BYTES_MAX` stays at 946,000. An earlier revision of this branch raised
the pre-rebase ceiling; after the rebase the constant's own note says the next
lane should SPLIT rather than raise, this branch has already split its bytes
into the narrow-`paths:` satellite, and raising in two consecutive lanes is the
ratchet that note exists to slow — so the ceiling is left alone, with the
decision recorded in the constant's comment and the residual (~3 KB of
headroom) flagged for the maintainer. Discussed in the PR thread.

`CORPUS_BYTES_MIN` 899,000 -> 908,000, re-measured rather than nudged, which is
the argument its own note makes: measured 942,951 B, keeping the 34,620 B of
slack the previous measurement set rather than widening it to 43,951 B.

I did NOT add the gutting-property floor that `tests/setup.ts` carries -- the
hooks rows deliberately do not have it. Those are calibrated numbers; say if you
would rather carve it differently.

**The lockfile diff is large and it is a dedupe.** Declaring
`@aws-sdk/credential-provider-node` a direct dependency collapses duplicate
`@aws-sdk/*` / `@smithy/*` versions: package entries go 1049 -> 969.
`vp install --frozen-lockfile` passes. `agent-base` is declared directly rather
than left transitive, since the routing agent subclasses it.

**`docs/changelog-cdkd.md` is deliberately untouched.** It reads as maintainer-
authored and I did not want to guess the house style for a per-PR entry. Happy to
add one in whatever shape you want.

## Verification

`vp check`, `vp run typecheck:test`, `vp run build` and `vp test run` are green
(852 files, 17,604 tests). Per the thread I have not run the integration tests.

Closes #2388


go-to-k:
Reviewed against the issue thread, with four parallel passes (spec fidelity, code, tests, security). The short version: this is a faithful build of the agreed design — 12 of the 13 agreed decisions verified in the code, the lockfile churn is a genuine dedupe with every added package checked against its upstream, and the CI change is the single fence step. Three things need a change before merge; none of them is structural.

## Changes requested

**1. `src/utils/aws-region-resolver.ts:122-139` — the shared-bag shape survives here.**

The `auth` bag calls `awsClientDefaults()` once and up to two `S3Client`s are built from it (the region-corrected rebuild), so both share one `requestHandler` — one routing agent, one credential chain. This is the exact shape the PR body describes fixing in `verifyAssetStorageExists`, and the `shared-defaults` rule exists to block. Today it is harmless in practice (the first client is destroyed before the second is built, and a destroyed Node agent can still open new sockets), but it is the invariant broken through the same back door, and the "auth FIRST" comment on the first construction currently reads as justifying the sharing. The fix is the same two-liner as before: call `awsClientDefaults()` inside each client's own literal.

**2. `scripts/check-aws-client-defaults.ts:289-312` — the hole that let (1) through.**

The shared-bag detection counts a bag only when it is passed as a bare identifier argument (`new Client(bag)`). A bag shared via spread —

```ts
const bag = { ...awsClientDefaults(), region };
new A({ ...bag });
new B({ ...bag });
```

— calls the helper once and shares the handler, yet both sites classify as `defaults-first` through the `opensWithDefaults` recursion. That is precisely the shape in (1), which is how it shipped past your own checker. Please extend the bag-usage count to leading spread identifiers inside object literals (exempting literals that also contain a direct `awsClientDefaults()` call), and add this shape to `tests/unit/scripts/aws-client-defaults-fence.test.ts` — the existing shared-bag probe covers only the bare-identifier form.

**3. The SDK-behaviour fences from the thread are missing, and the injection shape is unpinned.**

The review comment on the issue asked for a counting `requestHandler` stub asserting that (a) SSO portal traffic goes through the injected handler and (b) the STS hop inherits it via `parentClientConfig`. Neither test exists in the diff. Compounding it, the only assertion on the injected credentials is `typeof credentials === 'function'` (`tests/unit/utils/aws-client-defaults.test.ts`) — hard-coding `defaultProvider({})`, dropping both `profile` and `clientConfig: { requestHandler }`, passes every test in the PR. Right now the evidence that the SSO path is actually routed lives only in your machine's repro, so an SDK bump that changes the `clientConfig` coalescing would regress silently, which is the scenario the fence was asked for. Two independent review passes converged on this one.

## Smaller items — take or leave in this PR

- `tests/unit/cli/proxy-env-routing.test.ts:85`: the subprocess env clears `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` but not `AWS_SESSION_TOKEN`, so a real ambient token rides along with the fake keys. It can only reach `127.0.0.1:1` or genuine AWS endpoints, so it is not exploitable — but adding it to the cleared set makes the test fully hermetic.
- The every-spelling loop in `aws-client-defaults.test.ts` covers 4 of the 6 entries in `PROXY_ENV_VARS` — dropping `ALL_PROXY`/`all_proxy` from the source passes the suite.
- `src/utils/proxy-routing-agent.ts:62-67`: `requestUrl` does not bracket IPv6 literal hosts, so an IPv6 `NO_PROXY` entry can never match. Unreachable with AWS hostnames; noting it for completeness.
- A whitespace-only `HTTPS_PROXY=' '` counts as configured and then fails per request with a URL-parse error that names neither the variable nor cdkd. An early, named error in `isProxyConfigured` would be friendlier.
- Untested but delegated behaviours: a proxy URL carrying basic-auth, and lowercase-spelling routing end-to-end. Both are `proxy-from-env`'s to own — fine to leave, mentioned so the choice is recorded.

## Verified clean, for the record

- Every agreed design point except the fence in (3): `{}` unproxied, fresh agent per call with a per-instance inner cache and forwarded `destroy()`, per-request `NO_PROXY` via `getProxyForUrl` returning inner agents, per-client `defaultProvider` with only `requestHandler` in `clientConfig` and `profile` alongside, the seven-file scope with `config-loader`'s explicit credentials surviving the spread, the corrected `NO_PROXY` docs, the transitive region-fold fence, and the rule-corpus numbers as discussed.
- Supply chain: all five added packages resolve to registry.npmjs.org with integrity hashes and match their upstream repos (including the transitive `proxy-agent-negotiate@1.1.0`, which is the real TooTallNate/proxy-agents package behind the v9 agents); no install scripts, no overrides, no tarball/git resolutions; the 1939 deleted lockfile lines are the disclosed dedupe. The proxy URL — auth included — is never logged, thrown, or persisted anywhere in the diff, and nothing weakens TLS verification.
- The checker script is a pure offline AST walk (no network, no writes, no child processes), and the fence-test probes (spread order, shared bag, alias hole, stale allowlist) all bind to real failure shapes — (2) above is the one probe the set is missing.

On `docs/changelog-cdkd.md`: leaving it out was the right call; we will add the entry at merge time. The integration runs remain ours, as discussed.

masucamp:
All three addressed, plus the four smaller items. Rebased onto `origin/main`
(39c2e91) on the way, which changed the corpus numbers again -- see the last
section.

**1. `aws-region-resolver.ts` — fixed.** `auth` no longer carries the defaults;
each of the two clients calls `awsClientDefaults()` in its own literal, first, so
an explicit `credentials` from `auth` still wins. The "auth FIRST" comment that
read as justifying the sharing is gone, replaced by one saying why the bag must
NOT carry it.

**2. The checker hole — closed, and it was wider than the spread form.**
`sharedDefaultsSource` now counts a bag reached by spread as well as by bare
argument, exempting a literal that calls the helper directly (that exemption is
tested on the literal AT THE SITE, never on a resolved bag, or `new A(bag)` would
exempt itself).

Extending it surfaced a second thing worth knowing: the first cut of the wider
rule reported all 22 `aws-clients.ts` sites as gaps. What separates them from
`auth` is not the spelling but the EVALUATION -- `const bag = {...}` is computed
once and shared, a getter is recomputed per access and shares nothing. So
`isEvaluatedOnce` is the actual predicate, and a class FIELD is on the shared
side of it while a getter is not. Both directions are tested, and
`resolveObjectLiteral` learned to read a class field, without which that case
classified `missing` -- a louder, wrong verdict hiding the sharing.

Probes: restoring the real pre-fix `aws-region-resolver.ts` shape reports both
sites `shared-defaults`; the getter shape stays clean.

**3. The SDK fences — added, and the negative half is stated rather than faked.**
`aws-client-defaults-sdk-contract.test.ts` drives the real providers against a
counting `requestHandler` from a temp AWS home: an SSO profile with a live cached
token resolves through `clientConfig.requestHandler` and the handler sees
`portal.sso.us-east-1.amazonaws.com`; a `role_arn` profile on a client built with
that handler and NO injected chain sees `sts.us-east-1.amazonaws.com` before the
service call. Dropping either wiring fails -- and fails through this repo's own
"tests must never transact with AWS" fence, so they are hermetic by construction
rather than by my care.

I did NOT fence the negative (that a client's own handler does not reach the
portal). Every cheap way to observe it short-circuits before the portal client is
built, so the test would pass for the wrong reason; that is said in the file
rather than left implicit.

`aws-client-defaults-injection.test.ts` covers the other half you named: that
cdkd still PASSES `profile` alongside and nothing but `requestHandler` inside
`clientConfig`. `defaultProvider({})` now reds four cases there; against the old
suite it passed all twelve.

**Smaller items** -- all four taken. `AWS_SESSION_TOKEN` cleared. The spelling
loop now derives from the exported `PROXY_ENV_VARS` instead of copying it, with a
length assertion, so dropping `ALL_PROXY` fails. IPv6 hosts are bracketed in
`requestUrl`. A whitespace-only value throws naming the variable -- I chose an
error over treating it as unset, because unset sends the run direct and fails
with the certificate error this change exists to remove; it is a plain `Error`
rather than `CdkdError`, since importing that module would put a `.js` relative
import in the closure reachable from `aws-clients.ts`. The two delegated
behaviours are left to `proxy-from-env`, as you suggested.

**The corpus numbers moved under me, and one call needs yours.** `origin/main` is
now 35 files / 933,620 B (I re-derived it rather than reading the comment). This
branch is 36 / 942,951, so it FITS under the 946,000 you set, with 3,049 B.

I did not raise it again, and the reasoning cuts against the rule of thumb from
the issue thread, so it is recorded in the constant's comment rather than left as
a silent choice: my delta is 9,331 B, larger than the 3,049 B I leave, which by
that argument says raise -- but the note directly above the constant says the
next lane needing more should SPLIT rather than raise, and this one has: its ~9 KB
is already in a narrow-`paths:` satellite with only pointers left behind, so there
is nothing of mine left to move. Raising in two consecutive lanes is also the
ratchet that comment exists to slow. The residual is real: the next lane adding
more than ~3 KB fails there. Say the word and I will raise it in this PR instead.

`CORPUS_BYTES_MIN` 899,000 -> 908,000, re-measured at 942,951 to keep the 34,620 B
of slack the previous measurement set; leaving it would have widened it to
43,951 B.

Local: 852 files / 17,603 tests green, `gen:all-matrices` clean, the fence green,
and the live A/B unchanged (`ECONNREFUSED 127.0.0.1:1` inverted, a real AWS API
response through the corporate proxy).

go-to-k:
Re-reviewed the delta (052e9700). All three requests are addressed, and two of them better than asked: the evaluation-count predicate (`isEvaluatedOnce`, with the getter/class-field split and the `resolveObjectLiteral` PropertyDeclaration arm) is the right generalization of the spread hole, and the SDK-contract fences pin exactly the two behaviours the design rests on, with the un-fenceable negative stated in the file instead of faked. The real-file regression probes are anchor-guarded, and `defaultProvider({})` now fails four cases that it used to pass. Nothing blocking remains.

Five small things surfaced in the re-review. None blocks this PR; take the first two wherever convenient (here or PR 2), and the three checker bounds are recorded for PR 2, which exercises the checker against the remaining 61 files anyway:

1. `aws-client-defaults.ts` `isProxyConfigured`: `.some()` short-circuits at the first valid variable, so `HTTPS_PROXY=http://ok` plus a whitespace-only `http_proxy='   '` never throws — the typo'd var resurfaces later as exactly the unnamed URL-parse error the guard pre-empts. Validate all six before deciding presence.
2. `aws-client-defaults-sdk-contract.test.ts`: the cleared-env set misses the endpoint-shaping vars (`AWS_USE_FIPS_ENDPOINT`, `AWS_USE_DUALSTACK_ENDPOINT`, `AWS_ENDPOINT_URL` and its `_STS`/`_S3` forms). A machine setting them moves the hostnames your exact `handler.seen` assertions pin (and `sts-fips.` breaks the `startsWith('sts.')` body switch). Wrong-direction risk only — a false red, never a vacuous green — hence not a blocker.
3. Checker bound, for PR 2: a bare `new S3Client(this.opts)` over `private opts = { ...awsClientDefaults() }` escapes the shared-bag pass — the argument is neither an identifier nor a literal, so `sharedDefaultsSource` returns undefined while the PropertyDeclaration arm classifies it `defaults-first`. Mirroring the spread branch's `this.x` extraction in the bare-argument branch closes it.
4. Checker bound, for PR 2: `isEvaluatedOnce` is asked only of the top name while the defaults-credit recursion is transitive, so a getter returning `{ ...this.cached }` over a once-evaluated `private cached = { ...awsClientDefaults() }` shares a handler unflagged. No such shape exists in the tree today.
5. Checker JSDoc nit: the shared-bag `uses` count is file-global by name, so two functions each with a local `const opts = { ...awsClientDefaults() }` for one client each would both flip to `shared-defaults`. Fail-closed, unlike `resolveObjectLiteral`'s "never manufacturing a gap" claim it sits next to — worth one sentence saying so.

On `CORPUS_BYTES_MAX`: leave it at 946,000, exactly as you reasoned. The constant's own note says the next lane should split rather than raise, you have split, and the residual headroom is the next lane's problem to solve with its own measurement. Recording the decision in the comment was the right move.

CI runs are approved on the new head. Once they are green, what remains is on our side: the integration runs and the merge.

go-to-k:
One amendment to the "wherever convenient" above, now that the integration runs are about to start: please take items 1 and 2 in THIS PR after all. The reasoning is timing, not substance — we have not started the integ runs yet, so a push now costs nothing, whereas the same push after they run would invalidate them; and neither item touches the files PR 2's agreed scope covers (the migration sites and the allowlist), so deferring them there would stretch that PR's otherwise mechanical diff. Items 3-5 stay as recorded: PR 2 territory, where the checker is being exercised anyway.

Both are small: validate all six spellings before deciding presence in `isProxyConfigured`, and add the endpoint-shaping vars (`AWS_USE_FIPS_ENDPOINT`, `AWS_USE_DUALSTACK_ENDPOINT`, `AWS_ENDPOINT_URL`, `AWS_ENDPOINT_URL_STS`, `AWS_ENDPOINT_URL_S3`) to the contract test's cleared set. Once that push is up, we run the integs against it and merge.

masucamp:
Both taken, pushed as ae2bf5e. Nothing else changed, so the integs can run
against this head.

**1.** `isProxyConfigured` now walks all six spellings and decides presence after
the loop, so a valid variable no longer short-circuits past a typo'd one. Pinned
by a case using your exact ordering -- `HTTPS_PROXY` valid, `http_proxy` set to
whitespace -- which reverting to `.some()` reds.

**2.** The contract test clears `AWS_USE_FIPS_ENDPOINT`,
`AWS_USE_DUALSTACK_ENDPOINT`, `AWS_ENDPOINT_URL` and the `_STS` / `_S3` forms.
You are right that `sts-fips.` would also miss the body switch, which would have
turned a false red into a confusing one.

Items 3-5 stay recorded for PR 2.

Local on this head: 852 files / 17,604 tests green, `gen:all-matrices` clean,
the fence green, and the live A/B unchanged. One thing worth reporting rather
than hiding: the first full run exited 1 on a vitest `Worker exited unexpectedly`
with every test passing and nothing named. A re-run was clean, and I had seen the
same pool teardown noise earlier in the branch before either new test file
existed, so I am treating it as flaky rather than mine -- flagging it in case it
shows up in your runs.

go-to-k:
Merged — thank you for a first contribution of rare quality: the measurement-first debugging, the honest evidence at every round, and the discipline through three review cycles made this a pleasure to ship. The PR 2 allowlist sweep is warmly welcome whenever you get to it.

Issue #2150, opened 2026-08-21
fix(scrub): the secret-bearing literal scan sees BOTH arms of an Fn::If, so a conditional export can strand a consumer stack unclearably


go-to-k:
## Summary

`producerPublishesSecretExpression`'s secret-bearing test is a literal scan --
`JSON.stringify(subject).includes('{{resolve:')` -- so it sees BOTH arms of an
`Fn::If`. A producer whose output is

```json
{"Fn::If": ["IsProd", "{{resolve:secretsmanager:app/db:SecretString:pw}}", "plain"]}
```

deployed with `IsProd` false yields the verdict `declared`. The discriminator
then reads the producer's stored value, which in that deployment is the benign
string `plain`: it carries no dynamic reference, so scrub refuses with
`SCRUB_CROSS_STACK_PRODUCER_PLAINTEXT`.

That refusal is **unclearable**. No `cdkd scrub` of the producer can turn
`plain` into a `{{resolve:...}}` expression, because the deployed branch never
carried one. There is no bypass flag, so every other secret in the consumer
stack is stranded -- the same class go-to-k/cdkd#2133's own review had to
convert into a non-blocking finding for `CrossAccountSecretRefusalError`.

## Why this is not a go-to-k/cdkd#2146 regression

This is go-to-k/cdkd#2133's code, unchanged. `main` refuses the identical input
today. go-to-k/cdkd#2146 fixed the sibling instance in the new hop walk --
`collectReExportHops` now mirrors `resolveIf` and takes only the FALSE branch of
a well-formed 3-tuple, no branch of a malformed one -- and deliberately did NOT
narrow the older literal scan, because doing so changes which stacks an already
shipped refusal fires on. The asymmetry is recorded in that function's JSDoc
rather than left for a reader to trip over.

## The asymmetry, stated plainly

After go-to-k/cdkd#2146 the two halves of one question disagree:

| path | `Fn::If` handling |
| --- | --- |
| hop walk (`collectReExportHops`) | FALSE branch only; malformed yields no hop |
| literal scan (`producerPublishesSecretExpression`) | both arms |

## Why FALSE is the right selection here too

Every condition met on this path belongs to ANOTHER stack. Scrub evaluates
conditions only for the stack it is scrubbing, from that template's parameter
defaults (it takes no `--parameters`), so a producer's condition is not
evaluable. `resolveIf`'s unknown-condition arm selects FALSE, and the pre-pass
already mirrors that for exactly this reason.

[...]

PR #2414, opened 2026-09-01
chore(integ): rebuild the issue 2163 taken-branch Fn::If live arm with per-arm decline diagnostics

### LINKED ISSUE #2163: fix(scrub): the producer-plaintext refusal does not fire for an Fn::If export whose SELECTED branch carries the secret (real AWS; unit suite refuses correctly)
## Summary

`cdkd scrub`'s `SCRUB_CROSS_STACK_PRODUCER_PLAINTEXT` refusal does NOT fire, against real AWS, for a producer export declared as an `Fn::If` whose SELECTED branch carries the secret expression -- while the same shape refuses correctly in the unit suite. Measured while building the go-to-k/cdkd#2150 live arm; the divergence is in the live read path, not in go-to-k/cdkd#2150's change (which touches only the template-side scan).

The consequence is the class go-to-k/cdkd#2133 exists to close: `cdkd scrub <consumer>` exits 0 printing `No plaintext secrets found` while the producer's `state.outputs` still holds a plaintext for the export that consumer imports.

## What was measured

A fixture arm was built on `tests/integration/cross-stack-secret-import/` and run against real AWS FOUR times. It is NOT in the merged branch -- it was reverted once the divergence was isolated -- but it is reproducible from this description.

Setup (all premises asserted and PASSING in every run):

- the producer declares a second export whose `Value` is
  `{"Fn::If": ["<literal-false condition>", "<decoy literal>", "{{resolve:secretsmanager:<name>:SecretString:password::}}"]}`,
  i.e. the SECRET is in the FALSE branch -- the one scrub selects;
- the deploy stores that export REDACTED, verified through `cdkd state show`:
  `state.outputs["CdkdCrossStackConditionalTakenSecret"]` and
  `state.outputs["CrossStackConditionalTakenSecretOutput"]` both hold the `{{resolve:...}}` expression;
- the consumer reads it through an `Fn::ImportValue` nested in an `Fn::Join` in its SSM parameter `Description`, and the live description is asserted to carry the resolved value, so the read demonstrably happens;
- BOTH producer output keys are then seeded to the bare plaintext (the pre-go-to-k/cdkd#1899 shape) and read back THROUGH `cdkd state show` to prove scrub sees a plaintext there.

Result: `cdkd scrub <consumer>` exits **0** and prints `No plaintext secrets found in <consumer>`. Expected exit 2 with `SCRUB_CROSS_STACK_PRODUCER_PLAINTEXT`.

## Where it diverges from the unit suite

Three unit reproductions of the same shape all REFUSE correctly, so the template-side verdict is not the cause:

- the plain shape (`tests/unit/cli/commands/scrub-import-value-secret.test.ts`, "a producer declaring the expression in the TAKEN Fn::If branch STILL refuses");
- the same read nested inside an `Fn::Join`;
- two imports in one `Fn::Join`, the first verdicting `no` and the second `declared`.

## The evidence that narrows it

`--verbose` on the failing run shows the import resolving through the EXPORTS INDEX rather than through the producer's state.json:

```text
Resolved Fn::ImportValue: CdkdCrossStackConditionalSecret      (from index: <producer> / us-east-1; literal value)
Resolved Fn::ImportValue: CdkdCrossStackConditionalTakenSecret (from index: <producer> / us-east-1; redacted dynamic reference)
Resolved Fn::ImportValue: CdkdCrossStackSecretPassword         (from index: <producer> / us-east-1; redacted dynamic reference)
```

and NO `logger.debug` from any of the classified-and-declined arms in `makeCrossStackPrePass`'s `readOne`. That is the load-bearing observation: every arm that classifies and then declines logs a debug line, so the run took one of the three SILENT early returns instead --

- `if (!producer) return;` (`recordedProducer` found no recorded import), or
- `if (!producerTemplates.has(producer.stack)) return;`, or
- `if (verdict.kind === 'no') return;`

Note that `CdkdCrossStackSecretPassword` refuses correctly through the same "from index" path in the same fixture (step 10), so the index path is not blanket-broken -- the difference between the two exports is the `Fn::If`.

## What to do

Instrument the three silent returns (a debug line each would have settled this in one run rather than four) and determine which one fires, then fix the underlying cause. The instrumentation is worth keeping either way: three silent returns in the one function whose job is to decide whether a stack is reported clean is why this took four real-AWS runs to narrow and still is not resolved.

If the cause turns out to be the verdict, the fix belongs beside `producerPublishesSecretExpression`; if it is `recordedProducer`, beside `resolveImportValue`'s index arm.

Once fixed, the live arm reverted here should be restored as the missing negative control for go-to-k/cdkd#2150: the untaken-branch arm alone cannot tell a NARROWING of that refusal from a REMOVAL of it.

Session-fit: next (not this session) -- the cause is in the exports-index read path, which the go-to-k/cdkd#2157 / go-to-k/cdkd#2150 branch does not touch, and narrowing it needs instrumentation plus its own real-AWS cycle rather than a fix that can ride this one.
Severity: medium -- a consumer is reported CLEAN while the producer's state still holds the plaintext it imports, but it needs a producer that publishes a secret through an `Fn::If`, so the reachable population is narrower than the direct shape go-to-k/cdkd#2133 already covers.
Effort: medium (M) -- the fixture arm is written down above and rebuildable, but it needs its own instrumented real-AWS cycle plus a re-review round.
Estimate: ~2-3 h -- most of it is the diagnose loop: each hypothesis costs a ~4.5 min fixture round-trip, and four of them were already spent narrowing it this far.

go-to-k:
## Summary

Rebuilds the reverted issue-2163 live arm into `tests/integration/cross-stack-secret-import/` permanently, with the PR go-to-k/cdkd#2268 per-arm decline instrumentation wired into its failure diagnostics — so the single real-AWS run this PR requires before merge either proves the taken-branch `Fn::If` refusal now fires (closing the live divergence the issue measured) or names the declining arm in its own output, where four instrumentation-free runs could not.

Closes #2163

## Why this shape, and why no `src/**` change

The issue's 2026-08-26 findings comment supersedes the original body's premise: `readOne` had FIVE silent decline arms (not three), all instrumented by PR go-to-k/cdkd#2268; the quoted `from index` lines came from the DEPLOY, not the failing scrub (scrub deliberately supplies no `exportIndex`); and the exact shape REFUSES at unit level against main, reproduced from the templates `aws-cdk-lib` synthesizes for this fixture. The template-side verdict and `recordedProducer` are ruled out for the state.json scan path. What remains is a live-only divergence whose surviving hypotheses (stale bucket contents during the four measurement runs, or a non-final working-tree binary — the four runs predate every commit of PR go-to-k/cdkd#2164) cannot be separated statically. The findings comment's own prescription is: rebuild the arm and read the new debug lines — one run names the arm. This PR is that arm, made permanent because it is also the negative control issue go-to-k/cdkd#2150 asked for: the untaken-branch premise alone cannot tell a NARROWING of the producer-plaintext refusal from a REMOVAL of it.

## What the diff does

- `lib/producer-stack.ts` / `lib/shared.ts`: a second conditional export `CrossStackConditionalTakenSecretOutput` (`Export.Name: CdkdCrossStackConditionalTakenSecret`) whose `Fn::If` puts the secret expression in the SELECTED (false) arm and a plain decoy in the untaken arm — the mirror of the existing 2150 export. Decoy and export names stay in the fixture's `crossstack` vocabulary, disjoint from the plaintext needle per the fixture's naming rule.
- `lib/consumer-stack.ts`: the consumer reads the new export through the SSM parameter Description's `Fn::Join` (same placement as the 2150 read, so no resource-count assertion moves). The deployed description now ENDS with the resolved secret, which is what proves the read demonstrably happens.
- `verify.sh`:
  - step 2 gains the mirror premise (both taken keys store the `{{resolve:...}}` expression; the synthesized `Fn::If` arms checked in BOTH directions — selected arm carries the expression, untaken arm is the plain decoy);
  - the live-description assertion now checks the token-free / untaken-contains / taken-suffix triple, routing failures through `diag` (the description legitimately carries the resolved secret now, and `diag` withholds needle-bearing text);
  - step 10b: back up producer state, verify both taken keys hold the step-2 expression, seed BOTH to the run plaintext, read back through `cdkd state show`, run `cdkd scrub <consumer>` (rc captured), run the `--verbose` diagnostic scrub WHILE the seed is still in place (restore heals the condition, so afterwards is too late), restore, purge the seeded noncurrent versions, then assert: rc=2 specifically; the resolved-read refusal wording; NOT step 9's unresolvable wording; the producer named; the `declared` verdict naming `CdkdCrossStackConditionalTakenSecret` (fences the key-matching half — a `widened` fallback fails); the refusal position at the Description's `Fn::Join`; the remedy; the summary line; and the byte-identical-consumer-state / no-clean-summary secondary nets. Every failure branch prints the captured per-arm decline lines through `diag`.
- `tests/unit/cli/commands/scrub-import-value-secret.test.ts`: comment + fixture-string alignment only — the issue-2163 describe block's `PRODUCER_OUTPUTS` / `CONSUMER_PROPS` now mirror the REAL synthesized templates (verified against a local `cdk synth` of the updated fixture: `Fn::If` argument order `[condition, decoy, secret]`, Description `Fn::Join` with the taken import at `[1][3]`, real parameter name and description strings).
- `docs/cli-flag-coverage.md` + `docs/_generated/cli-flag-coverage.json`: regenerated (`vp run cli-flag-coverage`) for the fixture's new `scrub --verbose` invocation.

## Verification (no real AWS touched by this lane)

- `/check`: typecheck + lint + format, CI-parity `vp run check`, `vp run typecheck:test`, build, full unit suite — 852 files, 17603 passed, 1 skipped, rc=0.
- `/check-docs`: only docs-scope delta is the regenerated matrix; targeted review of `docs/testing.md` and `docs/scenario-coverage.md` fixture mentions found nothing stale.
- All 54 `tests/unit/scripts/` fixture-scanner suites pass over the modified `verify.sh` and fixture sources (2056 tests).
- Local `cdk synth` of the modified fixture confirms the synthesized shapes the arm and the unit fixture depend on.
- A stubbed dry-run exercised the new shell logic (case-statement polarity both ways, all six refusal-wording greps against the real message shape plus step-9 disjointness, the empty-diagnostic guard, and `diag`'s withhold-on-needle behavior).
- The refusal-wording greps are additionally grounded by the unit suite's own assertions on the real error message for this exact shape (the mutation-probed cases the 2026-08-26 session added: stripping instrumentation, forcing the discriminator early, and flipping the `Fn::If` selection each redden the corresponding unit cases).

## Required before merge (orchestrator-serialized)

`/run-integ cross-stack-secret-import` — the fixture's real-AWS run. Two outcomes, both useful:

- GREEN: the taken-branch shape refuses against today's main on real AWS; the issue-2163 divergence is gone (consistent with the environmental / non-final-binary hypotheses) and the arm stays as the standing regression net. Merge.
- RED at step 10b: the diagnostics print the named declining arm (`Scrub of <consumer>:` lines plus the scrub-side `Resolved Fn::ImportValue` lines) — the diagnosis the issue says is worth one run. Fix follows on this branch before merge.

Optional second run for arm-discrimination proof against real AWS (parent's call): mutate `selectTakenConditionalBranches` to select the TRUE arm, rebuild, re-run — step 10b must go RED (the same mutation reddens the unit refusal case, its negative control, and the untaken control together per the 2026-08-26 probes); then restore and rebuild. The mutated binary writes nothing (a clean-reporting scrub has no needle to rewrite), so the probe is AWS-safe.

## Review disposition

Two read-only reviewers (test-adequacy + code-quality) ran against the committed diff: CLEAN — zero blockers, zero should-fixes, three nits, all FIXED in this PR:

- test-adequacy nit: the step-10b diagnostic grep's `Scrub of <consumer>:` prefix was pinned by no green-path assertion — now pinned in the unit suite's negative-control case, so a prefix reword fails a unit run instead of degrading the live diagnostic when it is needed.
- code nit 1: the "matches an export by either spelling" rationale overstated the v9 read path — on a v9 record only the `Export.Name` alias is bindable (`importableOutputKeys`, go-to-k/cdkd#2193); all three comment sites (including the pre-existing 2150 premise copy the new text was modeled on) now state the alias-only binding and keep the twin-key check as an anti-divergence measure.
- code nit 2: the empty-diagnostic placeholder asserted "the pre-pass never classified this read", which a transient re-run failure also produces — now worded neutrally.

## Recorded won't-do

No `Fn::GetStackOutput` twin of the taken-branch live arm: the measured divergence was specific to the `Fn::ImportValue` live read path, the `Fn::GetStackOutput` decline arms are covered by the same instrumented `readOne` and extensive unit cases, and a new fixture family (CDK app + verify.sh + ledger + coverage rows) is large effort with no evidence of a live divergence on that route. Recorded here rather than filed, since there is no defect claim to file.

https://claude.ai/code/session_014ui4DstXi7fMNmGsYZadGs


go-to-k:
Won't-do record (merge-time, orchestrator): the optional AWS-safe discrimination probe (mutate selectTakenConditionalBranches to the TRUE arm -> step 10b must go RED -> restore) was deliberately skipped. Rationale: step 10b is a POSITIVE assertion — it requires an actual rc=2 refusal naming CdkdCrossStackConditionalTakenSecret with the 'declared' verdict sourced from the Description read, and the live run produced exactly that — so the vacuous-green failure mode the probe exists to rule out cannot apply to a green 10b (an inert arm yields exit 0 / no refusal, which 10b fails on). The probe would add a second full deploy/destroy cycle to re-prove what the assertion's own shape already guarantees. If 10b is ever weakened to a negative assertion (absence of plaintext only), build the probe then.

Issue #2179, opened 2026-08-22
fix(destroy): destroy-path deployment events persist unmasked error text


go-to-k:
## What is unmasked

Two `extractDeploymentEventError` call sites persist an UNMASKED message into `deployments/{runId}.jsonl`, the durable store that outlives `cdkd destroy`:

- `src/cli/commands/destroy-runner.ts:1449` and `:1463` — the `RESOURCE_FAILED` event for a delete that failed.
- `src/cli/commands/deployment-events-run.ts:144` — a run-level event.

Compare `src/deployment/deploy-engine.ts:3172`, where `recordEvent` routes every event through `maskSecretsInEvent`, and `src/deployment/rollback-executor.ts:210`, which masks via `maskedRollbackEventError`. Those two are the masked precedents; the three above are not.

## Why this is currently ACCEPTED rather than a live disclosure

`DeleteContext` deliberately carries no masker, and the reasoning holds (`.claude/rules/providers.md`, "delete() deliberately has no masker"): the secrets bag lives in `DeployEngine.perResourceSecrets` and is built by the DEPLOY that resolved the references. A `cdkd destroy` is a different command with no such bag, so a masker bound to it would mask nothing — the value a destroy sees is the STATE-borne one, which since the GHSA fix is a redacted `{{resolve:...}}` expression rather than plaintext. `deployment-events-run.ts:144` is run-level and its text is cdkd-authored.

## Why it is still worth tracking

The safety of the destroy path is entirely inherited from "state never holds plaintext". Issue #2012's state-redaction residuals are the coupling: any shape that leaves plaintext in a state record's `properties` becomes durably persisted the moment a delete of that resource fails, because the AWS error routinely quotes the offending value back and nothing on this path masks it.

So this is not "add a masker to destroy" — it is "if #2012 ever leaves plaintext in state, this is the sink that publishes it". Whoever closes #2012 should check this path as part of it.

Found by the security reviewer on the issue #2176 PR as an explicitly out-of-delta observation.

## Classification

Session-fit: next (not this session) — no live disclosure to fix; the actionable work is coupled to issue #2012's residuals and belongs with whoever closes that, not with #2176's provider-boundary sweep.
Severity: low — conditional on a state record holding plaintext, which the GHSA fix already prevents; it is a latent sink rather than a current leak.
Effort: medium (M) — one re-review round; the fix, if #2012 ever needs it, is threading a state-derived bag into the destroy path, which is a design question rather than an edit.
[...]

Issue #2036, opened 2026-08-19
fix(secret-redaction): a PUBLIC ssm mixed leaf is over-redacted on the empty-map readback paths


go-to-k:
## Context

Accepted consequence of the #1926 lane (PR #2025), recorded rather than papered
over. Found by the `secrets-dynamic-ref` integ, which caught the opposite defect
first: an earlier revision read "not in the verdict store" as "public" on paths
where nothing is ever recorded, and persisted a DECRYPTED SecureString.

## The behaviour

`refuseUncertifiedReadbackPositions` decides a MIXED source leaf — a
`{{resolve:...}}` reference embedded in surrounding text — and splits on whether
a secrets map exists:

- **Populated map.** A pass resolved this bag, so absence from the verdict store
  is real evidence the ssm parameter is public, and the RESOLVED value is kept.
  Precise, no over-redaction.
- **Empty map** (`cdkd state refresh-observed`, and the deploy's persist choke
  point for an UNCHANGED resource). Nothing was resolved, so nothing could be
  recorded and absence carries no signal. The leaf is treated as secret-bearing
  and the source expression is substituted.

On that second path a genuinely PUBLIC ssm mixed leaf is therefore
OVER-redacted: `observedProperties` gets the expression where AWS holds the
resolved value, so `cdkd drift` reports a phantom on that property until the
next resolving pass rewrites it.

## Why it is accepted

The asymmetry is the point. Under-redaction persists a decrypted secret — a
disclosure, and the thing this lane exists to prevent. Over-redaction produces a
baseline that does not match AWS — visible, recoverable, and not a disclosure.
With no evidence available, the recoverable failure is the right one to take.

Reachability is narrow. It needs a PUBLIC `{{resolve:ssm:...}}` expression to be
sitting in a persisted `properties` bag, and issue #1901 means that normally
cannot happen: a `String` / `StringList` parameter is stored RESOLVED, so a
surviving token is a SecureString by construction. The one documented hole in
that premise is `cdkd import`'s warn path (see `drift.ts`'s note), which can
leave a public expression in state.

## Direction

The honest closure is a real TYPE classification on the empty-map paths rather
than an inference from an empty store — i.e. give those paths a way to ask AWS
what the parameter is (`GetParameter` with `WithDecryption: false` is enough to
learn the type and yields no plaintext). That is the same plumbing issue
[#2012](https://github.com/go-to-k/cdkd/issues/2012) needs for its own family,
so the two should be built together rather than separately.

[...]

Issue #1926, opened 2026-08-18
fix(state): cdkd state refresh-observed performs NO redaction and persists secret plaintext (GHSA follow-up)


go-to-k:
## Context

Found by the Lane B implementer while building the #1917 / #1915 provenance
table (GHSA-p5qg-v9gv-hc7w follow-up set), and verified independently against
`main` at `8e33f535`.

`src/cli/commands/state.ts` imports **nothing** from
`src/deployment/secret-redaction.ts`:

```
$ grep -n "secret-redaction\|redactSecretsForState\|scrubResourceRecord" src/cli/commands/state.ts
(no matches)
```

## The gap

`cdkd state refresh-observed` writes the provider readback straight into the
state record and saves it, with no redaction pass of any kind
(`src/cli/commands/state.ts`, the `refresh-observed` walk):

```ts
const observed = await provider.readCurrentState(
  resource.physicalId, logicalId, resource.resourceType,
  resource.properties ?? {}, buildReadCurrentStateContext(state, logicalId)
);
if (observed === undefined) { unsupported++; return; }
resource.observedProperties = observed;   // <- no redaction, then saveState
```

A resource whose template carried `{{resolve:secretsmanager:...}}` (or a
`{{resolve:ssm:...}}` naming a `SecureString`) was deployed with the RESOLVED
value, so `readCurrentState` returns the plaintext and `refresh-observed`
persists it into `state.json`. This is the original advisory disclosure, in a
writer the #1910 sweep did not reach.

It is **worse than the array-shaped gap in #1915**: this path leaks SCALARS
too, since there is no redaction at all rather than a redaction that declines
to descend.

## Why the #1910 sweep missed it

That sweep's framing was "every writer now passes a POSITION source". This
writer passes no source *at all*, so it never appeared in a search for callers
of `redactSecretsForState`. It is the only observed-writer that does not reach
the redaction module.

## Direction

This is the #1900 shape, and the remedy already exists: the observed bag is an
AWS readback whose position source is the record's own `properties`, i.e.

```ts
redactSecretsForState(observed, secrets, resource.properties, STATE_SOURCED_READBACK_RULES)
```

Two things need deciding when it is built:

1. **Where the secrets map comes from.** `refresh-observed` does not synthesize
   and does not resolve, so `recordedSecretValues` is empty — exactly the
   #1900 / #1915 "empty map" case. The path pass carries the load here; whether
   a value scan can contribute anything depends on the same question #1915 is
   answering.
[...]

Issue #1915, opened 2026-08-14
fix(state): a secret nested in an ARRAY is still persisted in plaintext on the unchanged-resource path


go-to-k:
## Context

Recorded on (#1910) as residue and left open by its PR, which closed the SCALAR half. Filing separately so it survives that issue closing.

## Gap

A secret nested in an ARRAY is still persisted in plaintext on the UNCHANGED-resource path.

The two halves that would each have caught it are both switched off there, and that is by design rather than by oversight:

- The readback rules for a state-sourced `observedProperties` bag set `descendArrays: false`. That is deliberate -- AWS does not preserve list order (the reason `src/analyzer/drift-normalize.ts` exists), so positional descent would write an expression onto the WRONG element while leaving the real secret in plaintext.
- The fallback is therefore the VALUE scan -- but on the unchanged path the resource was never resolved this deploy, so its `perResourceSecrets` entry is EMPTY and the scan is a no-op. That empty map is exactly the (#1900) case the path pass exists to serve.

Verified by executing the branch module directly: `observedProperties.Containers[0].Env[0].Value` keeps the plaintext while `properties` correctly holds the expression.

This is NOT a regression -- main leaks scalars there too, and (#1910) fixed only the scalar half.

## Note on what (#1910) DID change here

(#1910) added `STATE_DERIVED_RULES` (`descendArrays: true` + `trustAnyExpression: true`) for the rollback-replay writer, whose bag genuinely IS produced by resolving its source and therefore has identical structure. That constant is NOT applicable to this case: an `observedProperties` bag is an AWS READBACK, so its element order is not guaranteed to match the source and the reason for `descendArrays: false` still stands.

## Direction

Two candidates, neither a one-liner:

1. **Key-match array elements by a non-secret sibling field** (`Name` / `Key`), so descent does not depend on order. This is the shape the drift normalizer already uses for keyed lists, so there is prior art in `canonicalizeUnorderedArraysAtPaths`.
2. **Seed the value scan from the source bag own expression set** rather than from the recorded map, so the unchanged path has needles to scan for even with an empty `perResourceSecrets`. This composes with the expression SET (#1910) introduced.

## Verification

Needs a fixture whose secret sits inside an array element -- an ECS task definition `ContainerDefinitions[].Environment[]` or `Secrets[]` is the natural shape -- deployed, then REDEPLOYED unchanged so the resource takes the unchanged path and the observed-capture refresh runs.

[...]

PR #2415, opened 2026-09-01
fix(secret-redaction): close the last two readback leaks with needles derived from the record's own certified positions

### LINKED ISSUE #2012: fix(secret-redaction): four readback shapes POSITION cannot certify keep their plaintext (GHSA follow-up)
## Context

Residual family of the GHSA-p5qg-v9gv-hc7w follow-up #1926 lane (PR #2025),
widened from the original single case by that PR's security review and by the
`#1915` fences.

## The gap

`redactSecretsForState`'s path pass substitutes only where the SOURCE leaf is a
WHOLE `{{resolve:...}}` token. On the readback paths the secrets map is EMPTY by
construction (nothing was resolved), so the value scan has no needles and
POSITION is the only mechanism. #2025 added
`refuseUncertifiedReadbackPositions`, which closes the row position can
justify — a MIXED leaf whose KEY the source carries, including inside an
identity-keyed element. Four rows remain, and they are ONE root cause: no
needle and no position, so nothing distinguishes a resolved secret from an
ordinary literal.

```text
  source leaf in the STATE record                  verdict
  -----------------------------------------------  ------------
  `['--pw', '{{resolve:...}}']` (no identity key)  LEAK
  `[{Field, Val: '{{resolve:...}}'}]` (no `Name`)  LEAK
  an UNPAIRED element beside a paired one          LEAK
  an observed KEY the source does not carry        LEAK
```

Reached by `cdkd state refresh-observed`, by `cdkd scrub`, and — the reason
this is not a niche path — by a plain `cdkd deploy`, whose
`drainObservedCaptures` baseline reaches the persist choke point with the same
configuration. Each row is pinned by a test in
`tests/unit/cli/state-refresh-observed.test.ts` named `RESIDUAL (issue #2012)`,
so a fix must flip those assertions deliberately rather than discover them.

## What does NOT work, measured

Taking the SOURCE subtree whenever the bag cannot be vouched for. #2025 built
it and the existing issue #1915 fences rejected it as a regression:

- `{Name:'', Value:'an-unrelated-literal'}` was rewritten to `Value: <expr>` —
  a FALSE redaction of a value that was never a secret.
- an AWS-reported `[{Value:'x'}]` became `[{Name:'db', Value:<expr>}]` —
  fabricating drift-baseline content AWS never reported.

`cdkd drift --revert` pushes the baseline to AWS, so both would be applied to
the live resource (the #1917 / #1498 class). **Redaction may not buy itself a
fabricated baseline** — that constraint is what makes this hard, and any
proposal has to clear it.

## Direction

The only mechanism that can decide these is a NEEDLE, so the candidate is to
give the readback paths a secrets map: resolve the record's own
`{{resolve:...}}` expressions once and pass the resulting
`RecordedSecretValues` in, turning the value scan back on. That closes all four
rows by VALUE rather than by position.

The cost is real and needs its own review: it makes `cdkd state
refresh-observed` (and the deploy's observed capture) FETCH secrets, which they
currently never do — a new IAM requirement, a new failure mode, and a new
place plaintext exists in memory. Weigh against masking by SPAN (#1935), which
is narrower but only helps the MIXED-leaf row already closed.

## Verification

Unit matrix flipping each of the four pinned RESIDUAL tests, plus the whole
`secret-redaction-array-identity.test.ts` fence set staying green (it is what
catches a fabricated baseline), plus an arm on
`tests/integration/secrets-dynamic-ref/verify.sh` Phase 1f.

Session-fit: now (do it in this session) -- claimed 2026-09-01 by the `fix/2012-2036-readback-redaction` lane (this session), bundled with go-to-k/cdkd#2036 per that issue's own instruction; the mechanism for the two surviving rows is being built now. (Previous value `next`, set 2026-08-27 after PR go-to-k/cdkd#2295 closed three of the four rows, stands superseded by this claim.)
Severity: high -- a resolved secret stays in plaintext in a durable, versioned state artifact on paths a plain `cdkd deploy` reaches, which is the class of the published GHSA this project already had. Unchanged: PR go-to-k/cdkd#2295 narrowed the reachable population, it did not change what the surviving rows leak.
Effort: large (L) -- re-rated from `medium` with the `now`. Medium described the anchor pass that has now shipped; the remainder is a behaviour change on the readback paths (they would begin FETCHING secrets: a new IAM requirement, a new failure mode, and a new place plaintext exists in memory), which needs its own PR and its own review round.
Estimate: ~3 h+ -- most of it is the IAM and failure-mode story for making `cdkd state refresh-observed` and the deploy's observed capture fetch secrets at all, not the substitution itself.

go-to-k:
Closes #2012.

**Issue (#2036) was in this lane and has been WITHDRAWN from it** — a security
review found that the fix drafted for it is a DISCLOSURE. See the section below;
the issue stays open with the constraint recorded on it.

On the readback paths the secrets map is EMPTY by construction — `cdkd state
refresh-observed` resolves nothing, and a plain `cdkd deploy` reaches the persist
choke point the same way for an UNCHANGED resource — so the value scan has no
needles and POSITION is the only mechanism. Where position could not certify a
leaf, the pass kept the plaintext, and two shapes had no position to argue from:
an UNPAIRED array element beside a paired one, and an observed KEY the source
does not carry. Both are places `redactByPath` ALREADY delegates to the value
scan; it is the scan that had nothing to say.

## Issue (#2012) — derived needles

The two surviving rows (an UNPAIRED array element beside a paired one; an
observed KEY the source does not carry) are exactly the two positions where
`redactByPath` ALREADY delegates to the value scan. They never lacked a needle in
principle: the same plaintext almost always sits at a position the pass DOES
certify, and certifying it IS the assertion that AWS's value there is that
expression's resolved form.

`deriveReadbackNeedles` runs the refusal walk once in LEARN mode over the same
two bags — the SAME function, so the two passes can never disagree about what
"certified" means — collects the `(plaintext -> expression)` pairs, and hands
them to `redactByPath` as the secrets map.

**Nothing is resolved and nothing is fetched.** The issue's own proposed
direction was to resolve the record's expressions, which would have made
`cdkd state refresh-observed` and every deploy's observed capture FETCH secrets:
a new IAM requirement, a new failure mode, and a new place plaintext lives in
memory. None of that is needed — AWS already handed us the plaintext, in the very
bag being redacted. **There is therefore NO new IAM permission and no new
failure mode on any readback path.**

Five bounds on what may become a needle, because a needle is a rewrite with a
blast radius rather than a per-leaf decision:

- only the readback-projected rules (a cross-generation or template source
  describes a different generation of the resource);
- only when the map is EMPTY — `crossStackAssociations` /
  `nestedStackParameterExpressions` are `WeakMap`s keyed by the
  RecordedSecretValues INSTANCE, so substituting a derived Map would silently
  lose every association that pass recorded; with an empty map there are none;
- only from an expression naming a SECRET-BEARING service and not PROVEN PUBLIC
  (`expressionMaySeedANeedle`) — `isSingleDynamicReferenceToken` accepts any
  `{{resolve:<anything>}}` spelling and the resolver's unsupported-service arm
  WARNS and returns the LITERAL, so `{{resolve:notaservice:/x}}`'s readback is
  ordinary data. `cdkd drift` already pins that rule for its own registration
  path, and this makes the two commands answer the same way;
- only at or above `MIN_NEEDLE_LENGTH`, since the whole-value arm matches at ANY
  length;
- a plaintext learned twice under DIFFERENT expressions is POISONED rather than
  collapsed (the issue go-to-k/cdkd#1910 wrong-reference class, on a path `cdkd drift
  --revert` and `resolveReplayProps` re-resolve against the live resource).

A MIXED leaf yields a needle too — the shape this module calls DOMINANT for CDK
(`Fn::Join` around `secretValueFromJson`) — guarded to EXACTLY ONE span whose
literal prefix and suffix both bracket the bag, anchored at the ENDS rather than
searched. The pair maps the extracted plaintext to the **TOKEN, never to the
whole leaf**: mapping it to the leaf writes a whole connection string over a
field AWS reported as a bare password, i.e. fabricated baseline content that
`--revert` then pushes. That was caught by its own test, not by reasoning.

## Issue (#2036) — drafted, then WITHDRAWN as a disclosure

The over-redaction issue (#2036) records was to be closed by a POSITIVE
`provenPublicExpressions` store: `forgetSecretExpression` recording the
resolver's DEFINITIVE public verdict instead of only deleting a memo, so the
empty-map arm could answer from evidence rather than from absence. No lookup
added — the writer is `pinSecretVerdict`'s retraction, which already ran.

**The security review measured the cost.** The store is keyed on the BARE
expression and lives for the whole process, so on a `cdkd deploy --all` spanning
regions:

- stack B, region B: `{{resolve:ssm:/app/token}}` is a plain `String` → recorded
  PROVEN PUBLIC;
- stack A, region A: the same parameter NAME is a `SecureString`, its resource is
  UNCHANGED, so it resolves nothing and never contradicts;
- stack A's readback then keeps the decrypted value.

```text
source  { Conn: 'Server=db;Password={{resolve:ssm:/app/token}};' }
before  Server=db;Password={{resolve:ssm:/app/token}};
after   Server=db;Password=THE-REAL-SECURESTRING-PLAINTEXT;
```

That is the UN-redacting direction — worse than the over-redaction it fixes —
and it falsifies the verdict store's own claim that a cross-region verdict "can
at worst store a public reference as an expression, never a secret as plaintext".
That claim held while only the SECRET store was process-global, because a stale
SECRET verdict can only over-redact; a PUBLIC store inverts it.

**Reverted rather than patched.** The fix is to key the verdict by SCOPE (region
+ account) and check it at the READ side — which threads a parameter through
`redactSecretsForState` and its ~6 callers including `deploy-engine.ts`, a
cross-cutting diff that arms the `integ-broad` gate and deserves its own review.
Poisoning on a second scope is NOT sufficient: the dangerous case is a verdict
written once and never contradicted. The constraint is recorded on issue (#2036)
and in `mixedLeafMayCarryPublicReference`'s own doc, and the fixture's Phase 1f3
now PINS the residual so it is covered by a real-AWS arm even though the fix is
not in.

## One anchor-gate relaxation, in the direction that never guarded anything

`anchorsCorroboratePairing` compared object key COUNTS, so an AWS-added `Arn` /
`LastModified` inside an unkeyed-array element refused the whole element and left
the secret in it as plaintext. It now requires CONTAINMENT — every SOURCE key
present and corroborating — while a source key MISSING from the bag still
refuses, which is the fabrication direction that stopped `[{Value:'x'}]`
## Composing the scan with the position passes — three failed attempts

This is where the PR spent its review rounds, and every attempt failed in a way
no unit test in the file caught. All three are now fenced.

**Attempt 1 — SCAN FIRST** (needles fed into `redactByPath` before the refusal
pass). The scan rewrites a frame LITERAL that happens to embed a learned
plaintext — a coinciding anchor, exactly issue (#2012)'s own population — so
`unkeyedArrayPairsByAnchors`, re-run against the SCANNED bag, finds the anchor no
longer deep-equal to its source. The WHOLE array refuses and a sibling MIXED leaf
that the SHIPPED code redacts BY POSITION persists in full plaintext.
**Under-redaction**, in the GHSA-p5qg-v9gv-hc7w direction. Measured on both
trees: `1a73e99d` redacts the connection string by position (the anchor leaks —
the known bug), attempt 1 fixed the anchor and leaked the connection string.

**Attempt 2 — SCAN LAST, unrestricted** over the position passes' output. The
scan then sees leaves whose content came from the SOURCE, and a needle in such a
leaf's literal FRAME gets replaced:
`postgres://appuser:{{resolve:secretsmanager:...}}@h/db` becomes
`postgres://{{resolve:ssm:/app/db-user}}:{{resolve:...}}@h/db` when `appuser` is
also some whole-token position's resolved value. A **fabricated baseline** — a
reference the template never had at that offset — which `cdkd drift --revert`
re-resolves and pushes, so the day that parameter changes the revert writes a
different user onto the live resource.

**Attempt 3 — MERGE, but keyed on `refused === bag`.** Scan the RAW bag, take
that result only where the position passes did not change the leaf. Right idea,
wrong test: value equality cannot distinguish an UNDECIDED position from one the
pass decided IN FAVOUR of the value already there. Two shapes hit it, both
fabricating attempt 2's baseline again — the resolver's unsupported-service arm
leaves an `{{resolve:ssm-secure:` token LITERAL so AWS echoes it back and source
EQUALS bag, and the empty-map arm that deliberately KEEPS a leaf returns `bag` by
design.

**Shipped — MERGE keyed on a MARK TREE.** `refuseUncertifiedReadbackPositions`
runs a second time in `mark` mode, returning a `POSITION_DECIDED` sentinel at
every position it decides and the bag's value elsewhere.
`preferPositionDecisions` then takes the scan's answer only at a STRING leaf the
mark tree does not claim. Running the REAL pass rather than a mirror of it is the
point: the pairing rules (identity keys, anchor corroboration, the refusals)
decide which positions exist, and a second copy of them would drift. A symbol
cannot be produced by any walk of JSON, so no readback value can impersonate it.

**Attempt 4 — and the widening the issue-(#2036) revert caused.** Removing that
half also removed `expressionMaySeedANeedle`'s proven-public guard, so a bare
`{{resolve:ssm:` token with no verdict — accepted as secret-bearing on the
(#1901) premise — could rewrite every leaf that merely CONTAINED its value:

```text
bag {Env:'production', Bucket:'my-production-logs'}   source {Env:'{{resolve:ssm:/app/env}}'}
before  Bucket: 'my-{{resolve:ssm:/app/env}}-logs'    <- a value AWS never reported
after   Bucket: 'my-production-logs'
```

If that parameter is in fact public, `cdkd drift --revert` re-resolves the
baseline and renames the live bucket the day it changes. The security review
proposed dropping bare `ssm` from the seeding set; that would have failed exactly
on `cdkd state refresh-observed`, the process issue (#2012) is reported against,
because no verdict exists there. So **evidence strength bounds the BLAST RADIUS
instead of admission**: `secretsmanager:` / `ssm-secure:` / a recorded verdict
keep the substring arm, an INFERRED bare-`ssm` needle rewrites a leaf only WHOLE.
Issue (#2012)'s two rows are whole-value positions, so the closure is untouched —
measured on all four shapes.

**Attempt 5 — the same region blindness, through a different store.** The first
cut of that split treated a RECORDED `SecureString` verdict as CERTAIN. But
`recordedSecretExpressions` is keyed on the bare expression and lives for the
whole process, so on a `cdkd deploy --all` a verdict pinned where the parameter
is a `SecureString` is inherited where it is a plain `String` — and
`skipDynamicReferences` skips the lookup on a `true` verdict, so the second
region never retracts it. That is exactly what issue (#2036)'s store was
withdrawn for, arriving through the store this PR kept. Inheriting a
secret-direction verdict is safe for ADMISSION (it can only over-redact the leaf
itself) and unsafe for BLAST RADIUS, so CERTAIN is now **spelling alone**, via an
explicit allowlist rather than "the admission list minus `{{resolve:ssm:`" — so
the next prefix someone adds defaults to the narrower radius.

**And one more the merge itself introduced.** The merge's OBJECT arm runs before
its leaf rule, `isPlainObject` admits a `Date`, and `Object.entries(new Date())`
is `[]` — so the merge rebuilt a readback `LastModified` as `{}`, newly extending
issue (#2427)'s flattening (pre-existing on the populated-map paths) to the
empty-map `drainObservedCaptures` / `refresh-observed` baseline that `drift
--revert` pushes. An earlier revision of this section claimed the leaf rule
prevented exactly that; it was measured wrong. The object arm is now guarded on
`hasPlainPrototype`, fenced in both the object and array routes.

**Separately: WHICH MAP the refusal pass sees.** The derived map goes to neither
position pass. `mixedLeafMayCarryPublicReference` SPLITS on whether a map exists,
and a derived map satisfies `size > 0` while proving nothing was resolved —
handing it over would read every `{{resolve:ssm:` mixed leaf as public and
persist the decrypted `SecureString`, i.e. the regression the
`secrets-dynamic-ref` integ caught before issue (#1926) shipped, arriving through
a new door. Pinned by a dedicated case.

**3. WHICH MAP the refusal pass sees.** The derived map goes to neither position
pass, and in particular not to `refuseUncertifiedReadbackPositions`.
`mixedLeafMayCarryPublicReference` SPLITS on whether a map exists, and a derived
map satisfies `size > 0` while proving nothing was resolved — handing it over
would read every `{{resolve:ssm:` mixed leaf as public and persist the decrypted
`SecureString`, i.e. the regression the `secrets-dynamic-ref` integ caught before
issue (#1926) shipped, arriving through a new door. Both are stated in the code
at the lines that decide them and each is pinned by a dedicated case.

## Deliberate assertion flips

Five pinned assertions are flipped on purpose, each with its old value named in
place:

| file | case | was | now |
| --- | --- | --- | --- |
| `state-refresh-observed.test.ts` | `RESIDUAL: an UNPAIRED element beside a paired one` | plaintext | expression |
| `state-refresh-observed.test.ts` | `RESIDUAL: an observed KEY ... INSIDE the anchor arm` | plaintext (both leaves) | expression (both leaves) |
| `secret-redaction-array-identity.test.ts` | `RESIDUAL: an observed KEY the source does not carry` | plaintext | expression |
| `secret-redaction-anchor-pairing.test.ts` | `REFUSES a bag carrying an EXTRA key` | refuse | pair |
| `secret-redaction-anchor-pairing.test.ts` | `REFUSES a bag key explicitly set to undefined` | refuse | pair |

A sixth kept its assertion and was re-documented: `the UNPAIRED row stays open
where the two arms DISAGREE` is now `does NOT over-redact an AWS-added element
holding an UNRELATED literal`. Its element is still untouched, which is the
discrimination separating a value-keyed needle from the blanket source-subtree
rewrite the issue (#1915) fences rejected.

## Mutation probes

Every claim below was executed against the tree, one mutation at a time, restored
between runs. Suites: the four secret-redaction / state-refresh / anchor-pairing
files plus `drift-secret-redaction.test.ts`.

| mutation | failing cases |
| --- | --- |
| `derived = secrets` (no derived needles) | 10 |
| derive FIRST, into `redactByPath` (the reverted order) | 1 |
| drop `learnNeedle`'s poisoned RE-CHECK | 1 |
| turn `learnNeedle`'s duplicate-pair no-op into a poison | 1 |
| drop `{{resolve:ssm-secure:` from the seeding prefixes | 1 |
| drop the mixed-leaf slice's own token guard | 1 |
| treat every seeded expression as CERTAIN (inferred needles regain the substring arm) | 1 |
| treat every seeded expression as INFERRED (certain needles lose it) | 4 |
| demote `{{resolve:ssm-secure:` to inferred | 1 |
| let an inferred needle overwrite a certain needle's rewrite | 1 |
| drop the whole-token arm's MARK | 1 |
| let a RECORDED verdict promote a bare `ssm` token to CERTAIN | 1 |
| drop `{{resolve:ssm-secure:` from the spelled-secret allowlist | 1 |
| drop the merge, scanning `refused` unrestricted | 1 |
| let the merge's leaf rule scan a position the passes DECIDED | 3 |
| drop `hasPlainPrototype` from the merge's object arm | 2 (object + array routes) |
| key the merge on `refused === bag` again instead of the MARK tree | 1 |
| `expressionMaySeedANeedle` returns `true` | 3, incl. `drift`'s own look-alike case |
| drop the needle POISONING | 1 |
| drop `MIN_NEEDLE_LENGTH` from `learnNeedle` | 2, incl. `drift`'s own short-value case |
| restore the anchor key-COUNT check | 3 |
| pass the DERIVED map to `refuseUncertifiedReadbackPositions` | 1 |
| mixed-leaf needle maps to the whole leaf instead of the token | 2 |
| drop the `isSingleDynamicReferenceToken(bag)` guard | 1 |
| drop the mixed-leaf prefix/suffix frame check | 1 |
| accept more than one span in a mixed leaf | 1 |

The bottom six rows are the review round's, and each kills exactly ONE case with
no collateral — the probe reports the failing titles, so a fence that happened to
overlap another would show up as a count above 1.

Three mutations deliberately SURVIVE and are reported rather than hidden, all
confirmed EQUIVALENT by the round-4 test review rather than merely asserted by
me: the refusal pass's KEEP-arm mark (unreachable — the merge only runs on an
EMPTY map, where the predicate is constant `false`), and `asChild`'s two
conjuncts plus `asIndex`'s guard (`marks` can only ever be a null-prototype
object, a bag by identity, or the symbol, so both readings agree). They are
structurally unfenceable, not untested branches, and the code says so beside
them.

**Two fixtures came back non-discriminating and were rebuilt** — reported because
the probe is what caught them, not review:
- the whole-token MARK fence first had NO needle in it at all, so
  `deriveReadbackNeedles` returned early, the merge never ran, and the mutant
  survived for a reason unrelated to the mark. It now carries an unrelated
  needle-yielding position to keep the merge alive.
- see also the two below, from round 1.

**Two of those fences were VACUOUS on the first cut and were rewritten**, which
is the part worth reading rather than the table:

- `pass the DERIVED map to the refusal pass` came back GREEN. With a mixed leaf
  whose frame matches, `redactByPath` value-scans it with the derived needle and
  produces the source form anyway, so the refusal pass's answer was a confluence
  point. The fixture now uses a leaf whose frame AWS normalised, which the needle
  cannot reach, so only the refusal decides it.
- `accept more than one span` came back GREEN too, because with two RESOLVED
  references the computed suffix contains a whole `{{resolve:...}}` token that a
  resolved readback cannot end with — the frame check refuses it independently.
  The fixture now uses a second reference cdkd leaves LITERAL (`ssm-secure:`,
  the unsupported-service arm), which is the one shape where the relaxed form
  really would extract. That also establishes the refusal is CONSERVATIVE rather
  than a correctness guard, and the test says so.

## Integ

`tests/integration/secrets-dynamic-ref` gains two env vars and three arms.

**The first cut of the issue (#2036) arms rested on a FALSE PREMISE and was
rebuilt** — this is the second review blocker, and it was measured rather than
argued. Phase 1f asserted that `cdkd state refresh-observed` leaves `PUBLIC_URL`
as the EXPRESSION and called that issue (#2036)'s residual; that arm fails on
fixed code AND on `main` (the pre-merge run at `e7b25d43`: rc=1 at exactly that
line, 332s, earlier phases green). Phase 1g's CONTROL 3, claimed as the issue
(#2036) closure, passes IDENTICALLY on `origin/main` — zero discrimination. Both
had one upstream cause: `properties` holds a public `ssm` String RESOLVED by
construction (issue (#1901)), so the POSITION SOURCE carries no reference at all,
`refuseUncertifiedReadbackPositions` never reaches its mixed-leaf arm, and the
readback passes straight through whatever the verdict store says.

- **Phase 1f2 (new)** — the issue (#2012) row against a real readback. It deletes
  `SSM_SECURE_COPY` from the record's persisted `properties` (AWS still reports
  it), stamps the decrypted value into the observed bag, refreshes, and asserts
  the key took the expression by derived needle while the public `SSM_VALUE` was
  NOT dragged along. `properties` is restored afterwards.
- **Phase 1f3 (new)** — issue (#2036)'s residual, on a source that actually
  carries it. A public `ssm` EXPRESSION survives in `properties` only via
  `cdkd import`'s warn path, so the phase STAMPS that shape (the same S3
  write/restore idiom Phases 1f / 1f2 already use), refreshes, and asserts the
  leaf takes the expression. It also pins the needle's BLAST RADIUS: once
  `PUBLIC_URL`'s source carries the reference, the needle learned from it also
  rewrites `SSM_VALUE`, which holds the same resolved value of the SAME
  parameter. It is a separate phase for exactly that reason — folding it into
  Phase 1f2 would destroy that phase's `SSM_VALUE was not dragged along` control.
  The blast-radius value is asserted EXACTLY rather than as an either/or, and it
  was settled at unit scale first rather than guessed at and paid for with a live
  run. **The blast-radius arm is also the one that DISCRIMINATES**: `main` leaves
  `SSM_VALUE` resolved. The residual assertion beside it is labelled a PIN, not
  coverage — `main` refuses that leaf too, by the older `secrets.size === 0` rule
  rather than by the absent PROVEN-public verdict. That is the same measurement
  that demoted Phase 1g's CONTROL 3, applied to this PR's own new phase.
  `properties` is restored BEFORE any assertion runs, so an assertion failure
  cannot leave the stamp behind for Phase 1g's deploy.
- **Phase 1f (corrected)** — `PUBLIC_URL` now asserts the RESOLVED value, with
  the reason stated in place: its position source carries no expression, so there
  is nothing to refuse.
- **Phase 1g (corrected)** — CONTROL 3 keeps its assertion and loses its claim.
  It is relabelled a sanity control (it still catches a mask/drop regression at a
  leaf whose correct answer is the plaintext, on the deploy path) and explicitly
  says it is NOT issue (#2036) coverage. Its SecureString twin `SSM_SECURE_COPY`
  must still hold the expression on the same deploy, unchanged.

There is no issue (#2036) CLOSURE direction to fence any more — the fix was
withdrawn (see above), so the residual Phase 1f3 pins is the SHIPPED behaviour
rather than a known-remaining gap in a closed issue. The end-to-end arm for a
future closure is tracked as issue (#2425), which carries its own measured
constraint: the deploy path cannot host it, because a properties-borne public
expression makes the resource read as CHANGED and the next UPDATE rewrites it
resolved.

## Live-test status: RUN AND GREEN

`/run-integ secrets-dynamic-ref` against real AWS at `547ad4e4`: **rc 0, 466 s,
zero FAIL lines**, ledger row committed. None of the integ markgate gates
activate for this diff (`secret-redaction.ts` / `drift.ts` sit outside the
`integ-destroy` / `integ-broad` / `integ-local` scopes), so this run was a
commitment rather than a gate — which is exactly why it is reported here.

The run BEFORE this one, at `e7b25d43`, FAILED (rc 1, 332 s) at the Phase 1f
`PUBLIC_URL` assertion, precisely as the test review predicted. That failure was
the fixture's premise, not the product: the arm asserted issue (#2036)'s residual
on a leaf whose position source carries no reference, and it failed on `main`
too. Both rows are in the ledger.

Live values from the rebuilt arms, all matching the unit-scale predictions:

```text
OK: observed PUBLIC_URL kept the resolved value (premise pin; its source carries no expression)
OK: the orphaned SSM_SECURE_COPY took the expression by DERIVED NEEDLE (#2012)
OK: observed PUBLIC_URL took the expression (#2036's residual — a PIN, same on main)
OK: SSM_VALUE took its OWN parameter's expression by value (the needle's blast radius)
OK: the decrypted SecureString is absent from the WHOLE state document
```

The blast-radius value was settled with a unit case FIRST and only then asserted
exactly in the fixture, rather than guessed at and paid for with a second AWS
cycle.

**Cleanup verified independently of the fixture's own sweep** (this fixture
deliberately seeds secrets into state, so "the object is gone" is not "the
content is gone"): 0 objects, 0 versions and 0 delete markers across the entire
`cdkd/CdkdSecretsDynamicRefExample` prefix, plus 0 IAM roles, 0 Lambda functions,
0 secrets (including planned-deletion) and 0 SSM parameters.

## Issue (#2179)'s verification row

Issue (#2179) asks whoever closes issue (#2012) to check the destroy-events sink, because a state
record holding plaintext becomes durably published the moment a delete fails.
**Checked on this tree, and the coupling does not exist**:

- `destroy-runner.ts:1570` / `:1584` (both `RESOURCE_FAILED` records inside
  `runDestroyForStack`'s per-resource `catch`) still persist an UNMASKED message into
  `deployments/{runId}.jsonl` — unchanged by this PR, and issue (#2179) stays open on its
  own terms;
- what a delete can quote back is bounded by what it SENDS, and
  `provider.delete` receives `resource.properties` and never
  `observedProperties` (which is read only by `countProtectedResources`, for two
  deletion-protection flags);
- `properties` is redacted against the TEMPLATE source, so
  `isReadbackProjectedFromState` is false for it and issue (#2012)'s residual rows never
  applied there at all.

So no plaintext reaches that sink through issue (#2012)'s residuals, and after this
change those residuals are closed in `observedProperties` too.

https://claude.ai/code/session_01UUQehEnrHAaJJrouKCPuwt


go-to-k:
# Review round 1 verdicts + full fix designs (evidence preservation — continuation next session)

Four reviewers ran against HEAD e7b25d43 (3-axis + security add-on). Spec: clean. Security, code, test: **two distinct blockers**, four minor mutation gaps, three nits. No merge this session; this PR is now a draft. Everything needed to finish is recorded below so the next session starts from designs, not re-derivation.

**The parent orchestrator's live `/run-integ secrets-dynamic-ref` run against e7b25d43 FAILED exactly as the test reviewer predicted** (Phase 1f `PUBLIC_URL` assertion, rc=1, 332s) — treat that as confirmation of Blocker 2, not as a product regression.

## Blocker 1 (security): pass ordering — the derived-needle scan can UN-CERTIFY a pairing

**Mechanism.** `redactSecretsForState` currently derives the needles and feeds them into `redactByPath` BEFORE `refuseUncertifiedReadbackPositions` runs. The value scan can rewrite a frame LITERAL that happens to contain a learned plaintext (a coinciding anchor — exactly the issue (#2012) population). The refusal pass then re-runs `unkeyedArrayPairsByAnchors` against the SCANNED bag: the rewritten anchor no longer deep-equals its source, the whole array refuses, and a sibling MIXED leaf that the SHIPPED code redacts by position is persisted in full plaintext. Regression of shipped redaction, GHSA disclosure direction.

**Probe that proved it** (security reviewer, run against BOTH trees): source array `['frame-SECRETVALUE123-x', 'postgres://{{resolve:...username}}:{{resolve:...password}}@h/db', '--flag']`, readback `['frame-SECRETVALUE123-x', 'postgres://admin:hunter2secret@h/db', '--flag']`, plus a whole-token sibling elsewhere in the record resolving to `SECRETVALUE123` (which seeds the needle). Base 1a73e99d: connection string redacted by position (anchor leaks — the known bug). Tip e7b25d43: anchor fixed but the connection string persists in plaintext. The unscannable-sibling classes that leak this way: a two-span mixed leaf (`learnMixedLeafNeedle` declines, spans != 1 — the dominant CDK shape), a plaintext below `MIN_NEEDLE_LENGTH`, and a poisoned duplicate value.

**Fix design (drafted and validated in-session; re-apply verbatim).** Reorder so BOTH position passes run first on the untouched bag — with an empty map `redactByPath`'s value arms are no-ops, so their combined output is byte-identical to the shipped pipeline by construction — and the derived-needle scan runs LAST, as a plain value pass over their output (`redactSecretsForState(refused, derived)` with no source). The scan can then only ADD rewrites, never un-certify a pairing. Replace the `source !== undefined` branch of `redactSecretsForState` with:

```ts
if (source !== undefined) {
  const positioned = redactByPath(bag, source, secrets, rules, new Set(secrets.values()));
  if (!isReadbackProjectedFromState(rules)) return positioned as T;
  // (keep the existing inheritance + scrub-exclusion comment, and the
  //  "`secrets`, NOT the derived map" hazard comment — both still hold)
  const refused = refuseUncertifiedReadbackPositions(positioned, source, secrets);
  // DERIVED NEEDLES (issue 2012), applied STRICTLY AFTER both position passes,
  // as a plain value scan over their OUTPUT. First revision fed them into
  // redactByPath BEFORE the refusal pass; the PR 2415 security review probed
  // the un-certification regression described in this thread. Position passes
  // first == byte-identical to shipped (empty map makes redactByPath's value
  // arms no-ops); the derived scan can then only ADD rewrites.
  const derived = deriveReadbackNeedles(bag, source, secrets, rules);
  return (derived === secrets ? refused : redactSecretsForState(refused, derived)) as T;
}
```

Notes that must ride along: (a) the learn pass stays on the RAW bag (`deriveReadbackNeedles(bag, source, ...)` — it needs pre-substitution values); (b) the derived map still never reaches `refuseUncertifiedReadbackPositions` (its `size > 0` split in `mixedLeafMayCarryPublicReference` would read it as "a pass resolved this bag"); (c) the value scan's own token guards protect already-substituted expressions; (d) update the object-walk comment in the refusal pass that claims the value scan already ran ("redactByPath has already value-scanned this position") — under the new order it runs AFTER; (e) update `deriveReadbackNeedles`' JSDoc ("hands them to redactByPath as the secrets map" becomes "applied as a final value pass"), the changelog entry, the `.claude/rules/layout-deployment-secrets.md` one-liner ("feeds `redactByPath` pairs" likewise), and the PR body's mechanism paragraph + mutation table row 1/9 wording. Add the regression test: `does NOT un-certify a pairing whose anchor embeds a learned plaintext` — the probe fixture above as a unit case, asserting the mixed leaf takes its source expression AND the anchor is rewritten by value AND no plaintext survives; it fails under the old order by construction.

## Blocker 2 (test + code + live integ): the issue (#2036) integ arms rest on a FALSE premise

A plain deploy persists `properties.Environment.Variables.PUBLIC_URL` RESOLVED (the public reference resolves at deploy; the `Fn::Join` source refuses the skeleton pass and the value scan passes through), so the readback walk's SOURCE leaf for this key is a resolved literal — the mixed-leaf arm is NEVER consulted. Consequences, both measured: Phase 1f's new `RO_PUBLIC_URL == EXPECTED_PUBLIC_URL_EXPR` assertion fails on fixed code AND on main (confirmed by the parent's live run: rc=1 at Phase 1f); Phase 1g's CONTROL 3 passes identically on origin/main (zero discrimination). Root cause: a public mixed EXPRESSION survives in `properties` only via `cdkd import`'s warn path, and the fixture does not create that shape.

**Fix design.** Fold a `properties.Environment.Variables.PUBLIC_URL = <expression form>` stamp into Phase 1f2's existing S3 write/restore cycle (same idiom, same restore), and move the residual-direction assertion there: after `refresh-observed`, the stamped-source leaf must take the EXPRESSION (empty-map refusal — the documented issue (#2036) residual; no verdict exists in that process). Phase 1f's current PUBLIC_URL assertion changes to expect the RESOLVED value with an honest comment (source carries no expression, nothing to refuse). Phase 1g's CONTROL 3 either drops its issue-(#2036) claim or is relabeled a non-discriminating sanity control — do NOT leave it labeled as coverage. The issue (#2036) CLOSURE itself (proven-public verdict admits the resolved value) is thoroughly unit-fenced (both polarities, both poisoning orders, every-vs-some, stray opener, derived-map door); the end-to-end closure site is `cdkd drift --revert` / `--accept` (the process that has both the expression-bearing baseline and in-process GetParameter verdicts), NOT the deploy path — building that arm is optional follow-up, not a merge requirement.

## Minor mutation gaps (test reviewer; each one small unit case)

1. `learnNeedle` poisoned re-check (`if (collector.poisoned.has(plaintext)) return;`) deletable with all suites green — add a THIRD certified position after the poisoning pair (`{A: EXPR, B: EXPR_B, C: EXPR, Copy: plaintext}`; Copy must stay plaintext).
2. `forgetSecretExpression`'s CONFLICTING-preserving disjunct deletable — add record -> forget -> forget: `isProvenPublicExpression` must stay false (de-poisoning would keep a decrypted SecureString).
3. Duplicate-pair no-op (`if (already === expression) return;`) mutated to poison stays green in units — add two certified positions sharing ONE (plaintext, expression) pair plus a Copy; Copy must take the expression.
4. `SECRET_BEARING_REFERENCE_PREFIXES`' `{{resolve:ssm-secure:` entry deletable — add: source `{Tok: '{{resolve:ssm-secure:/x}}'}`, bag `{Tok: 'cfn-resolved-plain', Extra: 'cfn-resolved-plain'}` (the import-migrated shape where CloudFormation resolved server-side); Extra must take the expression. (`'{{resolve:ssm-secure:'.startsWith('{{resolve:ssm:')` is false — the entry is load-bearing.)

## Nits (one-line dispositions)

- Security nit — `learnMixedLeafNeedle` lacks the bag-is-a-token guard its whole-token sibling has: add `if (isSingleDynamicReferenceToken(<extracted slice>)) return;` for symmetry (identity-rewrite-only today).
- Code nit — `learnNeedle`'s JSDoc justifies the `MIN_NEEDLE_LENGTH` floor via `buildNeedleRegex`, but the load-bearing reason is the whole-VALUE arm matching at any length; fix the comment.
- Spec nit — issue go-to-k/cdkd#2179 cites `destroy-runner.ts:1449/:1463`; current lines are `:1577/:1591`. PR body is correct; optionally note on that issue.

## Docs correction identified during synthesis

`docs/cli-reference.md`'s new issue-(#2036) paragraph claims "`cdkd deploy` and `cdkd drift` now reuse the parameter type" — the deploy-path claim is moot (a properties-borne public expression makes the resource read as CHANGED, and the next update rewrites it resolved; the unchanged+expression configuration is unreachable from a template-declared leaf). Reword to name `cdkd drift --revert` / `--accept` as the beneficiaries and `cdkd state refresh-observed` as the still-refusing residual. The fixture README's per-path table needs the same correction for the `PUBLIC_URL` row.

## Remaining steps to merge (in order)

1. Apply Blocker 1 reorder + regression test; Blocker 2 verify.sh/fixture-README fix; 4 minor fences; 3 nits; docs corrections (changelog, rules one-liner, cli-reference, PR body mechanism paragraph + mutation table).
2. Full gates: `/check` (full suite) + `/check-docs`; commit; push.
3. Delta re-review: security reviewer MINIMUM against the fix delta (its blocker's fix is new code no reviewer has seen); test reviewer recommended for the new fences.
4. `/run-integ secrets-dynamic-ref` with the FIXED verify.sh (the current one fails at Phase 1f by construction — do not re-run it unfixed).
5. Merge (squash) after CI green; auto-closes issues (#2012) and (#2036).

Review provenance: 3-axis + security tier (diff >= 1000 LOC + secret-flow surface); spec reviewer clean with citations; the five flipped pinned assertions were confirmed red on origin/main (not vacuous); the derived-map-not-evidence and two-span fences were rewritten in-session after mutation probes proved the first fixtures non-discriminating, and re-proved red.

go-to-k:
Parent-session closing note (2026-09-02T00:05Z), for the continuation session:

- The predicted integ failure was MEASURED live before the lane's warning landed: `/run-integ secrets-dynamic-ref` at `e7b25d43` failed at exactly the Phase 1f `PUBLIC_URL` assertion the test review called (rc=1, 332s; earlier phases green). Ledger row committed as `8791a9c0`. Post-run cleanup verified: all fixture AWS resources gone (secret + both SSM params + no roles/functions), S3 prefix fully purged (0 versions, 0 delete markers), and the leaked deployment-event files were grepped for all three seeded plaintexts before deletion (zero hits — no disclosure).
- Process lesson for the retro (stage 10, which rides the continuation session): the parent started this integ in parallel with the review dispatch to save deadline time, violating `/run-integ`'s own "review first, integ last" rule — the burned cycle is the cost that rule predicts. Do not repeat under deadline pressure; the review verdict arrived 4 minutes after the deploy started.
- Continuation checklist = the 5 steps in the comment above, plus: (6) run the /work-issues stage-10 retro for the 2026-09-01 run when this PR merges.

PR #2465, opened 2026-09-02
fix(utils): spread awsClientDefaults() at every client site, including the dynamic-import ones

masucamp:
The follow-up promised in (#2398). Every `new XClient(...)` under `src/` spreads
`awsClientDefaults()` first — 158 sites across 79 files — and
`tests/aws-client-defaults-allowlist.json` is empty.

## The checker was tightened first, and twice

Before a line of the migration, so the sweep is validated by the final rules
rather than the ones that let its predecessor through. Round one closed the three
bounds recorded on (#2398). Round two closed the hole review found: the walk read
only top-level `ImportDeclaration`s, so 18 destructured
`await import('@aws-sdk/client-*')` sites had no verdict at all — not a gap,
nothing — plus 6 more in the `new mod.XClient()` namespace form. `deploy.ts`'s
`GetCallerIdentity`, on every deploy, was among them.

## Completeness is measured, not asserted

An earlier version of this body claimed "there is no version of this change that
forgets one and still passes". Review rejected that, correctly: a green run
speaks only for the population the detector can SEE, and mine could not see a
quarter of it. The sentence is kept here as the thing that was wrong, because the
round-1 comment refers to it.

What replaces it is a check. `aws-client-defaults-fence.test.ts` reconciles the
AST walk against a method wrong in different ways — a text scan for the
construction itself. Every hit must be a site the walk found, or prose. Reverting
the destructured arm makes it name 18 unexplained sites. One shape defeats both
methods identically and is recorded as the bound on the premise: an aggregated
client (`import { S3 } from ...`) has no `Client` suffix for either to key on.

## Two fail-open exemptions removed

Self-review found the getter-arm guard in `sharedRoot` pinned by nothing —
mutating it away redded no test — and wrong for a constructible shape: a getter
that calls the helper and then spreads a once-evaluated bag hands out the bag's
`requestHandler` on later-key-wins. Its twin, the argument-level exemption, had
the identical defect. Both are gone, each with an accept/reject pair; restoring
either reds its own reject fixture and nothing else.

The checker's report is byte-identical to before either removal — which is what
shows the real sites never leaned on them, and what let the earlier verification
stand.

## The migration

Six shapes; 53 of the 158 are the
`this.providerRegion ? { region: this.providerRegion } : {}` idiom. The ternary
is wrapped rather than rewritten, so each provider's diff is an import plus a
wrap. Eight sites needed the profile binding their own scope uses; typecheck
found seven and review found the eighth.

`--profile` reaches these clients through `AWS_PROFILE`, which
`src/cli/program.ts` sets in a `preAction` hook, so an argument-less call
resolves the same identity the client already did — that is what makes the sweep
mechanical rather than a per-site judgement.

## Verification

878 files / 18,120 tests green, `gen:all-matrices` and `audit:coverage:check`
clean, fence at 158/158, no new lint warnings.

Live A/B, `cdkd gc --dry-run --state-bucket X`: `connect ECONNREFUSED
127.0.0.1:1` against a dead proxy, `self-signed certificate in certificate
chain` with the variables unset — the second being the original bug.
`deploy.ts`'s STS site is not reachable end to end without credentials; its
coverage is the checker plus (#2398)'s routing-agent tests, stated rather than
glossed.

Gates: `integ-destroy`, `integ-broad` and `integ-local` all activate; a broad run
covers the first, so two runs.


go-to-k (changes requested):
Thanks — the migrated 101 sites themselves review clean (defaults spread first everywhere, all 53 wrapped ternaries preserve the original expression verbatim, the 8 explicit-`profile` sites each thread their own scope's binding, and explicit `credentials` sites all spread after the defaults so site identity wins). The checker tightening for bounds 1 and 2 is genuinely discriminating — we ran the old and new checkers against your reject fixtures and the old one passes them.

One blocker and four smaller items before this can merge (4 reviewers, 3 of which converged on the blocker independently):

## Blocker: dynamic-import client sites are invisible to the checker and unmigrated

`sdkClientIdentifiers` (`scripts/check-aws-client-defaults.ts:111`) walks only top-level `ImportDeclaration`s, so `const { STSClient } = await import('@aws-sdk/client-sts')` sites contribute zero identifiers and never get a verdict. At the PR tip, 17+ such sites construct clients WITHOUT `awsClientDefaults()` (verified line-by-line):

- `src/cli/commands/deploy.ts:490` — bare `new STSClient({ region })` + `GetCallerIdentity` on **every deploy**; the original #2388 failure class survives on the mainline path
- `src/cli/commands/export.ts:2281` (`new ApiGatewayV2Client({})`), `:2336` (`new IAMClient({})`) — a file this PR already edits
- `src/cli/region-options.ts:225`
- `src/assets/asset-publisher.ts:230`, `src/assets/asset-redirect.ts:573`, `src/cli/commands/publish-assets.ts:156`
- `src/cli/commands/local-invoke.ts:1258,1483`; `local-start-api.ts:2618` (SSO credential resolution — the exact failure mode the helper's docstring names), `:2650`, `:3115`; `local-run-task.ts:429,454,683`; `local-invoke-agentcore.ts:1609,1857`
- `src/local/httpv2-service-integration.ts:144-169` — six `new mod.XClient({ region })` sites (SQS/SNS/EventBridge/Kinesis/SFN/SSM)

Meanwhile `src/deployment/intrinsic-function-resolver.ts:2342` and `src/cli/config-loader.ts:523` are dynamic-import sites that WERE migrated — so the class was known, and the diff is internally inconsistent (export.ts's two sites vs. intrinsic-function-resolver's one).

The consequence: the PR body's "no version of this change that forgets one and still passes" and the rule doc's totality claim hold only for the checker-visible population. The empty allow-list asserts a completeness the tool cannot measure, and a future author can permanently evade the ratchet via `await import`. The checker's own docstring (line 100-104) honestly scopes itself to static named imports but names only namespace imports as the deliberate non-coverage — dynamic imports are a silent hole, the opposite of that docstring's "would claim a coverage this does not have" standard.

Fix: migrate these sites AND teach `sdkClientIdentifiers` the destructured-`await import('@aws-sdk/client-*')` shape (plus a dynamic-import accept/reject pair in the fence test), so the totality claim becomes mechanically true. If you'd rather scope dynamic imports out, that needs the doc claim corrected + the exclusion fenced — but given `deploy.ts:490` is the commonest proxy-broken path left, migration is the right call.

## Smaller items

1. **Bound 3 (file-global `uses` count) has no test pair.** The PR body says each of the three bounds got an accept/reject pair; bounds 1 and 2 have theirs, but the uses-count polarity was closed by comment only (`scripts/check-aws-client-defaults.ts:441-449`). We verified the documented fail-closed behavior is real (two same-named single-use local bags both flip to `shared-defaults`) — that exact shape would make a good reject-side fixture.
2. **The "stops following at a getter that calls the helper itself" test (`tests/unit/scripts/aws-client-defaults-fence.test.ts:194`) does not exercise the branch its title names.** Its site literals spread `...awsClientDefaults()` directly, so `sharedDefaultsSource` exits at the argument-level exemption before `sharedRoot` runs — and its getter spreads `this.cached`, not the helper. The getter-stop branch (`scripts/check-aws-client-defaults.ts:370-371`) is actually covered by the `:209` test ("does NOT flag a getter spread"). Either retitle `:194` for what it pins, or make its getter call the helper with no site-level call.
3. **The exemption comment over-claims** (`scripts/check-aws-client-defaults.ts:305-309`): "a literal that calls `awsClientDefaults()` DIRECTLY is exempt whatever else it spreads: ... builds a fresh handler per client." If a later-spread once-evaluated bag itself carried `requestHandler`, the later spread would win and real sharing would be exempted. No such bag exists today, but the comment states an invariant the code does not enforce — worth a caveat sentence.
4. **`src/deployment/intrinsic-function-resolver.ts:2340`** spreads an argument-less `awsClientDefaults()` under a `credentialConfig` that can carry `profile`; correctness rests on the `AWS_PROFILE` mirror in `program.ts`. Passing `scoped.credentialConfig?.profile` into the helper would match the seven sibling sites and drop the hidden coupling.

Items 1-2 keep the checker's "each bound has a pair" story true and are cheap to add alongside the blocker fix; 3-4 are take-or-leave polish.

masucamp:
Blocker and all four smaller items addressed, pushed as 700b820. You were right
about the blocker and about what it meant for the claim I made.

## The blocker

`sdkClientIdentifiers` now knows the destructured dynamic form AND the
`const mod = await import(...); new mod.XClient()` namespace form -- the latter
needed a second arm, since the construction is a property access rather than an
identifier. The detected population went 134 -> 158 sites across 67 -> 79 files,
and the 23 newly-visible sites are migrated, `deploy.ts`'s STS call included.

The inconsistency you flagged has a cause worth recording: the one dynamic site
that WAS migrated sits in a file carrying a top-level
`import type { ServiceDiscoveryClient }`, which was enough for the old
`sdkClientIdentifiers` to record the name. It was detected by ACCIDENT, and I
read the inconsistency as noise instead of as evidence about the tool.

The docstring now enumerates what the walk still cannot see -- a static
namespace import, an inline `(await import(...)).XClient`, a `.then(m => m.X)`
chain, a re-export -- rather than naming one exclusion and going quiet about the
one that mattered.

## The claim, made mechanical

You put it exactly right: the empty allow-list asserted a completeness the tool
could not measure. Rather than re-word the claim, I made it checkable.

`tests/unit/scripts/aws-client-defaults-fence.test.ts` now reconciles the AST
walk against a method that is wrong in DIFFERENT ways -- a text scan for the
construction itself. Every hit must be a site the walk also found, or a line of
prose. Today: 171 text hits, 158 detected, 13 in comments, nothing unexplained.

It is probed the way you would want: reverting `sdkClientIdentifiers` to its
top-level-only form makes that test fail naming 18 sites, `deploy.ts:491` first.
That is the check that would have caught this before you had to.

The reconciliation is also the step I skipped. A `grep -c 'new [A-Za-z]*Client('`
takes seconds and disagrees with the walk in a way one method's green never can.

## Smaller items

1. Bound 3 has its reject pair now -- two same-named single-use local bags, both
   flipping to `shared-defaults`, exactly the shape you measured.
2. Fixed rather than retitled: the site literals no longer call the helper, so
   the getter is the only caller and `sharedRoot` actually runs. You were right
   that `:209` was covering the branch all along.
3. Caveat added: the exemption is a heuristic, not an enforced invariant, and a
   later-spread bag carrying its own `requestHandler` would win. Nothing
   produces one but the helper, which is why it holds today rather than by
   construction.
4. `scoped.credentialConfig?.profile` is passed now, so the site matches its
   seven siblings and the `AWS_PROFILE` coupling is gone.

## Verification

877 files / 18,044 tests green, `gen:all-matrices` and `audit:coverage:check`
clean, fence at 158/158, no new lint warnings.

One thing I could not do, said plainly rather than glossed: `deploy.ts`'s STS
site is not reachable end to end without real credentials -- it sits behind a
state-bucket check. The A/B I can run (`cdkd gc --dry-run --state-bucket X`)
gives `connect ECONNREFUSED 127.0.0.1:1` against a dead proxy and
`self-signed certificate in certificate chain` with the variables unset. That
site's coverage is the checker plus (#2398)'s tests that the helper attaches a
routing agent.

Gate scopes are unchanged from the last round: `integ-destroy`, `integ-broad`
and `integ-local` all activate, and a broad run covers the first, so two runs.

go-to-k (commented):
Round 2 verified — the substance is clean, and every round-1 finding was genuinely fixed, not narrated. We re-ran everything empirically at 700b820c:

- Checker at tip: 158/158 across 79 files, exit 0, allow-list empty. Independent checker-free reconciliation (our own text scan): 171 hits = 158 code + 13 prose, zero unmigrated sites in the tree.
- Your revert probe reproduces: destructured arm reverted → the reconciliation test reds with exactly 18 unexplained sites. The prose filter fails closed (a swallow needs a double fault that formatted TS never produces), and the multi-line shape the regex misses is still covered by the primary `report.gaps` sweep — no hole between the two methods.
- All four new dynamic-import fixtures red under the round-1 checker — real discrimination, both arms. The non-SDK `helpers.FooClient` refusal correctly passes under both.
- Bound-3 pair: a per-declaration-scoping mutant reds ONLY the new fixture (all four pre-existing shared-bag fixtures stay green under it) — it is the sole fence for the file-global polarity, exactly what was missing.
- The rewritten `:194` fixture reaches its titled branch: two getter-arm mutants both red it, while the round-1 fixture shape stays green under the same mutants — round 1's finding was real and the rewrite fixes it.
- The 23 new sites: defaults first at every one, no option lost, `profile` threaded at all four sites whose scope carries one (region-options, asset-redirect, local-invoke-agentcore:1610, local-start-api:2619) and at no site that lacks one. `intrinsic-function-resolver`'s `credentialConfig` spread order is right even when it carries static credentials (later-key-wins restores the site's identity, and static creds need no network hop).
- The deploy.ts residual note is accurate: the site sits behind `verifyBucketExists`, so no credential-free live probe reaches it; checker + reconciliation + #2398's routing-agent tests are the honest floor. (One correction to the reply: the revert probe's first-named site in walk order is `asset-publisher.ts:231`; `deploy.ts:491` is third.)

Two small things before merge, both prose-only:

1. **The rule doc disagrees with its own commit's checker.** `.claude/rules/proxy-support.md:45` still says "134 sites across 67 files" and "1726 sites where there are 134", and the paragraph this PR adds at `:53` says "all 134 client sites" — while the same commit's checker prints 158/79. `scripts/check-aws-client-defaults.ts:201` says "where there are 165" and `tests/unit/scripts/aws-client-defaults-fence.test.ts:75` says "where there are 134" — three mutually inconsistent counts at one commit. Since these counts drift by construction, consider stating them once as "158 as of this PR" (or "~160") and pointing at the checker as the live count. The doc should also record the reconciliation test as the completeness oracle — it is the piece that makes the totality claim mechanical, and the doc is where the next session learns why.
2. **The "stated exhaustively" exclusion list has one more member: the aggregated-client shape** (`import { S3 } from '@aws-sdk/client-s3'` / `new mod.S3()` — every SDK package exports a non-`Client`-suffixed convenience class). Both the AST walk and the reconciliation regex key on the `Client` suffix, so this is the one shape where the two methods are wrong in the SAME way and the reconciliation cannot catch it — worth naming precisely because it contradicts the "wrong in DIFFERENT ways" premise. Nothing in the tree uses one today; one line in the exclusion comment.

Optional, take or leave:

3. The destructured-dynamic ALIAS arm (`const { S3Client: Bucket } = await import(...)`) is handled correctly (we probed it — `['missing']`) but has no fixture; a one-fixture pin would close the untested branch.
4. `aws-client-defaults-fence.test.ts:383` says "17 `await import` sites had no verdict" — strictly 18 destructured sites had no verdict (config-loader.ts's migrated one included), plus the 6 namespace-form ones; the "17" is right where the prose says "sat unmigrated" (checker:144, test:283), so only this one spot conflates the two scopes.

For what it's worth, your "17 of them sat unmigrated" counts survived our attempt to red-pen them — 17 destructured-unmigrated is exact; the probe's 18 includes the already-migrated config-loader site, which is why the two numbers differ and both are right.

After the doc-count fix lands, this is mergeable from our side.

masucamp:
All four taken, pushed as 214547f. Prose and tests only — no behaviour change, so
the round-2 verification stands.

**1. The counts.** Stated once now, as of this commit, with the checker named as
the live source: the rule doc says "158 sites across 79 files at the sweep ...
run the checker for today's count rather than trusting this one", and the two
`1726 sites where there are N` lines both became "against a real population of
~160". The floor comment in the test says why it is a round number below the
real count rather than a fourth place for the count to drift.

The rule doc also records the reconciliation as the completeness oracle, with
the distinction spelled out — an empty allow-list says the checker found
nothing, not how much it can see — plus the revert-probe result and the
same-way-wrong bound below. That is the piece the next session needs and it was
missing.

**2. The aggregated-client shape.** Added, and thank you for framing it as the
bound on the premise rather than as one more exclusion — that is what makes it
worth naming. Both the AST walk and the reconciliation's text scan key on the
`Client` suffix, so `import { S3 } from '@aws-sdk/client-s3'` is the one shape
where the two methods are wrong IDENTICALLY and the reconciliation cannot
rescue it. Recorded in the exclusion list and in the rule doc, with "a first use
needs both sides widened".

**3.** Alias fixture added (`const { S3Client: Bucket } = await import(...)` →
`['missing']`).

**4.** Fixed. The reconciliation docblock now says 18 destructured sites had no
verdict, 17 of them unmigrated, plus 6 in the namespace form, and names
`config-loader.ts` as the difference — migrated by hand in the previous PR, so
it had no verdict while already being correct.

**Your correction is right and mine was wrong.** I re-ran the revert probe: walk
order is `asset-publisher.ts:231`, `asset-redirect.ts:574`, `deploy.ts:491`. I
had written "deploy.ts:491 first" from memory of which site mattered most rather
than from the output, which is the kind of thing this whole PR is about. The
commit message now states the order as measured.

877 files / 18,045 tests green, fence at 158/158, `gen:all-matrices` and
`audit:coverage:check` clean, and no count in the tree disagrees with the
checker any more.

go-to-k (commented):
Round 3 verified at 214547f: checker 158/158 across 79 files (exit 0), fence suite 45/45 green on an extracted copy of the tip (44 + the new alias fixture), and the delta is prose/tests only — round 2's verification stands, as you said.

Items 1-3 are all in the tree and check out:
- The rule doc now states the count once, dated to the sweep, names the checker as the live source, and records the reconciliation as the completeness oracle (with the revert-probe result and the same-way-wrong bound). The two `1726 vs N` lines agree at "~160", and the floors comment explains why the assertion numbers are round rather than a fourth copy of the count.
- The aggregated-client shape is in the exclusion list and the doc, framed as the bound on the reconciliation's premise, with "a first use needs both sides widened".
- The alias fixture is present and passes.

One discrepancy: **item 4's fix is not in the pushed commit.** Your reply says the reconciliation docblock now reads "18 destructured sites had no verdict, 17 of them unmigrated, plus 6 in the namespace form, and names `config-loader.ts`" — but at 214547f, `tests/unit/scripts/aws-client-defaults-fence.test.ts:397` still says "17 `await import('@aws-sdk/client-*')` sites -- ... -- had no verdict at all" (and `:285` still has the round-2 wording, which was already correct as "sat unmigrated"). It looks like the hunk was written but didn't make it into the commit — perhaps lost in an amend. The item was optional, so either push the missing hunk or tell us you're dropping it; both are fine. Everything else is merge-ready from the review side, and we'll bind the review marker to whichever sha you land on.

masucamp:
Item 4 is applied. One question before I push, and one thing I am fixing anyway.

**Question.** Self-review found the getter-arm guard in `sharedRoot`
(`if (spreads.some(isDefaultsCall)) return undefined`) is pinned by nothing —
mutating it away reds no test. It is also wrong for one shape: a getter that
calls the helper and then spreads a once-evaluated bag hands out the bag's
`requestHandler` on later-key-wins, and that measures `defaults-first`.

Removing the guard classifies that correctly, and I measured that it changes no
answer on the tree — 158/158, gaps 0, suite 18,112 green, byte-identical report.
Nothing in the tree uses the shape.

So: caveat only (what I have staged now, matching how you disposed of the
argument-level exemption), or remove it with an accept/reject pair? The
"changes nothing on the tree" measurement argues for the fix; not wanting to
move behaviour after your round-3 verification argues for the caveat. Your call.

**Fixing regardless.** The PR title and body are stale from before the blocker
fix — 134/101 sites, 18,037 tests, and the totality sentence you rejected is
still sitting in the body. I will refresh both, keeping a line that records the
correction so your round-1 comment still has a referent.

go-to-k (commented):
Remove it, with the pair. Three reasons, in order of weight:

1. The caveat route reproduces the exact class we flagged at round 2 item 3: a comment stating an invariant the code does not enforce. We accepted that disposition for the argument-level exemption because nothing but the helper produces a `requestHandler`-bearing bag — but here you have gone one better and named a constructible shape where the guard's verdict is WRONG, not merely unenforced. Once the failure shape is known and cheap to fence, the caveat is the weaker option.
2. The guard is fail-open and unpinned — the worst combination this PR has been systematically eliminating. Removing it moves the polarity the direction every other round-2 fix moved.
3. "Changes no answer on the tree" is the measurement that makes the move safe: byte-identical report means our round-3 verification stands materially, and the delta re-verification is mechanical (checker run + fence suite) on our side. Don't let the round-3 sign-off argue for freezing a known wrong branch — the sign-off was of the report, and the report doesn't change.

One thing to catch while you're in there: with the guard gone, the `:194` fixture ("stops following at a getter that calls the helper itself") passes for a DIFFERENT reason — there is no direct-call stop any more; it passes because `this.cached` carries no defaults. That title will be describing a mechanism that no longer exists, which is the same title/mechanism mismatch we fixed in round 1. Retitle it to what it now pins (a helper-calling getter whose other spreads are defaults-free stays accepted), and let the new reject fixture (helper-calling getter spreading a once-evaluated defaults-bearing bag → `shared-defaults`) carry the new behavior.

The title/body refresh sounds right, including keeping a line that gives the round-1 comment a referent. Bundle item 4's docblock hunk with this push and we'll verify the lot in one pass.

masucamp:
Guard removed with the pair, retitle done. One more before I push, because it is
your call rather than mine.

Applying your round-3 test to the sibling you left as a caveat — the
argument-level exemption in `sharedDefaultsSource` — gives the same three
answers:

```ts
const shared = { ...awsClientDefaults(), region: 'x' };
new S3Client({ ...awsClientDefaults(), ...shared });   // reports defaults-first
new ECRClient({ ...awsClientDefaults(), ...shared });  // both share shared's handler
```

Constructible and wrong, unpinned by any test, and removing the exemption leaves
the report byte-identical — 158/158, zero newly flagged.

I have not touched it. Round 2 was your disposition, and you named that shape
yourself before deciding a caveat was enough, so the round-3 reasoning ("you
have gone one better and named a constructible shape") does not obviously
transfer. If the distinction is that 158 real sites lean on this exemption while
the getter arm had none, that is a good reason and I would rather hear it than
infer it.

Same treatment, or is there a difference? Happy either way; I just did not want
to ship one arm fixed and its sibling untouched in the same file without asking.

go-to-k (commented):
Same treatment — remove it with the pair. And you caught a real inconsistency in my reasoning, so let me state it rather than paper over it: the "you have gone one better and named a constructible shape" distinction does not survive your comparison. Round 2's finding already named this exact shape for the exemption ("false if a later-spread once-evaluated bag itself carries `requestHandler`"). The honest account is that round 2's caveat was the weaker disposition, made when nobody had measured what removal does to the tree. What upgrades it now is exactly what upgraded the getter arm: your byte-identical measurement, which was the missing piece — plus the sibling argument you made yourself, which is the strongest one. Shipping one arm removed and its twin caveated in the same file would leave the file arguing with itself about what "direct call present" is worth.

It is not the 158-sites reason you offered me. The prescribed form does not lean on the exemption — with it removed, a `{ ...awsClientDefaults(), ...bag }` site is clean because its OTHER spreads resolve to nothing defaults-bearing, not because the direct call short-circuits. Your zero-newly-flagged run is the proof.

Two things to take care of in the same change:

1. The round-2 caveat comment (and the surrounding JSDoc block describing the exemption — "a literal that calls `awsClientDefaults()` DIRECTLY is exempt whatever else it spreads ... tested on the literal WRITTEN AT THE SITE") describes a mechanism that will no longer exist. Rewrite it to describe what replaces it: a direct call at the site no longer exempts anything by itself; every spread is resolved on its own, and the direct call simply contributes no shared root. The "never test the resolved bag" warning inside it dies with the exemption — the concern it guarded against (exempting `new A(bag)` because the call is inside `bag`) can no longer arise.
2. Keep the pair shaped like the getter arm's: reject = two sites spreading a once-evaluated defaults-bearing bag AFTER their direct call → `shared-defaults` at both; accept = the prescribed form with a defaults-free later spread stays `defaults-first` (which also pins that the 158 real sites do not depend on the exemption).

Push the lot — both removals, both pairs, the retitle, item 4's docblock, and the title/body refresh — and we'll verify it in one pass.

go-to-k (commented):
Verified at ad34ae01, all empirically:

- Checker at tip: 158/158 across 79 files, exit 0. Fence suite 48/48 green on an extracted tip copy.
- **Byte-identity independently confirmed**: ran the 214547f checker and the ad34ae01 checker over the same tip tree — `--json` reports compare byte-identical. Both removals are behavior-preserving on the real population, exactly as you measured.
- **Both new reject pairs discriminate**: restoring both guards on a scratch copy reds exactly the two new reject fixtures (46 others stay green) — the pairs pin precisely the removals and nothing else.
- The retitle, item 4's docblock (18/17/6 with `config-loader.ts` named), the JSDoc rewrites (including the companion warning's burial), and the title/body refresh — including keeping the rejected sentence as the referent for the round-1 comment — are all in and read right.

One sentence to fix, and then this is done: the retitled accept test's comment ends with "The SITE literals deliberately do not call the helper, or `sharedDefaultsSource` would exit at the argument-level exemption and `sharedRoot` would never run" (`aws-client-defaults-fence.test.ts`, the `accepts a helper-calling getter...` case). That exemption was removed in this same commit — with it gone, a site-level call no longer exits anything (your own new argument-level reject fixture proves it: its sites call the helper and still get flagged). The sentence preserves a true fact about WHY the fixture is shaped this way — it was reshaped in round 1, when the exemption existed — but states the dead mechanism in the present tense, which is the stale-rationale class this PR has now buried twice elsewhere. Reword to past tense ("...or, at the time this fixture was reshaped, `sharedDefaultsSource` would have exited...") or drop the clause; the fixture's shape needs no defense now that the site-level call is harmless either way.

Push that and we'll set the review marker on the final sha — nothing else remains from the review side.

masucamp:
Fixed and pushed as baf5541. You are right, and I verified all three parts of it
before touching the sentence: the exemption it names is gone from this commit,
and the argument-level reject fixture does carry a site-level call and is still
flagged.

Reworded to past tense rather than dropped, because the round-1 reshape is still
the reason the fixture looks the way it does — but the clause now says the
exemption is gone and that a site-level call is harmless either way, pointing at
the reject fixture below as the demonstration. I measured that last claim rather
than asserting it: adding a site-level call to that fixture's sites leaves both
`defaults-first`.

Diff is that comment and nothing else — 878 files / 18,120 tests green, fence
158/158, no main movement since the last push.

go-to-k (approved):
Verified at baf5541: comment-only delta, checker 158/158, fence 48/48 on an extracted tip copy. The reworded clause is accurate (and your added measurement — a site-level call added to that fixture leaves both verdicts defaults-first — matches what we probed on our side). Review marker is set, bound to baf5541. Nothing remains from the review side; merge still needs the integ gates (broad + local runs, as you noted).

go-to-k:
Merged and released as 0.285.13. Thank you for an exceptional contribution — not just the sweep itself, but the way every review finding came back verified rather than narrated: the reconciliation oracle, the measured guard removals, and catching my own round-2 inconsistency made this PR better than the review asked for. cdkd now works end to end behind a corporate proxy thanks to you.

PR #2476, opened 2026-09-03
fix(secretsmanager): send SecretString on update only when the value source changed (#2472)

### LINKED ISSUE #2472: fix(secretsmanager): update() regenerates GenerateSecretString on every in-place update, so a Tags-only or Description-only update silently replaces the live secret value
## Summary

`SecretsManagerSecretProvider.update()` re-runs `generateSecretString()` on **every in-place update** of an `AWS::SecretsManager::Secret` that carries a `GenerateSecretString` block, and sends the fresh value as `SecretString` in `UpdateSecretCommand`. It never compares `GenerateSecretString` against `previousProperties`. Any in-place update of this resource — a Tags-, Description-, KmsKeyId- or ReplicaRegions-only change — therefore also sends a newly generated `SecretString`, which **silently replaces the live secret value**: a new version is created and `AWSCURRENT` moves to it, without any rotation workflow running against the backing service.

CloudFormation creates a new version only when the `GenerateSecretString` block or `SecretString` itself changes. CDK's `new secretsmanager.Secret(...)` synthesizes `GenerateSecretString: {}` by default, so the common `DatabaseSecret` / `Secret` + consumer shape is exactly the one affected.

The literal branch has the same defect in a milder form: on every in-place resource update an unchanged `SecretString` is re-sent, creating a new version, moving `AWSCURRENT` to it and `AWSPREVIOUS` to the former current version; the older previous version becomes unlabeled.

(A no-change redeploy does not call `update()` and is not affected; the trigger is an in-place `UPDATE` of the secret resource.)

## Affected sites (current `main`, a8874977)

- `src/provisioning/providers/secretsmanager-secret-provider.ts:148-161` — `update()`:

```ts
const generateConfig = properties['GenerateSecretString'] as Record<string, unknown> | undefined;

if (generateConfig) {
  secretString = this.generateSecretString(generateConfig);   // fresh CSPRNG value, no previousProperties comparison
} else if (properties['SecretString']) {
  secretString = properties['SecretString'] as string;        // unchanged literal re-sent
}
...
if (secretString) updateParams.SecretString = secretString;
```

- The value is sent at `secretsmanager-secret-provider.ts:220` (`new UpdateSecretCommand(updateParams)`); tags and replicas are then handled by their own API calls (`:220-289`).
- `src/analyzer/replacement-rules.ts:469-480` lists `Description` / `KmsKeyId` / `SecretString` / `GenerateSecretString` / `ReplicaRegions` / `Tags` as `updateableProperties`, so a change to any of them is classified `UPDATE` (`src/analyzer/diff-calculator.ts:310`, union-of-keys comparison at `:791-823`), and the deploy engine hands the FULL resolved property bag — `GenerateSecretString` included — to `provider.update()` (`src/deployment/deploy-engine.ts:4800-4805`).

## Reachable paths

1. **`cdkd deploy`** — any in-place `UPDATE` of the secret resource (Tags-only, Description-only, KmsKeyId-only, ReplicaRegions-only).
2. **Rollback replay** — `src/deployment/rollback-executor.ts:2210-2216` resolves the previous state's full property bag (which still carries `GenerateSecretString`), and `:2296-2303` passes it through `updateWithRollbackRetry()` to the provider, so a rollback that touches the secret replaces its value again on the recovery path.

`cdkd drift --revert` is NOT affected by this path: `readCurrentState()` omits both value-source fields (`src/provisioning/providers/secretsmanager-secret-provider.ts:418-429`), `getDriftUnknownPaths()` marks them unknown for the comparator (`:468-476`), and `buildRevertNewProperties()` starts from the AWS-readback bag (`src/cli/commands/drift.ts:3894-3902`), so the revert bag never carries the generate block.

## Impact

- A user adds a tag to (or edits the description of) a secret whose value was consumed at create time and the deploy succeeds with no signal that the value changed. A database initialized from the old value (RDS `MasterUserPassword` via `{{resolve:secretsmanager:...}}`) keeps expecting it, while applications that subsequently fetch `AWSCURRENT` — or newly launched ECS tasks resolving `Secrets.ValueFrom` — receive the new value and fail authentication. Already-running ECS tasks retain the value injected at start.
- The same happens on a rollback that replays the secret's update.
- Literal `SecretString` secrets accumulate a new version per in-place resource update; each update advances `AWSPREVIOUS`, leaving the formerly previous version unlabeled.

## Not a duplicate

go-to-k/cdkd#1160 (absent-field removal silent drop across SDK providers) is the opposite class: a field REMOVED from the template that `update()` fails to reset. This issue is a field that is PRESENT and unchanged being re-applied with a freshly generated value. Searched open issues (titles and bodies) and closed-issue bodies for `generateSecretString(` / `UpdateSecretCommand` — no hit in this sense.

## Existing test coverage

- `tests/unit/provisioning/secretsmanager-secret-provider-roundtrip.test.ts` — the update bags carry neither `SecretString` nor `GenerateSecretString`, so nothing fences either branch.
- `tests/unit/provisioning/secretsmanager-secret-provider-readcurrentstate.test.ts:103-108` — covers only the drift-unknown declaration.

## Proposed fix

In `update()`, send `SecretString` only when the value SOURCE actually changed:

- `GenerateSecretString` present: call `generateSecretString()` only when the block deep-differs from `previousProperties['GenerateSecretString']` (CloudFormation semantics).
- `SecretString` literal present: send it only when it differs from `previousProperties['SecretString']`.
- Otherwise omit `SecretString` from `UpdateSecretCommand` (its merge semantics leave the current version untouched).

Regression tests: a Tags-only update and a Description-only update must produce an `UpdateSecretCommand` input WITHOUT `SecretString`, for both a generated and a literal secret; a changed `GenerateSecretString` block / changed literal must include it. Removing either comparison from the fix must make the corresponding unchanged-generated / unchanged-literal test fail.

Adjacent, same function, SEPARATE change: `generateSecretString()` (`:375-398`) samples with `bytes[i] % chars.length` (modulo bias over an 88-character alphabet) and ignores `RequireEachIncludedType` / `IncludeSpace`. `GetRandomPasswordCommand` exists in `@aws-sdk/client-secrets-manager`, but adopting it adds an IAM permission (`secretsmanager:GetRandomPassword`) and an asynchronous API call, and cdkd would still have to merge the returned password into `SecretStringTemplate` / `GenerateStringKey` locally — so it should not ride the same PR as the update guard.

## Verification plan

- Unit: the regression tests above.
- Integ: extend `composite-stack` (its `AppSecret` at `tests/integration/composite-stack/lib/composite-stack.ts:211-219` uses `generateSecretString`) with a `CDKD_TEST_UPDATE=true` branch that adds a tag to `AppSecret` only; capture the secret's `AWSCURRENT` `VersionId` after the first deploy, redeploy with the branch on, and assert the `VersionId` is unchanged. Today it changes. `secrets-rotation-schedule` also carries a generated secret but is documented CREATE + DESTROY only and leaves a deliberately incomplete rotation, so it is not the right fixture without extra isolation.

Dup-check: searched open issues (titles and bodies) and closed-issue bodies for GenerateSecretString / UpdateSecret / secret rotation / secretsmanager-secret-provider -- none covers this root cause; go-to-k/cdkd#1160 is the absent-field reset class
Session-fit: next (not this session) -- filed from a security audit; the fix needs the unit fence plus the `composite-stack` CDKD_TEST_UPDATE branch above, then `/run-integ composite-stack` with CDKD_TEST_UPDATE=true
Severity: high -- an in-place tags-only or description-only update replaces a live credential, and consumers seeded at create time stop authenticating with no signal in the deploy output
Effort: medium (M) -- edit plus unit tests, and a run of an EXISTING integ fixture with a new UPDATE branch asserting VersionId stability
Estimate: ~2-3 h -- the integ UPDATE branch (VersionId capture before / after a tags-only redeploy) is what eats the time; the provider edit itself is small

nix-tkobayashi:
## Summary

`SecretsManagerSecretProvider.update()` re-ran its local `GenerateSecretString` generator on **every** in-place update and re-sent an unchanged literal `SecretString`. `UpdateSecret` with a `SecretString` creates a new version and moves `AWSCURRENT` to it, so a Tags-only or Description-only deploy of the CDK-default `GenerateSecretString: {}` shape silently replaced a live credential (a database seeded from the old value then rejected every consumer reading the new one), and a literal secret stacked a redundant version per update, advancing `AWSPREVIOUS` off the real previous value. The same bag is replayed by the rollback executor, so the recovery path did it too.

The value now rides an in-place update **only when its SOURCE changed**, which is CloudFormation's semantics:

- `GenerateSecretString` present: a value is minted only when the block differs from `previousProperties` — compared over a JSON round-trip with `node:util` `isDeepStrictEqual`, so a key-order difference between the state-read bag and the resolver's is not a change and neither is an explicit `undefined` member. **The block is compared as STATE SPELLS IT**: a `{{resolve:...}}` inside `SecretStringTemplate` reaches the provider as plaintext while state holds the redacted expression, so a raw comparison would always differ and re-roll on every Tags-only update for exactly the shape that embeds a secret. `asPersisted()` rewrites the desired block through the same `redactSecretsForState` the state writer uses, with the pairs from the per-resource secrets scope the deploy engine and the rollback executor bind around the provider call (absent on `drift --revert` / import, where the raw comparison applies as before). The map is handed to the redaction helper only, never enumerated or logged; `resource-secrets-scope.ts`, `.claude/rules/provider-masking.md` and `.claude/rules/layout-deployment.md` now name this second reader and why a masking function cannot substitute (`***` never equals the persisted expression).
- literal `SecretString`: sent only when it differs from the previous literal. An empty string is a value, not "absent" — a change TO `''` goes on the wire and AWS accepts or rejects it, rather than the old value being silently kept. A non-string is refused by shape (`SecretString must be a string, got <type>`, never the value) instead of being forwarded. A literal that is itself a `{{resolve:...}}` is deliberately still re-sent on every in-place update: cdkd cannot tell whether the referenced value changed since the last deploy, CloudFormation re-applies it when it did, and a redundant version is the milder failure than a stale copy.
- switching source in either direction counts as a change.
- otherwise `SecretString` is omitted; `UpdateSecret` has merge semantics, so the current version is untouched.

`cdkd drift --revert` was never affected: its bag starts from the readback, which omits both value-source fields.

Closes #2472

## Changes

- `src/provisioning/providers/secretsmanager-secret-provider.ts` — new private `changedSecretValue()`; `update()` sends `SecretString` only when it returns a value.
- `tests/unit/provisioning/secretsmanager-secret-provider-update-value-source.test.ts` — 25 cases: Tags-only / Description-only updates send no value (generated and literal), key-order-only block difference is not a change, a changed block mints a value of the new shape, a changed literal is sent, `''` is sent, both source switches, `GenerateSecretString` precedence over a literal, a rollback-replay-shaped call (previous == desired) and the reversed revert bags send nothing, the CDK-default `{}` block, an explicit-`undefined` member, a bag with neither source, non-string / array-valued refusals (asserting the message never carries the value), and a scoped group bound with `withCurrentResourceSecrets`: a `SecretStringTemplate` embedding a reference is unchanged on Tags-only, regenerated when the template changed, still sends when unscoped, a pre-GHSA plaintext record matches the raw arm, the literal-is-reference re-send, and the accepted KNOWN EDGE below. Every keep-assertion fails against the old code; the round-2 test reviewer ran the mutants (drop either comparison arm, identity `asJson`, truthy literal gate, old send gate, throw -> return) and each is killed.
- `tests/integration/secretsmanager-update-value-source/` — new real-AWS fixture (generated + literal secret). Phase 1 deploy; Phase 2 Tags-only re-deploy asserting the tag reached AWS while `AWSCURRENT` `VersionId` and the version count stay unchanged on both secrets; Phase 3 positive control (block changed to `PasswordLength: 40`, literal changed) asserting both values DO move (new `AWSCURRENT`, previous current is `AWSPREVIOUS`, two versions each), so Phase 2 cannot pass on a provider that never sends a value; Phase 4 destroy with gone-probes and an S3 object-version sweep (the literal marker lands in state.json verbatim). No secret value is read anywhere in the fixture (`list-secret-version-ids` only).
- `tests/unit/scripts/integ-secret-fixture-sweep.test.ts` — the new fixture joins the audited seeding set (it declares `unsafePlainText`).
- `src/deployment/resource-secrets-scope.ts`, `.claude/rules/provider-masking.md`, `.claude/rules/layout-deployment.md` — the reader policy of the secrets scope names its second reader.
- `docs/changelog-cdkd.md`, `docs/testing.md` (the `generateSecretString` exemption paragraph now names the fifth fixture), regenerated `docs/integ-coverage.md` / `docs/scenario-coverage.md` / `docs/_generated/*`, and the integ ledger row.

## Verification

- `/check`: typecheck + lint + format, test-project typecheck, build, full unit suite — green.
- `/run-integ secretsmanager-update-value-source` against real AWS (us-east-1), twice — once on the first cut (176 s) and again on the final tree after the review rounds (179 s): PASS, destroy 0 errors, 0 orphans, 0 surviving state-bucket object versions. Recorded in `docs/_generated/integ-last-run.tsv`; `integ-destroy` marker set on the final tree.
- Review: Codex loop on the diff until zero findings (round 1: empty-string literal, fixed; round 2: `NO ISSUES`; round 3: the value-scan edge below, accepted and documented; rounds 4 and 5, the last on the final delta: `NO ISSUES`) plus the repo's 3-axis + security reviewer set: round 1 on the design, round 2 on the fix delta (test reviewer executed the mutation probes), round 3 on the last delta (clean). Remaining reviewer nits, all cosmetic and left as-is: the KNOWN-EDGE and upstream-rotation tests overlap with the scoped unchanged test (kept as named documentation pins); the `lock.json` key in `verify.sh` `cleanup` is hand-spelled rather than derived from `STATE_PREFIX`.

## Follow-up after the maintainer review (second commit)

- `changedSecretValue()` decides UNCHANGED for a literal (`isDeepStrictEqual` over a JSON round-trip) BEFORE judging its shape, so a record a pre-fix binary wrote with a non-string `SecretString` still takes a Tags-only update; the shape refusal fires only for a value that would be sent.
- `create()` routes `SecretString` through the same shared `requireSecretStringShape()` — skipping only `''` (the documented no-version create) — so `null` / `false` / `0` / an object are refused there too and such a record can no longer be created.
- JSDoc: the collapsed-pair residual (two references in one block resolving to the same plaintext) re-rolls on EVERY in-place update, not once, and why threading the previous block as a position source would erase a genuine reference switch; the precedence consequence for a previous bag carrying both a block and an ignored literal is recorded.
- Sweep-lint floors follow the added fixture (7 / 7 / 6).
- Tests: changed block beside a literal (a precedence pin for the other polarity — HEAD already behaved that way), legacy non-string literal takes an unrelated update, `create()` refuses an object / `null` by shape without echoing the value and still skips `''`.
- Verified again on the follow-up: full unit suite green, `secretsmanager-update-value-source` re-run against real AWS (PASS, 199 s, 0 orphans, 0 surviving state-bucket versions), Codex loop to zero findings (round 1: the precedence test's comment overstated it as a regression fence, reworded; round 2: `NO ISSUES`) and the repo's code-reviewer agent on the delta (clean; its three nits — the create-side falsy gate, a JSDoc paragraph split, a dash — are folded in).

## Accepted edges (documented in code)

- **Value-scan collision**: the redaction `asPersisted()` uses is a VALUE scan, so a literal in the block that equals a plaintext this resource resolved elsewhere is rewritten to that expression too; replacing a `{{resolve:...}}` with the literal it currently resolves to therefore compares equal and does not regenerate, where CloudFormation would. Kept because the state writer applies the same rewrite to that literal, so the two sides stay consistent, and the miss is in the safe direction (no unrequested re-roll). JSDoc on `changedSecretValue` + a pinning unit test.
- **verify.sh cleanup gate is a bash-only invariant**: `cleanup` removes state.json only when `state destroy` succeeded (otherwise the `noncurrent` purge would reap the just-removed current version). The `integ-s3-versions-harness` lint inspects only `all`-mode purges, so this gate is not mechanically fenced; extending that lint to every `noncurrent` caller is a repo-wide change left out of this PR.

## Not in this PR

- `generateSecretString()` itself samples with `bytes[i] % chars.length` (modulo bias) and ignores `RequireEachIncludedType` / `IncludeSpace`. Delegating to `GetRandomPassword` adds an IAM permission and an async call and still needs the local `SecretStringTemplate` merge, so it is a separate change (noted on the issue).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01GKcFssWnNJqi8znDi6taqi

**Maintainer-side note**: minor follow-ups from the merge-time delta review (a stale exemption-comment enumeration, one recorded won't-do) are tracked in (#2479).


go-to-k:
Thanks for this fix — the review came back clean on the substance. A 4-axis review (spec vs #2472, code, test adequacy, security flow-trace) found **no blockers**: the CloudFormation-parity semantics all check out on deploy / rollback-replay / `drift --revert`, the keep-assertions genuinely discriminate against the old code, and the security trace found no path where a value or the redaction pairing reaches a log, error message, or state. Six minor items, two of which are worth addressing in this PR:

**Worth fixing here:**

1. **`src/provisioning/providers/secretsmanager-secret-provider.ts:411-412` — the JSDoc understates the collapsed-pair residual.** The comment says a collapsed pair (two `{{resolve:...}}` expressions resolving to the same plaintext) "can still read as changed *once*". It actually reads as changed on **every** unrelated update: `asPersisted()` is a value scan with no position source, so it rewrites both leaves to the survivor expression, while the persisted previous holds each leaf's own expression (the positioned persist), and the post-update record is positioned again — the mismatch recurs, so that shape re-rolls its password on every Tags-only deploy. Rare shape, but the residual is permanent, not one-shot; please fix the comment (or thread a position source through).

2. **`secretsmanager-secret-provider.ts:455-459` — the non-string refusal fires before the unchanged check.** A legacy-deployed non-string `SecretString` (which `create()` accepted via the cast at :104 and state persisted) now makes every in-place update throw — including a Tags-only deploy that changes nothing about the value and previously succeeded. Comparing first and refusing only when the value would actually be sent keeps the refusal without breaking updates on legacy records.

**Nits (fine as follow-ups or one-liners here):**

3. `create()` (:99-105) still forwards a non-string `SecretString` via a bare cast while `update()` now refuses it — the same shape refusal in `create()` would close the asymmetry (and the class of records item 2 trips on).
4. A previous bag carrying BOTH `GenerateSecretString` and an ignored `SecretString: 'x'` masks a block→literal switch when the literal is still `'x'` (compares equal, nothing sent). Invalid shape in CloudFormation and cdkd-created only — a one-line note next to the precedence comment would do.
5. `tests/unit/scripts/integ-secret-fixture-sweep.test.ts:139/151/436` — the floors documented as "today's exact count — raise when a fixture is added" stay at 6/6/5 while the new fixture makes the true counts 7/7/6; the starvation fence is now one fixture looser than designed.
6. The precedence rule is only pinned in the unchanged-block direction — a case with a *changed* `GenerateSecretString` block alongside a literal on the same bag would cover the other polarity.

nix-tkobayashi:
Thanks for the thorough review — all six items are addressed in the follow-up commit.

**Fixed here**

1. JSDoc on `changedSecretValue()` now states the collapsed-pair residual is permanent (re-rolls on every in-place update), explains why (value scan vs. the state writer's positioned persist), and why threading the previous block as a position source would erase a genuine reference switch, so it stays open for that rare shape.
2. The literal branch decides UNCHANGED (`isDeepStrictEqual` over a JSON round-trip) before judging the shape; the non-string refusal fires only for a value that would actually be sent, so a legacy record still takes a Tags-only update. Pinned by a test with a legacy non-string previous.

**Nits, all taken**

3. `create()` routes `SecretString` through the same shared `requireSecretStringShape()` (skipping only `''`, which is the documented no-version create), so `null` / `false` / `0` / an object are refused there too — pinned by two tests (object and `null` refused without echoing the value; `''` still skipped).
4. Precedence consequence (a previous bag carrying both a block and an ignored literal masking a block-to-literal switch) is recorded next to the precedence rule.
5. Sweep-lint floors raised to 7 / 7 / 6 (counted independently by the review agent against the fixture tree).
6. Added the changed-block-beside-a-literal case; note it is a precedence pin (HEAD already took the block branch first), not a regression test.

Verification on the follow-up: full unit suite green (18,135 tests), `secretsmanager-update-value-source` re-run against real AWS — PASS, destroy 0 errors, 0 orphans, 0 surviving state-bucket versions — and the Codex review loop plus the repo's code-reviewer agent on the delta, both clean.

go-to-k:
Merged — thank you for the careful fix and for turning every review finding around so thoroughly (the follow-up killed all six items, and the maintainer-side re-run of the fixture against real AWS came back clean: destroy 0 errors, 0 orphans, 0 surviving state-bucket versions). The two remaining note-level items are tracked in (#2479). Much appreciated!

Issue #2485, opened 2026-09-03
fix(secret-redaction): a literal leaf EMBEDDING a {{resolve:...}} token persists a sibling's expression when two references share one plaintext (perpetual UPDATE)


nix-tkobayashi:
## Summary

A template leaf that EMBEDS a `{{resolve:...}}` token in a longer literal string (`postgres://app-svc:{{resolve:ssm-secure:NAME}}@db.internal:5432/app`) is persisted with a SIBLING's expression when another leaf of the same resource resolves to the same plaintext through a different spelling (`{{resolve:ssm-secure:NAME:1}}`, or a `secretsmanager` reference at a different version stage). For the measured value (length >= 4) no plaintext leaked -- the wrong EXPRESSION was written, and it stays wrong. For a 1-3 character plaintext the same path can persist the PLAINTEXT, because substring redaction deliberately ignores needles shorter than `MIN_NEEDLE_LENGTH` (4) and `learnMixedLeafNeedle` refuses them too.

Measured on real AWS on 2026-09-04 with the `ssm-secure-dynamic-ref` fixture from go-to-k/cdkd#2482, three env vars on one Lambda:

```text
template  SSM_SECURE_WHOLE      {{resolve:ssm-secure:cdkd-test-ssm-secure-<acct>}}
template  SSM_SECURE_EMBEDDED   postgres://app-svc:{{resolve:ssm-secure:cdkd-test-ssm-secure-<acct>}}@db.internal:5432/app
template  SSM_SECURE_VERSIONED  {{resolve:ssm-secure:cdkd-test-ssm-secure-<acct>:1}}
state     SSM_SECURE_EMBEDDED   postgres://app-svc:{{resolve:ssm-secure:cdkd-test-ssm-secure-<acct>:1}}@db.internal:5432/app
```

The two whole-value leaves persisted their own expressions; the embedded one persisted the versioned sibling's. The fixture now seeds a second parameter for the versioned form to stay out of this; the PR notes where.

## Mechanism (current `main`, 8cbcffdc)

- `RecordedSecretValues` is `Map<plaintext, expression>`, so a resource whose references resolve to one plaintext through two spellings keeps ONE entry -- whichever the resolver recorded LAST (`Map.set` overwrites; here the CDK env-var key order made `:1` the last).
[...]

Issue #2320, opened 2026-08-27
fix(secret-redaction): a child resource mixing an EMBEDDED and a WHOLE-VALUE leaf over two Parameters sharing one plaintext reports a perpetual UPDATE


go-to-k:
## The defect

A nested-stack CHILD resource that holds BOTH an EMBEDDED leaf and a WHOLE-VALUE
leaf, fed by two `Parameters` that resolve to ONE plaintext, has its persist side
and its diff side disagree forever: `cdkd deploy` reports an UPDATE on every run,
and a REPLACEMENT on every run for a create-only property.

This is a REGRESSION introduced by the go-to-k/cdkd#2291 PR, not a pre-existing
gap that PR failed to close. It is filed rather than fixed there because closing
it needs a new positioning concept (below), and taking that at round 3 of a
review cascade is how round 4 happens.

## Measured, with the `main` baseline that makes it a regression

One child resource, parameters `A` (`...:handoff::`) and `B`
(`...:handoff:AWSCURRENT:`) resolving to one plaintext. Driven through the REAL
`IntrinsicFunctionResolver` plus the real persist and diff paths, on
`main` (f56c2cf9) and on the go-to-k/cdkd#2291 branch:

| shape                                | main    | go-to-k/cdkd#2291 branch |
| ------------------------------------ | ------- | ------------------------ |
| `{Ref: A}` and `{Ref: B}`            | agree   | agree                    |
| `Fn::Sub 'x${A}'` only               | agree   | agree                    |
| `Fn::Sub 'x${A}'` **and** `{Ref: B}` | agree   | **DISAGREE**             |

```
persisted p1: x{{resolve:secretsmanager:prod/db/cred:SecretString:handoff:AWSCURRENT:}}
desired   p1: x{{resolve:secretsmanager:prod/db/cred:SecretString:handoff::}}
```

`main` agrees because BOTH halves take the collapsed survivor -- on the WRONG
expression, which is go-to-k/cdkd#2291 itself, but they match, so nothing
perpetually updates.

## Mechanism

The inherited bag is `Map<plaintext, expression>`, so it holds ONE entry per
plaintext -- whichever `Ref` resolved LAST in that resource. The DIFF side
(`DeployEngine.redactParametersForDiff`) answers PER PARAMETER, which
go-to-k/cdkd#2291 requires for the whole-value leaves. A WHOLE-VALUE leaf is
positioned by the parent's per-parameter association and never reads the bag, so
[...]

Issue #2298, opened 2026-08-26
fix(secret-redaction): a cross-stack secret embedded in surrounding text persists a siblings expression, because the key arm is whole-token only


go-to-k:
## The defect

A cross-stack secret reference embedded in a leaf that is NOT a whole token -- `pre-${Child.Outputs.Cur}-post` via `Fn::Sub`, or the equivalent through `Fn::Join` -- is persisted with the WRONG expression when a sibling leaf resolves to the same plaintext. Measured: the leaf resolving `...:AWSCURRENT:` persisted `pre-{{resolve:...:AWSPREVIOUS:}}-post`.

Consequence: `resolveReplayProps` re-resolves the persisted expression, so `cdkd drift --revert` / rollback would apply the WRONG secret version to the live resource. No plaintext leaks -- the value scan still redacts -- so this is a wrong-value write, not a disclosure.

## Why the go-to-k/cdkd#2270 fix does not reach it

go-to-k/cdkd#2270 gave `crossStackSourceKey` an `Fn::Sub` arm so a leaf that is EXACTLY one bare placeholder derives the same key the writer recorded, which is what lets `positionByCrossStackSource` put each leaf back on its own expression. That arm is deliberately whole-token only, and it has to be: `recordCrossStackExpression` records a whole-token expression, so a source-derived key for a partial leaf would certify something the writer never keyed -- and over-redaction on this path is worse than under-redaction, because `drift --revert` PUSHES the baseline to AWS.

So closing this needs the WRITER to record a span (or the reader to gain a span-aware key), not another reader arm.

## Scope: not only `Fn::Sub`

`Fn::Join` carries the identical residual on `main` today, and it predates go-to-k/cdkd#2270 -- that path never went through the whole-token arm at all. Whoever takes this should fix the mechanism rather than the `Fn::Sub` spelling, or the `Fn::Join` half will be a follow-up to the follow-up.

## Relationship to the neighbours

- go-to-k/cdkd#2102 asks for masking by SPAN in the drift preserve / revert paths, where there is no map entry to match. That is the same missing primitive seen from the drift side; whoever builds span-keying should read both.
- go-to-k/cdkd#2291 is the nested-stack `Parameters` handoff, a different cause (a plaintext-keyed inherited bag) with the same symptom.
- go-to-k/cdkd#2270 closed the whole-token `Fn::Sub` case and its 2-arg spelling.

## Reachability

[...]

Issue #1914, opened 2026-08-14
fix(drift): cdkd drift never redacts — it prints secret plaintext, reports permanent phantom drift, and --revert ships the literal {{resolve:...}} to AWS


go-to-k:
## Context

Found by the security review of the (#1910) PR, which traced every sensitive value from WRITE to every READER. `src/cli/commands/drift.ts` imports nothing from `src/deployment/secret-redaction.ts`. It is the one write-to-state surface the GHSA-p5qg-v9gv-hc7w work never reached, and it is worse than a write leak because it also FEEDS A LIVE AWS CALL.

Recorded on (#1910) as residue; filing separately so it survives that issue closing.

## Three distinct defects, one root cause

`runDriftForStack` builds its baseline as `observedProperties ?? properties`. Since PR (#1899) both hold `{{resolve:...}}` expressions, while the AWS-current snapshot holds the resolved plaintext. Nothing reconciles the two.

1. **`--revert` ships the literal expression to the live resource.** The expression-bearing baseline becomes `desiredProperties` and is passed to `provider.update`, so the env var / property is set to the literal string `{{resolve:secretsmanager:...}}`. This is exactly the GHSA rollback blocker in a second path -- `resolveReplayProps` re-resolves before the provider call, and drift has no counterpart.
2. **Every `cdkd drift` on a dynamic-ref stack reports permanent phantom drift**, because an expression never equals plaintext -- and it PRINTS the plaintext in the drift report as the AWS-current side.
3. **`--accept` persists the plaintext into `state.json`**, re-introducing the original advisory disclosure.

## Reproduction

Deploy `tests/integration/secrets-dynamic-ref`, then:

```
cdkd drift <stack>            # prints cdkd-known-pw-123 as the AWS-current value; exits 1
cdkd drift <stack> --accept   # writes the plaintext into state.json
cdkd drift <stack> --revert   # sets the Lambda env var to the literal {{resolve:...}} string
```

## Direction

The three need different remedies and the ordering matters:

- **`--revert`** must RE-RESOLVE the baseline before handing it to the provider, the way `resolveReplayProps` does for the rollback replay. This is the data-loss / live-breakage half and should land first.
- **The comparison** must compare like-for-like. The cheapest correct rule is to resolve the baseline for comparison ONLY (never persisting the result), which also stops the plaintext reaching the report.
[...]

PR #2492, opened 2026-09-03
fix(resolver): resolve {{resolve:ssm-secure:...}} instead of sending the literal token to AWS

### LINKED ISSUE #2482: fix(resolver): a {{resolve:ssm-secure:...}} reference is sent to AWS as the literal template string with only a warning, so the live credential can BE the template text
## Summary

A `{{resolve:ssm-secure:<name>}}` dynamic reference is **sent to AWS as the literal template string**. `resolveDynamicReferences` has arms for `secretsmanager` and `ssm` only; every other service hits an `else` that logs `Unsupported dynamic reference service: ssm-secure` at WARN level and `continue`s, leaving the token in the property bag. The provider then hands that text to the service API. Where the API accepts it — demonstrably a suitably short IAM `LoginProfile.Password`, and any free-form string field — the resource is created or updated with the token text as its live value and the deploy exits 0. Other APIs reject it on their own validation (RDS refuses `/` in a master password, so a hierarchical parameter name fails loudly there), which is the better outcome but still not the resolved value CloudFormation would have written.

cdkd never goes through CloudFormation, so nothing resolves the token server-side; the comment in `src/deployment/secret-region-classification.ts:24-37` ("the literal token is passed through to AWS and CloudFormation resolves it SERVER-side") describes CloudFormation's own deploy path, not cdkd's. The repo already knows the value is broken — `tests/integration/secrets-dynamic-ref/verify.sh:50-54` says `ssm-secure` "hits the `else` branch (warn + leave literal), so it would deploy a broken value" and skips it — but nothing treats that as a defect: the deploy succeeds, `docs/cli-drift.md` frames the survivor only as a `cdkd drift` limitation, and no issue tracks the deploy path.

## Affected code (current `main`, 8cbcffdc)

- `src/deployment/intrinsic-function-resolver.ts:6995-7040` — `resolveDynamicReferences`: the `secretsmanager` arm, the `ssm` arm (`resolveSSMReference(parts, decrypt)`, which classifies a `SecureString` off the same `GetParameter` response), then:

```ts
} else {
  continue;
}
```

- `src/deployment/secret-region-classification.ts:24-37, 99-101` — `REPLAY_SECRET_SERVICES` is `{secretsmanager, ssm}` on the false premise above, and `ssmParameterName()` is `inner.substring('ssm:'.length)`, which would turn `ssm-secure:/pw` into `secure:/pw`.
- No pre-flight refuses the spelling: `ProviderRegistry.validateResourceProperties` validates provider property coverage, never `{{resolve:` services.
- Everything downstream is built around the token SURVIVING unresolved: `src/cli/commands/drift.ts` (`isSecretBySpelling`, the `unresolvedToken` `notCompared` cause that is deliberately excluded from the exit code, the span-masking arms around the readback), `src/cli/commands/scrub.ts` (`SECRET_REFERENCE_OPENINGS`, the "an `ssm-secure` reference cdkd resolves for nobody" arm), `src/deployment/secret-redaction.ts` (`SECRET_BEARING_REFERENCE_PREFIXES` / `SPELLED_SECRET_REFERENCE_PREFIXES` already treat the spelling as secret-bearing and spelled-certain), `src/analyzer/outputs-diff.ts`, `src/deployment/outputs-export-alias.ts`, and `docs/cli-drift.md:178-253, 325-360`.

## Reachable paths

1. **`cdkd deploy`** — a resource property carrying the token, whole or embedded (`"jdbc:...password={{resolve:ssm-secure:/pw}}"`), reaches the SDK or Cloud Control provider with the token in place (`deploy-engine.ts:3970-4044`). Only a WARN line distinguishes it from a healthy deploy.
2. **Rollback replay** — the journal's bag re-enters `resolveDynamicReferences`, the token survives again, and the recovery path re-sends it.
3. **`cdkd import --migrate-from-cloudformation`** — the imported record holds the token (`import.ts:1570-1588`) while AWS holds the value CloudFormation resolved. The NEXT in-place update of that resource sends the literal and **overwrites the good value**; `docs/cli-drift.md:355-360` documents the embedded case doing exactly that "as `cdkd deploy` does".

`cdkd drift` cannot surface it: the token is classified `unresolvedToken` / "resolves for nobody", excluded from the exit code by design, and the CloudFormation-supported destinations are write-only, so nothing reads them back.

## Impact

For a CloudFormation-supported `ssm-secure` destination — `AWS::IAM::User.LoginProfile.Password`, `AWS::RDS::DBInstance.MasterUserPassword` and the rest of the documented allowlist — a template that is correct for CloudFormation deploys under cdkd with the literal handed to the underlying API; where that API accepts it, the credential IS the template text. That text sits in `cdk.out`, in git, in the synthesized template every reviewer sees; anyone with it can authenticate. Because the deploy exits 0 and the only signal is a WARN line, it is easy to ship. Severity is higher than the value-rotation class of go-to-k/cdkd#2472: this is a wrong credential, silently, on the main path — conditional on the destination API accepting the string, which the most common destination (an IAM console password) does.

## Not a duplicate

Searched open and closed issues (titles and bodies) for `ssm-secure` and `Unsupported dynamic reference` immediately before filing. go-to-k/cdkd#2102 (drift `--revert` masking by span for a readback that carries the token) and go-to-k/cdkd#2166 (`cdkd scrub` missing an `ssm-secure` token inside `Fn::Sub`) both take the surviving token as a given and handle its SURVIVAL in drift and scrub; neither names the deploy writing it as the credential. The GHSA-p5qg-v9gv-hc7w work (go-to-k/cdkd#1900 / go-to-k/cdkd#1901 / go-to-k/cdkd#1914) covers `secretsmanager` and `ssm` SecureString; `ssm-secure` was left as "cdkd resolves it for nobody".

## Proposed fix

Resolve `ssm-secure` through the existing SecureString path, with the verdict decided by SPELLING:

- In `resolveDynamicReferences`, classify `service === 'ssm-secure'` as `isKnownSecret` (today that predicate is `service === 'secretsmanager' || recordedSecretExpressions.has(fullMatch)`, so a first-seen `ssm-secure` reference would otherwise still perform a `GetParameter` on the comparison path), and add an `ssm-secure` arm beside `ssm`: `resolveSSMReference(parts, decrypt)` (the `name:version` selector already reaches `GetParameter.Name` as-is, which SSM accepts), `isSecret = true` unconditionally, `pinSecretVerdict(fullMatch, true)`. If `GetParameter` reports a definitive `String` / `StringList` type, refuse (`IntrinsicResolutionRefusalError`, `markNonRetryable`) rather than resolve a public value under a secret spelling — CloudFormation's `ssm-secure` is defined only for SecureString parameters.
- In `secret-region-classification.ts`, add `ssm-secure` to `REPLAY_SECRET_SERVICES`, generalize `ssmParameterName()` to strip the actual service prefix (`ssm:` or `ssm-secure:`), and rewrite the comment; cover the CloudFormation-supported name and numeric-version spellings in its tests, and label the ARN form — which the `ssm` classifier already accepts — as a cdkd-only extension if it is carried over.
- The recorded `plaintext -> {{resolve:ssm-secure:...}}` pair then flows through the existing GHSA machinery unchanged: state persists the expression, logs / events / errors are masked, rollback replays re-resolve it, `secret-redaction.ts` already lists the spelling as secret-bearing and spelled-certain.
- Sweep the sites that assumed `ssm-secure` NECESSARILY survives, while keeping the generic machinery: the `unresolvedToken` cause stays (any other unsupported service still survives), and `isSecretBySpelling('{{resolve:ssm-secure:')` stays (it masks legacy records and resolution-failure paths offline). What changes is the `ssm-secure`-specific wording and expectations — `drift.ts`'s "permanent, unclearable / resolves for nobody" arms and the `docs/cli-drift.md` text that describes them, `scrub.ts`'s "resolves for nobody" comment, the `secrets-dynamic-ref` fixture's explicit skip, `docs/architecture.md:367`, `docs/testing.md:1777-1779`, the three `.claude/rules/layout-*.md` mentions, and the unit tests that pin the survivor behaviour for this spelling (`drift-secret-redaction`, `drift-cross-region-secret`, `secret-redaction-derived-needles`, `scrub-import-value-secret`, `outputs-diff`, ...).

**Residual the fix does not repair by itself:** a stack cdkd deployed BEFORE the fix holds the literal on AWS and the same literal in `state.json`. After the fix the persisted record and the template both spell the expression, so the deploy diff (`deploy-engine.ts:2068-2099`, `skipDynamicReferences`) reads `NO_CHANGE` and the bad live value survives until that property is next updated. `cdkd drift` becomes the detector for types whose readback carries the property (drift resolves the state expression in memory and compares it with AWS, so the literal no longer matches), and `drift --revert` the remedy; write-only destinations (`MasterUserPassword`, `LoginProfile.Password`) need a deliberate touch. Worth a changelog note and a line in the deploy summary when an `ssm-secure` reference resolves for the first time.

The minimal interim alternative — refuse the spelling at resolution with a `markNonRetryable` error instead of warning — closes the disclosure in a few lines, but it blocks every create, changed update and rollback replay of a resource carrying the spelling (including a CloudFormation-imported stack's next real change) until the reference is re-spelled, while an unchanged imported resource stays `NO_CHANGE` and untouched. Only worth taking if the full path cannot land in one PR.

## Verification plan

- Unit: the new arm (resolves, records the pair, pins the verdict, refuses a non-SecureString, skips with no lookup on the comparison path), the region classification and name parser for `ssm-secure` (name and numeric-version forms; the ARN form only as a labelled cdkd-only extension), and the flipped `ssm-secure` expectations in the drift / scrub / redaction suites.
- Integ: a new fixture `ssm-secure-dynamic-ref` — `verify.sh` seeds a SecureString parameter with a known marker value; the stack consumes `{{resolve:ssm-secure:<name>}}` whole and embedded in a **Lambda environment variable**, chosen because it reads back (`get-function-configuration`). This is explicitly a cdkd-only resolver check, not CloudFormation parity: Lambda `Environment` is not on CloudFormation's `ssm-secure` destination allowlist, and cdkd does not enforce that list for any service. Assert the live value equals the parameter's plaintext and NOT the token, assert `state.json` holds the expression and NOT the plaintext, and on a Tags-only re-deploy assert that whatever the provider re-sends carries the plaintext and never the token (the Lambda provider compares the resolved desired `Environment` against the redacted previous, so a re-send is expected and is not the assertion). Sweep the state bucket's object versions (secret-seeding fixture rules). Today Phase 1 fails: the live value is the token.
- Broad integ `lambda` for the cross-cutting resolver change (`integ-broad` gate), plus `secrets-dynamic-ref` (which drops its explicit skip) and `drift-revert` for the drift arms.

Dup-check: searched open and closed issues (titles and bodies) for ssm-secure / "Unsupported dynamic reference" right before filing -- none names the deploy writing the literal as the credential; go-to-k/cdkd#2102 and go-to-k/cdkd#2166 handle the token's survival in drift and scrub respectively
Session-fit: now (do it in this session) -- the fix PR follows in this session; the verifier is `/run-integ ssm-secure-dynamic-ref` (new fixture) plus `/run-integ lambda`
Severity: high -- for an API that accepts the string (the IAM console password does) a CloudFormation-correct template deploys with its credential set to the public template text, and the deploy reports success
Effort: large (L) -- a cross-cutting resolver change, a sweep of the drift / scrub / redaction sites and their suites, a new integ fixture, a broad integ, and docs
Estimate: ~4-6 h -- the drift-side sweep (every `ssm-secure`-specific arm and its unit suites) and the new fixture are what eat the time; the resolver arm itself is small

nix-tkobayashi:
Closes #2482

## Summary

`{{resolve:ssm-secure:<name>}}` was sent to AWS as the literal template text: `resolveDynamicReferences` had arms for `secretsmanager` and `ssm` only, so the spelling fell into the unsupported-service `else` (WARN + leave the token), and where the service API accepted the string the live credential WAS the public template text, with the deploy exiting 0. The spelling is now resolved through the SecureString path, secret by spelling, and the drift side drops the arms that existed only because it used to survive.

## What changed

- **Resolver** (`src/deployment/intrinsic-function-resolver.ts`): `isKnownSecret` treats `ssm-secure` as secret by spelling, so the comparison path (`skipDynamicReferences`) leaves it unresolved with NO `GetParameter`, exactly like `secretsmanager`. A new arm resolves it via `resolveSSMReference(parts, true)` (the `<name>:<version>` selector reaches `GetParameter.Name` as-is), pins the verdict `true` and records the `plaintext -> expression` pair, so state persists the expression, logs / events / errors are masked and a rollback replay re-resolves it through the existing GHSA machinery. A parameter `GetParameter` reports as a definitive `String` / `StringList` is REFUSED (`IntrinsicResolutionRefusalError`, `markNonRetryable`, naming the type and never the value); an unclassifiable `Type` resolves as a secret.
- **Region classification** (`src/deployment/secret-region-classification.ts`): `ssm-secure` joins `REPLAY_SECRET_SERVICES`, so the name form is `ambiguous` under a foreign producer region and `local` without one, and the ARN form is `named-region`; `ssmParameterName` strips the actual service prefix (the fixed `'ssm:'.length` would have turned `ssm-secure:/pw` into `secure:/pw`).
- **Drift** (`src/cli/commands/drift.ts`): a `{{resolve:...}}` token that survives the pass is no longer any CloudFormation service, so `isSecretBySpelling` and the two arms it fed (marking a survivor's path secret-bearing; `preserveLiveValuesAtUnresolvedTokens` registering the live value it copied, whose `secrets` parameter is removed) go with the population. The `unresolvedToken` cause, its exit-code exclusion and whole-token live-value preservation stay for look-alike spellings — and their live values are now treated as ORDINARY data, so a surviving look-alike (or future-service) token's live value can appear unmasked in the human and `--json` drift reports, which is the documented intent (`docs/cli-drift.md`): masking text that merely looks like a reference left a path refused by `--accept` and pinned by `--revert` with no remedy. Comments in `scrub.ts` / `secret-redaction.ts` that described the old survivor are updated.
- **Tests**: 10 resolver / redaction cases in `dynamic-references.test.ts`, 4 classifier cases, and the drift / scrub suites — survivor cases now use a look-alike spelling (`{{resolve:notaservice:...}}`), and the `ssm-secure` cases flip to the resolved path (a CFn-migrated record compares CLEAN, a rotated value is masked by the recorded pair, `--revert` writes the decrypted value and persists the expression, the region-less form is refused under a foreign producer region and resolved locally without one, `cdkd scrub` records the needle through a scrubbed producer).
- **New integ fixture** `tests/integration/ssm-secure-dynamic-ref/`: seeds two SecureString parameters out of band (CloudFormation cannot create one), consumes the whole / embedded / `:1`-versioned forms in a Lambda environment variable (chosen because it READS BACK — a cdkd-only resolver check, not CloudFormation destination-allowlist parity), captures the deploy output and scans it for the markers before showing it, asserts the live values are the plaintext and never the token, that `state.json` holds the expressions and never a plaintext, that `diff --fail` / `scrub --dry-run` / `drift --json` are clean, that a Tags-only re-deploy keeps all of that, then destroys, sweeps the log group, deletes the parameters, purges the state bucket's object versions and asserts zero survive.
- **Docs**: `docs/cli-drift.md`, `docs/testing.md`, `docs/changelog-cdkd.md`, `.claude/rules/layout-drift.md`, the `dynamic-reference-resolution` scenario description, and `tests/integration/secrets-dynamic-ref` (its explicit skip now points at the new fixture); matrices regenerated.

## Residual (documented, not repaired by this PR)

A stack cdkd deployed BEFORE this change holds the literal token on AWS and the same token in state, so the deploy diff reads `NO_CHANGE` and the upgrade does not rewrite the live value. For a property the provider reads back, `cdkd drift` now reports it and `--revert` writes the resolved value; a write-only destination (`MasterUserPassword`, `LoginProfile.Password`) needs that property to be updated once. Stated in `docs/cli-drift.md` and the changelog.

## Found by the fixture, filed separately

The first real-AWS run had the versioned form on the SAME parameter as the other two: three references to one plaintext in one resource, and the EMBEDDED leaf persisted the versioned sibling's expression (`...{{resolve:ssm-secure:NAME:1}}@...` for a template that spells `NAME`), which the next deploy diffs forever. That is the value scan's pre-existing one-entry-per-plaintext behaviour for embedded leaves (the residual go-to-k/cdkd#1904 accepted; the same primitive go-to-k/cdkd#2298 / go-to-k/cdkd#2320 ask for), not the new arm — filed as go-to-k/cdkd#2485. The fixture seeds a second parameter for the versioned form and says why.

## Verification

- `vp check` / `vp run typecheck:test` / `vp run build` / full `vp test run`: all green — 879 files, 18,170 tests passed (1 skipped), no type errors, root-asserted in the PR worktree after the rebase onto 5d8d93b1.
- `/run-integ ssm-secure-dynamic-ref`: PASS (208 s, re-run on the rebased build; 162 s on the first green run) — 2 deleted, 0 errors, 0 orphans, 0 surviving S3 object versions, both parameters removed.
- `/run-integ lambda` (broad, `integ-broad` gate — `intrinsic-function-resolver.ts` is cross-cutting): PASS (99 s, re-run on the rebased build; 107 s before it) — 9 deleted, 0 errors, 0 orphans. (An account-wide IAM-to-Lambda propagation outage from about 17:20Z to 17:53Z on 2026-09-03 failed this and the two fixtures below on their first attempts with `The role defined for the function cannot be assumed by Lambda` after the full 47.75 s propagation budget; a hand-made role could not be used by `create-function` for 120 s either, so the runs were repeated once the probe recovered. Every failed run's rollback completed and its state prefix was swept.)
- `/run-integ secrets-dynamic-ref` (the sibling fixture whose skip note changed): PASS (487 s) — 4 deleted, 0 errors, 0 orphans, 0 surviving state versions.
- `/run-integ drift-revert` (the drift arms changed): PASS (225 s) — 13 deleted, 0 errors, 0 orphans.
- Codex review loop on the diff to NO ISSUES (2 rounds: the deploy-log `tee` ordering in verify.sh and a stale "masked by position" sentence in `docs/cli-drift.md` were the round-1 findings), plus the repo's 3-axis + security reviewers: spec (clean), code, test (mutation probes RUN against scratch copies), security (no blocker) — their findings were applied and the fix delta re-reviewed by code + test reviewers and by Codex (5 rounds in total, ending NO ISSUES). Dropped from this PR as pre-existing: an embedded 1-3 character secret is below the redaction needle floor (same for `secretsmanager` today, accepted in go-to-k/cdkd#1914).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01GKcFssWnNJqi8znDi6taqi


go-to-k:
Merged — thank you @nix-tkobayashi for a thorough fix and an exemplary PR write-up.

Pre-merge verification on our side: 3-axis + security review (spec / code / test / security, all clean — the security pass traced the newly-resolved plaintext from `resolveSSMReference` through state, events, logs, errors, diff, both drift reports, rollback replay, export and scrub), full local suite (879 files / 18,170 tests), and fresh real-AWS re-runs of `ssm-secure-dynamic-ref` (PASS, 0 orphans, 0 surviving state versions) and `lambda` (broad set, PASS, 0 orphans) on the rebased head.

The only findings were nit-level (three low test-coverage candidates, two cosmetic items, one fixture scan-symmetry note) — bundled as follow-up in (#2501), none blocking.

Issue #2491, opened 2026-09-03
test(scripts): aws-client-defaults-fence AST-walk case times out at 5000ms on slow CI runners (once-leak-detect flake)


go-to-k:
## What happened

On the dependabot-config PR (#2486) — a `.github/dependabot.yml`-only diff, which cannot affect test behavior — the `once-leak-detect` CI job failed once and passed on rerun of the same commit:

- Failed run: https://github.com/go-to-k/cdkd/actions/runs/33788268409/job/100758262353
- Passing rerun (same sha `1cdd824a`): https://github.com/go-to-k/cdkd/actions/runs/33788268409/job/100762212428

The failure:

```
Error: Test timed out in 5000ms.
 > tests/unit/scripts/aws-client-defaults-fence.test.ts:488:5
   it('finds no construction the AST walk did not, except in prose', ...)
Test Files  1 failed | 876 passed | 2 skipped (879)
```

## Why it flakes

The test at `tests/unit/scripts/aws-client-defaults-fence.test.ts:488` does a repo-wide AST walk under the default 5000ms `testTimeout`. The failing run's summary shows the runner was under heavy load (`Duration 562.33s`, `import 520.46s`, `tests 932.24s` for the whole suite) — on a slow-enough runner the walk crosses 5s. The rerun on the same commit passed (9m53s job time), and main's three most recent CI runs were green, so this is load-dependent, not a regression.

## Suggested fix

Give that `it()` (and any sibling AST-walk cases in the same file near the budget) an explicit timeout (e.g. 30_000) — the test is legitimately long-running, not hung.

Session-fit: next (not this session) — independent of the session's diff (dependabot.yml only); needs its own PR + check cycle
Severity: low — a CI job reddens occasionally on slow runners; rerun is a full workaround, no shipped behavior affected
Effort: small (S) — a one-line per-test timeout, riding the ordinary unit-test cycle
Estimate: ~30 min — confirming which sibling cases in the file also sit near the 5s budget is the only real work
Dup-check: searched open issues for "flake", "once-leak", "timeout", "aws-client-defaults-fence" — no existing issue names this test (the closest, (#2444), is a different test's wall-clock race)

nix-tkobayashi:
[...]

Issue #1933, opened 2026-08-18
fix(deployment): dynamic-reference cache is keyed without region, so a cross-region run resolves one region's secret into another


go-to-k:
## Context

Found by the security review of the #1914 lane (GHSA-p5qg-v9gv-hc7w follow-up
set), and verified independently against `main` at `c4fd97bc`. Pre-existing, not
a regression -- but `cdkd drift --revert` makes it a LIVE-WRITE consumer for the
first time, which is why it is worth its own issue rather than a comment.

## The gap

Dynamic-reference resolution caches by the EXPRESSION STRING ALONE, with no
region component:

```ts
// src/deployment/intrinsic-function-resolver.ts:843
const cachedDynamicReferences: Record<string, string> = {};
// :4315
if (fullMatch in cachedDynamicReferences) { ... }
// :4383
cachedDynamicReferences[fullMatch] = resolved;
```

while the lookups themselves go through the AMBIENT clients:

```ts
// :4489  (secretsmanager)   and  :4684  (ssm)
const awsClients = getAwsClients();
const client = awsClients.secretsManager;   // / .ssm
```

`getAwsClients()` is process-ambient, and the cache is module-global for the
process. So in any run that spans regions, the FIRST region to resolve
`{{resolve:secretsmanager:mysecret:...}}` wins, and every later stack in every
other region silently reuses that value.

Secrets Manager secrets and SSM parameters are regional and independent: the
same NAME in `us-east-1` and `ap-northeast-1` is two different values, and
routinely holds two different credentials. So this substitutes one region's
credential into another region's resource.

## Why it matters more now

Before, the affected consumers were `cdkd deploy` (writes the wrong secret to
the created/updated resource) and the resolver's diff-time uses. `cdkd drift
--revert` adds a THIRD consumer that writes to a live resource, and it is
reached without an explicit deploy — a `cdkd drift --all --revert` across
regions can push one region's secret into another's resource while reporting
success.

`src/cli/commands/drift.ts` already carries a comment acknowledging the
module-global cache and the ambient-region lookup as a hazard, but there was no
issue number to point at.

## Direction

The cache key has to carry whatever the lookup actually depends on -- at
minimum the region, and arguably the resolved account/credential identity too,
since a cross-account run has the same shape. Two candidates:

1. **Key the cache by `${region}|${fullMatch}`** (or by the resolved client
   identity), which is the smallest change and keeps one process-global map.
2. **Move the cache off module scope** onto the resolver instance, so its
[...]

PR #2511, opened 2026-09-04
fix(secret-redaction): position a literal leaf embedding one token by its own span

nix-tkobayashi:
Refs go-to-k/cdkd#2485 — fixes the 4+ character half (the wrong EXPRESSION persisted for an embedded leaf); the 1-3 character plaintext residual that issue also names is carried by go-to-k/cdkd#2516, filed from the review below, so the issue is not closed over a disclosure it names.

## Summary

`RecordedSecretValues` is keyed by PLAINTEXT, so two `{{resolve:...}}` expressions resolving to one value keep a single entry — whichever the resolver recorded last. A WHOLE-token leaf was immune (the position pass copies its own source), but a literal leaf that EMBEDS a token (`postgres://app-svc:{{resolve:ssm-secure:NAME}}@db/app`) was redacted by the value scan, which could only write the map's survivor. Measured on real AWS (go-to-k/cdkd#2492's first fixture run): the embedded leaf persisted `...NAME:1}}@...` for a template that spells `NAME`, so the deploy diff reported that leaf on every run (a REPLACEMENT every run for a create-only property), and a rollback or `drift --revert` re-resolved the version-pinned expression to a stale value after rotation.

## What changed

- **Evidence** (`src/deployment/secret-redaction.ts`): a `WeakMap` side table keyed by the map INSTANCE holds the uncollapsed `expression -> plaintext` pairs the resolver recorded into that map (`recordResolvedPair`; an expression seen with two values becomes conflicting and vouches for nothing). It is exactly as pass-local as the map: a map a resolver populated — the deploy's `perResourceSecrets` entry, and equally drift's, scrub's and import's own resolution maps — carries only that pass's pairs (the scrub outputs pass shares one map between its `Export.Name` and value loops, and a moved value can conflict a pair there: go-to-k/cdkd#2531), while a map no resolver populated (a derived needle map, a nested-stack inheritance copy, a `new Map(secrets)` copy) starts with none and keeps the old fall-through. `mergeResolvedPairs` carries it into the one sanctioned copy that can position a leaf — the deploy engine's `outputSecrets` accumulation (`src/deployment/deploy-engine.ts`), so a literal Output embedding a reference is covered too. The engine's other entry-by-entry copy, an `Export.Name`'s secrets into the pass map, deliberately does NOT merge pairs: a name never positions a leaf, a value re-using the same token records its own pair at the seam, and the merge could only mark a pair CONFLICTING (a non-cacheable `{{resolve:ssm:X}}` whose value moved between the value pass and the name's resolution) and destroy positioning the value pass had earned.
- **Recording** (`src/deployment/intrinsic-function-resolver.ts`): both seams (cache hit and fresh resolution) record the pair beside `recordedSecretValues.set`.
- **The arm** (`positionByEmbeddedSpan`, called from `redactByPath`'s string arm before the value-scan fall-through): the source has exactly ONE span; the bag carries the source's prefix and suffix around a non-empty middle that is not itself a token; the middle EQUALS what this pass recorded the source token resolving to; and the ordinary value scan of the bag equals `prefix + <the map's survivor> + suffix` — i.e. the scan would rewrite the middle and nothing else. Then `prefix + token + suffix`. That last condition is what makes the substitution a CHOICE among this pass's expressions rather than a new claim: the arm never rewrites a leaf the scan would leave alone (an embedded 1-3 character secret stays the scan's documented residual — go-to-k/cdkd#2516; a whole leaf that is another recorded plaintext, or a needle starting in the prefix, keep the scan's precedence). The scan runs once per string leaf, inside the arm, which returns it on every refusal — so the compared value provably comes from the same bag and map. A public `ssm` token and a mask-only `NoEcho` value are never recorded, an already-redacted middle is refused, and an `Fn::Sub` / `Fn::Join` source is an object and never reaches this arm (the placeholder primitive go-to-k/cdkd#2298 / go-to-k/cdkd#2320 ask for). The frame is copied from the source, not scanned, for the reason the whole-token arm returns its source unscanned.
- **Tests**: `tests/unit/deployment/secret-redaction-embedded-span.test.ts` (positive in both map orders; the outputs-bag merge incl. a conflict; the same-plaintext previous-generation middle pinned as the scan's class of answer; fall-throughs: two spans, frame mismatch, unrecorded cross-generation middle, previous-generation middle equal to a sibling secret while the token resolved elsewhere, mask-only middle, already-redacted middle, conflicting pair, map copy, whole-leaf-is-another-plaintext, overlapping needle, 1-3 character middle, empty middle — each asserted EQUAL to the value scan's own answer), a real-resolver case in `tests/unit/deployment/dynamic-references.test.ts` (both key orders), a real-resolver case in `tests/unit/deployment/secret-redaction-intrinsic-source.test.ts` for an `ssm` reference whose `Type` came back unclassifiable, in both key orders: NOT pinned process-wide, but still pass-local evidence, so a literal leaf embedding it beside a whole-value sibling sharing the plaintext is positioned by its own span (when the sibling loses the slot it is the empty-frame shape the span arm writes back as itself; when it holds the slot the whole-token arm copies it), and an engine case in `tests/unit/deployment/deploy-engine-outputs-generation-rules.test.ts` pinning that the `Export.Name` resolution's pairs are NOT merged: its mock resolver returns a moved value on the second resolution of the token (modelling a re-asked unclassifiable `ssm` reference, with that file's stub and spellings), which would otherwise conflict the evidence the value pass earned and push the embedded Output back to the sibling's spelling.
- **Live arms** (two fixtures). `tests/integration/secrets-dynamic-ref`: a new `DB_DSN_LITERAL` env var embedding the same `:password` reference `SECRET_PASSWORD` holds whole, built with a CONCRETE account (`CDK_DEFAULT_ACCOUNT`, which cdkd's synth exports to the app) so it synthesizes as a plain string rather than the `Fn::Join` every other env var of that fixture is, and placed ABOVE `SECRET_PASSWORD_STAGED` so the collapsed survivor is the staged spelling. `verify.sh` asserts the live value, a premise guard that the synthesized leaf is a string, and exact equality of the persisted leaf with its own expression (a `:AWSCURRENT` spelling is named as the go-to-k/cdkd#2485 collapse), on top of the existing `diff --fail` guard. `tests/integration/ssm-secure-dynamic-ref` (go-to-k/cdkd#2492, the fixture whose first real-AWS run measured this defect): a fourth env var `SSM_SECURE_VERSIONED_SAME` restores the exact collision shape (whole + embedded + version-pinned references to ONE parameter, declared after the embedded key), with an exact-spelling guard on the embedded leaf that names the `:1}}@` collapse and a premise guard on the synthesized key order.
- **Docs**: `docs/changelog-cdkd.md`; a sentence in `.claude/rules/layout-deployment-secrets.md`. (An earlier revision corrected a drift.ts comment; go-to-k/cdkd#2492 removed that block first, so drift.ts is untouched after the rebase.)

## Deviation from the issue's proposal, deliberate

The issue proposed gating on "the middle equals this pass's recorded plaintext for the token" and asked for positive 1-3 character coverage. Review (Codex, 5 rounds) showed that gate alone still lets the arm rewrite a leaf the value scan would NOT touch (a middle below the needle floor, or a leaf where another needle wins the scan's whole-value / leftmost precedence) — a new claim on a bag this pass did not necessarily produce. The gate is therefore equivalence with the scan's own answer, and the 1-3 character case stays a residual, stated in the docstring, the changelog and the unit case that pins it. The maintainer review measured that dropping the `scanned === bag` refusal alone would close the residual while keeping the three interference cases; a second review found that it would also drop the arm's generation bound (`recorded === middle` proves what this pass resolved the token to, not that this pass produced the bag — a previous generation's or readback 1-3 character middle coinciding with today's plaintext would be persisted as today's expression, where the scan makes no claim below its floor). go-to-k/cdkd#2516 asks for a bag-level generation marker before the relaxation.

## Verification

On the round-2 review-fix commit (the scan folded into the arm):

- `vp check --fix` / `vp run check` / `vp run typecheck:test` / `vp run build`: 0 errors; full `vp test run` at the default timeout: 882 files, 18,228 passed (1 skipped), no type errors, root-asserted in the PR worktree.
- `/run-integ secrets-dynamic-ref`: PASS (482 s) — 4 deleted, 0 errors, 0 orphans, 0 surviving state versions; the `DB_DSN_LITERAL` guard holds at persist, after the rollback replay and in the re-captured baseline.
- `/run-integ ssm-secure-dynamic-ref`: PASS (154 s) — 2 deleted, 0 errors, 0 orphans, versions swept.
- `/run-integ lambda` (broad): PASS (103 s) — 9 deleted, 0 errors, 0 orphans.
- Codex review loop on the delta to NO ISSUES (1 round), on the go-to-k/cdkd#2531 body to NO ISSUES (3 rounds — round 1 caught that the scrub comment's cache mechanism does not apply to a never-cached reference, round 2 that "never plaintext" needed the go-to-k/cdkd#2516 qualifier), and on this reply (2 rounds).

On the review-fix commit (the delta the maintainer review asked for):

- `vp check --fix` / `vp run check` / `vp run typecheck:test` / `vp run build`: 0 errors; full `vp test run` at the DEFAULT timeout (no `--testTimeout` override, since go-to-k/cdkd#2513 landed): 882 files, 18,228 passed (1 skipped), no type errors, root-asserted in the PR worktree on the rebased tree.
- `/run-integ secrets-dynamic-ref`: PASS (516 s) — 4 deleted, 0 errors, 0 orphans, 0 surviving state versions; the `DB_DSN_LITERAL` guard holds at persist, after the standalone rollback replay and in the re-captured baseline.
- `/run-integ ssm-secure-dynamic-ref`: PASS (156 s) — 2 deleted, 0 errors, 0 orphans, versions swept; `SSM_SECURE_VERSIONED_SAME` beside `SSM_SECURE_EMBEDDED` keeps the embedded leaf on its own spelling.
- `/run-integ lambda` (broad, `integ-broad`): PASS (110 s) — 9 deleted, 0 errors, 0 orphans.
- Mutation probes on the delta: dropping the arm fails both new unclassifiable-`ssm` cases (and 4 others); re-adding the `Export.Name` pair merge fails the new engine case; `<=` → `<` survives, as documented in the empty-middle case (equivalent under the current bound).
- Two repo reviewer agents (code, test) on the delta: the rules-file sentence, the ssm test's comment and second order, the `Export.Name` pin and the `cacheable` wording above came from them. Codex review loop on the delta to NO ISSUES (3 rounds; round 1 found one false test comment), on the go-to-k/cdkd#2516 body to NO ISSUES (6 rounds — each round tightened where a generation marker can honestly be placed: not per caller or per record, not on a provider-substituted `effectiveProperties` bag, and covering `observedProperties` and `outputs`), and on this reply (4 rounds).

On the original commit:

- `vp check` / `vp run typecheck:test` / `vp run build` / full `vp test run`: all green — 881 files, 18,204 tests passed (1 skipped), no type errors, root-asserted in the PR worktree on the rebased commit. Run with `--testTimeout=20000`: the two tree-walk cases of the `aws-client-defaults-fence` critic (go-to-k/cdkd#2491, untouched by this change) need 5.0–7.7 s on this 16-core WSL host under the default parallelism and were cut off at 5 s in four default-budget runs; they pass in the default budget when run alone.
- `/run-integ secrets-dynamic-ref` from this worktree's build, on the final rebased commit: PASS (487 s) — 4 deleted, 0 errors, 0 orphans, 0 surviving state versions; the new assertions pass at persist (Phase 1), after the standalone rollback replay (Phase 1e) and in the re-captured baseline (Phase 1g), and the key-order premise holds (`DB_DSN_LITERAL` index 2, `SECRET_PASSWORD_STAGED` index 3). Three earlier green runs on this branch before the rebase.
- `/run-integ ssm-secure-dynamic-ref` (the fixture whose first run measured the defect; the same-parameter `SSM_SECURE_VERSIONED_SAME` arm added here): PASS (234 s) — 2 deleted, 0 errors, 0 orphans, S3 versions swept; the embedded leaf keeps its own spelling beside the version-pinned sibling (premise: index 3 follows index 1).
- `/run-integ lambda` (broad, `integ-broad` gate — `deploy-engine.ts` and the resolver are cross-cutting): PASS (102 s) — 9 deleted, 0 errors, 0 orphans.
- The same fixture against a PRE-FIX binary (origin/main built in a scratch worktree, dist swapped in): FAILS at the new guard because the embedded leaf collapsed onto the staged sibling's expression (go-to-k/cdkd#2485), so the live arm discriminates (81 s, cleaned up).
- Codex review loop on the diff to NO ISSUES (9 rounds; findings in rounds 1–4 and 6 — the outputs-bag copy losing the evidence, the same-plaintext cross-generation argument, the needle-floor and scan-precedence gaps that turned the gate into scan-equivalence, a `SECRET_MASK` special case, and a Phase 1g assertion that was described as a go-to-k/cdkd#2485 fence but is an empty-map positional invariant — followed by three consecutive clean rounds, the last on the rebased commit plus the second live arm), plus the repo's spec (clean), code, security (no blocker) and test reviewers (mutation probes RUN in scratch copies) — their findings applied: a shared `singleSpanFrame` helper, accurate doc wording about which maps carry evidence, conflict-rule / from-side-merge / cache-hit-seam / outputs-merge cases, pinned fall-through answers, and the fixture's key-order premise asserted from the synthesized template.


go-to-k:
## Review: 3-axis + security

Tier from the `/review-pr` heuristic: 12 non-generated files (>= 10) puts this at `3-axis`, and `src/deployment/secret-redaction.ts` is a secret-handling surface, so the security reviewer was dispatched on top of spec / code / test.

**Verdict: no blocker.** One design decision below is worth settling before merge; the rest are a stale comment and two nits.

### What held up

- **The scan-equivalence gate is sound.** No reviewer could construct an input that makes the arm write a wrong expression. `survivor` is always a complete token and `middle` never is, so equality proves the scan rewrote exactly the middle and nothing in the frame. Cross-generation bags cannot escape: either the middle is an already-persisted token (refused at `secret-redaction.ts:2559`), or this pass did not produce it (`recorded !== middle`), or it did — in which case the pre-fix scan wrote an expression at that position too.
- **The WeakMap really is pass-local.** Every producer keeps the instance (`buildResolverContext` -> `perResourceSecrets.set`); nested-stack inheritance copies associations, not entries; `scrub`'s union and the derived needle map are consumed source-less. No path carries pairs onto a map they do not describe.
- **Every claim in the PR body checks out against the code** — both resolver seams, the `Export.Name` and outputs merges, `Fn::Sub` / `Fn::Join` being unreachable (the arm sits inside `isDynamicReferenceString(source)`), copies losing the evidence. The `singleSpanFrame` extraction is behaviour-equivalent (only refusal order moved).
- **Security: no new plaintext path.** Both `recordResolvedPair` calls sit under the identical gate as the pre-existing `secrets.set`, so no new value class enters the store. `resolvedPairsOf` is module-private, never exported, serialized or logged; `outputSecrets` is reset per `deploy()`. Every reader that re-applies (rollback replay, `drift --revert`) now re-resolves the leaf's own token instead of the map's survivor.
- **The integ arms discriminate.** Exact string equality plus the `:1}}@` / `:AWSCURRENT}}@` negatives, and both premise guards (synthesized shape is a plain string, key ordering) `exit 1` rather than passing silently.
- **Mutation sweep over `tests/unit/deployment`:** 15 mutations, 13 killed — dropping scan-equivalence, dropping `recorded !== middle`, dropping the evidence gate, dropping the not-a-token and single-span checks, inverting the conflict rule in either direction, skipping the `mergeResolvedPairs` conflict branch, removing the outputs-bag merge, removing either resolver seam's record, and removing the arm call.
- **Post-merge check.** CI is green and the branch merges cleanly into current `main`. I built the post-merge tree and ran `rule-file-payload` / `docs-site-links` / `check-scope-checker-inputs` on it (288 passed), so #2512's `.claude/rules` growth does not collide with this branch's `layout-deployment-secrets.md` addition.

### 1. The 1-3 character residual is closable here, and `Closes #2485` currently narrows the issue

#2485 names plaintext persistence for an embedded 1-3 character secret in its Summary, its Impact and its `Severity:` line — as a **disclosure**, not a cosmetic residual. This PR leaves it and pins it as intended at `tests/unit/deployment/secret-redaction-embedded-span.test.ts:63`, whose final assertion (`:76`) is `expect(persisted).toEqual({ Dsn: leaf })` — the plaintext itself. No follow-up issue exists (searched open and closed).

The deviation argument for condition 5 is **correct**, and I measured it rather than taking it on faith. Dropping `secret-redaction.ts:2649` entirely (the four-condition gate #2485 proposed) fails three cases: the whole leaf being another recorded plaintext, the literal frame itself equalling a recorded plaintext, and a needle starting in the prefix and overlapping the middle. So the fifth condition earns its place.

But the 1-3 character drop does not follow from it. Relaxing `:2649` to

```ts
const scanned = redactSecretsForState(bag, secrets);
if (scanned !== bag && scanned !== prefix + survivor + suffix) return undefined;
```

still refuses all three of those cases, and across `tests/unit/deployment` plus the drift / scrub suites (2,083 tests) the only test that flips is the residual pin at `:63`. The needle floor exists because a value scan cannot tell a short secret from coinciding text; this arm holds positive pass-local proof (`recorded === middle`) that the middle IS this token's resolved value, so that ambiguity is not present here.

Two ways to settle it: relax the condition in this PR, or change `Closes #2485` to `Refs` and file the follow-up. Either is fine — leaving it as-is closes the issue over the disclosure it names.

### 2. `src/deployment/intrinsic-function-resolver.ts:7127-7128` is now false, and contradicts the new docstring

> Such a pair therefore still falls back to the value scan, i.e. to today's behavior.

`recordResolvedPair` at `:7115` is gated only on `isSecret && resolved`, so an unclassifiable-`Type` `ssm` reference does get pass-local evidence and can be positioned. `positionByEmbeddedSpan`'s own docstring (`src/deployment/secret-redaction.ts:2608-2613`) describes exactly that shape and says it "is written back as itself". Two comments in one PR give opposite answers for the same input.

That shape also has no test — `secret-redaction-ssm-collision.test.ts` passes with the arm removed and fails only when the evidence gate is removed. Worth one case plus a corrected sentence.

### 3. nit: `src/deployment/deploy-engine.ts:5792` buys nothing

`mergeResolvedPairs(nameSecrets, context.recordedSecretValues)` is strictly non-positive — it cannot add a pair the output-value seam did not already record (the only source handed to `redactOutputs` is `output.Value`), but it can mark an expression `CONFLICTING` and destroy positioning the value pass had earned, e.g. a non-cacheable `{{resolve:ssm:X}}` whose value changes between PASS 1 and the `Export.Name` resolution. Deleting the call changed no test. The direction is safe, so this is a nit either way — but the code comment already says it is not a reachable difference.

### 4. nit: the refuse path scans twice

`positionByEmbeddedSpan` runs `redactSecretsForState(bag, secrets)` at `:2649`, then `redactByPath` runs it again at `:2934` on the fall-through, each rebuilding a fresh `RegExp` over every needle (`buildNeedleRegex` is not memoized). Only string leaves whose source contains `{{resolve:`, so negligible today — noted so a future widening of the arm's entry condition does not make it a persist-path cost.

### Two smaller notes

- Two mutations survived. `secret-redaction.ts:2550`'s `<=` -> `<` (admitting an empty middle) has no test, but looks like an equivalent mutant — the arm would need `recorded === ''` and `learnNeedle`'s floor drops `''`. The other is item 3 above.
- Some fall-through case names claim more than they pin: the TWO-spans case is refused by the frame check rather than the span-count clause, the already-a-token case by the evidence gate rather than the token clause, and the unvouched-middle case by `survivor === undefined`. All three clauses are pinned elsewhere (the derived-needle, conflict and map-copy cases), so there is no coverage gap — only labels that overstate.

I did not merge: the `verify-pr` / `pr-review` / `integ-*` markers are per-worktree and I ran none of them here, and item 1 is a call for the author to make.

nix-tkobayashi:
Thanks for the review — every item is addressed in the follow-up commit, with one design call explained below.

### 1. The 1-3 character residual — kept as a residual, re-scoped to `Refs`, follow-up filed as go-to-k/cdkd#2516

I measured the relaxation you proposed and it behaves exactly as you describe: the three interference cases stay refused, and only the residual pin flips. I still did not apply it, because of what `recorded === middle` proves and what it does not. It proves that THIS pass resolved the source token to the middle; it does not prove that this pass produced the bag. Under `TEMPLATE_DERIVED_RULES` / `TEMPLATE_SOURCED_RULES` (`sourceIsSameGeneration: false` on both, for the reason written beside them: `redactStateForPersist` walks every record while `perResourceTemplateProps` is set before the provider call) the bag can be a previous generation's record or an AWS readback walked against today's template. For a middle of 4+ characters the value scan already rewrites such a bag onto one of this pass's expressions regardless of generation, so the arm only changes WHICH expression is written — that is the argument in the docstring's "what this evidence does not claim" paragraph, and the bound the scan-equivalence condition keeps. Below the floor the scan makes no claim, so `scanned === bag` would let the arm make a new one: a 1-3 character readback or previous-generation middle that coincides with today's plaintext (`port:0` where AWS returned a default and today's secret resolved to `0`) is persisted as today's expression. It round-trips until the secret rotates; after that `cdkd drift` reports a change that never happened and `drift --revert` writes the secret over a literal the template never put there.

What narrows the residual to something stateable is a bag-level generation marker that the rules cannot supply — installed on the bags this pass produced on its success path, never at resolution time: `propertiesToRecord`'s result when it IS the resolved bag (a provider-substituted `effectiveProperties` bag is excluded as mixed provenance — the DynamoDB global-table provider carries previous-state values into it), the resolved `outputs` bag, and the `drainObservedCaptures` readback for the same freshly written resource. The readback is not something this pass wrote, and that is where a residual stays: a field AWS normalized to a value coinciding with the 1-3 character secret at that offset would be persisted as the expression — the same class the value scan already accepts for a 4+ character coincidence in a readback, and stated as such. That is its own review cascade, so go-to-k/cdkd#2516 carries it with the constraints, the negatives and the integ arm spelled out, and this PR is `Refs go-to-k/cdkd#2485` rather than `Closes`, so the issue is not closed over the disclosure it names. If you would rather take the relaxation here with the residual stated, say so and I will switch — the measurement is not in dispute, only the bound.

### 2. `intrinsic-function-resolver.ts` comment — corrected, plus the missing test

You are right: the pair is recorded under `isSecret && resolved`, so an unclassifiable-`Type` `ssm` reference IS pass-local evidence and is positioned; only the process-wide pin is withheld. The comment now says exactly that (and the same sentence in `.claude/rules/layout-deployment-secrets.md`, which a reviewer caught), and `secret-redaction-intrinsic-source.test.ts` gained a resolver-driven case in both key orders: an unclassifiable `ssm` reference embedded in a literal leaf beside a whole-value sibling sharing the plaintext. When the sibling resolves last it holds the collapsed slot and the whole-token arm copies it; when it resolves first it loses the slot, the whole-token arm cannot vouch for it, and it is the empty-frame shape the span arm writes back as itself — the case the corrected comment promises. In both orders the embedded leaf persists its own expression and the process-wide store stays empty. Dropping the arm fails both.

### 3. `deploy-engine.ts` `Export.Name` pair merge — removed

Agreed that it is strictly non-positive. The entries copy stays (the "survives every exit" invariant is about entries); the pairs are not merged, and the comment at the site and the `mergeResolvedPairs` docstring now say why: a name never positions a leaf, a value re-using the token records its own pair at the seam, and the only thing the merge could do is mark a pair conflicting. Since that is now a reachable harm rather than "no difference", it has a pin: an engine case in `deploy-engine-outputs-generation-rules.test.ts` whose MOCK resolver returns a moved value on the second resolution of the same token (the one the `Export.Name` triggers). That models the shape — the real resolver never caches an unclassifiable `ssm` reference, so a name re-asks AWS — with that file's `secretsmanager` spellings and a stub, not the real resolver; the real-resolver half of the story is the ssm case in item 2. Re-adding the merge fails it. One pre-existing sentence a few lines above the site ("a LATER output resolving the same reference cache-hits, records nothing") predates the resolver's `cacheable = false` rule for that shape — the entries merge it motivates is still needed for the throw-partway path, so I left it, but it reads as contradicting the new comment and is worth a separate tidy.

### 4. Double scan — removed

`redactByPath`'s string arm now computes `redactSecretsForState(bag, secrets)` once and hands it to `positionByEmbeddedSpan` as `scanned`; the arm compares against it and never scans on its own, and every refusal falls through to the same value.

### The two smaller notes

- `<=` → `<`: kept `<=`. Under the current bound the mutant is equivalent (an empty middle has no map entry, so `survivor` refuses it too), and I did not want to claim a pin I do not have — the new empty-middle case in `secret-redaction-embedded-span.test.ts` says so in its comment and exists so the invariant has its own case when go-to-k/cdkd#2516 relaxes the bound and the frame refusal becomes the only thing standing there.
- The three fall-through names: both readings are true and I left the names. Read as "which clause the arm reaches first", they are accurate — the TWO-spans case is refused at `spans.length !== 1` before any frame comparison, the already-a-token case at `isSingleDynamicReferenceToken(middle)` inside the same helper before the evidence lookup, and the unvouched cross-generation case at `recorded !== middle` (its map has no entry for that middle, so `survivor` would refuse it next). Read as "which clause a mutant needs", they overstate exactly as you say: delete the named clause and the next one still refuses the leaf. Since the clauses have their own pins elsewhere (the derived-needle, conflict and map-copy cases) and the names describe the input shape rather than a probe, I kept them; happy to rename if you prefer the probe reading.

### Verification

- `vp check --fix` / `vp run check` / `vp run typecheck:test` / `vp run build`: 0 errors; full `vp test run` at the DEFAULT timeout (no `--testTimeout` override, since go-to-k/cdkd#2513 landed): 882 files, 18,228 passed (1 skipped), no type errors, root-asserted in the PR worktree on the rebased tree.
- `/run-integ secrets-dynamic-ref`: PASS (516 s) — 4 deleted, 0 errors, 0 orphans, 0 surviving state versions; the `DB_DSN_LITERAL` guard holds at persist, after the standalone rollback replay and in the re-captured baseline.
- `/run-integ ssm-secure-dynamic-ref`: PASS (156 s) — 2 deleted, 0 errors, 0 orphans, versions swept; `SSM_SECURE_VERSIONED_SAME` beside `SSM_SECURE_EMBEDDED` keeps the embedded leaf on its own spelling.
- `/run-integ lambda` (broad, `integ-broad`): PASS (110 s) — 9 deleted, 0 errors, 0 orphans.
- Mutation probes on the delta: dropping the arm fails both new unclassifiable-`ssm` cases (and 4 others); re-adding the `Export.Name` pair merge fails the new engine case; `<=` → `<` survives, as documented in the empty-middle case (equivalent under the current bound).
- Two repo reviewer agents (code, test) on the delta: the rules-file sentence, the ssm test's comment and second order, the `Export.Name` pin and the `cacheable` wording above came from them. Codex review loop on the delta to NO ISSUES (3 rounds; round 1 found one false test comment), on the go-to-k/cdkd#2516 body to NO ISSUES (6 rounds — each round tightened where a generation marker can honestly be placed: not per caller or per record, not on a provider-substituted `effectiveProperties` bag, and covering `observedProperties` and `outputs`), and on this reply (4 rounds).

go-to-k:
## Round 2 on the delta (`0cda05aa..7c2ac6a8`)

Re-reviewed the fix delta on its own — code, security and test axes, with mutation probes — rather than re-running the size heuristic. **No blocker. Three low-severity items and one out-of-scope observation, all below.**

### 1. I withdraw the relaxation. Your bound is right and mine was wrong.

I verified your argument rather than taking it: `TEMPLATE_DERIVED_RULES` and `TEMPLATE_SOURCED_RULES` both carry `sourceIsSameGeneration: false` (`src/deployment/secret-redaction.ts:1758` and `:1778`), and `buildNeedleRegex` filters every needle shorter than `MIN_NEEDLE_LENGTH` out of the alternation unconditionally (`:1550`).

That is the asymmetry I missed. At 4+ characters the scan already rewrites a cross-generation bag onto one of this pass's expressions, so the arm only chooses WHICH — that is the argument I accepted in round 1 and it does not extend downward. Below the floor the scan makes no claim at all, so `scanned === bag` would have the arm make a new one, on a bag whose generation it cannot establish. And the failure is not merely possible there, it is *likelier* than at full length: `recorded === middle` on a three-character middle is a coincidence a readback hits routinely (`port:0`, `:80`) where a 20-character secret never does. So the very property I argued made the short case safe — positive evidence for the middle — is weakest exactly where I wanted to lean on it.

`Refs` plus go-to-k/cdkd#2516 is the right landing. I read go-to-k/cdkd#2516 and it is scoped better than my finding was: it establishes from `buildNeedleRegex` that the sub-floor leaf is unredacted **with or without a same-plaintext sibling**, which go-to-k/cdkd#2485 did not say, and it names the three bags a generation marker has to cover rather than one. No change requested.

I added the missing `severity:medium` / `effort:medium` labels to go-to-k/cdkd#2516 and to go-to-k/cdkd#2485 — both were unlabelled, which is what `pr-inherit-issue-labels.yml` reads, so this PR was inheriting nothing. That was an unhandled item from my round 1 and I should have said so then rather than only listing it.

### 2-4. Verified fixed

The scan hoist, the `Export.Name` pair-merge removal and the corrected resolver comment all check out, and the pins are real rather than asserted. Mutation results on the delta:

```text
positionByEmbeddedSpan(bag, source, secrets, bag)              KILLED  9 tests / 4 files
scanned = redactSecretsForState(source, secrets)               KILLED  31 tests / 9 files
call the scan twice and throw if unequal (determinism probe)   never fired -> the 2-to-1 collapse is safe
re-add mergeResolvedPairs(nameSecrets, ...)                    KILLED  exactly the new engine case
const spanned = undefined (span arm disabled)                  KILLED  both new intrinsic-source cases
add `|| service === 'ssm'` to the pinSecretVerdict gate        KILLED  both new cases + the pre-existing "does NOT pin"
```

Two specifics worth recording. The whole-token arm's exits both close above the hoist (`secret-redaction.ts:2926` / `:2932`, inside the block that ends at `:2933`), so nothing that used to be returned unscanned is scanned now, and every path through the arm performs exactly one scan where the worst case used to perform two. And the `Export.Name` removal drops only the pairs: `resolvedPairsOf` has exactly one read path (`resolvedPlaintextOf` -> `positionByEmbeddedSpan`), while the entries copy at `deploy-engine.ts:5785` is what the masking, journal and exports-index readers consume, so the removal can change which expression is written and never expression-to-plaintext.

The `<=` decision holds up too, and for a stronger reason than "uncovered": no writer can put `''` into a `RecordedSecretValues` map (the resolver seam gates on `isSecret && resolved`, `recordMaskOnlyValue` gates on `MIN_NEEDLE_LENGTH`), so `secrets.get('')` is always `undefined` and `survivor` refuses independently. The mutant is genuinely equivalent, and the empty-middle case is honest documentation of that. Keeping `<=` is right.

### New: the fourth parameter has no type-level tie to the bag it scanned

`positionByEmbeddedSpan(bag, source, secrets, scanned)` (`secret-redaction.ts:2637-2642`) takes `scanned: string`, and the arm's entire safety bound is `scanned === prefix + survivor + suffix`. Nothing in the signature says that string has to be `redactSecretsForState(bag, secrets)`; a second caller passing a scan of a different bag or map defeats the bound with no type error and no test failure — the `bag`-instead-of-`scanned` mutant above kills 9 tests today, but only because there is one caller doing the right thing.

`scanned` has exactly two uses (the arm call at `:2944` and the fall-through `return`), so folding it back and returning the scan on refusal — `string` rather than `string | undefined` — collapses the caller to

```ts
return positionByEmbeddedSpan(bag, source, secrets);
```

That keeps the one-scan property this round bought and makes the compared value provably come from the arm's own `(bag, secrets)`. Nit, not a blocker. If you keep the parameter, the docstring should say the caller must pass the scan of THIS bag with THIS map.

### New: two comment/fixture accuracy items in the engine case

Both in `tests/unit/deployment/deploy-engine-outputs-generation-rules.test.ts`, both low, neither a false pass — the case's load-bearing assertion (`expect(saved.outputs['Dsn']).toBe(DSN)`) is solid and re-adding the merge kills it.

- **`:319-320`** — the comment says the moved value "is still an ENTRY of the pass map: nothing of it reaches state", but `expect(JSON.stringify(saved)).not.toContain(MOVED_PLAINTEXT_A)` does not verify that. Deleting the entries loop leaves this case green: the `Export.Name` is refused, so `MOVED_PLAINTEXT_A` never reaches a bag and the assertion is vacuously true. The behaviour IS pinned — in `deploy-engine-outputs-export-name-collision.test.ts` ("a REFUSED export name still RECORDS its secret for the rest of the pass"). Worth pointing the comment there rather than claiming this assertion does it.
- **`:43-50`** — the comment justifies the moving value with "the real resolver never caches an `ssm` reference whose `Type` came back unclassifiable (`cacheable = false`)", but the expression put in `movesAfterFirst` is `EXPR_A_PLAIN`, a `secretsmanager` reference (`:37`). `cacheable = false` is set only inside the `ssm` arm (`intrinsic-function-resolver.ts:7042`), so the real resolver would cache this one and could not return a moved value on a second resolution in one deploy; the stub models no cache at all. Documentation realism rather than a wrong-reason pass, and I appreciate it is awkward to fix — the same-plaintext plain/staged collision the case needs is inherently `secretsmanager`-shaped. A sentence saying the stub deliberately models the shape rather than a reachable resolver state would close it.

### Out of scope, for a separate issue if you agree

The conflict hazard this delta closes on the deploy side is still reachable in `cdkd scrub`. `src/cli/commands/scrub.ts:3202` resolves each `Export.Name` with `recordedSecretValues: outputSecrets` — the SAME map the value loop fills — and the name loop runs FIRST. The entries side of that is deliberate and load-bearing (the comment at `:3205-3212` explains why: an unpinned `ssm` reference must be visible when the value loop meets it, or `--dry-run --fail` reports CLEAN on a leaking stack), so the deploy-side remedy of not merging does not transfer: there is no separate `nameSecrets` map to withhold, the resolver seam records the pair directly. So a non-cacheable unclassifiable `ssm` reference resolved in the name loop and again in the value loop with a moved value marks the pair conflicting and costs the value leaf its positioning. Bounded to expression-versus-expression — it falls back to the value scan's answer, never to plaintext — so a follow-up rather than something for this PR.

### On the three fall-through case names

Agreed, leave them. Your "which clause the arm reaches first" reading is the one the names are written in, and I checked the clauses do have their own pins elsewhere. Renaming would trade an accurate description of the input shape for an accurate description of a probe, and the probe reading is the one a mutation run gives you anyway.

nix-tkobayashi:
Thanks — the three items are in the follow-up commit, and the out-of-scope observation is filed as go-to-k/cdkd#2531.

### `scanned` folded back into the arm

Taken as proposed. `positionByEmbeddedSpan(bag, source, secrets)` now computes `redactSecretsForState(bag, secrets)` once itself and returns that scan on every refusal (return type `string`), so the bound `scanned === prefix + survivor + suffix` provably compares against the scan of the same bag and map; the string arm of `redactByPath` is the one-liner you wrote. The docstring records why the parameter went away (a second caller could have defeated the bound with no type error), and the changelog sentence was updated to match. Byte-equivalent for every input — the one-scan property from round 1 is kept, just moved inside.

### The two engine-case comments

- The stub's header comment (your `:43-50`) now says the stub models the SHAPE and not a reachable resolver state: the expression it moves is this file's `secretsmanager` spelling, which the real resolver would cache, and the stub caches nothing.
- The assertion comment (your `:319-320`) now calls `not.toContain(MOVED_PLAINTEXT_A)` a leak guard rather than a pin of the entries copy, and points at the case that does pin it ("a REFUSED export name still RECORDS its secret" in `deploy-engine-outputs-export-name-collision.test.ts`).

### scrub — go-to-k/cdkd#2531

Agreed it is reachable and agreed it is a follow-up. One thing came out of writing it up that is worth a sentence here: the shared-map comment at `scrub.ts:3206-3212` grounds the entries sharing on the cache-hit arm re-recording only what it can still prove is secret, but the unclassifiable-`Type` reference this hazard needs is never cached at all (`cacheable = false`), so the value loop re-asks AWS and records its own entry either way; and since go-to-k/cdkd#1933 the cache carries the verdict beside the value, so the hit arm re-records a cached secret too. The issue therefore does not lean on that mechanism — it proposes the deploy engine's shape (a private `nameSecrets` map, an entries copy in a `finally`, no pair merge), which keeps every entry the shared map would have held by construction, and asks the fix PR to re-read that comment against go-to-k/cdkd#1933 rather than deciding it here. Severity low: expression-versus-expression for a 4+ character plaintext; the sub-floor case is go-to-k/cdkd#2516's regardless.

### Verification

- `vp check --fix` / `vp run check` / `vp run typecheck:test` / `vp run build`: 0 errors; full `vp test run` at the default timeout: 882 files, 18,228 passed (1 skipped), no type errors, root-asserted in the PR worktree.
- `/run-integ secrets-dynamic-ref`: PASS (482 s) — 4 deleted, 0 errors, 0 orphans, 0 surviving state versions; the `DB_DSN_LITERAL` guard holds at persist, after the rollback replay and in the re-captured baseline.
- `/run-integ ssm-secure-dynamic-ref`: PASS (154 s) — 2 deleted, 0 errors, 0 orphans, versions swept.
- `/run-integ lambda` (broad): PASS (103 s) — 9 deleted, 0 errors, 0 orphans.
- Codex review loop on the delta to NO ISSUES (1 round), on the go-to-k/cdkd#2531 body to NO ISSUES (3 rounds — round 1 caught that the scrub comment's cache mechanism does not apply to a never-cached reference, round 2 that "never plaintext" needed the go-to-k/cdkd#2516 qualifier), and on this reply (2 rounds).

go-to-k:
## Round 3 on the delta (`7c2ac6a8..ced41e18`)

Code and security axes on the delta. **Clean — no blocker, no minor. One optional nit, and it predates this round.**

The fold is byte-equivalent. Four refusal points (`src/deployment/secret-redaction.ts:2646`, `:2649`, `:2661`, `:2662`) all return `scanned` and no fifth exists; the success return is untouched; round 2's caller computed `scanned` in the same basic block immediately before the call, so `spanned ?? scanned` is literally what the new body evaluates. Still exactly one scan per string leaf, and the whole-token block still returns unconditionally on both arms above it, so nothing that used to be returned unscanned reaches the scan. `positionByEmbeddedSpan` has one caller, is not exported, and no test imported it, so nothing consumed the `undefined` signal.

Worth recording that the collapse **removes** a hazard rather than trading one away. While `scanned` was a parameter, a second caller could have bound the arm's safety condition to a scan of a different bag with no type error — the arm would then position a leaf the scan for THIS bag would not rewrite, which is disclosure-shaped. That is now unreachable by construction. The residual ambiguity (a caller cannot tell positioned from refused) resolves toward the more-redacted answer, and round 2 had the same property anyway: `spanned` was equally an already-positioned string that must not be rescanned.

Mutation probes confirm the internal scan is pinned rather than merely moved: `const scanned = bag` fails 8+ cases across the embedded-span and dynamic-references suites, and refusals `:2646` / `:2649` / `:2662` fail 2 / 5 / 3 respectively.

### Nit — `secret-redaction.ts:2661` is unreachable, and pre-dates this round

`if (survivor === undefined) return scanned;` never fires. Replacing it with a `throw` and running all of `tests/unit/deployment/` (119 files, 2001 tests) leaves the suite green.

It is unreachable by construction, which I checked rather than inferred from the probe: `recorded === middle` implies the resolver ran `secrets.set(middle, expr)` beside `recordResolvedPair` at both seams (`intrinsic-function-resolver.ts:6978` and `:7107`, entry first in both), `mergeResolvedPairs` has exactly one call site and the entries loop sits immediately above it in the same block (`deploy-engine.ts:5908-5915`), and no `.delete()` on any `RecordedSecretValues` exists anywhere under `src/`. So once `recorded === middle`, `secrets.get(middle)` is always defined.

`secrets.get()` is typed `string | undefined`, so the guard cannot simply go — the honest fix is a sentence saying it is a type-narrowing formality that `recorded === middle` already implies, not a reachable refusal. Keeping the fail-closed shape over a non-null assertion is right. It arrived in round 2 (returning `undefined` then, equally unobservable), so it is not a regression here and I would not hold the PR for it. Do not add a test — the state cannot be constructed through the public seam.

### The two comment corrections

Both accurate, verified against the code rather than read: the stub does model a `secretsmanager` spelling while the real `cacheable = false` state is `ssm`-only, and the `not.toContain(MOVED_PLAINTEXT_A)` assertion does hold with the entries loop deleted, so "leak guard, not a pin of the entries copy" is the correct description — and `deploy-engine-outputs-export-name-collision.test.ts:656` ("a REFUSED export name still RECORDS its secret for the rest of the pass") is where that copy is actually pinned.

### go-to-k/cdkd#2531 and the correction in it

Your correction lands, and I verified both halves: `cacheable = false` is set only in the unclassifiable-`Type` branch (`intrinsic-function-resolver.ts:7042`), so the reference this hazard needs is never cached and the value loop re-asks AWS regardless; and the cache-hit arm re-records on `cached.secret && cached.value` (`:6978`), so a cached secret is re-recorded too. My round-2 comment took `scrub.ts:3206-3212` at face value as the justification for the shared map, and that comment's stated mechanism does not cover the case — so the entries sharing needs a different grounding than the one written there. Asking the fix PR to re-read it against go-to-k/cdkd#1933 rather than settling it in the issue is the right call, and proposing the engine's shape (a private name map with an entries copy in a `finally`) keeps every entry by construction instead of by that argument.

I added the missing `severity:low` / `effort:small` labels to go-to-k/cdkd#2531.

### Verdict

No further blocking findings from me. The three integs, the full suite at the default timeout and the delta probes cover this round; `check-build-test` is the only check still running as I write. Nothing here needs another round of review — the `:2661` comment is optional polish you can take or leave.

nix-tkobayashi:
Took the `:2661` nit, as a comment only: the guard now says it is a type-narrowing formality that `recorded === middle` already implies (both seams set the entry beside the pair, the one `mergeResolvedPairs` caller copies the entries first, nothing deletes from a `RecordedSecretValues` map), kept in the fail-closed shape rather than a non-null assertion, with no test since the state cannot be built through the public seam. Comment-only change; `vp check` / `vp run check` / `typecheck:test` / `build` 0 errors, full `vp test run` green at the default timeout, Codex NO ISSUES on the sentence. Nothing else from my side — over to you for the merge.

go-to-k:
## Merge-gate review: 3-axis + security on `1a5f29c0`

Re-ran the full tier for the merge rather than only the last delta — the gate binds to HEAD and this is a fork PR, so the maintainer-side markers have to be rebuilt from scratch anyway. `+1058 / -52 = 1104 LOC` after excluding `docs/_generated/**`, 14 files, 3 fix-back commits on the branch: `3-axis`, with `pr-security-reviewer` additive because the PR is a redaction / sensitive-value-persistence change.

**No blockers on any axis.** Spec: compliant, and `Refs` + go-to-k/cdkd#2516 reads as an honest boundary rather than a silent narrowing. Security: clean, and the arm provably cannot emit less redaction than the code it replaces — the bound makes the frame byte-identical to the value scan's output, and the only slot that can differ carries a complete token. Code: no blockers. Test: 20 mutants, 15 killed by the PR's own suite, including both resolver seams, both merge decisions and every evidence-bearing clause.

Two things below I would take before merging, and a list of nits I would not hold the PR for. **A test-only round costs no real AWS**: `integ-broad` / `integ-destroy` are scoped to `src/**` paths, so my maintainer-side integ runs survive a `tests/**` push. I am running them now in parallel, so taking these costs you one push and me one `check` re-run.

### Worth taking (1) — the frame anchors have no test in either consumer

`src/deployment/secret-redaction.ts:2554`:

```ts
if (!bag.startsWith(prefix) || !bag.endsWith(suffix)) return undefined;
```

Dropping **either** disjunct survives the entire unit suite (18,186 tests; the mutant's failure set is byte-identical to the baseline, compared by diff rather than by count).

Inside `positionByEmbeddedSpan` they are genuinely equivalent — the `scanned !== prefix + survivor + suffix` bound masks a misalignment. They are **not** masked in the other consumer. `learnMixedLeafNeedle` destructures `middle` straight into `learnNeedle` as the plaintext, and after the anchors the only remaining guards are the length floor and `isSingleDynamicReferenceToken`. So a bag that ends with the suffix but does not start with the prefix — `mysql://user:pass@host:SECRETVALUE@db/app` against a `postgres://app-svc:{{resolve:...}}@db/app` source — yields a misaligned slice that clears `MIN_NEEDLE_LENGTH` and is learned as a needle mapping arbitrary text onto the expression.

Pre-existing on main (the same two lines lived inline in `learnMixedLeafNeedle`), so not a regression. But the stated reason for extracting `singleSpanFrame` is "so the two refusals cannot drift apart", and an extracted predicate with no case in either consumer only half-earns that. One case in `secret-redaction-derived-needles.test.ts` closes it.

### Worth taking (2) — the integ premise guards pin one pair, not the invariant

Each guard pins a single ordering (`DSN_IDX < STAGED_IDX`; `SAME_IDX > EMB_IDX`), but the invariant the assertion needs is that the colliding key is the **last of ALL keys sharing that plaintext**. In `ssm-secure-dynamic-ref`, `SSM_SECURE_WHOLE` at index 0 also resolves to `MARKER`; move it below `SSM_SECURE_VERSIONED_SAME` and the survivor becomes the embedded leaf's own token, the assertion goes vacuous, and the premise guard stays green. Same shape in `secrets-dynamic-ref` for `SECRET_PASSWORD`. Asserting the collision key is the max index among the keys valued at that plaintext is a couple of `jq` lines and makes the guard say what it means.

Everything else about the fixtures checks out: both premise guards `exit 1` rather than warn, both fail flags are consumed (`redaction_fail` → `exit 1`; `assert_state_redacted` ends on `[ "${fail}" -eq 0 ] || exit 1`), pre-fix the survivor really is the sibling's spelling in both fixtures so both guards go red, and the Phase-G empty-map arm is labelled a non-fence rather than sold as one.

### Nits I would not hold the PR for

- `secret-redaction.ts:137-141` — the conflict path's "target map does not exist yet" branch is unexercised (replacing it with `continue` survives). Reachable: the first stack merges a conflicted pass map into a still-empty `outputSecrets`, and a later stack's valid pair would then wrongly restore vouching. Every conflict case merges into a `to` that already has pairs.
- `secret-redaction.ts:2211` — `CONFLICTING_PLAINTEXT` is declared ~2100 lines after its first use. Correct at runtime, awkward to read.
- `secret-redaction.ts:137-145` — `mergeResolvedPairs`'s symbol branch re-implements `recordResolvedPair`'s lazy init; a `pairsFor(map)` helper would leave one place to forget the `set`.
- `intrinsic-function-resolver.ts:6978` / `:7106` — `context?.recordedSecretValues?.set(...)` followed by `if (context?.recordedSecretValues) { recordResolvedPair(...) }`; binding once keeps the two from diverging.
- `secret-redaction.ts:3952-3968` — `learnMixedLeafNeedle`'s JSDoc still enumerates the frame guards that moved into `singleSpanFrame`, and omits the middle-is-a-token refusal it now inherits.
- `docs/changelog-cdkd.md:25` — the entry lists the `secrets-dynamic-ref` live arm but not the `ssm-secure-dynamic-ref` one; `ssm-secure-dynamic-ref` appears only as the fixture whose first run measured the defect.

If you would rather not take any of these, say so and I will file them as one follow-up issue and merge as-is — none of them changes behaviour.

### One thing I am handling on the merge side

`docs/_generated/integ-last-run.tsv` in this PR rewrites the `lambda` / `secrets-dynamic-ref` / `ssm-secure-dynamic-ref` rows with PASS timestamps I did not observe. It gates nothing (markgate uses local sentinels), but `/pick-integ` ranks staleness from that file, so merging marks three fixtures fresh for 14 days on an outside claim. I am re-running them maintainer-side now — plus `export` and `cross-stack-references`, which the `Export.Name` change touches and which are 8d and 26d old — and I will land the observed rows as a follow-up ledger PR right after the merge, since I cannot push to the fork branch.

nix-tkobayashi:
Thanks for landing both items directly and merging — `e214a9bb` covered exactly what I had staged (frame-anchor cases and survivor-pinning guards), so I dropped my duplicate. Your anchor fixtures are the better ones (a frame SHORTER than the source's makes the misaligned slice a substring of the plaintext, so the rewrite shows without a planted sibling), and matching the template's expression is the right key for the guards where my live-value `contains` form was a superset. One thing missed the merge by a few minutes: the changelog nit (naming the `ssm-secure-dynamic-ref` live arm and what the guards assert) is on the fork branch as `cf6d5b85`, docs-only; I will bring it as a small docs PR. For the five `src/` nits, please file the one follow-up as offered — I will pick it up with go-to-k/cdkd#2531 unless you get to it first.

go-to-k:
Already filed — go-to-k/cdkd#2546, at 10:10Z, about ninety seconds before the merge, which is probably why you missed it. It carries exactly the five `src/` items as numbered entries (declaration order of `CONFLICTING_PLAINTEXT`, the duplicated lazy init in `mergeResolvedPairs`, the doubled optional chain at both resolver seams, `learnMixedLeafNeedle`'s JSDoc still enumerating the guards that moved into `singleSpanFrame`, and the unexercised empty-target branch on the conflict path), with the four classification fields, the `Dup-check:` line and `severity:low` / `effort:small`.

It also carries the changelog nit, as a comment rather than a sixth numbered item — I had decided not to spend an 11-minute CI cycle and a `pr-review` re-bind on a sentence. Since you already have `cf6d5b85` staged, take it as a docs PR and I will strike that comment from go-to-k/cdkd#2546 when it lands, so the issue is only the five `src/` items. Worth widening it slightly while you are there: the entry also predates the strengthened premise guards, so "what the guards assert" is now two things, the live arm on `ssm-secure-dynamic-ref` and the last-key-wins invariant both fixtures pin.

On the anchor fixtures — the useful part was the failure, not the result. My first pair broke one end each and still did not kill either mutant: a same-length mismatch slices the plaintext out correctly by accident, and a merely-different frame yields a slice nothing else in the record contains, so the needle it learns matches nothing. Only the shorter frame makes the misaligned slice a substring, which is what makes the rewrite observable in the sibling leaf that was already there. Worth remembering for the next offset guard: the mutant survives not because the guard was skipped but because skipping it produced something inert.

Taking go-to-k/cdkd#2546 alongside go-to-k/cdkd#2531 sounds right — they touch the same machinery from the two sides, and go-to-k/cdkd#2531's private-name-map shape is what item 2's `pairsFor(map)` helper wants to exist first. Both are yours; I am not picking them up.

Issue #2516, opened 2026-09-04
fix(secret-redaction): an embedded 1-3 character secret persists in plaintext below the value scan's needle floor


nix-tkobayashi:
## Summary

A literal template leaf that EMBEDS one `{{resolve:...}}` token whose resolved value is 1-3 characters long (`port:{{resolve:secretsmanager:prod/db:SecretString:pin}}` with `pin` = `42`) persists the PLAINTEXT in `state.json` — as a resource property, as its AWS readback (`observedProperties`), and as a `CfnOutput` value (`outputs`, walked by `DeployEngine.redactOutputs` under `TEMPLATE_SOURCED_RULES`, `src/deployment/deploy-engine.ts` around lines 1248-1275 and 1472-1483). No same-plaintext sibling is needed for this: the substring arm of the value scan filters every needle shorter than `MIN_NEEDLE_LENGTH` (4) out of its alternation when it builds it (`buildNeedleRegex`, `src/deployment/secret-redaction.ts` around lines 1537-1550), so an embedded 1-3 character secret — resolver-recorded or derived — is never redacted by the scan at all, with or without a sibling. go-to-k/cdkd#2511 (issue go-to-k/cdkd#2485) positions the embedded leaf by its own span but binds that arm to the value scan's own answer, so it leaves the sub-floor leaf exactly where the scan leaves it. The residual is stated in `positionByEmbeddedSpan`'s docstring and pinned by the `leaves an embedded 1-3 character secret to the value scan` case in `tests/unit/deployment/secret-redaction-embedded-span.test.ts` (whose sibling only decides which expression survives the value-keyed map), so closing it is a deliberate change rather than a side effect.

go-to-k/cdkd#2485 named this residual as a disclosure in its Summary, Impact and Severity lines; go-to-k/cdkd#2511 was re-scoped to `Refs` so that this issue carries it.

## Why the arm did not close it

The maintainer review of go-to-k/cdkd#2511 measured that relaxing the fifth condition of the arm to

```ts
const scanned = redactSecretsForState(bag, secrets);
if (scanned !== bag && scanned !== prefix + survivor + suffix) return undefined;
```

[...]

Issue #2134, opened 2026-08-21
fix(resolver): classify a dynamic reference's REGION after intrinsic assembly, not before


go-to-k:
Follow-up to issue #2109 (PR lane `fix/2109-2099-scrub-secret-resolution`).

## What is wrong

The region classification `cdkd scrub` and the rollback replay share
(`classifyReplaySecretRegion`) runs on a COMPLETE `{{resolve:...}}` expression.
scrub applies it as a PRE-PASS over the RAW template leaf, before intrinsic
resolution, and the shared token scan is `\{\{resolve:[^}]+\}\}` -- a class that
cannot cross a `}`. Measured on this branch:

```text
  "{{resolve:secretsmanager:my-secret:SecretString:pw}}"        1 opening, 1 token
  "{{resolve:secretsmanager:${Env}-db:SecretString:password}}"  1 opening, 0 tokens  (Fn::Sub)
  "{{resolve:secretsmanager:"  (one Fn::Join part)              1 opening, 0 tokens
```

With zero tokens there is nothing to classify, so the leaf was returned by
identity and `resolveJoin` / `resolveSub` then called
`resolveDynamicReferences` on the PRIMARY resolver once the parts were joined --
issue #2109 verbatim, silently, with the `ambiguous` refusal never firing.

## What shipped instead, and what it leaves open

PR-side fix (fail-closed, no resolver change): a leaf that OPENS more references
than the scan matched raises `SCRUB_SECRET_REFERENCE_UNCLASSIFIABLE`, but ONLY
when a foreign producer region is on record. With none there is no cross-region
question to get wrong, and refusing there would reject every ordinary `Fn::Sub`
template.

The residual is therefore: **an assembled reference naming a FOREIGN ARN, in a
stack with no cross-stack read on record, is still resolved in the stack own
region** -- the wrong-region needle, so scrub reports the stack clean over
surviving plaintext. The rollback replay has the same shape wherever a persisted
bag can hold an assembled expression.

## The structural fix

Classify AFTER assembly, inside `src/deployment/intrinsic-function-resolver.ts`:
`resolveDynamicReferences` already has the complete expression in hand at the
point it builds its client, which is exactly where the region question can be
answered without any pre-pass. That is a change to a path shared with deploy, so
it needs its own PR and its own review.

Evidence lives in `tests/unit/cli/commands/scrub-cross-region-secret.test.ts`
[...]

Issue #1901, opened 2026-08-14
fix(state): redact SecureString SSM parameters resolved via {{resolve:ssm:...}} (GHSA follow-up)


go-to-k:
## Context

Follow-up to the dynamic-reference secret-redaction fix (GHSA-p5qg-v9gv-hc7w, PR #1899), which redacts `{{resolve:secretsmanager:...}}` plaintext out of persisted state.

## Gap

A SecureString SSM parameter referenced via the plain `{{resolve:ssm:...}}` form resolves to plaintext (`WithDecryption: true`) and is NOT redacted. This is the same disclosure class as the secretsmanager case, but was out of the reported advisory's scope.

The shipped fix deliberately left it: `resolveSSMReference` returns a plain string, and plain `ssm` is treated as public config that state stores RESOLVED (so a secret-bearing resource is not a perpetual UPDATE and diff makes no secret fetch). There is an in-code NOTE at the `return paramValue` site in `src/deployment/intrinsic-function-resolver.ts` and a "Known limitation" note in `docs/cli-reference.md`.

## Why it is not a one-line change

Redacting SecureString correctly requires the diff / no-op path to SKIP it too (like the secretsmanager `skipDynamicReferences` handling), which in turn requires knowing the parameter TYPE (`String` / `StringList` / `SecureString`) WITHOUT leaking the fetched value. Plain `String` / `StringList` must stay resolved in state; only `SecureString` should persist the expression. That is a design change to the resolver's ssm path, not a redaction-map tweak.

## Verification

Needs a real-AWS integ arm with a `SecureString` parameter: deploy a resource referencing it, assert state holds the expression (not the decrypted value), assert diff makes no spurious UPDATE, and assert a plain `String` parameter still resolves in state.

Session-fit: next (not this session) -- resolver ssm-path design change + a new real-AWS integ (SecureString + String parameters); independent of the shipped secretsmanager choke-point fix / Effort: ~1-3 h (design + one integ run + a review round)

go-to-k:
Working on this in branch `fix/1901-securestring-ssm-redaction` (worktree `.claude/worktrees/fix-1901-securestring-ssm-redaction`) — touching `src/deployment/intrinsic-function-resolver.ts`, `src/deployment/deploy-engine.ts` (dynamic-reference skip set), `docs/cli-reference.md`, plus unit tests and a real-AWS integ arm with a SecureString + String parameter pair.

[...]

PR #2562, opened 2026-09-04
fix(scrub): resolve each intrinsic Export.Name through a view of the outputs pass map that keeps the resolved pairs to itself

### LINKED ISSUE #2531: fix(scrub): an Export.Name resolving a moved unclassifiable ssm reference conflicts the pair a literal Output embedding it needs
## Summary

`cdkd scrub` can lose the span positioning of a literal `Output` value that embeds a `{{resolve:ssm:...}}` reference, when that reference is ALSO resolved by an `Export.Name` in the same run and its value moved between the two resolutions. The outputs pass then persists the value scan's answer for that leaf instead of the leaf's own token — for a plaintext of four or more characters, one of this run's expressions (never plaintext); a 1-3 character plaintext is below the scan's needle floor and is left in plaintext with or without this hazard, which is go-to-k/cdkd#2516's subject, not this one's. Found by the maintainer's round-2 review of go-to-k/cdkd#2511 as the scrub-side twin of a hazard that PR closed on the deploy side.

## Mechanism (branch of go-to-k/cdkd#2511, `src/cli/commands/scrub.ts`)

- `outputSecrets` (declared around line 2883) is ONE `RecordedSecretValues` map for the whole outputs pass. The `Export.Name` loop (around line 3188) resolves each declared name with `recordedSecretValues: outputSecrets` (around lines 3203 and 3214), and it runs FIRST; the value loop (around line 3302) resolves each `Output.Value` with the same map (around lines 3346-3349).
- Sharing the map is deliberate for the ENTRIES: the comment at lines 3206-3212 gives the reason (a reference first resolved in the name loop warms the resolver's cache, and the comment says the hit arm re-records only what it can still prove is secret, so an unpinned `ssm` reference recorded into a dropped private map would be invisible to the value loop and `--dry-run --fail` would report CLEAN on a leaking stack). Two things about that reason are for the fix PR to settle rather than this issue: the unclassifiable-`Type` reference this issue is about is never cached at all (`cacheable = false`, so the value loop re-asks AWS and records its own entry at the fresh-resolution seam, `intrinsic-function-resolver.ts` around line 7106), and since go-to-k/cdkd#1933 the cache stores the secret verdict alongside the value, so the hit arm re-records a cached secret too. Whatever the shared map still buys, an explicit entries copy keeps it; this issue does not depend on the comment's mechanism being current.
- Since go-to-k/cdkd#2511 the resolver's recording seam also writes the uncollapsed `expression -> plaintext` pair into the same map (`recordResolvedPair`, `src/deployment/secret-redaction.ts`), and a second write of the same expression with a DIFFERENT plaintext marks the pair conflicting — after which `positionByEmbeddedSpan` refuses to position any leaf embedding that expression and falls back to the value scan.
- An `ssm` reference whose `Type` came back unclassifiable is never cached (`cacheable = false`, `src/deployment/intrinsic-function-resolver.ts` around line 7042), so the value loop re-asks AWS. If the parameter was rotated between the name loop's read and the value loop's read — or the name loop's read and the value loop's read simply disagree — the pair is marked conflicting in `outputSecrets`, and a literal `Output.Value` such as `postgres://u:{{resolve:ssm:/p/unclassifiable}}@h/db` is written from the value scan: the map's surviving expression for the plaintext, which for a same-plaintext sibling is the sibling's spelling (the go-to-k/cdkd#2485 defect, re-opened for this one shape).

The deploy engine does not have this hazard: `DeployEngine` resolves each `Export.Name` with a PRIVATE `nameSecrets` map, copies its entries into the pass map in a `finally`, and — since go-to-k/cdkd#2511 — deliberately does NOT merge its resolved pairs, because a name never positions a leaf and the merge could only mark a pair conflicting.

## Impact

Bounded to expression-versus-expression for a plaintext of four or more characters: the leaf falls back to the value scan's answer, which is one of this run's recorded expressions, so nothing scrub exists to remove survives that the scan would have removed. (A 1-3 character plaintext survives the scan regardless — the go-to-k/cdkd#2516 residual — and the conflict changes nothing for it, since the arm refuses a sub-floor middle either way.) The wrong expression is the go-to-k/cdkd#2485 consequence at scrub scale — the next `cdkd deploy` diffs that output against the template's spelling and re-persists it — and it needs an unclassifiable-`Type` `ssm` reference that appears both in an `Export.Name` and embedded in a literal `Output.Value`, with the value moving inside one scrub run. Rare, and a follow-up rather than a blocker.

## Proposed fix

Mirror the deploy engine's shape: resolve each `Export.Name` with a private `nameSecrets` map, copy its ENTRIES into `outputSecrets` on every exit (the `finally` invariant the deploy engine documents — the resolver records and caches as it goes, so a resolution that throws partway has already recorded and, for a cacheable reference, warmed the cache), and never merge its pairs. Every entry the shared map would have held is still an entry of `outputSecrets` when the value loop runs — so whatever the comment at lines 3206-3212 protects is kept by construction, without re-deriving whether its mechanism is current — while the one write that can conflict a pair is withheld. The name loop still runs first; only the map the resolver records into changes. The fix PR should also re-read that comment against go-to-k/cdkd#1933's verdict-alongside-value cache and correct it if it is stale.

## Verification plan

- Unit: a scrub outputs pass with a literal `Output.Value` embedding an unclassifiable-`Type` `ssm` reference beside a whole-value sibling sharing the plaintext, and an `Export.Name` (`Fn::Sub`) resolving the same reference, with the mock resolver returning a moved value on the name's resolution; the value leaf must persist its own expression (today it persists the sibling's). Same shape with the name's resolution THROWING partway, asserting every entry the name's resolution recorded before the throw is an entry of `outputSecrets` afterwards — a pin of the `finally` entries copy itself, discriminating against a success-path-only copy, not a claim about the cache.
- Unit, unchanged: the existing scrub cases for an unpinned `ssm` reference first seen in an `Export.Name` keep passing (the entries copy).
- Integ: `/run-integ secrets-dynamic-ref` (its scrub phases) after the change; the moved-value case itself is not reproducible against real AWS without a rotation inside the run, so the unit case is the discriminator.

Dup-check: searched open and closed issues for `scrub Export.Name`, `outputSecrets conflict`, `recordResolvedPair scrub`, `CONFLICTING_PLAINTEXT` right before filing -- go-to-k/cdkd#2511 closes the deploy-side twin and names this one in its review; go-to-k/cdkd#2485 is the collapse the value scan falls back to; nothing covers the scrub name loop
Session-fit: next (not this session) -- named by the maintainer as a follow-up to go-to-k/cdkd#2511, and the fix touches `scrub.ts`'s outputs pass rather than the PR's redaction arm
Severity: low -- wrong EXPRESSION for one output shape under a value that moves inside a single scrub run; no plaintext beyond the pre-existing go-to-k/cdkd#2516 residual, and only for an unclassifiable-`Type` `ssm` reference used both as an `Export.Name` and embedded in a literal `Output.Value`
Effort: small (S) -- a private map for the name loop plus an entries copy in a `finally`, mirroring `DeployEngine`, and two unit cases
Estimate: ~1-2 h -- the `finally` invariant and the existing scrub outputs tests decide the shape; the change itself is a dozen lines

nix-tkobayashi:
Closes #2531

## What was wrong

`cdkd scrub`'s outputs pass resolves each intrinsic `Export.Name` FIRST, and did so into the same `outputSecrets` map the value loop fills afterwards. Since go-to-k/cdkd#2511 (issue go-to-k/cdkd#2485) the resolver also records the uncollapsed `expression -> plaintext` pair beside each entry, and a second write of one expression with a DIFFERENT plaintext marks the pair conflicting — after which `positionByEmbeddedSpan` refuses every literal leaf embedding that token. An `ssm` reference whose `Type` came back unclassifiable is never cached (`cacheable = false`), so a name and a value resolving it in one run each ask AWS; a value that moved in between conflicted the pair through the shared map, and a literal `Output` embedding the token was written from the value scan — the sibling's spelling, the go-to-k/cdkd#2485 class re-opened for this one shape. Bounded to expression-versus-expression for a plaintext of four or more characters; found by the maintainer's round-2 review of go-to-k/cdkd#2511.

## What changes

`src/cli/commands/scrub.ts`, the `Export.Name` loop:

- **`SharedEntriesSecrets`** (exported) — a `Map` subclass with its own identity whose every read and write (`size`, `get`, `has`, `set`, `delete`, `clear`, `keys`, `values`, `entries`, `forEach`, the iterator) goes to `outputSecrets`. The name is resolved through it (`pinCrossRegionSecrets`, `resolverContext`, `resolveCrossStackReads`, `resolver.resolve`), so the ENTRIES stay shared live in both directions while the resolved PAIRS — recorded by `recordResolvedPair` against the instance — hang off the view and die with it. A name never positions a leaf; `mergeResolvedPairs` is not called. Two residuals of discarding what the view recorded, both bounded to the choice of expression and stated in the class doc: a value leaf keeps its own positioning only when its own read through the primary resolver succeeds (otherwise it drops to the value scan, where the name-recorded entry still scrubs it — whole at any length when the leaf IS the plaintext, by the substring arm at `MIN_NEEDLE_LENGTH` or longer when embedded, a shorter embedded one being go-to-k/cdkd#2516; a failed read through the cross-region pin is a refusal, not a fallback), and the cross-stack associations the pre-pass records die with the view (recorded only for a plaintext already in the map, so the loss degrades the expression, never the redaction).
- **Why a view and not a private map with a copy** (the deploy engine's `finally` shape): the resolver records as it goes and `Fn::Join` resolves its parts concurrently, so a part that rejects while a secret part is still pending ends the block before that part records. An end-of-block copy misses the late entry the shared map used to keep; a private map seeded from the pass map is a snapshot that a LATER name's cross-region pin and cross-stack pre-pass — which mask their own messages against the map they are handed — would not see the late entry through either. Both shapes were tried and are the mutants the late-write case kills. The view lets a late write land whenever it happens but cannot order it before the next consumer; the honest fix for that is a resolver-level drain of concurrent parts, filed as go-to-k/cdkd#2563 together with the deploy engine's own copy-loss (its `finally` shape has the same hole — its comment now names go-to-k/cdkd#2563 as that residual).
- The swallowed resolver error is warned about on a boolean sentinel rather than `nameError !== undefined` (`undefined` is a legal thrown value; the old catch warned for it too). A region-ambiguous refusal still rethrows; the pin's and the pre-pass's own throws still propagate.
- The comments that grounded the sharing on "the cache-hit arm re-records only what it can still prove is secret" predated go-to-k/cdkd#1933's verdict-alongside-value cache (the hit arm re-records a cached secret too, and the never-cached reference re-asks AWS regardless). Rewritten in scrub, in the deploy engine's sibling block above its own `nameSecrets` and in its `finally` (comment only — `lambda` re-run for the `integ-broad` marker), in the scrub test fixture, and in `.claude/rules/layout-deployment.md` (which now says INTRINSIC `Export.Name` and names go-to-k/cdkd#2563); `mergeResolvedPairs`'s docstring and the changelog entry (under 2026-09-07, within the 2000-character cap) name the deploy engine explicitly where they describe its copy.

## Tests

`tests/unit/cli/commands/scrub-export-name-collision.test.ts` — the mock resolver gained a "moving" reference (first resolution returns one plaintext, later ones another; records the entry AND `recordResolvedPair`, as the real seam does, through one function shared by the primary resolver's `resolve` and the pinned sibling's `resolveDynamicReferences`), top-level `Fn::Join` resolved with `Promise.all` (as the real resolver does) with a `__THROW__` part that throws echoing the parts resolved before it, a `__THROW_UNDEFINED__` part, and a `__LATE_UNPINNED__` part that records only when the resolver's NEXT call begins — a deterministic "after that block, before this one". Six cases:

- **moved value on the name's resolution**: `Dsn: pre-{{resolve:ssm:/p/moving}}-post` beside `Whole: {{resolve:ssm:/p/moving:1}}` (the same-plaintext sibling, resolving last) and an `Export.Name` resolving the same token first with a moved value; movement pinned by the returned values. `Dsn` persists its own token. Fails with the shared map; fails with a pair merge.
- **the same through the cross-region pin** (maintainer's Major 1): the moving parameter spelled as a foreign-region ssm ARN, a `named-region` verdict with no `producerRegions` evidence, so the token never reaches `resolver.resolve` — the pin resolves it through the `eu-west-1` resolver and records into the map it is handed. Premise pinned by the constructing region of both `resolveDynamicReferences` calls and the returned values in order; `Dsn` keeps its own ARN token, and an unaccounted `Stale` key holding the MOVED value (resolved only by the name's pin) is repaired to the ARN expression, so the pin's entry write-through is load-bearing. Fails when the pin is handed `outputSecrets` (on `Dsn`); fails when it is handed a fresh map (on `Stale`); fails with the raw pass map.
- **throw part-way, masked**: `Fn::Join ['', [UNPINNED, __THROW__]]` — the error echoes the plaintext the first part resolved; the warn must carry `***` and not the plaintext, and `Leaky` is scrubbed. Fails with a map that does not write through.
- **late write after the rejection**: `Fn::Join ['', [__LATE_UNPINNED__, __THROW__]]` then a later name; the later name's resolver context must carry the plaintext and `Leaky` is scrubbed. Fails with an end-of-block copy and with a seeded snapshot that forwards writes.
- **seed / view**: a later name is resolved against every entry recorded so far, through every read surface (fails with a fresh private map, and with any one read method delegating to its own storage).
- **`throw undefined`**: a fence for the boolean `nameFailed` sentinel (a truthiness sentinel skips the warn), not a go-to-k/cdkd#2531 repro — it passes on the pre-fix code too.

`tests/unit/cli/commands/scrub-shared-entries-secrets.test.ts` (new) — fences the view directly: every own key of the test runtime's `Map.prototype` (names and symbols) is an own key of the class's prototype, allowing exactly `Symbol.toStringTag` and `constructor` (so a future `getOrInsert` arriving with a Node bump surfaces as a non-empty difference); the view stores nothing of its own (the original `Map.prototype` members applied to the view see empty slots); `set` returns the view; `has` hit and miss; `delete` / `clear` reach the target; `forEach` passes the view and honours `thisArg`; every iteration spelling.

The unpinned reference's record-once modelling is documented as a MODEL (the real resolver never caches it and re-records), which is what makes the name loop's recording observable in these cases.

`tests/unit/scripts/secrets-dynamic-ref-name-fallback-sync.test.ts` (new) — binds the fixture's two absence-grep strings (read off its `for NAME_FALLBACK in ...` line; exactly two, distinct) to the literal RUNS of the scrub source's `logger.warn` calls about an `Export.Name`, read off the TypeScript AST (each string and template piece a run of its own, never joined; receiver exactly `logger` / `<x>.logger`), one call per string with the string inside one run — so a reword that keeps the old wording in a comment of any kind, demotes the warn to `debug`, or splits the phrase across a substitution fails it; a self-probe pins the reader's own clauses. The two warn substrings also carry sentinel comments at their unit assertions.

## Verification

- `vp check --fix` / `vp run check` / `vp run typecheck:test` / `vp run build`: 0 errors; full `vp test run` at the default timeout on the final tree: 913 files, 19,446 passed (1 skipped), no type errors, root-asserted in the PR worktree (rebased onto current `main` first).
- `vp run gen:all-matrices` / `vp run audit:coverage:check`: nothing dirty.
- **Live arm** on `tests/integration/secrets-dynamic-ref/`: no scrub-exercising fixture declared an INTRINSIC `Export.Name`, so its scrub phase never entered the changed arm (a Codex round caught the earlier claim that it did). The fixture declares `FunctionNameExport` with `exportName: Fn.sub('${AWS::StackName}-function-name')` (public value — a secret-bearing name is a different, refused arm). Guard 7b: premise 1, the synthesized `Export.Name` is an `Fn::Sub` whose body carries `${AWS::StackName}`; premise 2 (deploy-side — the only scrub in the fixture is `--dry-run`, which writes no state), the deploy keyed the alias under the resolved name; the assertion, the `scrub --dry-run` output carries neither name-fallback warning (`could not be resolved during scrub` / `did not fully resolve during scrub`). The silent non-string fallback has no observable at default verbosity and stays pinned by the unit suite. `/run-integ secrets-dynamic-ref` on the final fixture: PASS (505 s) — all three guard lines printed, `scrub --dry-run` clean, 4 deleted, 0 errors, 0 orphans, 0 surviving state versions. The moved-value shape itself needs a rotation inside one run and is not reproducible live; the unit cases are the discriminator.
- `/run-integ lambda` (broad; `src/deployment/deploy-engine.ts` touched, comment only): PASS (101 s) — 9 deleted, 0 errors, 0 orphans; event-log versions swept. Round 2 (class-doc comment edit in `scrub.ts`): `secrets-dynamic-ref` PASS (498 s, the three Guard 7b lines printed, 4 deleted, 0 errors, 0 orphans, 0 surviving versions) and `lambda` PASS (88 s, 9 deleted, 0 errors, 0 orphans).
- `/run-integ export` (broad, on the `Export.Name` path): PASS (255 s) — cdkd deployed the fixture's 22 resources, `cdkd export` handed them to CloudFormation, post-export `cdk diff` clean with no REPLACEMENT, CloudFormation stack deleted, no orphans on the post-run scan, state versions swept. A first run the same day failed on an account precondition rather than the code (CloudFormation refused the IMPORT changeset with `Unable to fetch parameters [/cdk-bootstrap/hnb659fds/version]` — the test account's `CDKToolkit` was a 2020 legacy v1 bootstrap); `cdk bootstrap` modernised it and the re-run passed. The ledger row carries the PASS; the earlier FAIL row stays in the branch's history (commit 414d1309).
- Mutation probes, each killed by the case named beside it and the tree restored byte-exact between probes: raw pass map (both moved-value cases), pair merge (moved-value), pin handed `outputSecrets` (pin case), ARN spelling dropped from the mock's moving set (pin case — the movement premise), fresh private map (view case), private map + end-of-block copy (late-write), seeded snapshot that forwards writes (late-write), map that does not write through (throw-part-way), truthiness sentinel (throw-`undefined`), `delete` / `clear` / `forEach` override removed (fence case, plus the delete-clear / forEach cases; for `forEach` also the view case), `super.delete` / `super.has` / `set` returning `this.target` (storage and delete-clear cases), each read method delegating to its own storage (view case). Round 2: pin handed a fresh map (pin case, `Stale`); for the sync test — warn demoted to `debug`, reword with the old text in a standalone or a TRAILING comment, fixture string duplicated, fixture loop line renamed, both strings folded into one warn, the real warn split across a substitution, and on the reader itself the warn-name guard, receiver guard (and its Identifier test), Export.Name filter, plain-string collection, substitution traversal, run joining and a receiver suffix match each dropped. Moving the warn back into the catch is an equivalent mutant (nothing runs between).
- Reviewers: `pr-code-reviewer` (round 0: two nits taken; round 1: one minor — the pairs residual over-generalized "drops to the value scan" to the pin seam, whose failure is a refusal — and two nits, all taken), `pr-test-reviewer` (round 0: the read-surface pins, the order guard and two comment notes; round 1: clean, one note — sentinel comments at the unit assertions on the warn substrings the fixture greps — taken; it re-measured the pin mutant and two more of its own, all killed), four maintainer-checklist proxy passes to NO ISSUES across the two rounds.
- Codex review loop on the delta to NO ISSUES — round 0: 11 rounds, plus 1 on the fixture arm and 2 on the body; round 1: 2 rounds on the delta (bounding "scrubbed either way" to `MIN_NEEDLE_LENGTH`; the sentinel comment over-claimed what the truthiness mutant skips), then the maintainer-checklist proxy pass to NO ISSUES in 3 rounds (movement asserted by value, `set` / `has` fenced, the `Fn::Sub` premise, no value in the diagnostic, "intrinsic" and the arms named in the changelog and rules; the fixture grepped one of the two fallback warnings) plus a fourth after the reviewer fixes.


go-to-k (changes requested):
Reviewed at `ed47df12`. Three reviewers (code / test / security) plus an inline pass. **The implementation is correct** — no defect found in the fix itself, and the reasoning behind the view holds up in source. Requesting changes for the test gap under Major: it leaves a mutant that re-opens this very issue with nothing going red, so it is worth closing before merge rather than after. The four minor items are cheap and in the same files.

Both halves of the obsolescence claim check out against the code, not just the body: the cache-hit arm re-records the entry *and* the pair off `cached.secret` / `cached.value` (`src/deployment/intrinsic-function-resolver.ts:6978-6985`), and the unclassifiable-`Type` `ssm` arm sets `cacheable = false` (`:7042`, gated at `:7103`), so it re-asks AWS and re-records every pass. The #1901 leak class the old comment guarded does not re-open, and the live entry sharing is the load-bearing half either way.

The delegate is complete on the pinned Node 24.19: `Object.getOwnPropertyNames(Map.prototype)` is exactly `clear, constructor, delete, entries, forEach, get, has, keys, set, size, values`, own symbols `Symbol.iterator` / `Symbol.toStringTag`. Every data-bearing member is overridden; `new Map(view)`, spread, `Array.from`, `Object.fromEntries` and `forEach` all route through the overridden iterator. Dropping the pairs and the cross-stack associations cannot leave a plaintext behind — `recordCrossStackExpression` is gated on `recordedSecretValues?.has(reresolved) === true` (`intrinsic-function-resolver.ts:5436`), so the needle is already in `outputSecrets` and only the *choice of expression* degrades to the value scan. `deploy-engine.ts` and `secret-redaction.ts` are comment-only (comment-stripped diffs against `main` are empty). The `nameFailed` refactor is behaviour-preserving: nothing runs between the `catch` and the `if`, the region-ambiguous rethrow still precedes the sentinel, and the `try` scope is byte-identical to `main`.

---

### Major

**1. The `pinCrossRegionSecrets` map argument is unpinned.**
`src/cli/commands/scrub.ts:3305` passes `nameSecrets`, and the pin records pairs through the same seam into `ctx.recordedSecretValues` (`scrub.ts:1207-1223`). No case in `tests/unit/cli/commands/scrub-export-name-collision.test.ts` uses `producerRegions` — zero occurrences in the file — so a mutant that passes `outputSecrets` there re-opens the exact #2531 conflict for a cross-region name with nothing going red. Worth a case that drives a `named-region` verdict through the pin and asserts the leaf still persists its own token.

### Minor

**2. The new class orphans an existing JSDoc.**
On `main` the 34-line block starting `Is this the issue #2134 refusal…` sits directly above `isRegionAmbiguousRefusal`. `SharedEntriesSecrets` is inserted between them, so that doc now ends above the class and `isRegionAmbiguousRefusal` is left undocumented — while the new comment in the name loop tells the reader to "see `isRegionAmbiguousRefusal`". Moving the class below the function restores both.

**3. The integ alias assertion measures the deploy, not the scrub.**
`tests/integration/secrets-dynamic-ref/verify.sh:680-684`. The only scrub in the file is `--dry-run`, which writes no state, so `state.outputs['<stack>-function-name']` is purely the deploy's output; the failure message ("did not survive the scrub pass") claims a scrub-side property the run cannot exercise. The discriminating guard here is the `could not be resolved during scrub` grep — which is good, and does positively bind. Either relabel this one as a deploy premise or run a non-dry-run scrub before it.

**4. The rewritten engine comment now overclaims, and it is the line that should carry the residual.**
`src/deployment/deploy-engine.ts:5877` ends with "The merge keeps this resolution's entries either way." Under #2563's late-write shape the `finally` copy does *not* keep a late entry — which is exactly what motivated the view on the scrub side. Since this PR rewrites that comment, it is the natural place to name #2563 as the residual rather than leave a claim that #2563 contradicts.

**5. #2563 is not named in the changelog entry or in `.claude/rules/layout-deployment.md`.**
The changelog contract asks for issue / PR *plus residual numbers*, and #2563 was filed alongside this PR precisely as the engine-side residual. Related: the `SharedEntriesSecrets` doc asserts "a value leaf re-resolves and records its own", which holds only when the value loop's own read succeeds — worth stating as a residual rather than as the general case.

### Nits

- Nothing fences `Map.prototype`'s member set against the overrides. A future addition (the stage-3 `getOrInsert` / `getOrInsertComputed`) would read and write the permanently empty backing store, and a needle inserted through it would never reach `outputSecrets` — making the warn's mask a no-op for that plaintext. Unreachable on Node 24.19, but nothing catches the bump. A test asserting `getOwnPropertyNames(Map.prototype)` plus symbols is a subset of the class's own prototype keys would hold it. `structuredClone` is the same class and equally unreachable today.
- The `throw undefined` case is a refactor fence rather than a #2531 repro — `main`'s warn sat inside `catch (err)` and already fired for `undefined`, so it passes pre-fix. Fine to keep, just not evidence for this issue.
- `delete` / `clear` have no production reader on the view (the only `delete` on a secrets map is in `drift.ts`), and are untested and unfenced.
- `verify.sh` re-spells `cdk.out/${STACK}.template.json` instead of reusing `SYNTH_TEMPLATE`.

### Coverage notes

The record-once modelling of the unpinned reference is genuinely divergent from the real resolver (`cacheable = false` means it re-asks and re-records), so the `Leaky` / `secretsFound` assertions are non-discriminating on their own — but each case carries a discriminating assertion beside it, and the divergence cannot make a case pass that reality would fail. The `__LATE_UNPINNED__` release is a pure microtask with no timers, so the ordering is deterministic and both the end-of-block-copy and seeded-snapshot mutants really do die. The moved-value case does kill both named mutants: `resolvedPlaintextOf` returns `undefined` for `CONFLICTING_PLAINTEXT`, so a shared map — or `mergeResolvedPairs`, which propagates the conflict — drops the leaf to the value scan and writes the sibling's expression. The comment claiming the pin and the pre-pass mask against exactly that context over-states slightly: for the later name both ran before the release, and only the resolver's context is asserted.

Cross-stack associations dying with the view has no case of its own; worst outcome is a wrong-expression fallback, never a plaintext, so a residual note rather than a test would do.

nix-tkobayashi:
Thanks — all six items taken (Major 1, Minor 2-5), plus every nit and the three coverage notes; nothing deferred. Pushed as two commits on top of a rebase onto current `main` (the second is the late-write comment fix below, caught by a Codex pass on this reply).

### Major 1 — the `pinCrossRegionSecrets` map argument
Taken. New case in `scrub-export-name-collision.test.ts`: the same moving parameter spelled as a foreign-region ssm ARN (`{{resolve:ssm:arn:aws:ssm:eu-west-1:…:parameter/p/moving}}`), which `classifyReplaySecretRegion` verdicts `named-region` with no `producerRegions` evidence at all — so the token never reaches `resolver.resolve`; the pin resolves it through the resolver built for `eu-west-1` and records entry + pair into the map it is handed. The mock gained `resolveDynamicReferences` on that seam (records the entry AND `recordResolvedPair`, as the real one does at your `:7115`), and the moving logic is one function shared by both seams. Premise pinned three ways: both `resolveDynamicReferences` calls came from a resolver constructed with `eu-west-1` (a `local` verdict would have left the token to `resolve`), the counter says two resolutions, and the RETURNED values are asserted in order (`first`, then `settled`) — a mock that never moved would leave nothing to conflict and let a shared map pass. Measured: passing `outputSecrets` to the pin fails exactly this case (the primary-seam case stays green, which is the gap you named); the raw pass map fails both.

### Minor 2 — the orphaned JSDoc
Taken: `SharedEntriesSecrets` now sits below `isRegionAmbiguousRefusal`, so the #2134 block is back on its function. The class is exported for the fence file below.

### Minor 3 — the integ alias assertion
Taken as a relabel: it is now "premise 2, deploy-side", placed before the scrub grep, with the failure text saying the deploy did not key the alias (and printing the value's length rather than the value). A non-dry-run scrub would not have made it a scrub property either — a clean stack writes nothing (`recordsChanged > 0 && !opts.dryRun` gates `saveState`). The remaining assertion is stated as what it pins: the absence of BOTH name-fallback warnings (`could not be resolved` and `did not fully resolve`) on an intrinsic name the premises prove entered the arm; the silent non-string fallback has no observable at default verbosity and stays pinned by the unit suite's non-string case. Premise 1 was tightened from "an object" to "an `Fn::Sub` whose body carries `${AWS::StackName}`", and `SYNTH_TEMPLATE` is reused.

### Minor 4 — the engine comment
Taken: it now says the merge keeps every entry recorded before the block ended and cannot keep one a still-pending `Fn::Join` part records after a sibling rejected, naming #2563 as that residual.

### Minor 5 — #2563 and the doc's over-claim
Taken. #2563 is named in the changelog entry (which moved under a 2026-09-07 heading and was trimmed under the 2000-character cap — it had been sitting at 3.4k under the 2026-09-04 heading, which the cap's date partition would have let through) and in `layout-deployment.md`, both now saying INTRINSIC `Export.Name` (a literal one is never resolved). The class doc states two residuals, both bounded to the choice of expression: the pairs (a value leaf keeps its own positioning only when its own read succeeds; when a read through the PRIMARY resolver fails — that call sits in the value loop's best-effort catch — the leaf drops to the value scan, and the name-recorded entry still scrubs a plaintext of `MIN_NEEDLE_LENGTH` or longer, the #2516 residual below that, name loop or not; a failed read through the cross-region pin is not a fallback but a refusal, `SCRUB_CROSS_REGION_SECRET_UNRESOLVED`), and the cross-stack associations (recorded only for a plaintext already in the map, per the gate at your `:5436`, so losing one degrades the expression, never the redaction; no case of its own, as you suggested).

### Nits
- Runtime fence: new `scrub-shared-entries-secrets.test.ts` asserts every own key of the test runtime's `Map.prototype` (names and symbols) is an own key of the class's prototype, allowing exactly `Symbol.toStringTag` and `constructor` — an exact allowance, so `getOrInsert` arriving with a Node bump surfaces as a non-empty difference. Beside it: the view stores nothing of its own (the original `Map.prototype` members applied to the view see empty slots), `set` returns the view, `has` hit and miss, `delete` / `clear` reach the target, `forEach` passes the view and honours `thisArg`, every iteration spelling. `structuredClone` is named in the class doc as the same unreachable class.
- `throw undefined`: retitled as a fence for the boolean `nameFailed` sentinel, not a #2531 repro, with the comment saying the truthiness mutant skips only the WARN (the non-string check that follows still marks the bag untrusted), so the warn assertion is the discriminating one.
- `delete` / `clear`: pinned in the fence file (`super.delete` and dropping either override both fail it).
- `SYNTH_TEMPLATE`: reused.
- Coverage note on the late-write case: the comment claiming the pin and the pre-pass mask against exactly that context over-stated, as you said — for the later name both ran before the release. It now says the resolver seam is the last of the three readers and the only one asserted; the pin's and the pre-pass's own masking stays pinned by their suites. The cross-stack-association loss has a residual note in the class doc rather than a case, per your suggestion.

### Verification
- `vp check --fix` / `vp run check` / `vp run typecheck:test` / `vp run build`: 0 errors; full `vp test run` at the default timeout on the final tree: 913 files, 19,446 passed (1 skipped), no type errors, root-asserted in the PR worktree.
- `/run-integ secrets-dynamic-ref` (the live arm, re-run after the fixture edit): PASS (505 s) — the three Guard 7b lines printed (`Fn::Sub` premise, deploy-side alias premise, neither fallback warning), `scrub --dry-run` clean, 4 deleted, 0 errors, 0 orphans, 0 surviving state versions.
- `/run-integ lambda` (broad; `src/deployment/deploy-engine.ts` touched, comment only): PASS (101 s) — 9 deleted, 0 errors, 0 orphans; event-log versions swept.
- `/run-integ export` (broad, on the `Export.Name` path): FAIL (81 s) on an account precondition rather than the code — CloudFormation refused the IMPORT changeset with `Unable to fetch parameters [/cdk-bootstrap/hnb659fds/version]`; my test account's `CDKToolkit` is a 2020 legacy v1 bootstrap with no SSM version parameter. cdkd's deploy and the cleanup destroy were clean (22 deleted, 0 errors, 0 orphans, versions swept). Recorded in the ledger as such; your 2026-09-04 PASS stands as that fixture's last green run.
- Mutation probes on the delta, each restored byte-exact between probes: pin handed `outputSecrets` → pin case; raw pass map → both moved-value cases; ARN spelling dropped from the mock's moving set → pin case (premise); `nameError !== undefined` → sentinel case; `delete` / `clear` / `forEach` override removed → fence case plus the delete-clear / forEach cases (and, for `forEach`, the existing view case); `super.delete`, `super.has`, `set` returning `this.target` → the storage / delete-clear cases.
- Reviewers: `pr-code-reviewer` and `pr-test-reviewer` on the delta (the code reviewer found one minor — the pairs residual said "drops to the value scan" of both seams, while a failed read through the pin is a refusal (`SCRUB_CROSS_REGION_SECRET_UNRESOLVED`), now stated — plus two nits taken (the pin case renamed to what its assertions pin; the "masked" parenthetical applies to one of the two warnings); the test reviewer re-measured the pin mutant and two of its own (all killed) and asked for sentinel comments at the unit assertions on the two warn substrings the fixture greps, added).
- Codex review loop on the delta to NO ISSUES (2 rounds — round 1: bound the "scrubbed either way" sentence to `MIN_NEEDLE_LENGTH`, and the sentinel comment over-claimed what the truthiness mutant skips), then the maintainer-checklist proxy pass to NO ISSUES (3 rounds — round 1: movement asserted by value, `set` / `has` fenced, `Fn::Sub` premise, no value in the diagnostic, "intrinsic" and the arms named in the changelog and rules; round 2: the fixture grepped one of the two fallback warnings), and on this reply (2 rounds — round 1: the Minor 5 sentence over-generalized the pin seam, and your late-write coverage note had not been taken).

Nothing further from my side — over to you for the merge.

nix-tkobayashi:
Follow-up on the `export` integ in the reply above: the FAIL was the account, not the code, and it is now green. My test account's `CDKToolkit` was a 2020 legacy v1 bootstrap with no `/cdk-bootstrap/hnb659fds/version` SSM parameter, which is what CloudFormation wanted for the IMPORT changeset; after `cdk bootstrap` modernised it, `/run-integ export` on the same tree passed in 255 s — cdkd deployed the fixture's 22 resources, `cdkd export` handed them to CloudFormation, the post-export `cdk diff` was clean with no REPLACEMENT, the CFn stack was deleted, the post-run scan of Lambda / IAM / VPC / S3 / SNS found no orphans, and 12 state versions were swept. The ledger row now carries the PASS (one ledger-only commit on top of the two above) and the PR body's Verification section says the same. Nothing else changed.

go-to-k:
Re-reviewed at `e8e6029e` (spec / code / test / security, plus an inline pass). **Round 1 closes every item**, and the two follow-up commits close both coverage notes: `fb55166d` corrects the late-write case's reader claim, and `e8e6029e` records the `export` re-run — thanks for chasing the bootstrap down rather than leaving the FAIL row standing. I checked this account independently (`/cdk-bootstrap/hnb659fds/version` present at `32`, `CDKToolkit` updated 2026-04-30), which is why the original attribution looked odd from here; a legacy v1 on your side explains it and the PASS settles it.

Verified rather than taken on faith this round: the pin case really does drive a `named-region` verdict *through* `pinCrossRegionSecrets` (the two `resolveDynamicReferences` calls on the `eu-west-1` resolver are what proves the pin arm ran — a `local` verdict would leave the token to `resolve`), so reverting `scrub.ts:3354` to `outputSecrets` conflicts the pair and turns the `Dsn` assertion red; the `Map.prototype` fence derives its key set at runtime and pins each override's *delegation* with a negative control, not just its presence; the engine comment now matches `resolveJoin`'s actual `Promise.all`; and the `verify.sh` rework asserts the absence of *both* fallback warnings with `SCRUB_OUT` positively bound non-empty and the arm's reachability proved by the premises above it. `src/` is unchanged since the round-1 commit, so all of that still holds at the tip.

One item left, and it is a single sentence.

### Minor — the residual paragraph contradicts this file's own two statements of the same rule

`src/cli/commands/scrub.ts:2335-2337`:

> a plaintext of four or more characters is scrubbed either way (the scan's substring bound, `buildNeedleRegex` — **a shorter one is the #2516 residual, name loop or not**)

The parenthetical is unconditional, and the sub-4 half is not right. `scrub.ts` already states the correct rule twice, in almost the same words both times — `:1762-1764` and `:2748-2749`:

> `allRecordedSecrets` DROPS every needle below `MIN_NEEDLE_LENGTH`, **while the whole-value arm of `maskSecretsInText` matches at any length**

That carve-out applies here. `MIN_NEEDLE_LENGTH` bounds the substring arm only (`secret-redaction.ts:1550`, `.filter((v) => v.length >= MIN_NEEDLE_LENGTH)` inside `buildNeedleRegex`); `wholeValueExpr` is `s === '' ? undefined : secrets.get(s)` with no length bound, under a comment that says so outright — "Even below the needle threshold, a NON-EMPTY whole-value match must still be redacted" (`secret-redaction.ts:4600-4606`). And the path this paragraph is about persists through `redactSecretsForState(state.outputs, outputSecrets, …)` (`scrub.ts:3679`) — the raw pass map, not the sub-4-filtered `allRecordedSecrets` — so the whole-value arm is fully live on it.

So a 1-3 character plaintext that is the ENTIRE leaf is still redacted here; #2516 is the embedded case only. Worth fixing because this paragraph's job is to delimit a redaction gap, and a reader who takes it at face value will believe a covered case is uncovered. Something like: the substring arm loses a sub-`MIN_NEEDLE_LENGTH` plaintext embedded in a larger leaf (#2516), while a leaf that IS the plaintext is still redacted whole at any length.

### Nits, none blocking

- The pin's *entries* write-through direction is unpinned. `not.toContain(MOVING_FIRST)` is vacuous — no fixture state key holds that value — so a mutant handing the pin a fresh `new Map()` while the resolver keeps the view stays green. An extra state output holding `MOVING_FIRST` would make the pin's write-through load-bearing.
- The two absence greps in `verify.sh` are bound to the `scrub.ts` warn strings by comment only; the unit test documents the coupling but nothing enforces it, so a reword makes the greps silently green.
- Pre-existing, outside this diff: `handleOutputResolutionFailure` (`src/deployment/deploy-engine.ts:6347`) logs `String(error)` unmasked, and the `--strict-getatt` arm inlines `error.message`. Nothing this PR makes worse — #2563 does not change it either, since no masking is applied there at all — but probably worth its own issue.

CI is green at `e8e6029e` and the tree is otherwise ready; fix the one sentence and I will approve.

nix-tkobayashi:
Thanks — the one sentence is fixed, and all three nits are taken as well (the third as its own issue). Pushed as one commit on top of `e8e6029e`; `src/` changes are the class doc comment only.

### Minor — the residual paragraph
Taken, as you spelled it: the sentence now says the plaintext is scrubbed either way — whole, at any length, when the leaf IS the plaintext (`redactSecretsForState`'s whole-value arm is unbounded and the positioned pass hands it the raw pass map), and by the substring arm at `MIN_NEEDLE_LENGTH` or longer when the plaintext is EMBEDDED in a larger leaf; a shorter embedded one is the go-to-k/cdkd#2516 residual, which has nothing to do with the name loop. The code reviewer traced the described leaf (a declared output or its literal alias, so accounted) to the POSITIONED pass — `positionByEmbeddedSpan` refuses for want of a pair, the value scan runs against the raw `outputSecrets`, and the whole-value arm sees a sub-4 plaintext there; the widened pass, which does drop sub-4 needles through `allRecordedSecrets`, is not on that leaf's path.

### Nits
- **The pin's entries write-through**: the pin case's state now carries `Stale: MOVING_FIRST` — a key today's template cannot account for, holding the MOVED value only the name's pin ever resolved — asserted to be repaired to the ARN expression. Measured: handing the pin a fresh `new Map()` while the resolver keeps the view fails the case on that assertion alone (the round-1 mutant, the pin handed the pass map, still fails on `Dsn`; the two are complementary). The comment credits the value scan of the positioned pass first, with the widened pass re-deriving it — both read the pass map, so a fresh map fails both.
- **The absence greps, bound by comment only**: new `tests/unit/scripts/secrets-dynamic-ref-name-fallback-sync.test.ts`. It reads the fixture's `for NAME_FALLBACK in "..." "..."` line (exactly two DISTINCT strings, or the floor fails) and the scrub source through the TypeScript AST, collecting the string-literal and template pieces of every `logger.warn(...)` call's arguments as separate RUNS (receiver exactly `logger` or `<x>.logger`), keeping the calls one of whose runs names an `Export.Name of output`; each fixture string must sit inside one run of its own such call, one call per string — runs are never joined, so a phrase that only exists across a `${name}` boundary (which the emitted line interrupts) does not count. Comments are trivia the AST never carries into a literal, so a reword that leaves the old wording in a comment — standalone, trailing after `+`, or block — cannot satisfy it, and neither can a `debug` call (which the default-verbosity run never prints). Measured kills: warn demoted to `debug`; reword with the old text in a standalone comment; reword with it in a TRAILING comment; the real warn split across a substitution; the fixture repeating one string; the fixture loop line renamed; both strings folded into one warn. The reader's own clauses are pinned by a self-probe (a `logger.debug` with matching wording, `other.warn`, `otherlogger.warn`, `this.otherlogger.warn`, an unrelated `logger.warn`, and a `this.logger.warn` mixing plain strings, a template and a literal inside a substitution, asserting the exact collected text): dropping the warn-name guard, the receiver guard (or its Identifier test — a `'logger'.warn` control), the Export.Name filter, plain-string collection, substitution traversal, joining the runs, or widening the receiver back to a suffix match each fails it.
- **`handleOutputResolutionFailure` unmasked**: filed as go-to-k/cdkd#2728, with what the real resolver's errors actually carry (`resolveJoin` propagates a part's error unchanged — the plaintext-echo is the scrub mock's shape; `resolveDynamicReferences`' own messages echo the token, secret id / JSON key, SSM parameter name, or a propagated SDK error, so the exposure is a substituted value landing in one of those), the map-timing bound (entries recorded before the handler runs — go-to-k/cdkd#2563's late write is outside it), the fix (mask both arms against the outputs pass map, `cause` kept), and a deterministic real-resolver reachability case (a JSON key assembled from the resolved secret against the existing JSON secret, so the resolver's own `key '…' not found` message necessarily carries the plaintext — an SSM not-found is not usable, `ParameterNotFound` propagates unchanged). Labels do not stick from a fork, so please add `severity:medium` / `effort:medium` if you agree with the body's lines.

### Verification
- `vp check --fix` / `vp run check` / `vp run typecheck:test` / `vp run build`: 0 errors; full `vp test run` at the default timeout on the final tree: 914 files, 19,453 passed (1 skipped), no type errors, root-asserted in the PR worktree.
- `/run-integ secrets-dynamic-ref` (the fixture the sync test binds; `scrub.ts` touched, comment only): PASS (498 s) — the three Guard 7b lines printed, `scrub --dry-run` clean, 4 deleted, 0 errors, 0 orphans, 0 surviving state versions.
- `/run-integ lambda` (broad): PASS (88 s) — 9 deleted, 0 errors, 0 orphans; event-log versions swept.
- Mutation probes on the delta, each restored byte-exact between probes: the sixteen above, plus a first cut of the sync test whose JSDoc quoted a literal block-comment terminator and was silently not collected — caught by the baseline case count, and every probe re-measured with the file collected.
- Reviewers: `pr-code-reviewer` (the residual sentence verified against the positioned pass; one nit — the `Stale` comment credited the widened pass; taken) and `pr-test-reviewer` (the `Stale` discriminator confirmed sole-source; the same comment nit; taken).
- Codex review loop on the delta to NO ISSUES (1 round), then the maintainer-checklist proxy pass to NO ISSUES (8 rounds — inline comments not stripped; the reader's own clauses unpinned; a receiver suffix match; the Identifier test; runs joined across a substitution — each closed with a measured control), on the issue body (3 rounds), and on this reply (2 rounds — round 1: a probe count).

Nothing further from my side — over to you for the approve.

go-to-k (approved):
Approving at `c7720314`. The sentence is right, and the two nits you took on top of it are real improvements rather than paperwork.

Verified this round rather than taken from the reply:

- **The residual paragraph, clause by clause.** The outputs pass does hand `redactSecretsForState` the raw `outputSecrets` (`scrub.ts:3546-3552`) — `allRecordedSecrets` is only reached by the widened pass and by `maskSecretsInError`. `wholeValueExpr` is unbounded (`secret-redaction.ts:4604`). And the widened pass cannot clobber the positioned pass's sub-4 whole repair: for an accounted key it `continue`s, and for an unaccounted one it writes only when `next !== value` on the STORED value. So the new sentence is exact in both directions, and it now agrees with what `scrub.ts:1762-1764` and `:2748-2749` already said.
- **`src/` is comment-only.** Transpiled both revisions with `removeComments` — byte-identical.
- **`Stale: MOVING_FIRST` is genuinely load-bearing.** `makeState` has `resources: {}`, so the union is `outputSecrets` alone; a pin handed a fresh map leaves that plaintext with no needle and fails both the new assertion and the `not.toContain` that was vacuous before this round. Your comment's own admission that either pass alone repairs it is accurate.
- **The sync fence works.** I had it re-implemented and measured against the real files rather than reading the claim: 8 `logger.warn` calls collected, 2 carrying `Export.Name of output `, owners `[0,1]` — so the floor sits over a non-empty real-file collection, not the vacuous parse this repo keeps shipping. All seven claimed kills reproduced, including the direction I raised (reword `scrub.ts` and the fixture's greps go silently green), which now fails loudly and names the string and the file. `typescript-v6` matches the convention of the 18 other importers, and the parse costs 21 ms once at module eval.

Nothing blocking is left. Everything I still have is on the fence you added voluntarily to close a nit, and I have filed it as #2732 rather than spend another round there — the main item is that the fence is one-sided (it asserts both fixture strings are owned, but never that every literal-bearing name-fallback warn IS grepped, so a third arm would go unwatched with the fence green), plus a few nits and one asymmetry in `scrub-export-name-collision.test.ts:591`, where the twin non-pin case still has the vacuous `not.toContain(MOVING_FIRST)` the `Stale` key fixed on the pin side.

The reason those go to an issue instead of another round: this would be the third consecutive round finding items inside the previous round's fix, and the fix itself has been clean since round 1. Past that point iterating on test scaffolding costs more than it buys.

#2728 is a good catch and a well-scoped body — I have added `severity:medium` / `effort:medium` as you asked. Its premise checks out: `resolveJoin` resolves parts with a bare `Promise.all` and no `catch`, so a part's error does propagate unchanged, which is also why the plaintext echo in the scrub unit mock is a deliberate stress shape rather than the resolver's own behaviour.

Thanks for the thorough work on this one — the mutation probes and the measured kills in each reply made the review much faster than it would otherwise have been.

go-to-k:
Merged as a7e1df92 — thank you for this one. Three review rounds, every finding taken with a measured mutation probe beside it, and you filed the two residuals (#2563, #2728) yourself rather than leaving them for me to find. The class doc that came out of it is now the clearest statement in the file of what the pairs residual actually costs. Much appreciated.

Issue #2624, opened 2026-09-05
fix(state): four more places where cdkd reports data removed while AWS keeps a readable copy - scrub's rewrite, deploy's implicit scrub, the deployments/ purge, and EFS automatic backups


go-to-k:
Dup-check: searched open and closed issues for `scrub noncurrent`, `saveState noncurrent`, `deployments prefix noncurrent versions`, `EFS automatic backup recovery point`, `state.json noncurrent purge` and `version-blind supersede`. Nearest neighbours, none covering these: go-to-k/cdkd#2346 is the umbrella for the seven state-bucket DELETE sites and its population is derived from `grep -rn 'new DeleteObjectCommand\|new DeleteObjectsCommand' src/`, so a PUT that supersedes and a delete routed through `deleteRawObjects` are both outside it by construction; go-to-k/cdkd#2107 is integ-fixture sweep hygiene; go-to-k/cdkd#2274 is the `attributes` plaintext channel. MISS.

Session-fit: next (not this session) -- found by the sibling-shape sweep of the go-to-k/cdkd#2447 lane, which fixed the S3-replication half of this shape on the six purging paths. Two of the four sites need a maintainer decision that is the same unsettled recovery-vs-disclosure trade-off blocking go-to-k/cdkd#2346 sites 1-3, and the other two are in subsystems (`deployments/**`, EFS) that lane did not touch. Next session starts with `gh issue view <this>` and the greps pasted below.
Severity: high -- `cdkd scrub` exists to make a plaintext secret unreadable and reports `The plaintext is no longer stored there`, while the pre-scrub body stays readable with `GetObject --version-id` to anyone who can read the state bucket. That is the exact class go-to-k/cdkd#2340 and go-to-k/cdkd#2447 were filed for, on the command whose whole purpose is the removal.
Effort: large (L) -- four sites in three subsystems, two of them blocked on a decision rather than on an edit, and the honest minimum (correcting what the messages claim) is separable from the full remedy (purging).
Estimate: ~1 h for the message/doc corrections at all four sites; unbounded for the purge decisions at sites 1 and 2, since what costs is the recovery-vs-disclosure call, not the code.

---

## Summary

The go-to-k/cdkd#2447 lane fixed one instance of a shape: **cdkd tells the user data is gone while an AWS feature silently keeps a readable copy.** A sweep for siblings found four more, none documented anywhere in `docs/**`.

All four were verified against `main` by grep, not from memory. The counts are pasted below rather than asserted.

[...]

Issue #2346, opened 2026-08-27
fix(state): seven state-bucket delete sites leave the object body readable as a noncurrent version, one of them with measured plaintext credentials


go-to-k:
Dup-check: searched open issues for `noncurrent`, `delete marker`, `version purge`, `versioned state bucket`, `DeleteObjectVersion` and `rollback journal plaintext`, plus a body-window scan for the same terms. Three neighbours, none covering this root cause: go-to-k/cdkd#2107 is about integ FIXTURES not sweeping object versions (test hygiene, and its own body frames go-to-k/cdkd#2096 as having closed the disclosure half for fixtures), go-to-k/cdkd#2252 is the scrub-side remediation gap for nested-stack CHILD state records, and go-to-k/cdkd#2274 is the `state.json` `attributes` channel. This is cdkd's own PRODUCT delete paths leaving a readable body behind after the user runs a command that reports success -- a different root cause from all three.

[...]

PR #2625, opened 2026-09-05
fix(state): warn when S3 replication kept the bodies a noncurrent-version purge just removed

### LINKED ISSUE #2447: fix(state): the noncurrent-version purge is invisible to S3 replication, so a purged secret survives in the replica
## Summary

`purgeNoncurrentKeyVersions` (`src/state/s3-noncurrent-version-purge.ts`) is the mechanism that makes cdkd's "the object is deleted" claim true on the versioned state bucket: a bare `DeleteObject` leaves every prior body readable through `GetObject` with a `VersionId`, so five call sites now purge the noncurrent versions -- the rollback journal (measured holding a literal `"MasterUserPassword"`), the custom-resource response sidecar (the handler's full cfn-response including `Data`), the transient CloudFormation template, the bootstrap marker, and `lock.json`.

**S3 replication does not replicate version DELETES.** Replication propagates PUTs and, optionally via `DeleteMarkerReplication`, delete markers -- but a `DeleteObjects` call naming a specific `VersionId` is not replicated. That is deliberate on AWS's side, so that a delete on the source cannot destroy data on the destination.

So on a state bucket configured with Cross-Region Replication (CRR) or Same-Region Replication (SRR), the purge removes the secret-bearing bodies from the SOURCE bucket and they survive indefinitely in the REPLICA. cdkd reports success and warns about nothing, because from its point of view the purge succeeded -- which it did, on the only bucket cdkd knows about.

## Why this is worth recording rather than shrugging at

The whole point of the purge mechanism is that a delete-only cleanup "reports success while the secret stays retrievable". Replication reproduces exactly that shape one bucket over, and it is invisible from every signal cdkd emits. A user who enabled replication for durability has, without being told, opted out of the disclosure fix while keeping the reassuring output.

## What this is NOT

- Not a defect in the purge: nothing cdkd can send would delete the replica's versions, short of cdkd discovering the replication configuration and issuing cross-bucket deletes, which is a much larger and more dangerous capability.
- Not specific to any one call site. It is a property of all five, and equally of any future one.
- Not introduced by go-to-k/cdkd#2346 site 5 -- it has been true since the mechanism shipped in go-to-k/cdkd#2340. Found while reviewing the site 5 lane.

## Options, in rough order of cost

1. **Say it.** A paragraph in `docs/state-management.md` beside the existing "Without the two grants, nothing fails" section, and in the purge module's JSDoc. Cheapest, and it converts a silent hole into a known one.
2. **Detect and warn.** `GetBucketReplication` on the state bucket once per run; if a rule is enabled, warn that purged versions survive in the destination. Costs one API call and one more IAM action (`s3:GetReplicationConfiguration`), on a mechanism whose whole design principle is that missing permissions degrade to a warning rather than a failure -- so the detection itself has to degrade quietly, which weakens it.
3. **Nothing, recorded as a decision.** Defensible if the maintainer judges a replicated cdkd state bucket to be outside the threat model; it should still be written down rather than left implied.

## Verification

No probe was run -- this is a claim about S3 behaviour, not about cdkd code. Confirm against the S3 replication documentation ("What's replicated" / "What isn't replicated") before acting, and note that the answer differs for delete markers (replicable, opt-in) versus version-id deletes (never replicated).

Dup-check: searched open and closed issues for replication / CRR / replica / noncurrent version purge -- MISS. go-to-k/cdkd#2346 covers the delete sites and go-to-k/cdkd#2353 the sidecar lifecycle question; neither mentions replication.

Session-fit: next (not this session) -- it is a property of the whole five-caller mechanism rather than of the site 5 lane, and option 2 needs a maintainer decision about a new IAM action on a path designed to need none. The next session verifies it against the S3 replication documentation, not with a command.
Severity: medium -- a secret cdkd reports as removed stays readable in the replica, but only for users who configured replication on the state bucket, and the primary copy IS purged.
Effort: small (S) -- edit plus unit tests if a warning ships; no new fixture either way.
Estimate: ~1 h -- most of it is deciding what to say and where, not writing it.

go-to-k:
## Summary

Closes #2447.

`purgeNoncurrentKeyVersions` is what makes "cdkd deleted the object" mean the body is unreadable: `cdkd bootstrap` turns versioning on for the state bucket, so a plain `DeleteObject` leaves every prior body readable through `GetObject` with a `VersionId`. The purge closes that by deleting specific `VersionId`s.

**S3 never replicates a version-id delete.** PUTs replicate; delete MARKERS replicate when `DeleteMarkerReplication` is enabled; a delete naming a version id is deliberately excluded so a delete on the source cannot destroy data on the destination. So on a state bucket with Cross-Region or Same-Region Replication the purge is SOURCE-ONLY: the replica keeps its own copy of every version and the secret stays retrievable there — while cdkd reports the purge complete, because on the only bucket it knows about, it was.

Before this PR, `grep -n replicat` returned zero hits in both `src/state/s3-noncurrent-version-purge.ts` and `docs/state-management.md`. The user was not merely unprotected; they were REASSURED.

## What I chose to build, and what I did not

The issue offered three options. This implements **1 (say it) AND 2 (detect and warn)**.

**Option 1 — documentation.** A new section in `docs/state-management.md` beside the least-privilege policy and the five-object purge table, a `## What this CANNOT reach` section on the purge module's own JSDoc, and a new `.claude/rules/state-version-purge.md` satellite so the next reader of the CODE learns it too. The three other user-facing pages that describe the purge — `docs/cli-gc.md`, `docs/import.md`, `docs/cli-force-unlock.md` — link to it.

**Option 2 — detect and warn.** A purge that removed a BODY — or that could not settle whether there was one — ends with one `GetBucketReplication` on the state bucket, and warns, through the CALLER's own sink, naming the destination, the mechanism, and the remedy.

**Not built, deliberately:**

- **cdkd does not delete the replica's copies.** That would mean discovering a replication configuration and issuing deletes into a bucket cdkd was never pointed at — a far larger and more dangerous capability than cleaning up after itself. The warning says so (`cdkd cannot delete them`) so the user knows the remedy is theirs.
- **No new integ fixture.** The live proof is in the Test plan; a permanent fixture would need a second bucket, a replication IAM role and a replication-lag wait on every run.
- **No CLI flag to silence the warning.** The dedupe below reduces it to one line per (bucket, object kind, destination set) per run, which is the information content; a new user-facing flag is not justified here.

## Design notes

Six decisions are load-bearing. Most are CORRECTIONS the review rounds forced, and each is recorded in `.claude/rules/state-version-purge.md` with the failure it closes.

1. **Scoped to keys a BODY was really removed for — decided at LISTING time.** The first cut probed every requested key, so a routine green deploy announced that the rollback journal's copies "survive and remain readable" in the replica, about an object that had never existed (`deleteRollbackJournal` runs on every success). Scoping it to "we purged something" was not enough either: on a versioned bucket that delete writes a MARKER even when the key is absent, so the next deploy found the previous marker noncurrent, removed it, and said the same false thing one deploy later. Only entries from `resp.Versions` count. Keys whose listing FAILED or was TRUNCATED are included as unsettled, because there the provenance is genuinely unknown and over-warning is the safe direction.
2. **The probe is cached, the warning is deduped, on different keys.** The cache follows `src/provisioning/create-only-properties.ts` (in-flight promise stored, definitive answers kept, transient failures evicted) and is keyed on the bucket AND the asserted `ExpectedBucketOwner` — a mismatched-owner 403 cached under the bare bucket name silenced every correctly-scoped caller. The warning dedupes on `(bucket, object description, sorted destinations)`: the bucket alone lets `lock-manager.ts`'s `debug`-demoted release line eat the journal's warning, and without the destinations a `cdkd/dev/…` purge burns the slot so a later `cdkd/prod/…` purge never names the SECOND replica holding it. The slot is claimed only after a successful emit, so a throwing sink cannot lose it, and the destination list is sorted once so the warn line and the repeat line name the same set in the same order.
3. **Two arms are ANSWERS, not failures, and both are silent.** `ReplicationConfigurationNotFoundError` is S3's ordinary reply for an unreplicated bucket, delivered as an ERROR rather than an empty body. `AccessDenied` on `s3:GetReplicationConfiguration` is `debug`-only — most buckets are not replicated, so warning there would demand a permission from everyone to inform almost nobody. Both are pinned by CACHING rather than by silence; asserting silence alone left both green under a mutation that removed the arm.
4. **A credential-shaped 403 is transient**, checked before the blanket 403 rule, and classification reads BOTH `name` and `Code` with no precedence — either field alone could mask the other, in opposite directions.
5. **Coverage errs toward warning.** A tag-filtered rule reduces to a whole-bucket prefix; a rule with no `Status` counts as enabled; and a `Disabled` rule is kept AND flagged, because disabling stops FUTURE replication without removing what it already copied.
6. **No rejection ever reaches the cache.** A rejected promise there would be re-thrown to every later caller and swallowed by the outer catch — a silently dead detector.

`s3:GetReplicationConfiguration` joins the recommended bucket policy as its one **diagnostic** entry: remove it and everything still works, you simply stop being told.

## Test plan

`vp run check`, `vp run typecheck:test`, `vp run build` and the full unit suite pass; CI is green.

**Every branch this PR adds was mutation-probed, one mutation per probe, tree restored byte-exact between them.** The probes that mattered are the ones that came back GREEN, because each produced a change rather than a note:

- `ReplicationConfigurationNotFoundError` and an absent `ReplicationConfiguration` not special-cased — green, because the tests asserted only SILENCE and an unhandled error is silent too. Both now assert CACHING as the discriminator.
- Four provenance and display behaviours added in later rounds (the truncated-page unsettled arm, the absent-`IsLatest` body arm and its marker control, the no-`VersionId` arm, the unsorted repeat line, the unsanitized `--bucket` in the truncation tail) — each green when added, each now fenced.
- Two gates added at the `DeleteObjects` failure arms — green because a body-bearing key enters `purged` at LISTING time, so a failing delete can neither un-purge it nor add a marker. They were DELETED as dead code, the invariant stated in their place, and a positive case added so that moving `purged.add` into the delete-success path is no longer a green mutation.

**Live-tested against real AWS** (`us-east-1`), through the built `dist/cli.js`, both polarities, re-run after each code round, all resources torn down and verified gone (`list-buckets` / `list-roles` filtered on the probe prefix return `[]`):

1. *Replicated bucket.* A versioned source and destination bucket plus a replication IAM role, rule scoped to `Filter.Prefix: cdkd/`. Seeded two versions of a `rollback-journal.json` carrying a literal `LIVE-PROBE-SECRET-V2`, waited for both to replicate, then ran `node dist/cli.js state destroy <stack> --state-bucket <src> --yes`. The run printed the warning naming the real destination bucket; the SOURCE was left with only its delete marker (0 versions); the DESTINATION kept both versions; and `get-object --version-id` against the replica returned `LIVE-PROBE-SECRET-V2`.
2. *Unreplicated bucket.* Same flow on a plain versioned bucket: the purge left `versions: 0, markers: 1`, and no replication warning.
3. *Error shape.* `GetBucketReplication` against the real state bucket returned `name=ReplicationConfigurationNotFoundError Code=ReplicationConfigurationNotFoundError status=404`, confirming the arm the code keys on.

The `>3`-destination truncation tail and the sanitized `--bucket` are unit-covered only; the single-destination warn line is what the live runs exercised.

No integ gate applies: the diff touches `src/state/**`, docs, `.claude/rules/**` and `tests/unit/**` only — none of the `integ-destroy`, `integ-broad`, `integ-local` or `integ-schema-migration` scopes.

## Notes for the reviewer

- **Ten sibling test suites are in the diff for a mechanical reason.** Five hand-write a `vi.mock('@aws-sdk/client-s3')` factory listing every command the purge sends — a missing export throws inside the probe's own try and is swallowed, so those suites would have stayed green while the transient-CFn-template call site had zero coverage, the same false-green trap `upload-cfn-template.test.ts`'s own header documents for go-to-k/cdkd#2346. Three update pinned command sequences. Two gain only a comment and a `clearReplicationProbeCache()`, so their command streams stop depending on which test ran first.
- **`CORPUS_BYTES_MIN` / `CORPUS_BYTES_MAX` in `tests/unit/scripts/rule-file-payload.test.ts` were re-derived**, upward, with the reason recorded beside each. The new rules satellite put the corpus over the old ceiling after already being trimmed once; the floor had drifted to 79 KB of slack and was re-measured to the ~34 KB every previous setting used. Re-measured comments are scoped to the rows this PR actually MOVED — the three `src/state` / `src/types` rows the new satellite and the `layout-misc.md` pointer touch, plus the four the pointer's +291 B shifted (`synthesizer.ts`, both `src/assets` rows, `vite.config.ts`). The file's other rows were already stale on `main` and are left alone; correcting them is not this PR's to do.

## Follow-up

The sibling-shape sweep this lane ran — other places cdkd reports data removed while an AWS feature keeps a readable copy — found four more sites, none of them replication and none fixed here: go-to-k/cdkd#2624 (`cdkd scrub`'s rewrite, `cdkd deploy`'s implicit scrub, the `deployments/` purge, EFS automatic backups).


go-to-k:
Review record, for the reviewer who arrives at this PR cold.

Six rounds ran (3-axis + the additive security reviewer each time, per the size tier). The closing round on `152dc947` came back **clean from all four axes** — code, security, spec and test each reported nothing above nit level, and the security reviewer said so explicitly: *"review marker OK to set; PR is merge-ready from the security lens."*

Survivor rate across the mutation sweeps went **8 of 60 probes → 2 of 74 → 0**.

Three things the rounds caught that are worth knowing about, because each was a defect *inside a fix for the previous round's defect*:

1. Scoping the replication check to "we purged something" was not enough. On a versioned bucket `deleteRollbackJournal` writes a delete MARKER even when no journal exists, so the next deploy found the previous marker noncurrent, removed it, and announced a surviving rollback journal for a stack that never had one — the same false warning, one deploy later. Only entries from `resp.Versions` count now.
2. The gates I then added at the `DeleteObjects` failure arms were **dead code**: a body-bearing key enters `purged` at listing time, before any delete, so a failing delete can neither un-purge it nor add a marker. The two tests that "proved" them passed for a different reason. Both gates were deleted, the invariant stated in their place, and a positive case added so that moving `purged.add` into the delete-success path is no longer a green mutation.
3. A truncation assertion was **vacuous**: it asserted `not.toContain('more)')` while this module emits `... more; aws s3api ...`, so the mutation it was written to catch ran green. That needle came from the sibling module, whose tail does close its parenthesis.

**The final commit (`152dc947..a820454e`) is prose-only** and is the reviewers' own requested wording: one changelog number (the case count, which had been wrong three times because I was measuring it by grepping added `it(` lines instead of counting cases at runtime — 65 is the measured figure) and two code comments that contradicted a third comment about the exact string in item 3. `git diff 152dc947..HEAD` filtered to non-comment lines yields the single changelog line and nothing else. The repo's recursion rule stops at prose, which is why no seventh round was run.

Deliberately not fixed here, each recorded rather than implied: `evictIfOwn`'s identity check is unfenced (its only mid-flight clearer is a test-only helper); `safeBucket`'s placement inside the `try` is behaviourally inert and kept structurally; the truncation arm's `startsWith` over-reach names a sibling key whose own walk completed; and `main`'s 20 pre-existing stale payload-budget rows are left alone — this PR re-measured only the seven it moved.

Follow-ups filed rather than bundled: go-to-k/cdkd#2624 (four more sites where cdkd reports data removed while an AWS feature keeps a readable copy) and go-to-k/cdkd#2635 (a pre-existing raw-interpolation debug line the security reviewer surfaced as out of scope, since that file sits in the `integ-destroy` gate scope).

Issue #2613, opened 2026-09-05
fix(provisioning): 60 of 77 SDK providers drop UpdateContext, so the update path's assertRegionMatch guard never runs


go-to-k:
Issue go-to-k/cdkd#2301 item 1 added `expectedRegion` to `UpdateContext` and threaded it from `deploy-engine.ts`, both rollback replay arms and `drift --revert`, so the UPDATE path could run the same `assertRegionMatch` guard the DELETE path runs. The hazard it names is real: a provider is handed a `physicalId` read out of a state record, and most physical ids are NAMES, so if the client's region disagrees with the record's, the call does not fail -- it succeeds against a same-named resource in the client's region. On delete that is unrecoverable; on update it is a misapplied configuration on a resource cdkd does not manage.

**The caller side landed. The receiver side mostly did not.** `ResourceProvider.update`'s `context` parameter is OPTIONAL, so a provider that simply does not declare it compiles fine and the argument is dropped on the floor -- silently, with no type error and no lint.

Measured on `main` (each provider's own `async update(` signature, from the parameter list up to the `): Promise` line):

- **17** providers declare an `UpdateContext` parameter
- **60** do not

`AWS::Logs::LogGroup` is one of the 60 (`src/provisioning/providers/logs-loggroup-provider.ts`): `deploy-engine.ts` passes `expectedRegion: this.stackRegion` at the `updateProvider.update(...)` call, `LogsLogGroupProvider.update` declares five parameters, and `assertRegionMatch` is reached only from `delete()`. Found while reviewing the `LogGroupClass` refusal change (go-to-k/cdkd#2611, closing go-to-k/cdkd#2579) -- pre-existing, not introduced there.

## Why this is worth its own issue rather than a per-provider fix-as-you-go

The failure mode is invisible by construction. Nothing reds: the type allows the omission, the engine keeps passing the value, and the guard just never runs. A provider gaining an `assertRegionMatch` call one PR at a time leaves no way to tell the guarded from the unguarded except by re-running this count, and the count silently drifts DOWN every time a new provider is scaffolded from a template that omits the parameter.

## Candidate remedies

1. **A fence, first and cheapest.** A unit test that walks `src/provisioning/providers/*.ts`, parses each `update()` signature, and asserts the declaring set matches a recorded allow-list -- so the number can only move deliberately, in either direction. This is what stops the regression while the backfill proceeds. `tests/unit/scripts/` already holds several list-sync fences of this shape.
[...]

Issue #2301, opened 2026-08-26
fix(provisioning): CloudControlProvider still acts on a state record without confirming its region on the update path and on every non-S3 delete


go-to-k:
Dup-check: searched open issues for `CloudControlProvider region identity` and ran a body-window scan for `assertRegionMatch` / `CloudControlProvider` / `expectedRegion` across all open issues. Hits were go-to-k/cdkd#2283 (the S3 delete hook being closed now), go-to-k/cdkd#2245 (the SDK-provider twin), go-to-k/cdkd#1730 and go-to-k/cdkd#1710 -- none names this root cause. No open issue covers `CloudControlProvider`'s remaining region-identity gaps.

## Summary

go-to-k/cdkd#2283 added a per-type bucket-identity probe to `CloudControlProvider.delete()`, keyed on `CC_DELETE_IDENTITY_CHECKED_TYPES`. Three gaps in the SAME mechanism were found by that PR's review rounds and deliberately left out of it, each for a stated reason. One root cause: **`CloudControlProvider` acts on a state record's `physicalId` without confirming the record's region, everywhere except the one hook go-to-k/cdkd#2283 added.**

- [x] **`update()` has no guard, and the reason is a MISSING FIELD rather than a missing parameter.** `ResourceProvider.update` DOES take `context?: UpdateContext` (`src/types/resource.ts:135-142`); what it lacks is a region. `expectedRegion` is declared on `DeleteContext` only (`src/provisioning/region-check.ts:27`), while `UpdateContext` carries `desiredFromAwsReadback?` and nothing regional. So the remedy is NOT an interface change across every provider -- it is one optional field on an interface that already exists, plus threading it from the callers that already hold `state.region`. Consequence of the gap is misapplied configuration against a foreign-region resource, which is recoverable where the delete's is not; that asymmetry is why go-to-k/cdkd#2283 scoped to delete. **DONE -- shipped in go-to-k/cdkd#2378.**
[...]

Issue #2245, opened 2026-08-26
fix(s3): the issue 2227 region guard is create-only, so an already-poisoned state record still lets update() reconfigure and destroy() DELETE a foreign-region bucket


go-to-k:
## The gap

go-to-k/cdkd#2227 added `S3BucketProvider.assertExistingBucketRegion`, which
refuses to ADOPT an already-owned bucket that lives outside the deploy's region.
That guard is on the CREATE path only. A state record that already names a
foreign-region bucket -- written by any cdkd build before that fix -- is never
detected, and two paths then act on it:

- **`update()` has no region check at all.** It applies the stack's bucket
  configuration to whatever `physicalId` the state record names.
- **`delete()`'s `assertRegionMatch` fires only on the `NotFound` branch**
  (`src/provisioning/providers/s3-bucket-provider.ts`, the delete path). S3's
  region-redirect middleware follows the 301 for body-bearing operations, so a
  `DeleteBucket` against a foreign-region bucket does NOT come back NotFound --
  it succeeds. The region check is never reached, and the destroy deletes a
  bucket in another region.

So go-to-k/cdkd#2227 stops NEW poisoning; it does not heal or refuse an EXISTING
poisoned record. Found by the round-2 code review on that PR.

## Why this is not obviously urgent

Reaching it needs a state record written before go-to-k/cdkd#2227 whose bucket
name collided with an account bucket in another region. The collision itself is
not exotic -- `generateResourceName` produces `{stackName}-{logicalId}` with no
region or account, so one stack deployed to two regions collides by
construction -- but the record has to predate the fix.

The destroy consequence is the serious half: deleting a live bucket in another
region is not recoverable, where the update consequence (misapplied
configuration) is.

## What a fix needs, and the open question

The mechanism is settled and can be lifted from the create path: read
`x-amz-bucket-region` off the error when there is one, else `GetBucketLocation`,
fold empty -> `us-east-1` and `EU` -> `eu-west-1`.

What is NOT settled is the POLICY, and it differs per path:

- **update**: refuse, or heal the record? Refusing is safe but strands the
  stack; healing means rewriting a physical id, which is a state mutation.
- **delete**: refusing means a destroy cannot complete without manual
  intervention, which is its own foot-gun. But silently deleting another
  region's bucket is worse.

A pre-flight region probe on every S3 update/delete also costs a round trip on
a hot path, so the cheap version probably keys on something already in state
rather than probing unconditionally.

[...]

PR #2685, opened 2026-09-06
chore(provisioning): fence which SDK providers drop update()'s UpdateContext, and name the 67 that do

go-to-k:
Refs #2613 — **remedy 1 only** (the fence). Remedies 2 (backfill the parameter plus the `assertRegionMatch` call, per provider) and 3 (drop the `?` on the interface) stay open on the issue, which is why this says `Refs` and not `Closes`.

No `src/` change and no behavior change: two new files plus a rules-index entry and a changelog entry.

## The defect this fences

`ResourceProvider.update`'s `context` parameter is optional, so a provider that never declares it compiles fine, `deploy-engine.ts` / `rollback-executor.ts` (both replay arms) / `drift --revert` keep passing `expectedRegion` (issue go-to-k/cdkd#2301 item 1), and `assertRegionMatch` simply never runs for that provider. Nothing reds — the type allows the omission, and the split drifts silently every time a provider is scaffolded from a template that omits the parameter.

## The population, re-derived — and it is not the issue's

The issue recorded **17 declaring / 60 omitting over 77 files**, from `grep "async update("`. Re-derived per CLASS with the TypeScript parser:

```
$ node scripts/check-provider-update-context.ts
82 files, 85 provider classes: 18 declare update()'s UpdateContext parameter, 67 do not.
```

Both corrections come from the same place — the population rule:

- the unit is the **class**, not the file: `glue-provider.ts` declares seven providers, which the file-level count collapsed into one (`85 = 77 + 6 + 2`);
- `async` is **optional**, and two providers spell `update()` without it. The grep saw neither, so both fell out of *both* of its buckets: `ServiceDiscoveryProvider`, which **does** declare the parameter (so 17 → 18), and `EFSProvider`, which does not (60 → 67, with the six extra Glue classes).

## Why this shape, and not the alternatives

**Not a count.** A bare "67 omit it" rots on every provider added or removed and names nobody. `OMITS_UPDATE_CONTEXT` is the exact set, so the three interesting transitions each fail with their own message: a NEW provider without the parameter (message leads with "declare the parameter", not "add it to the list"), a provider that GAINED it (`good news — … delete them from OMITS_UPDATE_CONTEXT`, so the fence only ever ratchets tighter), and an entry naming a provider that no longer exists. The declaring set is *derived* rather than pinned as a second literal.

**Not a grep.** Following the lesson from this session's sibling lane: a text scan over a signature always has another hole. This parses each class with the TS compiler (`typescript-v6`, which 13 scripts under `scripts/` already import on `origin/main`, 7 of them `check-*` critics) and reads the parameter list structurally.

**Not a compile-time type constraint either, and this was the real design question.** A `.test-d.ts` asserting `Parameters<P['update']>['length']` per provider would need 85 hand-written imports — and a hand-maintained list is exactly what cannot derive its own population, so a new provider would simply be absent and unfenced. The load-bearing half here is the *population*, which a type-level check cannot enumerate from the filesystem.

**The population is derived from a relation a provider cannot omit.** Not `implements ResourceProvider` (optional — a structurally-assignable class registers without it) and not `async` (optional, and demonstrably the thing that made the issue's own count wrong), but MEMBERSHIP: `create` + `update` + `delete` all present, as methods or as function-valued properties. A class with two of the three, or with `implements ResourceProvider` and not all three, is REFUSED rather than classified.

**Everything unmodelled is refused, not bucketed.** The 6th parameter must be a plain identifier annotated `UpdateContext` (or a union with `undefined` / `null`) whose name is bound by an import RESOLVING to `types/resource.ts`. `context?: unknown`, `context?: DeleteContext`, an aliased import, a destructured or rest parameter, a missing annotation and a 7th parameter all stop the check instead of passing through it.

**What it does NOT claim** (stated in the script header, and in the test that pins the declaring set): declaring the parameter is not the same as CALLING the guard. `S3BucketProvider.update` is on the declaring list and its own region check compares against the CLIENT's region rather than `context.expectedRegion` — its comment says consuming that field is the remaining half of go-to-k/cdkd#2245. That is remedy 2, and no signature check can see it.

## Two independent instruments, which must agree

1. **Static** — the checker above.
2. **Runtime** — the unit suite runs `registerAllProviders` against a recording stand-in and reads `update.length` off every provider OBJECT it builds. No formatting, comment, line break or `async` keyword can fake that, and it is the only instrument that would see a provider inheriting `update()` from a base class.

A disagreement fails. The unregistered-class set is pinned by name (`CustomResourceProvider`, owned by `ProviderRegistry`'s own constructor), so a provider dropping out of registration is a failure rather than a silent exemption.

Collapse defences follow the sibling critics: `FLOORS` (files / classes / declaring) plus a hard failure on any parse diagnostic for collapse toward ZERO; `SELF_PROBE_CASES` — fixed sources with known verdicts INCLUDING refusals, each refusal case naming a substring of ITS OWN arm, count pinned as a literal by the suite — run before the tree is read, for collapse toward GREEN — with `CDKD_SELF_PROBE_FORCE_FAIL=1` proving the SPAWNED binary still consults them, since the unit tests call `runSelfProbes` directly and `main()` dropping the call would otherwise be unobservable.

## Probes — every one reverts PRODUCTION code, none re-types the logic

Ten failure probes run in the suite; the ones that mutate a provider run against a scratch COPY through the `--providers-dir=` seam and the rest read the real tree or a temp directory, so no probe writes to `src/`. Each asserts rc != 0 and the finding kind:

| Mutation | Verdict |
| --- | --- |
| a new provider file with a 5-parameter `update()` | RED `[new-omitting] BrandNewProvider` |
| `s3-bucket-provider.ts` loses its `context?: UpdateContext` | RED `[new-omitting] S3BucketProvider` |
| `logs-loggroup-provider.ts` gains it (import + parameter) | RED `[newly-declaring] good news … LogsLogGroupProvider` |
| `KMSProvider` renamed to `KMSKeyProvider` | RED `[stale-entry] KMSProvider` + `[new-omitting] KMSKeyProvider` |
| that 6th parameter typed `unknown` | RED `[refusal]`, and NOT `[newly-declaring]` |
| an empty providers directory | RED `[floor]`, not a vacuous pass |
| `CDKD_SELF_PROBE_FORCE_FAIL=1` | RED `self-probe FAILED` before the tree is read |
| a mistyped `--porviders-dir=` / an empty `--providers-dir=` | RED exit 2, no stdout — never a green about a tree nobody named |
| a provider hidden in a SUBDIRECTORY of the providers root | RED `[refusal] nested/ … not recursive` |
| two provider classes sharing a NAME | RED `[duplicate-name] KMSProvider` |

Plus TWO probes against the REAL tree, one per direction, each restored byte-exact (sha256 re-checked, `git status --porcelain` and `git diff --stat` both empty) — mutations, RED verdicts and restores recorded in a comment on this PR.

## Sibling sweep (`create` / `delete`), and why nothing is filed

The same "provider drops the context" shape exists on the other two methods, so it was measured rather than assumed:

```
create: declares=27 omits=58
update: declares=18 omits=67
delete: declares=82 omits=3
```

- **delete** — the unrecoverable twin, and the three omitters are `AgentCoreBrowserProvider`, `AgentCoreCodeInterpreterProvider` and `WaitConditionHandleProvider`, whose `delete()` bodies are each a single `logger.debug` no-op issuing no AWS call. No hazard, nothing to file.
- **create** — `CreateContext`'s constraint is CONDITIONAL (a create-side pre-flight refusal must downgrade on a rollback replay; a message interpolating a resolved value must be masked). The masking half is already fenced per SITE by `scripts/check-provider-secret-mask.ts`, and the five providers `docs/provider-rules.md` names as re-creating inside their own `update()` are still exactly those five (`grep -rn "this\.create(" src/provisioning/providers/` — the sixth hit, in `glue-provider.ts`, is a comment saying not to). A candidate population, not a proven defect, so it is reported here rather than filed as one.

## Gates

`git diff origin/main...HEAD --name-only` is four files — `scripts/check-provider-update-context.ts`, `tests/unit/scripts/provider-update-context.test.ts`, `.claude/rules/layout-scripts.md`, `docs/changelog-cdkd.md` — and none matches any `integ-*` gate's activation regex, so no real-AWS gate applies to this PR.

## Review round (3-axis: code + test + spec)

`loc >= 1000` put this at the `3-axis` tier. All three reviewers ran read-only against the worktree and reported a clean tree at start and end. No blockers. The code and test reviewers between them PROVED **four fail-open holes**, each by running the checker against a scratch copy — all four are now closed, each with its own named self-probe case, and each re-verified with the reviewer's own probe:

| Hole (all PROVED by a reviewer, then re-proved after the fix) | Was | Now |
| --- | --- | --- |
| `...context: UpdateContext` — a REST 6th parameter | classified `declares`, exit 0 | `[refusal] … destructured or a rest element` |
| `export const P = class { … }` — a class EXPRESSION | invisible to the population, exit 0 | classified; an unrecorded one reds `[new-omitting]` |
| a class declaring ONLY `update` (create / delete inherited) | neither classified nor refused, exit 0 | `[refusal] declares 1 of create/update/delete` |
| `import … from '../../mytypes/resource.js'` | accepted by the unanchored suffix as the shared type | `[refusal] … is not `UpdateContext` imported from /types/resource.ts` |

The first two were the ones the doc comment already CLAIMED were refused, which is the worse kind: a stated limit that was not enforced. Both are now enforced and the header states what the module test actually is (a path-suffix test on a resolved specifier, not the compiler's module resolution).

Also taken from the round: a subdirectory under the providers root is REFUSED rather than scanned past (the scan is not recursive, so a provider there was invisible); each refusal self-probe now pins a substring of its OWN arm, since a bare `refused` verdict could not tell the five arms apart; the runtime-vs-source disagreement loop no longer skips a name missing from the source scan (it was vacuous under a total collapse); `spawnSync` carries an explicit timeout, because `it`'s bound cannot interrupt a synchronous spawn; `updateParams` is asserted rather than write-only; and the self-probe count is a literal instead of a slack `>= 12`. The spec reviewer independently re-derived every number in this body and confirmed the four-file scope, `Refs` not `Closes`, and that no `integ-*` activation regex matches the diff; its three nits (a stale `14` in the rules entry, this body over-stating "against a scratch COPY", and understating the `typescript-v6` consumer count) are corrected above.

## Second review round (coordinator's independent 3-axis + security)

No blockers, and the population + sibling-sweep numbers were independently reproduced twice. One MAJOR and eight smaller items, all addressed.

**MAJOR — "refuses everything unmodelled" was FALSE; six shapes were SILENTLY DROPPED.** `memberSignature` returned `undefined` for any member whose name was not an identifier, or whose initializer was not a function expression, so `members.has('update')` was false and the refusal never fired: `classes: []`, `refusals: []`, **rc=0**. Reproduced before fixing — five provider-shaped classes added to a scratch tree produced `rc=0` with an empty stderr, and a `static update` inflated the counts to 86 classes / 19 declaring.

Fixed in CODE, not with a header note (the standalone CLI is what a contributor runs, and instrument 2 only covers REGISTERED providers):

| Shape | Was | Now |
| --- | --- | --- |
| `update = makeUpdate()` (non-function initializer) | dropped, rc=0 | `[refusal] … initializer is not a function expression` |
| `async 'update'(…)` (string-literal name) | dropped, rc=0 | name RESOLVED, classified |
| `async ['update'](…)` (computed over a literal) | dropped, rc=0 | name RESOLVED, classified |
| `async [KEY](…)` (unresolvable computed) | dropped, rc=0 | `[refusal] … computed name this checker cannot resolve` |
| `static update(…, context?: UpdateContext)` | credited as the INSTANCE verdict | `[refusal] … is not the instance member` |
| `class P extends Base` short of the triple | dropped, rc=0 | `[refusal] … extends a base class` |
| a provider in a `.mts` / `.cts` file | skipped outright, rc=0 | scanned |

Plus a **defaulted** parameter (`context: UpdateContext = {} as UpdateContext`) is now refused: it lowers `update.length` below the declared arity, so it was the one input that could break the two-instrument equivalence the suite asserts.

**Calibration, not blanket refusal.** Ungated, the `extends` and computed-name arms refused `HostedZoneNameNotFoundError` (`extends Error`) and a helper class — neither could ever be a provider. Both arms are gated on the class being a plausible provider (declares at least one of the three, or says `implements`). **The one limit that remains is NAMED rather than claimed away** in the header: a class that extends a base, declares NONE of the three and carries no `implements` is indistinguishable at this layer from every ordinary error subclass, so it is ignored — the runtime half covers it, since such a provider once registered fails "every registered provider is one the source scan found", and unregistered it is dead code no deploy can reach.

**The other eight.** The stale-entry message no longer fires for a REFUSED provider (measured: it told you to delete a live worklist entry for a provider still in the tree) and the test now asserts its ABSENCE. `updateParams` is derived from the parsed parameter list — and, since deriving it was **necessary but not sufficient** (an arity-classifier mutation still left the arm green over the real tree, because no provider there has an arity outside 5/6), a new case supplies the discriminating input: a 7-parameter provider in a scratch copy must be REFUSED, never classified with arity 7. Every spawn now goes through one helper carrying the timeout and the `proc.error` assertion. The floor derives from `UNREGISTERED_PROVIDER_CLASSES.length`. The test header's "every failure probe runs against a scratch COPY" is corrected to the claim that actually holds — no probe writes to `src/`. The stale PR comment above is edited, and this body's `typescript-v6` count is corrected to `origin/main`'s 13 / 7.

**Recurring shape, fifth time this session: a comment asserting a guarantee the code does not provide.** Swept my own doc comments for universals and weakened three that were false — the cross-check covers every REGISTERED provider (84 of 85), not "every verdict"; and "never silently dropped" became "refused rather than silently dropped", since limit 7 above is exactly the exception.

Self-probe corpus 23 -> 32 cases, each refusal case pinning a substring of its OWN arm. Every fix probed by reverting PRODUCTION code from a GREEN baseline and watching the corresponding case go RED, each restored byte-exact by sha256.

## Third review round (execution-based delta review)

Four items, all closed in code; nothing filed.

**1. The refusal arms OVER-REFUSED ordinary helper classes — and the suggested fix was not sufficient.** Reproduced first: a `static create()` factory, a `get create()` accessor, an options bag with `delete = true`, `class LruCache extends Map { delete() }` and `class WidgetFactory extends BaseFactory { create() }` were all refused, and appending that cache to a copy of the real `kms-provider.ts` made the shipped binary exit 1. Gating the three `unmodelled` arms (the proposed fix) would have closed only three of the five rows: the cache and the factory came through the `extends` arm, which was *already* gated. The predicate itself was wrong.

So there is now ONE gate on every arm, and it asks a different question — not "does this class touch one of the three names" but "is this a provider candidate": it says `implements ResourceProvider`, OR declares **two** of the three (a provider missing only `update` still has `create` and `delete`, both equally required), OR names `update` at all, in any spelling including an unreadable one. `delete` on a cache and `create` on a factory are single incidental hits and fall outside; `update` counts alone because it is the member this check is about. All five helpers are now ignored, `kms-provider.ts + LruCache` exits 0, and each is pinned as a named CONTROL case. The complement is pinned too: `class P extends Base { update = makeUpdate() }` is still refused.

**2. A new false universal, in the sentence added to BOUND the fence.** "for a REGISTERED provider, which is every one a deploy can route to" — false, and contradicted by this PR's own `UNREGISTERED_PROVIDER_CLASSES`. Deleted. The re-grep it prompted found **two more** in the very limit-7 bullet the earlier sweep had written: "every ordinary `extends Error` class in the tree" (no class under `providers/` extends `Error`; the real one extends `ProvisioningError`) and "unregistered, it is dead code no deploy can reach" — the same `CustomResourceProvider` error again, one bullet away from where I had just fixed it. Two header paragraphs also still described the superseded split gate. All rewritten against the code. The lesson recorded in the header: a universals sweep is not self-verifying, and its own output needs the same check.

**3. The `.mts` / `.cts` widening was untested** — narrowing the filter back leaves both the self-probes (which bypass `buildReport`'s filter) and the real-tree run green. Now fenced by its own arm: a `.mts` provider written into a scratch copy must appear in `report.files` and `classes`, and red the run end to end.

**4. Wording.** The bare no-substitution-template branch in `memberName` was unreachable dead code — a template literal in a member-name position is a syntax error (3 parse diagnostics, measured) — so it is deleted and the spelling list corrected to what is actually handled, including the previously unmentioned private identifier. The rest-element claim is narrowed to the sixth position (an earlier rest element classifies `omits` with `update.length` of 4; the verdicts still agree, which is what the arm compares). `extends Error` corrected to `extends ProvisioningError`. The 7-parameter test's comment now says it discriminates for the `updateParams` assertion specifically, not for the arity mutation as a whole, which the pre-existing self-probe already caught.

Corpus 32 -> 38 cases. Probes, each reverting production code from a green 28-test baseline: reverting the unified gate reds all five CONTROL cases; narrowing the file filter leaves the checker at rc=0 while the new `.mts` arm reds — which is exactly the gap it was added to close. The dead-branch removal is the one change with no red probe available, by construction: restoring the branch keeps rc=0 because the branch is unreachable, which is why it was removed.


go-to-k:
## Real-tree mutation probes

The suite's failure probes never write to `src/`: the ones that MUTATE a provider run against a scratch COPY through the `--providers-dir=` seam, while others read the real tree unmodified, the default root, or an empty temp directory. (Corrected: this paragraph originally said "eight failure probes all run against a scratch COPY" — both halves were wrong, and the count has since moved anyway. The safety claim, no probe writing to `src/`, is the one that holds.) The two below mutate the REAL `src/` files instead, one at a time, each restored from a byte-exact `cp` backup taken before the edit — because a fence probed only against copies has never been watched going red on the tree it actually guards.

### Probe A — an OMITTING provider gains the parameter (the ratchet direction)

`src/provisioning/providers/logs-loggroup-provider.ts`: added `UpdateContext` to the `types/resource.js` import block and a sixth `context?: UpdateContext` parameter to `update()`.

```
$ node scripts/check-provider-update-context.ts
RC=1
[newly-declaring] good news - 1 provider(s) now declare `context?: UpdateContext`:
LogsLogGroupProvider. Delete them from OMITS_UPDATE_CONTEXT in
scripts/check-provider-update-context.ts so the list keeps ratcheting.
```

The fence went red on **six** cases, and the two that matter are one per instrument — the static `the recorded omitting set is exactly what the tree holds` AND the runtime `the runtime omitting set matches the recorded worklist`. The runtime witness is therefore live, not decorative.

### Probe B — a DECLARING provider loses the parameter (the regression direction)

`src/provisioning/providers/s3-bucket-provider.ts`: deleted the `context?: UpdateContext` parameter from `update()`.

```
$ node scripts/check-provider-update-context.ts
RC=1
[new-omitting] 1 provider(s) drop `update()`'s `context?: UpdateContext` parameter
and are not recorded as known: S3BucketProvider. Declare the parameter (and call
assertRegionMatch ahead of every mutating step - issue go-to-k/cdkd#2613 remedy 2);
only add the name to OMITS_UPDATE_CONTEXT in
scripts/check-provider-update-context.ts if you deliberately cannot.
```

Red on five cases, again including both instruments.

### Restores

Each probe was reverted with `cp <backup> <file>` and verified three ways, not one:

```
$ shasum -a 256 src/provisioning/providers/logs-loggroup-provider.ts
ac64110ad0aef8340742c7543e6f07540a384a676d2ac592abc3468246f82c4f   # pre-mutation hash
$ shasum -a 256 src/provisioning/providers/s3-bucket-provider.ts
fd9136f5d9058393aeecd513364dd99aa0893afd333b6ed80ba4db3c5b0da171   # pre-mutation hash
$ git status --porcelain && git diff --stat && echo TREE-CLEAN-OK
TREE-CLEAN-OK
$ node scripts/check-provider-update-context.ts >/dev/null; echo $?
0
```

Both probes were taken AFTER the round's real work was committed, so `git diff` separates a probe mutation from unfinished work by construction.

## Sibling-method sweep

Reported in the PR body rather than filed, because measuring it showed no defect to file: the three providers whose `delete()` omits `DeleteContext` (`AgentCoreBrowserProvider`, `AgentCoreCodeInterpreterProvider`, `WaitConditionHandleProvider`) each have a single-line `logger.debug` no-op `delete()` that issues no AWS call, so `assertRegionMatch` has nothing to guard there.

Issue #2582, opened 2026-09-04
test(integ): a verify.sh assertion of the form 'printf | grep -q' is fail-open under pipefail


go-to-k:
Integration fixtures assert on captured CLI output with

```bash
if printf '%s' "${OUT}" | grep -q "<phrase>"; then
  echo "FAIL: ..." >&2
  exit 1
fi
```

Every `verify.sh` runs under `set -euo pipefail`. When the capture is large enough, `grep -q` finds its match and exits immediately, `printf` fills the pipe and takes SIGPIPE, and `pipefail` makes the PIPELINE report 141 instead of grep's 0. So the `if` reads FALSE -- "the phrase is not present" -- at exactly the moment the phrase IS present.

For a NEGATIVE check (the shape above, where a match means FAIL) that is FAIL-OPEN: the fixture prints PASS over the condition it exists to catch. For a POSITIVE check (`if ! ... ; then FAIL`) it fails LOUDLY instead, which is merely noisy.

## Measured, not inferred

bash 5.3.9, macOS (arm64), needle on line 1, five attempts per size:

| payload | pipeline non-zero DESPITE a match |
| --- | --- |
| 3 KB | 0/5 |
| 31 KB | 0/5 |
| 155 KB | 5/5 |
| 310 KB | 5/5 |
| 620 KB | 5/5 |
| 1.5 MB | 5/5 |

So the threshold on that host sits between roughly 32 KB and 155 KB -- consistent with the pipe buffer plus grep's read buffer, and therefore host- and grep-dependent rather than a fixed number. A single-line payload of 4 MB did NOT reproduce it (grep cannot decide before EOF without a newline), so the shape needs BOTH a line-oriented payload and enough of it.

A herestring returns 0 in every case above: there is no pipeline, so there is nothing for `pipefail` to report.

## A SECOND arm, with no size threshold (2026-09-07)

The measurement above explains one arm: the writer blocked on a full pipe. It
does NOT cover the two live false FAILs seen on `secrets-dynamic-ref` while
PR #2742 was being written, whose captures are about 1 KB -- far under the 32
to 155 KB range. The mechanism reported in
https://github.com/go-to-k/cdkd/issues/2582#issuecomment-5571368018, and
consistent with every number in it:

- bash's BUILTIN `printf` writes a multi-line argument in several write-family
  calls (counted as `/proc/<pid>/io` `syscw` deltas on bash 5.2.21: roughly one
  call per short line and one per 4 KB within a line).
[...]

Issue #2563, opened 2026-09-04
fix(resolver): a concurrent Fn::Join part that rejects surfaces before a sibling's secret has recorded, and the deploy engine's Export.Name copy loses that late entry


nix-tkobayashi:
## Summary

`IntrinsicFunctionResolver.resolveJoin` resolves a join's parts with `Promise.all`, so a rejecting part surfaces before a sibling part that is still recording a secret has recorded. `DeployEngine`'s `Export.Name` resolution copies the entries of its private `nameSecrets` map into the pass map in a `finally`. `Fn::Join` resolves its parts concurrently (`Promise.all` in `IntrinsicFunctionResolver.resolveJoin`), so a part that rejects while a secret-resolving sibling part is still pending ends the block — and runs the copy — before that sibling records. The late entry lands in the private map after the copy and never reaches the pass map: a plaintext the pass then does not know as a needle. `cdkd scrub` had the same shape and go-to-k/cdkd#2562 (issue go-to-k/cdkd#2531) replaced it there with a view of the pass map (`SharedEntriesSecrets`); the deploy engine still has the copy shape.

## Mechanism (`src/deployment/deploy-engine.ts`, the `Export.Name` block of the outputs pass)

- The name is resolved with `const nameSecrets: RecordedSecretValues = new Map()` so that the exposure decision (`exportNameSecretExposure` — is THIS name secret-bearing?) is exact, and the entries are merged into `context.recordedSecretValues` in a `finally` ("survives every exit").
- `resolveJoin` resolves the join's parts with `Promise.all`. When one part rejects, the returned promise rejects at once; a sibling part whose dynamic-reference resolution is still in flight keeps running and, when it completes, records `plaintext -> expression` into `nameSecrets` — after the `finally` has already copied.
[...]

Issue #1903, opened 2026-08-14
fix(state): nested-stack Parameters persist resolved secret plaintext into the CHILD state (GHSA follow-up)


go-to-k:
## Context

Follow-up to the dynamic-reference secret-redaction fix (GHSA-p5qg-v9gv-hc7w, PR (#1899)) and its sibling (#1901). Found by the 3-axis review of the (#1901) PR — reported independently by two reviewers.

**Pre-existing for `{{resolve:secretsmanager:...}}`** (i.e. a gap in the ALREADY-SHIPPED fix, not a regression of (#1901)). Issue (#1901) widens the affected value set to a `{{resolve:ssm:...}}` naming a `SecureString` parameter.

## Gap

The redaction rests on the resolver recording `plaintext -> {{resolve:...}} expression` into `recordedSecretValues`, which the deploy engine reads at its state-save choke point. A nested stack breaks that chain because the PARENT resolves the child's `Parameters` and hands the child the resulting **literals**:

- `src/provisioning/providers/nested-stack-provider.ts` passes `parameters: childParameters` — already-resolved plaintext — into the child `DeployEngine`.
- The child's `buildResolverContext` allocates a FRESH `recordedSecretValues` which never sees a `{{resolve:` expression, so the child's `perResourceSecrets` is empty.
- Result: the child's `state.json` persists the **resolved secret plaintext**, with no expression to redact back to.

## The two halves are COUPLED — do not fix the diff side alone

`src/cli/commands/diff-recursive.ts`'s `resolveChildStackParameters` builds its resolver context with neither `skipDynamicReferences` nor `recordedSecretValues`, so `cdkd diff --recursive` also resolves such a parameter to plaintext at plan time and prints it.

The obvious one-line fix (add `skipDynamicReferences: true` there) is **wrong on its own**: because the deploy path above persists plaintext into the child state, resolving the desired side to plaintext is what currently makes the comparison self-consistent. Skipping only on the diff side would compare expression-vs-plaintext and report a spurious perpetual change on every run.

So the child-state half has to land first (or together): the child needs to learn WHICH incoming parameter values are secret-bearing — e.g. by threading the parent's `recordedSecretValues` into the child engine, or by passing the UNRESOLVED expressions down and letting the child resolve them itself.

## Verification

[...]

PR #2742, opened 2026-09-07
fix(deploy-engine): mask output-failure error text and resolver lookup echoes

### LINKED ISSUE #2728: fix(deploy-engine): handleOutputResolutionFailure logs and rethrows the resolver's error text unmasked
## What is wrong

`DeployEngine.handleOutputResolutionFailure` (`src/deployment/deploy-engine.ts`, the private method above `resolveOutputs`) reports an output resolution failure with the resolver's error text UNMASKED, on both of its arms:

- the default arm logs `Failed to resolve output ${outputKey}: ${String(error)}` at `warn`;
- the `--strict-getatt` arm throws `new Error(\`Failed to resolve output ${outputKey}: ${error.message} (--strict-getatt promotes ...)\`, { cause: error })`, and that message is what the CLI prints.

Neither passes through `maskSecretsInText` / `maskSecretsInError`, which the engine already imports and applies at its other resolver-facing sites (the per-resource failure formatting, `maskSecretsInError(error, this.perResourceSecrets.get(logicalId) ?? EMPTY_SECRETS)`, and the events-record masking of `next.error.message` / `next.reason`).

What the real resolver's errors carry, precisely: `resolveJoin` propagates a rejecting part's error UNCHANGED (it does not append a sibling's resolved value — that echo is a shape the `scrub` unit mock manufactures, not the resolver's), and `resolveDynamicReferences`'s own failures echo the offending reference token (`Refusing to resolve the secret reference ${fullMatch}: ...`, `Refusing to resolve ${fullMatch}: the parameter is a ...`), the secret id and JSON key (`key '${jsonKey}' not found in secret '${secretId}'`), the SSM parameter name (`SSM parameter '${parameterName}' not found or has no value`), or a propagated AWS SDK error. So a plaintext reaches this log when a SUBSTITUTED value has become part of one of those echoed fields — an `Fn::Sub` / `Fn::Join` that assembles the reference token or the parameter name out of a value the same pass resolved from a secret (the assembled string is what `resolveSub` / `resolveJoin` re-enter `resolveDynamicReferences` with), and the lookup or classification then fails. Merely resolving a secret sibling before another part fails does not expose anything. `String(error)` on a non-`Error` thrown value is the same exposure with no `.message` indirection.

The needle set for this site is the outputs pass's own map (`context.recordedSecretValues`, merged into `this.outputSecrets` in the pass's `finally`): it holds every plaintext recorded BEFORE the handler runs, including an `Export.Name` resolution's entries copied back by the inner `finally`. It does not hold what a still-pending concurrent part would have recorded (go-to-k/cdkd#2563's late-write shape), so masking here is bounded by that residual the same way the rest of the engine's masking is.

Found by the maintainer's round-2 review of go-to-k/cdkd#2562 as pre-existing and outside that diff; go-to-k/cdkd#2563 does not change it either — no masking is applied at this site at all.

## Proposed fix

Mask both arms against the outputs pass map at the point of formatting — `maskSecretsInText(error instanceof Error ? error.message : String(error), context.recordedSecretValues ?? EMPTY_SECRETS)` for the message text, with the `cause` chain kept as it is (the `--strict-getatt` arm's `cause` is load-bearing for `isMarkedNonRetryable`; masking the WRAPPER's message does not touch it). The method takes `error, outputKey, outputs` today; it needs the secrets map (or the resolve context) as a fourth argument from its two call sites in `resolveOutputs`.

Tests, two layers because the exposure depends on what the REAL resolver echoes:

- mock-resolver unit cases, one per arm: a resolver rejecting with a message that embeds a recorded plaintext; assert the warn (default arm) and the thrown message (`--strict-getatt` arm) carry `***` and not the plaintext, and that the strict arm's `cause` is still the original error;
- a real-resolver case with faked SDK clients (the shape of `tests/unit/cli/commands/scrub-cross-region-secret.test.ts`): an output whose `Fn::Sub` first resolves a JSON-key secret into a variable and then uses that VALUE as the JSON key of a second reference to the same secret — `{{resolve:secretsmanager:<id>:SecretString:${Resolved}}}` — so the lookup succeeds and the resolver's own deterministic `Dynamic reference: key '${jsonKey}' not found in secret '${secretId}'` error necessarily carries the recorded plaintext; assert the output-failure warning carries `***` and not the plaintext. (An SSM not-found is NOT a usable shape: a missing live parameter raises the SDK's `ParameterNotFound`, which propagates unchanged, and the resolver's own name-bearing `SSM parameter '...' not found or has no value` message runs only on a successful response lacking `Parameter.Value` — a fake returning `{}` would exercise that defensive branch, not the live behaviour.)

Verification a fresh session can run: the cases above; `/run-integ lambda` for the `integ-broad` marker (`src/deployment/deploy-engine.ts` is in that scope). No current integ fixture reaches this warn — `secrets-dynamic-ref` resolves its references in resource properties and its outputs hold names, not secret values — so a live arm would be a new output in that fixture with the same JSON-key shape (the fixture already deploys a JSON secret whose `password` key it resolves), deployed under the default arm (warn + skip, the deploy still succeeds), asserting the `Failed to resolve output` line in the deploy log carries `***` and not the password.

## Classification

Dup-check: searched open and closed issues for `handleOutputResolutionFailure`, `Failed to resolve output plaintext`, `deploy-engine log secret mask output failure` — no hit; go-to-k/cdkd#2516 / go-to-k/cdkd#2320 / go-to-k/cdkd#2346 are redaction issues on other paths (value-scan floor, drift, S3 versions), and go-to-k/cdkd#2563 is the late-write copy loss at the same block, which leaves this log site untouched.
Session-fit: next (not this session) — the maintainer named it as pre-existing and outside go-to-k/cdkd#2562's diff, which is at "fix one sentence and I will approve"; the fix and its real-resolver case touch `src/deployment/deploy-engine.ts` and a new test file, none of the files that round has open.
Severity: medium — a resolver failure on an output can print a resolved secret plaintext to the terminal / CI log at default verbosity; condition-specific (a reference token, JSON key or parameter name assembled from a resolved secret, whose resolution then fails inside the resolver), and the state record itself is unaffected.
Effort: medium (M) — an edit with three unit cases (two mock arms, one real-resolver reachability case) plus one re-review round; `lambda` is an EXISTING fixture for the broad marker, and the live arm is a new output on the existing `secrets-dynamic-ref` fixture rather than a new fixture.
Estimate: ~2-3 h — the real-resolver case (faked SSM / Secrets Manager clients through the real `IntrinsicFunctionResolver`) and the fixture arm eat most of it; the two integ runs are ~15 min of it.

nix-tkobayashi:
Closes #2728

## What was wrong

`DeployEngine.handleOutputResolutionFailure` reported an Output resolution failure with the resolver's error text UNMASKED on both arms — the default `Failed to resolve output <key>: <error>` warn and the `--strict-getatt` throw — while every other resolver-facing site in the engine masks. The resolver's own failures echo the offending reference token, a secret id / JSON key or an SSM parameter name, and `resolveSub` / `resolveJoin` re-enter `resolveDynamicReferences` with the ASSEMBLED string, so a reference built from a value the same pass resolved out of a secret puts that plaintext into the echoed field: an `Fn::Sub` whose variable resolves the secret's `password` and whose body uses that value as the JSON key of a second reference fails with `key '<password>' not found in secret '<id>'`. (`resolveJoin` propagates a part's error unchanged — the "sibling's plaintext echoed" shape in the scrub unit mock is that mock's, not the resolver's.) The same assembled plaintext also reached the log one line earlier, at `--verbose`, through the resolver's `Resolving dynamic reference: secretsmanager:<id>:SecretString:<jsonKey>...` echo, and — for an SSM name assembled the same way — through the throttle-retry label and the unrecognized-`Type` warn (found by the security reviewer on this PR); a value landing in the SERVICE position (`{{resolve:${Pw}}}`) reached the `Unsupported dynamic reference service:` warn at default verbosity (found by the maintainer's review round 1). Named by the maintainer's round-2 review of go-to-k/cdkd#2562 as pre-existing there.

## What changes

`src/deployment/deploy-engine.ts`:

- `handleOutputResolutionFailure` takes the outputs pass map and the inherited bag as parameters and masks against BOTH, inherited first — the resolver's `maskSecretsForLog` order and reason (issue go-to-k/cdkd#1903 round 2: a nested-stack child's parent-decrypted parameter plaintext is in `inheritedSecrets` only until a `{Ref: <Param>}` resolution copies it across; no shape reaching this handler before that copy was constructed, so the inherited bag is defense in depth). The text is `error.message || error.name` (the name is where an SDK error keeps its code when the message is empty) masked with `maskSecretsInText`; both arms use it. The strict arm's `cause` is masked as an OBJECT through `maskSecretsInError` — a clone of each `Error` link within `errorCauseChain`'s bound (`ERROR_CAUSE_MASK_MAX_DEPTH`; a non-`Error` link is kept verbatim), symbols included, so `isMarkedNonRetryable`'s non-enumerable marker survives (the same reason `provisionResource` uses it); a thrown STRING is masked as text; a thrown value of any other type is not threaded at all (`maskSecretsInError` would hand it back unmasked, and `markNonRetryable` marks `Error` instances only, so it can carry no marker of ours). The cause is masked on the strict arm only — the warn arm never uses it, and the clone reads every link's `stack` accessor. No deploy-path sink renders past one level today (`formatError` prints `Caused by:` for a `CdkdError`'s direct cause only), but `cdkd scrub`'s `describeFailure` renders a chain — safe because its boundary masks with the same helper and it walks the same bounded chain — so that was a property of reachability, not of the sinks; masking the cause here gives a renderer within that bound nothing to leak. The default arm drops the `Error: ` class prefix so both arms report the same text; the two consumers of the prefix-less `Failed to resolve output <key>` string (`getatt-fallback-guard`, `raw-cfn-conditions-params`) are unaffected.
- Both `resolveOutputs` call sites pass the same two bindings — `context.recordedSecretValues` (everything recorded before the handler runs, an `Export.Name` resolution's entries included since its `finally` merges them back before the `catch`; the `?? EMPTY_SECRETS` is for the type, `buildResolverContext` always sets the map) and `context.inheritedSecrets ?? EMPTY_SECRETS` (set on a nested-stack child engine only, so that fallback IS taken on every top-level stack). What a still-pending concurrent part would have recorded is go-to-k/cdkd#2563's residual, the same bound every other masking site has.

`src/deployment/intrinsic-function-resolver.ts`: `resolveSecretsManagerReference` and `resolveSSMReference` take the resolve `context` as an optional last parameter (threaded from all three call arms — `secretsmanager`, `ssm`, `ssm-secure`) and pass their `Resolving dynamic reference:` debug echo, their `sendWithThrottleRetry` label and the SSM unrecognized-`Type` warn through `maskSecretsForLog(text, context)`; the `Unsupported dynamic reference service:` warn in `resolveDynamicReferences` goes through it too (`service` is the text before the first `:` of the assembled token, and the arm's `continue` then leaves the literal span in the value — go-to-k/cdkd#2743, pre-existing and out of scope here).

## Tests (`tests/unit/deployment/deploy-engine-output-failure-masking.test.ts`, 26 cases)

Two layers, because the exposure depends on what the real resolver echoes:

- **Mocked resolver** (a sentinel records a plaintext into the pass map, then fails with a message embedding it): the default warn is the exact line `Failed to resolve output Leak: Dynamic reference: key '***' not found in secret 's'`; the strict throw carries `***`, its `cause` is a masked clone (exact masked message, not the thrown instance) on which `isMarkedNonRetryable` holds, as it does on the wrapper — identity is deliberately NOT asserted, so a later hardening costs no test edit — and the premise that the instance the resolver threw carried both the plaintext and the marker is measured at the mock seam; a plaintext known only to the INHERITED bag (an engine built with `inheritedSecrets`, nothing recorded into the pass map) is masked on both arms, cause included; the strict arm with a non-`Error` thrown value renders the exact masked prefix with the string masked as `cause`, and with a thrown plain object renders `[object Object]` and threads no `cause`; the alias pass (an intrinsic `Export.Name`) is masked against its own resolution's entries on the default arm and, under `--strict-getatt`, aborts the deploy from that call site with the masked clone and the marker; a non-`Error` thrown value renders as the exact `raw failure mentioning ***`; an empty message under an SDK-style name renders the name; a value nothing recorded prints as is (the mask is a needle set, not a blanket).
- **Real `IntrinsicFunctionResolver` over faked `@aws-sdk/client-secrets-manager` / `client-ssm`** (the real class is wrapped at the mock seam so its top-level `resolve` rejection is captured, since the engine's `cause` no longer carries the plaintext): a premise case measures the password on the error the resolver threw (both `GetSecretValue` calls counted); the default and strict arms are masked, the strict `cause` included; `formatError` on the strict error and on a `CdkdError` wrapping it renders `***` and never the password; a value in the SERVICE position (`{{resolve:${Pw}}}`) — the arm warns `Unsupported dynamic reference service: ***` at default verbosity, `continue`s, and the output resolves, so that line is the whole exposure; the `Fn::Join` re-entry seam; for each of `ssm` / `ssm-secure`: the debug echo of an assembled name (the real SDK's `ParameterNotFound` names nothing — the echo is the exposure; the fake names the parameter, so the engine warn on that shape is asserted as the exact masked line too, which the file header calls what it is: redundancy over the mocked-resolver cases, not a live SSM shape; the fake refuses any command but `GetParameter`, like the Secrets Manager one), the throttle-retry label (`Retrying <service>:/probe/***`, the first lookup answered with a throttle and the backoff replaced through the exported `dynamicReferenceRetryDelays.sleep` seam) and the unrecognized-`Type` warn (a lookup that succeeds with `Type: 'Weird'`); the Secrets Manager retry label with the password as the secret ID; every log level is checked for the password on each of those paths except the `Fn::Join` case, which checks the failure warn only. The `Fn::Sub` template is built per case: `resolveSub` resolves the variable map IN PLACE, so a shared literal would carry the resolved password into the next case.

**Live arm** (`tests/integration/secrets-dynamic-ref/`): under `CDKD_TEST_OUTPUT_LEAK=true` the stack declares `OutputFailureLeak` — `Fn.sub('{{resolve:secretsmanager:<literalSecretName>:SecretString:${Pw}}}', { Pw: '{{resolve:secretsmanager:<literalSecretName>:SecretString:password}}' })` — for ONE `--verbose` probe deploy verify.sh runs right after Phase 1 (a no-change deploy, so only the outputs pass does work; the engine warns, skips the output and exits 0; a non-zero exit is reported with the log routed through `diag_output`). Guard 1b: premise, the synthesized `Value` equals that `Fn::Sub` pair by EQUALITY against this run's secret name (an inert look-alike, a wrong secret, `${Pw}` anywhere but the key position, or a CDK token turning the body into an `Fn::Join` object all fail it); the `Failed to resolve output OutputFailureLeak` warn is present (the sentinel that the arm ran) and isolated as one line (a consistency branch distinguishes "not isolated" from "no mask"); it does not carry `EXPECTED_PASSWORD`; it carries `***`; the whole `--verbose` log is password-free; and the resolver's `Resolving dynamic reference: secretsmanager:<name>:SecretString:***` echo of the assembled reference is in it (the premise that the echo is emitted at all, and the key position masked) — the throttle-retry label and the SSM `Type` warn need a throttle / an SSM shape this deploy does not produce and stay unit-pinned only. The log is captured, not shown — it reaches the terminal only through `diag_output`, which withholds it when it carries any of the fixture's three secrets. That helper and the guard's own checks use bash substring tests, and the fixture's 42 other `printf '%s' "${VAR}" | grep -q <needle>` sites are now here-strings (`grep -q <needle> <<< "${VAR}"`, same flags and needle — the remedy go-to-k/cdkd#2582 prescribes; the canonical `gone_probe` block is the one site kept verbatim, as that issue and the `integ-verify-probe-not-found` fence require), because that pipeline shape reports false negatives and the mechanism is now pinned rather than guessed (your item 4). It is the SMALL-payload arm of go-to-k/cdkd#2582, whose measurement covers the large one: bash's BUILTIN `printf` writes a multi-line argument in several write-family calls (`/proc/<pid>/io` syscw deltas of the writing bash alone: the 13-line plan text → 8, thirteen one-character lines → 13, a 10 KB single line → 3, a 60 KB single line → 15), `grep -q` (GNU 3.11) exits at the first matching complete line and closes the pipe, and a write the printf subshell still has pending after that takes SIGPIPE — `PIPESTATUS` reads `141 0`, `set -o pipefail` makes the pipeline 141, `if !` takes the FAIL branch and a leak check `if printf | grep -qF "${EXPECTED_PASSWORD}"` skips it. So it needs no capture over a pipe-buffer threshold: the race can occur on any multi-line capture with the needle before the last line, and a 1 KB, 13-line capture did. Seen live twice on this fixture (Guard 1b's first run, and this round's first integ run, which failed the pre-existing `SKIPPED` plan guard over a log that visibly carried the line); measured offline on that 1045-byte plan text at 67/3000, 31/2000, 6/2000 and 0/2000 in different load windows, every miss `141 0` (Codex reproduced it from the posted harness, 2/2000). A single-line payload did not race at any size tried, consistent with grep waiting for the newline or EOF before matching — which is why the 5 MB single-line probe answered `0 0` on both sides; an external `/usr/bin/printf`, a here-string and a `[[ == * ]]` test never missed in 2000 runs each. The measurements are posted on go-to-k/cdkd#2582; the repo-wide conversion stays that issue's. The output is gated because `cdkd diff` resolves outputs under `skipDynamicReferences`, where the body assembles without throwing, so a permanently declared output would be a phantom `ADD` under the fixture's later unchanged-stack `diff --fail` guard (a pre-existing class for any output that fails under the default arm — filed separately). `SYNTH_TEMPLATE` moved above Phase 1b; the probe deploy re-synthesizes `cdk.out`, which later premise guards now read.

## Verification

- `vp check --fix` / `vp run check` / `vp run typecheck:test` / `vp run build`: 0 errors; full `vp test run` at the default timeout on the final tree: 934 files, 20,031 passed (1 skipped), no type errors, root-asserted in the PR worktree after the rebase onto `f068f105` (933 files, 20,028 passed on the same content rebased onto `c328db66` an hour earlier). Runs alongside another session's suite on the same host had `tests/unit/scripts` critics time out at their 5 s / 60 s budgets; each such critic passes alone and the run reported here had no timeout. One critic on the rebased base, `check-scope-checker-inputs` ("every existing out-of-tree read target is inside the check include"), fails on any checkout that has a gitignored `.claude/settings.local.json` (main's new `settings-bash-first-optout.test.ts` reads it); this branch does not touch either file, the local settings file was set aside for the run, and the gap is filed as go-to-k/cdkd#ISSUE_SETTINGS.
- `vp run gen:all-matrices` / `vp run audit:coverage:check`: the CLI flag-coverage matrix gained `secrets-dynamic-ref` under `--verbose` (staged); nothing else dirty.
- `/run-integ secrets-dynamic-ref` (the live arm): PASS (582 s) on the final fixture and build (ledger row) — Guard 1b under `--verbose` printed the masked warn and found the masked resolver echo, the whole log was free of all three fixture secrets, the pre-existing plan guards passed on their here-string form, `scrub --dry-run` and the later `diff --fail` guard clean, 0 errors, 0 orphans, 0 surviving state versions. Four runs across the two review rounds: the first (pipeline-form guards) passed Guard 1b and then failed the pre-existing `SKIPPED` plan guard over a log that carried the line — the race described under the live arm; the second, on the here-string form, passed end to end at 832 s under suite load; the third (678 s) was round 1's ledger row; the fourth is this one, on round 2's build.
- `/run-integ lambda` (broad; `deploy-engine.ts` and `intrinsic-function-resolver.ts` are in scope): PASS (108 s) on round 2's build — 0 errors, 0 orphans; event-log versions swept (108 s on round 1's build). `getatt-fallback-guard` (the `--strict-getatt` output-failure path live, on a GetAtt miss that names no secret): PASS (225 s) on the PR's first round; that fixture's non-secret GetAtt failure keeps the same message and promotion behavior under this round's strict arm, and the round's changes to it — the masked cause, the inherited bag, the `error.name` fallback — are unit-pinned.
- Mutation probes on round 1's delta, each restored byte-exact and re-measured on the 25-case file as it stood at the end of round 1 (the final file has 26; round 2 added the alias strict case): unsupported-service warn unmasked (1 case fails); cause kept raw (5); string cause unmasked (1); inherited bag dropped (4); recorded bag dropped (14); the value-pass site handed `EMPTY_SECRETS` as the inherited bag (3), the alias-pass site (1); name fallback dropped (1); cause replaced by a same-message copy (2); cause replaced by `new Error(masked)` (2); cause masked only while still the original (1); detail masked only until the first `***` (1); a non-Error object cause stringified (1); the text left unmasked on both arms (17). Round-2 probes on the final 26-case file: a non-Error object threaded as the cause after all (1); the alias-pass catch swallowing its error instead of calling the handler (3); the strict arm's cause left raw (6). The `?? EMPTY_SECRETS` on the recorded bag replaced by a cast survives — the equivalent mutant the comment documents, since `buildResolverContext` always sets the map. First-round probes (the alias/value sites handed `EMPTY_SECRETS`, each resolver echo and label, `context` dropped per arm, Guard 1b's jq against six wrong shapes) were taken on the first-round file and are unchanged by this round.
- Reviewers on round 1's delta: `pr-code-reviewer` (no blocker; a false test comment and the "not isolated" comment's reason, both taken), `pr-security-reviewer` (clean at blocker and should-fix; flow trace of the masked cause through `isMarkedNonRetryable` / `formatError` / `handleError` / `extractDeploymentEventError` / `util.inspect`; the whole-log negative widened to all three fixture secrets), `pr-test-reviewer` (three gaps — the inherited bag on the alias pass, both bags on one error, a thrown object — all now cases and re-probed). First-round reviewers as before.
- Codex: round 1's delta to NO ISSUES (2 rounds — round 1: the "not isolated" comment overclaimed split-line detection), the maintainer-checklist proxy to NO ISSUES (2 rounds — round 1: the handler doc comment still said "every link" / "whole chain" where `maskSecretsInError` is bounded), round 2's delta to NO ISSUES (3 rounds — round 1: "the classifiers walk `Error` links only" was false, `isMarkedNonRetryable` follows `.cause` through any object; the fixture comment's "every other content check is a here-string" missed three checks that grep a file), the go-to-k/cdkd#2582 comment (5 rounds — counts, the write-count attribution, hedging the single-line explanation), this body (5 rounds) and the replies (2 rounds each).


go-to-k (changes requested):
Reviewed at the **3-axis tier plus the security add-on** (4 parallel read-only reviewers: spec / code / test / security).

Tier derivation, for the record: 818 LOC after excluding `docs/_generated/**` (824 raw - 6) across 9 files = `1-reviewer` by size, raised one step by the recent-defect trigger — `src/deployment/intrinsic-function-resolver.ts`'s last three commits on the base are all `fix:` (#2511 secret-redaction, #2492 ssm-secure, #2465), and `src/deployment/deploy-engine.ts`'s are 2 of 3 (#2562, #2695). The security reviewer is additive because this is a secret-masking fix.

## What holds

The core fix is correct, and the load-bearing claims in the body check out:

- **The "every sink renders one level" invariant is true on this path today.** Every reader of the strict arm's `new Error(masked, { cause: RAW })` was traced: `formatError` (`src/utils/error-handler.ts:767`) prints `Caused by:` only for a `CdkdError`, and this is a plain `Error`; `handleError:802` logs the top link's stack, whose first line is the masked message; `extractDeploymentEventError` (`src/types/deployment-events.ts:327`) takes the outer `message` and walks deeper only for `$metadata`/`Code`; `retryClassificationText` needs `markRedactedCause`, which is absent; and `src/cli/index.ts:77`'s `util.inspect` — the repo's own documented multi-level sink — is unreachable because `deploy.ts:1290` wraps the action in `withErrorHandling` and `--strict-getatt` is set only at `deploy.ts:923`.
- `context` is threaded at all three arms (`intrinsic-function-resolver.ts:7015`, `:7023`, `:7080`) and those are the only call sites of the two lookup methods.
- The alias-pass binding is correct: the `finally` at `deploy-engine.ts:6746` mutates the same `Map` object `outputsPassSecrets` aliases at `:6641`, so the merge lands before the `catch` at `:6760`.
- `?? EMPTY_SECRETS` is genuinely unreachable (`buildResolverContext:1199` always sets it).
- The gated `OutputFailureLeak` output cannot reach the later `diff --fail` guard (`resolutionFailed` suppresses the persist), and the fixture's `EXIT`/`INT`/`TERM` trap destroys the stack and the out-of-band SecureString on every Phase-1b `exit 1`.
- **Unit tests discriminate.** Reading the assertions rather than trusting the body: removing the engine's `detail` mask reds exactly the cases at `tests/unit/deployment/deploy-engine-output-failure-masking.test.ts:279, 288, 311, 327, 335, 392, 402, 422, 540` = 9; strict-arm-only = `288, 335, 422` = 3; default-arm-only = the other 6. That matches the body's matrix. Every `toContain('***')` has a companion `not.toContain(PASSWORD)`, the control at `:348` asserts `not.toContain('***')`, there are zero `*Once` primers, and the mocked message at `:166` reproduces `intrinsic-function-resolver.ts:7269` verbatim rather than fabricating a shape.

## Blocker

### 1. One echo in the same function is still unmasked, at default verbosity

`src/deployment/intrinsic-function-resolver.ts:7095`:

```ts
```

`service = inner.split(':')[0]`. This is the same assembled-token class the PR closes at `:7234`, `:7253`, `:7486`, `:7501` and `:7547` — `resolveSub` re-enters `resolveDynamicReferences` with the assembled string, so a template that lets a resolved value land in the SERVICE position makes `service` BE the plaintext:

```yaml
Fn::Sub:
  - "{{resolve:${Pw}}}"
  - Pw: "{{resolve:secretsmanager:MySecret:SecretString:password}}"
```

`Pw` resolves and records the password; the re-entry matches `{{resolve:<password>}}` and this line prints it. It is strictly worse than the echoes the PR did mask, which are `logger.debug` and therefore `--verbose`-only — this one is a `warn`, emitted at default verbosity. `context` is in scope; it is used two statements later at `:7107`.

Fix is one line — the same wrapper the five sibling sites already use:

```ts
this.logger.warn(
  this.maskSecretsForLog(`Unsupported dynamic reference service: ${service}`, context)
);
```

(The same arm's `continue` also leaves the literal `{{resolve:<password>}}` span in the value, where `redactOutputs`' "strictly inside a complete span" carve-out spares it into `state.json` — that half is pre-existing and out of this PR's scope; filed as #2743.)

## Should fix before merge

### 2. The `cause` does not have to stay unmasked — `maskSecretsInError` exists and preserves the marker

`src/deployment/deploy-engine.ts:6591` keeps the raw error as `{ cause }`, and the doc comment argues this is forced because `isMarkedNonRetryable` walks the chain for a non-enumerable symbol. That constraint is already solved in this repo: `maskSecretsInError` (`src/deployment/secret-redaction.ts:4994`) clones every link via `Reflect.ownKeys`, and its own comment says so — *"Symbols are included, which is what carries `markNonRetryable`'s marker."* The precedent for the identical constraint is 2700 lines up in this same file, at `deploy-engine.ts:3836`, where it is used specifically so `$metadata` and a nested `cause` survive for the classifiers.

Two reasons to take it even though no sink renders two levels today:

- The invariant as written in the new doc comment is broader than the code supports. `src/cli/commands/scrub.ts:142` `describeFailure` **is** a full-chain renderer that exists right now; it is safe only because `scrub.ts:3744` masks the whole chain at its boundary. So "every sink renders one level" is already false repo-wide, and the deploy path's safety rests on reachability, not on a property of the sinks.
- The new tests at `deploy-engine-output-failure-masking.test.ts:307` and `:432` **pin** the unmasked cause with `toBe`. A later hardening now costs a test edit, which is the wrong direction for a GHSA-p5qg-v9gv-hc7w-class file.

Suggested: `{ cause: maskSecretsInError(error, secrets) }`, and relax those two assertions to check the marker rather than object identity.

### 3. The new mask ignores `inheritedSecrets`, while the resolver's mask does not

`src/deployment/deploy-engine.ts:6571` masks against `recordedSecretValues` only. The resolver's `maskSecretsForLog` (`intrinsic-function-resolver.ts:6776-6795`) masks against **both** bags, `inheritedSecrets` first, and its comment names exactly this hazard from issue #1903 round 2: on a nested-stack child, `context.parameters` holds parent-decrypted plaintext that is not in `recordedSecretValues` until `recordInheritedParameterSecrets` fires, so *"Every log line emitted BEFORE that moment ... printed the secret"*.

The security reviewer could not construct a shape that reaches this handler before that recording (every route through `${Param}` goes via `resolveRef`, which records), so this is defense-in-depth rather than a demonstrated leak. But two masking sites in one flow should not argue opposite sides of the same question, and `context.inheritedSecrets` is on this very context (`deploy-engine.ts:1259`). Three of the four reviewers raised it independently.

### 4. The `diag_output` rewrite rests on a measurement that does not reproduce, and 43 sibling sites contradict it

`tests/integration/secrets-dynamic-ref/verify.sh:189-198` states that `printf | grep -q` yields a false negative on secret-bearing text via SIGPIPE under `pipefail`, "measured offline with a password followed by 1 MB of filler". The code reviewer could not reproduce it: bash 5.3.9 aarch64-darwin24.6, needle at byte 0 plus 5 MB of filler, `PIPESTATUS=0 0`, rc=0 (match).

Meanwhile **43** `printf '%s' ... | grep -q` sites remain in this same file, roughly 15 of them leak assertions with the identical shape and stakes — `:687`, `:696`, `:708`, `:732`, `:736`, `:1162`, `:1166`, `:1342`, `:1350`, `:1526`, `:1532`, `:1725`, `:1731`, `:1855`. Either the class is real, in which case those are silent false-negative leak checks and the PR fixed one instance of a live bug, or the comment asserts something untrue. Please pin the reproducer (exact bash and grep builds, filler size, `PIPESTATUS`) or convert the leak assertions wholesale; one site rewritten on a rationale the other 43 contradict is the worst of the three states.

### 5. The live arm cannot exercise the resolver-side half of the fix

`verify.sh:305` runs the probe deploy **without `--verbose`**, so the resolver's debug echoes (`:7234`, `:7486`) are never emitted, and the retry labels (`:7253`, `:7501`) require a throttle. The comment at `verify.sh:368-371` nonetheless says the whole-log negative covers "the resolver's own lookup echo, retry label and Type warn" — it cannot; those lines are suppressed.

Guard 1b itself is sound and non-vacuous for what it does cover (premise by equality, the warn sentinel, the `***` positive and the whole-log password negative), so this is about the claim, not the guard. Either add `--verbose` to the probe deploy — which would make the arm cover the resolver echoes for real — or narrow the comment to the engine warn.

## Minor

- `deploy-engine.ts:6572` — `error.message` can be empty where `String(error)` was not: `new Error()` now yields `Failed to resolve output X: `, and for an AWS SDK error the class NAME often carries the code while the message is generic. `error.message || error.name` keeps the strictly-better half of the change.
- `verify.sh:305` — a failing probe deploy aborts under `set -e` with the log captured and never printed, so the integ dies with zero diagnostic. Route it through `diag_output` on a non-zero status.
- `verify.sh:357-377` — if the loop isolates no line, `OUTPUT_FAILURE_WARN` stays empty and `:377` reports "carries no mask" rather than "warn line not isolated". Distinguish the two.
- `tests/unit/deployment/deploy-engine-output-failure-masking.test.ts:124` — the fake SSM error is `ParameterNotFound: GetParameterCommand`, which deliberately names nothing, so on both SSM cases the ENGINE warn has nothing to mask and `:476` only asserts `toBeDefined()`. Putting the parameter name in the fake message and asserting the masked warn would close that.
- `.claude/rules/layout-deployment-secrets.md` — the edit at `:63-70` is accurate, but the file still records neither `maskSecretsForLog` nor the new masking sites. The changelog entry is 1995 chars against the 2000 cap, so any wording addition needs a trim elsewhere.

## Verification notes

`docs/changelog-cdkd.md` and `.claude/rules/layout-deployment-secrets.md` were read for parity; the two prefix-less consumers of `Failed to resolve output <key>` were checked and are unaffected (`tests/integration/getatt-fallback-guard/verify.sh:242` needles the strict arm; `tests/integration/raw-cfn-conditions-params/verify.sh:120` is an absence grep). No deletion-logic, credential-surface or injection change in the diff. The fixture's "secret" is the committed constant `cdkd-known-pw-123`, so even a CI-log leak of the probe would disclose nothing.

Happy to re-review once (1) is in; the rest are judgement calls I would like to see answered rather than silently kept.

go-to-k (changes requested):
Round 2, re-reviewed at **3-axis plus the security add-on** (4 parallel read-only reviewers).

The tier moved on its own this round: 1213 LOC after excluding `docs/_generated/**` (1220 raw - 7) across 11 files is `3-axis` by SIZE now, where round 1 reached that tier only through the recent-defect bias. The second `fix:` subject adds the commit-count up-bias on top (clamped).

## Round-1 dispositions

Seven of eight taken, verified independently:

| Round-1 item | Disposition |
|---|---|
| **blocker** — `:7095` unsupported-service warn unmasked | **FIXED** — `intrinsic-function-resolver.ts:7102-7104` wraps it in `maskSecretsForLog(..., context)` |
| **major** — unmasked `{ cause }` on the strict arm | **FIXED** — `deploy-engine.ts:6598-6603`, `:6624`. The marker survives the clone: `markNonRetryable` defines an own non-enumerable `Symbol.for`, `maskSecretsInError`'s `Reflect.ownKeys` + `Object.create(getPrototypeOf(...), descriptors)` carries it, `instanceof` is preserved, and the double-mask re-clones from a clone that still owns the symbol. `$metadata` / `Code` / `name` copied the same way |
| **major** — handler ignored `inheritedSecrets` | **FIXED** — `deploy-engine.ts:6599` iterates `[inheritedSecrets, secrets]`, byte-for-byte the order at `intrinsic-function-resolver.ts:6791-6794`; the bag read at `:6671` is THIS engine's (`buildResolverContext:1258-1261`), i.e. the child's, not the parent's |
| **major** — the `printf \| grep -q` split | **FIXED and complete within the file** — 42 content checks converted; the only surviving pipeline is `verify.sh:83` inside `CANONICAL_HELPER_BLOCK`, still byte-identical to `scripts/check-integ-probe-not-found.ts:57`. Every converted needle is a non-empty literal or a constant with no newline, so no `-F` constituent can be empty and `<<<`'s appended newline cannot flip a verdict |
| **major** — live arm could not exercise the resolver echoes | **FIXED** — `--verbose` at `verify.sh:338`, and the arm now EXERCISES rather than merely enables: `:431` is a positive sentinel requiring `Resolving dynamic reference: secretsmanager:${SECRET_NAME}:SecretString:***`, which fails pre-fix. The over-claiming comment is narrowed at `:399-405` to concede the retry label and `Type` warn are unit-pinned only. Ledger row: `secrets-dynamic-ref ... PASS 678` |
| **nit** — `error.message` can be empty | **FIXED** — `deploy-engine.ts:6597` `error.message \|\| error.name`, pinned by the `__nameless__` case |
| **nit** — failing probe deploy died with no diagnostic | **FIXED** — `verify.sh:335-343` |
| **nit** — unisolated warn line misreported | **FIXED** — `verify.sh:396-404`, distinct message |
| **nit** — strict arm on the ALIAS call site untested | **still open** — `Export.Name` failures are exercised on the default arm only (`test:390`, `:453`); nothing pins that `--strict-getatt` aborts from `deploy-engine.ts:6797` |

Doc parity is clean: `cli-flag-coverage.json` and `.md` compared programmatically — 37 `--verbose` fixtures, identical sets, every other flag row matching, the sole delta being `secrets-dynamic-ref` as the correct consequence of `verify.sh:338`. Changelog entry 1968 chars against the 2000 cap. `.claude/rules/layout-deployment-secrets.md:62-73` now records `maskSecretsForLog` and every new site.

## Two corrections to my round-1 review

**Round-1 finding 4 is withdrawn.** I wrote that the `diag_output` rewrite rested on a SIGPIPE measurement that "does not reproduce". That non-reproduction was run with the wrong shape — needle at byte 0 of a single-line payload plus filler — which is precisely the configuration your go-to-k/cdkd#2582 comment measures at 0 misses. Your second arm (builtin `printf` issuing several write calls; `grep -q` closing stdin at the first matching COMPLETE line; the pending write taking SIGPIPE; `PIPESTATUS` reading `141 0`) is the mechanism, and it needs no size threshold. The class is real and my objection was not.

For the record, I re-ran your harness verbatim on this host — bash 5.3.9 aarch64-apple-darwin24.6, `/usr/bin/grep`, needle on line 12 of the 13-line 1045-byte payload — and got 0/2000 idle and 0/2000 under eight busy-loop CPU hogs, for the pipeline, the here-string and the substring test alike. The BSD grep build does not close stdin early enough to lose the race. That is now recorded in go-to-k/cdkd#2582 so a later "cannot reproduce on the Mac" is not read as a disproof, and that issue's `Severity` is raised to `high` on the strength of your recount — 14 of the bare-`if` sites grep for a SECRET-valued needle across 5 fixtures this PR does not touch.

**The residual half of that finding is answered**, by the split you proposed rather than by a wholesale conversion here: 42 sites in this fixture, the rest with go-to-k/cdkd#2582 plus its classifier. That is the right boundary.

## Blocker

### The PR body was not rewritten for round 2, and now contradicts the code it describes

This is the only thing standing between the branch and a merge. The repo squash-merges, so the body becomes the permanent commit message — a body asserting the opposite of what shipped is not a cosmetic problem.

| Body says | Tip does |
|---|---|
| "The strict arm **keeps `{ cause: error }` unmasked**" | `deploy-engine.ts:6598-6603` masks the cause per bag; `:6624` throws the clone |
| "takes the outputs pass map as a **fourth** parameter … masks with `maskSecretsInText`" | fifth parameter `inheritedSecrets` (`:6589`), a loop over **two** bags, inherited first (`:6599`) |
| "the default arm prints `error.message`" | `error.message \|\| error.name` (`:6597`) |
| "**18 cases**", and every mutant count measured on that file | 22 `it(` cases; the strict-arm cause cases now assert clone + marker, not `toBe` identity |
| `/run-integ secrets-dynamic-ref` "PASS (516 s)", `lambda` "PASS (118 s)" | ledger rows say 678 s and 108 s |
| silent | the `:7095` fix, the `--verbose` arm and its positive sentinel, the 42-site conversion and the go-to-k/cdkd#2582 split |

Please rewrite it against the tip. The verification section in particular should describe the runs that actually gate this branch, not round 1's.

## Out of scope — filed separately, no action needed here

The security reviewer re-enumerated the class from scratch rather than diffing against round 1's list, and found one more site of the same shape: `intrinsic-function-resolver.ts:2602-2605`, `evaluateConditions`' per-condition catch, printing a raw resolver `error.message` at **warn** level. A `Conditions` entry whose `Fn::Sub` assembles a reference out of a value the same pass resolved from a secret reaches it, and `Dynamic reference: key '<password>' not found in secret 's'` (`:7278`) lands on an ordinary `cdkd deploy` log with no flag set. `context` is the method's own parameter, so the fix is the same one-liner.

Filed as go-to-k/cdkd#2748 rather than asked for here: it is a catch-all warn in a different method, rendering any error rather than a lookup echo, and this PR is already at 1213 LOC. The issue also raises the question worth deciding once — after two misses, whether these should mask where the text is BUILT (`:7278`, `:6899`, `:7087` all embed `fullMatch`) instead of at each sink.

## Minor and nits

- **`deploy-engine.ts:6598-6603`** — a thrown value that is neither an `Error` nor a string is threaded as `cause` UNMASKED (`maskSecretsInError` returns a non-`Error` by identity), and `deploy-engine-output-failure-masking.test.ts`'s `__leak_object__` case PINS that with `toEqual({ code: 'NotAnError', detail: LEAKED })`. My reviewers split on whether it reaches a sink, so I settled it directly: it does **not**. `withErrorHandling` (`src/utils/error-handler.ts:823-833`) catches and calls `handleError`, which ends in `process.exit` (`:814`), so `main().catch`'s `console.error('Fatal error:', error)` at `src/cli/index.ts:75` is unreachable for a deploy, and `formatError` never descends to a non-`Error` cause. So this is not a live leak. It is still worth dropping, because threading it buys nothing — `isMarkedNonRetryable` walks `Error` links only — and the test currently pins the unsafe shape as if it were the contract. `...(cause instanceof Error || typeof cause === 'string' ? { cause } : {})`.
- **`deploy-engine.ts:6598-6603`** — the `cause` masking runs on BOTH arms and is discarded on the default one. `maskSecretsInError` clones the whole chain and reads each link's `stack` accessor, forcing V8 trace materialization, once per bag per failed output, on the warn path where the cause is never used. Move it inside the `strictGetAtt` branch.
- **`intrinsic-function-resolver.ts:7099`** — the comment reads as if the leftover span were cosmetic. Measured against this tree, `redactSecretsForState({Leak:'{{resolve:cdkd-known-pw-123}}'}, secrets)` returns the plaintext unchanged via the strictly-inside carve-out at `secret-redaction.ts:4611-4700`, so it persists into `state.outputs`. go-to-k/cdkd#2743's body says this; the comment a future reader lands on does not. Add "…which redaction spares into `state.json`".
- **`verify.sh:195-197`** — the comment says this file's content checks "are here-strings … every site but the canonical `gone_probe` block", but six checks in Guard 1b (`:382`, `:405`, `:416`, `:424`, `:431`, `:436`) and all of `diag_output` are `[[ == * ]]` builtin tests. State the invariant that actually holds: no content check is a `printf | grep` pipeline.
- **`deploy-engine-output-failure-masking.test.ts:136`** — the SSM fake throws `ParameterNotFound: parameter '<name>' not found`, a message the real SDK does not produce (disclosed in the file's own header). The engine-side assertion at `:646` is therefore coverage of a shape that cannot occur live; the genuine engine masking is already pinned against realistic shapes by the mocked-resolver cases, so this is redundancy rather than a false pass — but the disclosure should say so.
- **nit** — the SSM fake accepts any command name, unlike the Secrets Manager fake, which refuses unexpected commands; an unmodelled SSM call would silently answer `ParameterNotFound`. The throttle fakes set only `name`, where a real AWS throttle also carries `$metadata.httpStatusCode` (`isThrottlingError` matches either, so behaviour is right).
- **nit** — round-1's last item is still open: no case pins the strict arm on the alias call site.

## Test adequacy

Every fix-round site has a discriminating test, and the discrimination was checked by reading assertions rather than by trusting the body: `:7102` unsupported-service warn is pinned by a full-line `toBe('Unsupported dynamic reference service: ***')` plus a whole-log negative; the masked cause by five cases including the marker-survival pair; the `inheritedSecrets` bag by four (value pass, alias pass, both bags on one error, strict cause); `message || name` by the `__nameless__` case. Every `toContain('***')` has a companion plaintext negative, no case rests on `toBeDefined()` alone, and each whole-log negative is guarded by a positive sentinel (`toHaveLength(1)` / `echoes.length >= 2`) so none can pass on an empty log. No `*Once` primers at all; the hand-rolled once-semantics are drained in `beforeEach` and proven consumed. No shared mutable fixture — `assembled()` and `ssmAssembled()` are factories, so `resolveSub`'s in-place mutation cannot cross cases.

---

Code-wise this is ready. Rewrite the body against the tip and I will re-review and clear it.

nix-tkobayashi:
Thanks — the body is rewritten against the tip (it was drafted for round 1's response and sat in my working notes instead of on the PR; that is on me), and the minors are taken in 9052acca (one flattened commit, rebased onto main for the changelog conflict; the fork tip was mine both times). Round 1's answer is the comment above this one; the round-2 items:

### Blocker — the PR body
Rewritten end to end against the tip: the five-parameter handler with the two bags, the masked cause and the arms it is masked on, `error.message || error.name`, the `:7095` warn, the `--verbose` arm with its positive sentinel, the 42-site conversion and the go-to-k/cdkd#2582 split, 26 cases with the mutant counts re-measured on that file, and a Verification section that describes the runs gating THIS branch (ledger rows 582 s and 108 s, the four `secrets-dynamic-ref` runs across the two rounds and what each did). Codex reviewed the body to NO ISSUES in 3 rounds.

### Minors
- **Non-`Error`, non-string thrown value** — not threaded any more: the options argument is `cause instanceof Error || typeof cause === 'string' ? { cause } : {}`, with the reason in the comment (`maskSecretsInError` would hand it back unmasked, and `markNonRetryable` marks `Error` instances only, so it can carry no marker of ours). The `__leak_object__` case now pins `not.toHaveProperty('cause')`; threading it unconditionally fails that case.
- **Cause masking on the default arm** — moved inside the `strictGetAtt` branch; the warn arm masks the text only. Leaving the strict cause raw fails 6 cases.
- **`:7099` comment** — says the leftover span is spared into `state.json` by redaction's strictly-inside carve-out (go-to-k/cdkd#2743).
- **`verify.sh:195-197`** — states the invariant that holds: no content check in the file is a `printf | grep` pipeline; the helper and Guard 1b are `[[ == * ]]` tests, every other check on a captured value is a here-string, the three checks that read a file grep the file, and the canonical `gone_probe` pipeline stays verbatim.
- **`:136` SSM fake** — the header now says the engine-side assertion on the SSM shapes is redundancy over the mocked-resolver cases, not coverage of a live shape; the fake refuses any command but `GetParameterCommand`, like the Secrets Manager one. The throttle fakes keep `name` only, as you note `isThrottlingError` accepts.
- **Strict arm on the alias call site** — a new case: an `Export.Name` `Fn::Sub` that fails under `--strict-getatt` aborts the deploy; it asserts the masked message prefix up to `(--strict-getatt` (no plaintext), the cause's exact masked message, and `isMarkedNonRetryable` on the thrown error. A mutant that makes the alias catch swallow its error fails it and the two default-arm alias cases; the raw-cause mutant fails it too.

go-to-k/cdkd#2748 (`evaluateConditions`' catch) noted; not touched here.

### Verification
- `vp check --fix` / `vp run check` / `vp run typecheck:test` / `vp run build`: 0 errors; full `vp test run` at the default timeout on the final tree: 934 files, 20,031 passed (1 skipped), root-asserted in the PR worktree after the rebase (933 / 20,028 before it, same content).
- `/run-integ secrets-dynamic-ref`: PASS (582 s), 0 errors, 0 orphans; `/run-integ lambda`: PASS (108 s), 0 orphans — both on this round's build.
- Codex on the delta to NO ISSUES (3 rounds — round 1: "the classifiers walk `Error` links only" was false, and the fixture comment missed three file-input greps), on this reply (5 rounds).

Nothing further from my side — over to you.

go-to-k (approved):
Round 3, re-reviewed at **3-axis plus the security add-on** (1255 LOC after excluding `docs/_generated/**`, 11 files; two distinct `fix:` subjects add the commit-count up-bias, clamped). The original design is accepted — this round is scoped to the fix delta, which you flattened and rebased, so the reviewers took it per file against `46cc76f6`.

**No blockers, no majors. The delta is correct and the body is accurate.** Every round-2 item is verified fixed, with a citation:

| Round-2 item | Verified |
|---|---|
| PR body described the opposite of what shipped | Rewritten; every load-bearing claim checked against the tip — the two bags masked inherited-first (`deploy-engine.ts:6599`), `error.message \|\| error.name` (`:6598`, now the only occurrence in `src/`, so the earlier round's "not anywhere" is genuinely superseded), the `ERROR_CAUSE_MASK_MAX_DEPTH` bound (`secret-redaction.ts:4880` = 20), the strict-arm-only masking, the `?? EMPTY_SECRETS` claim, **26 cases** (23 `it(` plus 3 inside a 2-element service loop), and the 42-site conversion (measured: 46 `printf\|grep` lines at the base → 1 at the tip, `gone_probe:83`) |
| Non-`Error` cause threaded unmasked and pinned by a test | Not threaded (`deploy-engine.ts:6634`); the old `toEqual({… detail: LEAKED })` **asserted the leak**, so replacing it loses no mutation — a legitimate exception to keep-both. `new Error(m, {})` installs no own `cause`, so `not.toHaveProperty('cause')` is sound |
| Cause masking ran on both arms | Inside `if (strictGetAtt)` (`:6600-6610`); the warn arm consumes only `detail`, still masked with both bags at `:6599` |
| `:7099` comment understated go-to-k/cdkd#2743 | Now states the span is spared into `state.json`; verified — `DYNAMIC_REFERENCE_INNER` (`secret-redaction.ts:1988`) is byte-identical to the resolver's `([^}]+)`, so `{{resolve:<plaintext>}}` is a complete span and a shorter needle is strictly inside → KEPT |
| `verify.sh:195-197` contradicted the file | Restated as the invariant that holds |
| SSM fake accepted any command | Refusal added (`test:134-136`); no existing case changes meaning |
| No strict-arm case on the alias call site | New case (`test:446-465`) — and it discriminates: the prefix is `^`-anchored across the whole masked detail, with a companion `not.toContain(LEAKED)`, so a regression appending plaintext after the prefix is caught; a swallowing alias `catch` leaves `thrown` undefined and fails `:460` |

**Security: clean, no findings.** The flow trace confirms the thing that matters most here — both call sites loop `for (const bag of [inheritedSecrets, secrets])`, so there is **no arm that masks against only the empty bag**. On a top-level stack `inheritedSecrets` really is `EMPTY_SECRETS` and `recordedSecretValues` does all the work; the map is captured by REFERENCE, so the resolver's in-place `.set()` during the pass is visible to a later failure, and the alias `finally` merges `nameSecrets` into that same object before the outer `catch`. A **string** cause is masked with both bags at `:6606-6609` before being threaded, so nothing unmasked reaches `formatError`, `handleError`'s stack log, `util.inspect`, `extractDeploymentEventError`, or scrub's `describeFailure`.

That check was not routine this round. A sibling branch (go-to-k/cdkd#2755) found the same helper is a SILENT no-op when the context carries no bag: masking the line there passed five unit cases and four mutation probes while a live `cdkd diff` printed the password in full, because two callers passed bagless contexts. Your two call sites do not have that shape.

## Nits — none blocking, take or leave

- **`tests/.../deploy-engine-output-failure-masking.test.ts:129-136`** — the command refusal sits *after* the throttle branch, so the fake does not quite "refuse any command but `GetParameterCommand`": one carrying a primed `input.Name` throttles first. The Secrets Manager twin puts its refusal first (`:85-87`). Unreachable today; moving it above `:129` matches the twin. The SSM fake's doc comment (`:107-115`) also does not mention the refusal, while the SM one does (`:69`).
- **`:443`** — `not.toHaveProperty('cause')` is structural. A mutant attaching the raw object under a different key would leak and still pass; its two siblings (`:462`, `:514`) carry whole-string plaintext negatives and this one does not. `expect(error.message).not.toContain(LEAKED)` would close it.
- **`tests/integration/secrets-dynamic-ref/verify.sh:196-200`** — the inventory reads as exhaustive and is not: about ten `case "${VAR}" in` glob checks on captured values (`:483, 603, 611, 617, 627, 712, 727, 1163, 1191, 1307`) are none of the three listed forms, and `:1163` is itself a `printf | jq` pipeline. One clause naming the `case` checks would make it true.
- **`deploy-engine.ts:6601-6603`** — "clones the whole chain and reads each link's `stack` accessor" reads as unconditional; `maskSecretsInError` returns by identity when the bag is empty or no message changed (`secret-redaction.ts:4994-4997`). The rationale for hoisting is right, the wording overstates it.
- **Body** — the mutation-probe bullet says the round-1 probes were re-measured on the final **25**-case file while the heading and your reply say **26**; "final" is used for both files. Worth one word so a later reader is not comparing two different files.

`20,031 passed` and the mutant fail-counts are the only claims not independently reproducible without running the suite and the probes; the file count (934), the ledger rows (582 s / 108 s) and the 42-site conversion all are, and all check out.

Approving. `pr-review` marker set against `9052acca`. I am not merging it — it is your contribution and the merge is the maintainer's call; say the word and I will run the merge gates.

One note for later, not for this PR: go-to-k/cdkd#2748 (`evaluateConditions`' catch, the site my security pass found while reviewing this) is fixed in go-to-k/cdkd#2755. That branch conflicts with this one by construction — same `layout-deployment-secrets.md` paragraph, same changelog anchor, same `lambda` / `secrets-dynamic-ref` ledger rows; the `src/` hunks do not overlap. Its rule paragraph was cut to 1,186 B specifically so both fit under `CORPUS_BYTES_MAX`, which had 393 B of headroom. Whichever of us lands second rebases — no action needed from you.

nix-tkobayashi:
Thanks for the round-3 pass and the approval. Please go ahead and merge — 9052acca is the commit you approved, and I am deferring the four optional worktree changes to preserve that commit.

On the five nits, one is taken and four are deferred, each recorded here:

- **Body, "25-case file"** — taken: the mutation-probe bullet now says round 1's probes were re-measured on the 25-case file as it stood at the end of round 1, and the round-2 probes on the final 26-case file. A body edit does not move the SHA your `pr-review` marker is bound to.
- **SSM fake: refusal after the throttle branch, and its doc comment** — deferred. Unreachable today, as you note. I will move the refusal above the throttle branch and mention it in the doc comment the next time that file is touched.
- **`:443` structural `not.toHaveProperty('cause')`** — deferred, and agreed: a whole-message `not.toContain(LEAKED)` beside it would give this case the plaintext negative its two siblings carry; a raw object attached under some other property is a shape none of the three message assertions would see, so that one stays a review-time check. Same next-touch promise.
- **`verify.sh:196-200` inventory omits the `case` glob checks and the `printf | jq` at `:1163`** — deferred; a one-clause fix I will fold into the next change to that fixture. The narrower invariant the comment states — no content check is a `printf | grep` pipeline, `gone_probe` excepted — still holds: the `case` matching is builtin, though the one at `:1163` takes its input through `printf | jq`; the inventory is what is incomplete.
- **`deploy-engine.ts:6601-6603` "clones the whole chain" reads as unconditional** — deferred; the identity return for an empty bag or an unchanged chain is real and the comment should say "clones ... when a message changes". Same next-touch promise.

Noted on go-to-k/cdkd#2755: whichever of us lands second rebases; if that is me, the `layout-deployment-secrets.md` paragraph, the changelog anchor and the `lambda` / `secrets-dynamic-ref` ledger rows are the known conflict sites and I will resolve them keep-both.

Nothing further from my side — over to you for the merge.

go-to-k:
Thanks — merging.

The four deferred nits are filed as go-to-k/cdkd#2756 so the promise outlives this thread. Not a disagreement with deferring them: all four are unreachable or cosmetic and none is worth moving the commit I approved. It is that "next time that file is touched" is not one of the three dispositions this repo tracks (fixed / issue / recorded won't-do), and a note in a merged PR's comments is the one place a follow-up reliably evaporates. The issue carries the file:line for each, your suggested phrasing for the `maskSecretsInError` comment, and `severity:low` / `effort:small`.

The body edit for the 25-vs-26 wording confirmed at `9052acca` — head unmoved, so the `pr-review` marker still binds to the commit that was reviewed.

Before the merge I am re-running the broad integ on YOUR tree rather than trusting the markers in my worktree. Your ledger rows are specific and post-date the round-2 fixes (`lambda` 108 s, `secrets-dynamic-ref` 582 s with the `--verbose` Guard 1b and the here-string conversion, `getatt-fallback-guard` 225 s), so this is not doubt about your run — it is that `integ-broad` and `integ-destroy` read the LOCAL working tree, and merging from a worktree holding a different branch's content is a false green by construction. That is the cwd-race signature this repo has been bitten by, and the gate is only worth having if it attests to what is actually merging.

go-to-k/cdkd#2755 rebases onto this once it lands; the conflict sites are the ones we both named.

Issue #2752, opened 2026-09-07
test(check-scope-checker-inputs): a gitignored .claude/settings.local.json reds the unit suite locally since #2738, as a false checker input


nix-tkobayashi:
## Summary

Since go-to-k/cdkd#2738, `vp test run` goes RED on any checkout that carries a `.claude/settings.local.json` (gitignored, so CI never has one and stays green). `tests/unit/scripts/check-scope-checker-inputs.test.ts`'s case `every existing out-of-tree read target is inside the check include` reports:

```text
checker inputs outside the check include — add the path to .markgate.yml `check.include` (see the issue #2364 comment there), or record it as a repo-wide-scanner known limit
+ ".claude/settings.local.json (read by tests/unit/scripts/settings-bash-first-optout.test.ts)"
```

The literal comes from `settings-bash-first-optout.test.ts`'s `const LOCAL_SETTINGS = '.claude/settings.local.json'`, which the `keeps the local override out of the repo` case hands to `git check-ignore -v` (asserting the path IS ignored, and by this repo's `.gitignore`) and to `git ls-files -- <path>` (asserting it is not tracked). Neither reads the file's contents -- the test's own header says the file "cannot see `.claude/settings.local.json`". The scope test's literal scan cannot tell a content read from a git-metadata probe (its header lists that as limit (f)), and its "existing" filter turns the verdict into a property of the developer's checkout: the file is absent in CI and on a fresh clone, present on any machine that enabled a local plugin or a per-checkout permission set (the documented purpose of that file), so the same commit is green for one contributor and red for the next. Measured 2026-09-07 on the go-to-k/cdkd#2516 lane rebased onto `c328db66`: 1 failed / 20036 passed with the file present, 934 files green with it moved aside, nothing else changed.

Adding the path to `.markgate.yml`'s `check.include` is the wrong remedy: the file is gitignored and per-machine, so a marker digesting it would stale on every developer's private edit and would still see a different file set per checkout.

## Proposed fix

[...]

Issue #2745, opened 2026-09-07
fix(secret-redaction): a 1-3 character embedded secret still persists in plaintext for an Fn::Join / Fn::Sub source, a nested-stack parameter value that embeds it, and cdkd import's own resolution


nix-tkobayashi:
## Summary

go-to-k/cdkd#2516 closes the sub-floor (1-3 character) embedded-secret residual for a LITERAL leaf on the bags the deploy pass produced, by marking those bag objects same-generation (`markSameGenerationBag`, `src/deployment/secret-redaction.ts`) and letting `positionByEmbeddedSpan` accept the value scan's silence (`scanned === bag`) only under that mark. Three sites the same class reaches were measured during that PR's review and deliberately left out of it, because each is a different arm with its own evidence question. They are one issue because the defect is one: a plaintext shorter than `MIN_NEEDLE_LENGTH` (4) is dropped from the value scan's needle alternation (`buildNeedleRegex`), so wherever the position arms refuse, the plaintext persists.

- [ ] **The `Fn::Join` / `Fn::Sub` source shape** — the dominant CDK shape (`secret.secretValueFromJson(...)` renders the ARN as a `Ref`, so an L2-reached secret is a join). `positionByIntrinsicSkeleton` (`secret-redaction.ts`, its condition 1: `if (bag === '' || !secrets.has(bag)) return undefined;`) positions a bag only when the WHOLE bag is a recorded plaintext, so a join rendering `port:` + token with a 1-3 character resolved value falls to the value scan and persists `port:<pin>` in `properties`, in the readback and in an output. The #2516 mark is threaded through `redactByPath` and available at that call site; what is missing is a skeleton-side arm that, on a marked bag with exactly ONE candidate expression, an `Fn::Join` whose literal parts frame the middle, and a resolved pair equal to the middle, writes the token into the frame. The #2320 placeholder primitive is the other route.
[...]

Issue #2553, opened 2026-09-04
fix(deploy): the stateful guard was never swept over the Cloud-Control-routed population


go-to-k:
`STATEFUL_TYPES` (`src/provisioning/stateful-types.ts`) has two mechanical lower bounds -- the `DeletionPolicy: Snapshot`-capable union in `src/provisioning/final-snapshot.ts`, and the providers whose `delete()` consumes `DeleteContext.forceDataDelete`. Everything else is hand-curated, and the hand-curated passes so far (PR https://github.com/go-to-k/cdkd/pull/2519, and go-to-k/cdkd#2548 for its residual) were both derived by walking `src/provisioning/providers/**`. That derivation can only see types cdkd has an **SDK provider** for.

go-to-k/cdkd#2514 is about the population with no SDK provider at all -- the Cloud-Control-routed one -- and that population was never swept. It is the larger half by two orders of magnitude: `docs/_generated/provider-coverage.json` reports tier 1 (SDK provider) at 134 types and tier 2 (Cloud Control, deployable today) at 1372.

Verified against `docs/_generated/provider-coverage.json` and the `STATEFUL_TYPES` set at PR 2519's HEAD -- every one is tier 2, absent from tier 1 and tier 3, and absent from the guard list:

- [ ] `AWS::MemoryDB::Cluster`
- [ ] `AWS::MSK::Cluster`
- [ ] `AWS::Timestream::Database`
- [ ] `AWS::Timestream::Table`
- [ ] `AWS::Backup::BackupVault`
- [ ] `AWS::DocDBElastic::Cluster`
- [ ] `AWS::OpenSearchServerless::Collection`

Each holds user data that a DELETE + CREATE does not migrate, none has an SDK provider (so a replacement routes through Cloud Control's DELETE -- exactly the path PR 2519 hoisted the guard onto), and none has a `ReplacementRulesRegistry` entry, so `create-only-properties.ts` resolves the registry schema at diff time and a rename lands as a property-driven replacement a plain `cdkd deploy` reaches with no flag. That last step is the part this issue must confirm per type before adding it: read each type's createOnly set from its registry schema (the checked-in pattern is `tests/fixtures/cfn-schemas/AWS-<Service>-<Type>.json`) rather than assuming a `Name` property is create-only, the way `AWS::EMR::Cluster` was confirmed in PR 2519.

[...]

Issue #2453, opened 2026-09-02
fix(state): a NoEcho custom-resource value EMBEDDED in a longer property still reaches state.json


go-to-k:
## Summary

go-to-k/cdkd#2274 redacts a `NoEcho` custom resource's `Data` out of `state.json` by registering each value as a MASK-ONLY needle (`recordMaskOnlyValue` in `src/deployment/secret-redaction.ts`). That class is deliberately **whole-leaf only**: it is withheld from `redactSecretsForState`'s substring arm (`substringNeedlesOf`), so a persisted leaf is replaced by `***` only when the leaf EQUALS the value.

A template that EMBEDS the attribute in a longer string therefore keeps the plaintext in state:

```ts
new ssm.StringParameter(this, 'P', {
  stringValue: `postgres://user:${cr.getAttString('Password')}@db.example.com/app`,
});
```

resolves to `postgres://user:<the real value>@db.example.com/app`, which is not a whole-leaf match, so the state record holds it verbatim. `Fn::Sub` / `Fn::Join` around a `Fn::GetAtt` is an ordinary CDK spelling (building a connection string is the canonical use), so this is reachable rather than theoretical.

## Why the whole-leaf bound exists, and why widening it is not a one-liner

An EXPRESSION substituted into a longer leaf is lossless -- the persisted string still names a value every downstream reader can re-resolve. A MASK is not: an inline `***` cannot be told apart from a literal `***` a user wrote, so nothing downstream can recognise it, and `carriesSecretMask` (the recognition test `cdkd drift --revert`, `--accept` and the rollback replay all refuse on) is whole-leaf equality by design. Substring-masking a persisted value would therefore trade a disclosure for an UNRECOGNISABLE corrupt baseline that `--revert` and `resolveReplayProps` would push to AWS -- the go-to-k/cdkd#1498 / go-to-k/cdkd#1501 class.

Two candidate directions, neither free:

1. **Mask the WHOLE leaf on containment.** Keeps every mask recognisable and closes the leak, at the cost of destroying the public part of the leaf (over-redaction bounded by the per-resource scope and `MIN_NEEDLE_LENGTH`), plus a guard so a leaf that also carries a `{{resolve:...}}` span is not flattened (that would destroy a re-resolvable expression).
[...]

PR #2753, opened 2026-09-07
fix(secret-redaction): persist an embedded 1-3 character secret as its token on the bags the deploy pass produced

### LINKED ISSUE #2516: fix(secret-redaction): an embedded 1-3 character secret persists in plaintext below the value scan's needle floor

nix-tkobayashi:
Closes #2516

## What

A literal template leaf that embeds one `{{resolve:...}}` token whose resolved value is 1-3 characters long (`port:{{resolve:secretsmanager:prod/db:SecretString:pin}}` with `pin` = `q7`) persisted the PLAINTEXT in `state.json` -- as a resource property, as its AWS readback (`observedProperties`) and as a `CfnOutput` value -- because the value scan drops every needle shorter than `MIN_NEEDLE_LENGTH` (4) from its alternation and #2511's span arm bound itself to the scan's own answer. No same-plaintext sibling is needed.

The arm now accepts the scan's silence (`scanned === bag`) ONLY on a bag whose generation the engine proved.

- `src/deployment/secret-redaction.ts`: a module-private `WeakSet` (`sameGenerationBags`, the shape of the `resolvedPairsOf` side table) and the exported `markSameGenerationBag(bag)`; `redactSecretsForState` reads the mark for the object it is handed and threads `bagIsSameGeneration` through `redactByPath`'s four recursion sites into `positionByEmbeddedSpan`, whose fifth condition gains `if (bagIsSameGeneration && scanned === bag) return prefix + token + suffix;` after the existing scan-equivalence acceptance. The three interference cases (a whole leaf that is another plaintext, a frame that equals a plaintext, an overlapping needle) all make the scan rewrite something, so `scanned !== bag` and they stay refused.
- `src/deployment/deploy-engine.ts`, three installed bags marked after the provider call: `propertiesToRecord` marks the resolved bag only when the provider returned no `effectiveProperties`; `drainObservedCaptures` marks each readback before installing it; `resolveOutputs` marks the outputs bag it returns. Two never-installed COPIES of the resolver's own output are marked as well: the update arm's no-change re-check (`{ ...resolvedProps }`, without which a stored token compares unequal to the re-resolved plaintext and every deploy issues a redundant UPDATE -- not named by the issue, found while writing the tests) and a failed op's journaled `attemptedProperties` (a persisted S3 artifact of its own; found by the code and security reviewers). The completed op's journaled `properties` is the same marked object as the record's, so the journal inherits the token with no change.
- Stated residuals (on `positionByEmbeddedSpan` and in `.claude/rules/layout-deployment-secrets.md`): a provider's `effectiveProperties` replacement (mixed provenance, never marked); a readback AWS rewrote to a coinciding value; an interference refusal leaves the sub-floor middle in plaintext; and the `Fn::Join` / `Fn::Sub` source shape, which `positionByIntrinsicSkeleton` positions only for a whole-value plaintext. The last one, a nested-stack parameter whose value embeds the secret, and `cdkd import`'s own resolution are tracked by #2745. A separate local-suite friction found while rebasing (a gitignored `.claude/settings.local.json` reds `check-scope-checker-inputs.test.ts` since #2738) is filed as #2752, unrelated to this change.

## Tests

- `tests/unit/deployment/secret-redaction-embedded-span.test.ts`: the old residual pin is rewritten (an unmarked bag keeps the plaintext under both template rules constants); positive 1/2/3-character middles alone and beside a whole-value sibling in both map orders; threaded down the object walk and both array descents, while a sub-bag walked on its own, a copy of the bag and a copy of the map stay unmarked; no pair, a conflicting pair, an empty middle (the frame refusal is load-bearing there), and the three interference cases with the mark present.
- `tests/unit/deployment/deploy-engine-short-embedded-secret.test.ts` (new): CREATE persists the token in `properties`, in the readback and in the outputs; the completed-op journal and the failure persist's record; a failed op's `attemptedProperties`; a provider-substituted `effectiveProperties` bag stays unmarked; a previous-generation record on a failed UPDATE keeps its plaintext (`noRollback`), and so does the record the in-deploy automatic rollback re-persists; the no-change re-check issues no UPDATE against a stored token; the re-check marks a copy, not the resolved bag; the auto-refresh readback of a resource the re-check reduced to no-change converges on the stored token.
- `tests/unit/deployment/deploy-engine-outputs-generation-rules.test.ts`: the outputs bag this pass resolved is marked; the previous deploy's bag a resolution failure keeps is not.

Mutation probes (each mutant run alone, sources restored byte-exact between them; the named cases red):

| mutant | red |
| --- | --- |
| drop the mark from the arm's condition (`scanned === bag` alone) | 8 unmarked negatives |
| test the middle's LENGTH instead of the scan's silence | the 3 interference cases |
| write `prefix + survivor + suffix` instead of the token | sibling cases (versioned survivor), CREATE, no-change re-check |
| `isSameGenerationBag` -> `true` | 8 unmarked negatives |
| stop threading the flag at the object walk / the keyed array arm / the positional array arm | the threading and array cases (each site its own probe) |
| `singleSpanFrame`'s `<=` -> `<` | the marked empty-middle case |
| mark the `effectiveProperties` bag | the effectiveProperties and copy-not-bag cases |
| stop marking in `drainObservedCaptures` / `resolveOutputs` | CREATE readback / CREATE outputs + the outputs-file positive |
| mark `resolvedProps` in place at the re-check | the copy-not-bag case |
| stop marking at the re-check | the no-change re-check case |
| stop threading the flag at the keyed arm's unpaired-element positional fallback | the array-descent case |
| drop the per-deploy reset of `attemptedResolvedProps` | the reused-engine journal case |

## Live test

`tests/integration/secrets-dynamic-ref/`: the secret JSON gains `pin: 'q7'`; `DB_PORT_LITERAL` (a literal env var embedding the pin token, a plain string via the concrete account) sits above its whole-value sibling `SECRET_PIN_STAGED` (the map's survivor), and a `PortLiteral` `CfnOutput` embeds the same token. Guard 3c guards the synthesized shape and order (`exit 1`), asserts exact equality of the persisted `properties`, `observedProperties` and `outputs` leaves with the leaf's own expression, and greps the framed plaintext out of the whole state document; Phase 1e asserts the post-rollback record; Phase 1g the empty-map re-capture; the existing `diff --fail` guard and the zero-surviving-versions sweep cover the rest. `mask` now withholds the head of any value of four characters or fewer and `diag_output` withholds the framed form.

Re-run on this head's tree (2026-09-08, `us-east-1`, ledger rows in `docs/_generated/integ-last-run.tsv`). Every rebase that moved `src/deployment/**` got a fresh pair, because the merged `verify.sh` and the merged `propertiesToRecord` are forms no earlier run exercised:

| fixture | result | notes |
| --- | --- | --- |
| `secrets-dynamic-ref` | PASS, 550 s | Guard 3c: `properties`, `observedProperties` and `outputs.PortLiteral` all hold `port:{{resolve:...:pin}}`, the framed plaintext absent from the whole document; Phase 1e post-rollback and Phase 1g re-capture green; `diff --fail` clean; clean destroy, 0 surviving versions |
| `ssm-secure-dynamic-ref` | PASS, 194 s | the sibling span-arm fixture, re-run because the `m1` copy changes the drained readback for every fixture, not just this lane's; taken two commits earlier, at 5cc82973, which nothing since touches `src/` or any fixture |
| `lambda` (broad set) | PASS, 101 s | clean destroy, 0 errors; the events log swept after |

`integ-destroy` and `integ-broad` markers set from these runs; the state bucket holds only `_index/` afterwards.

## Verification

- Codex delta review: 13 rounds until `NO ISSUES` held across every later change (r1: the mark's contract and the `resolveOutputs` failure-path comment; r3: the `attemptedProperties` contract and the residual/follow-up wording; r5: `attemptedResolvedProps` never reset on a reused engine; r6, r8: the fixture's `mask` / `diag_output` self-checks; r7, r9-r13: NO ISSUES). Follow-up issue body: 3 rounds. Maintainer-checklist proxy pass: 9 rounds, converging on fixture hygiene (bare-value withholding, `mask` boundary, self-checks for every withholding arm, the ordering premise rendering `Fn::Join` with its delimiter and, inside the env, failing closed on any other intrinsic or non-pseudo `Ref`; outside the env, on a same-key reference, on any `Fn::Sub`, and on a `Ref` to a template parameter). Ten rounds in total; each later item was a hypothetical fixture mutation, fixed when cheap, and the loop was stopped there rather than run to a NO ISSUES it does not converge to. Since the maintainer review: 2 more delta rounds (r14 found the `m3` case's coverage claim false; r15 NO ISSUES) and 3 more proxy rounds (10 items in total, every one acted on with a measurement; r3 NO ISSUES).
- Reviewer agents (security / code / test / spec) on the working-tree diff: the journal `attemptedProperties` gap and the drain comment's wrong safety argument came from them and are fixed above; the spec reviewer confirmed the re-check copy is inside the issue's constraint (the mark never reaches a bag the pass did not produce).
- `/check`: 942 files / 20470 tests green, run alone on the final tree (runs taken while other sessions' suites held this host above load 12 time out tree-scanning script tests at their 5 s bound, the #2553 class, none of them touched here); `/check-docs`: consistent; `/verify-pr`: all checks pass.
- Pre-existing, not changed here and recorded rather than fixed: `maskSecretsInText` cannot mask a 1-3 character value inside provider error text or events (the #2453 class).

## Review response (2026-09-08)

`m1` applied as `markSameGenerationBag({ ...observed })`, and pinned from outside: after a deploy in which the provider returns its `properties` argument by identity, the object it handed back is walked through `redactSecretsForState` under the DEFAULT rules, where the object's mark alone decides the sub-floor middle. Removing the spread reds that case. `STATE_SOURCED_READBACK_RULES` would NOT discriminate, because it claims the generation itself.

`m2` and `m3` taken in-PR rather than deferred, both as unit cases. `m2` pins that an empty map admits no pair at all, on the docstring's own coincidence shape; its comment states exactly which leaf the fallback mutant moves. `m3` drives `propertyChanges[].requiresReplacement`, which is the property-driven replacement site: mutating THAT site reds it, while mutating the ordinary update site reds a different, pre-existing case (`wasReplaced: true` alone stays on the ordinary site, which review of this very case found).

Four of the five nits taken: `cdkd scrub` named as a residual in the docstring, the rules file and the changelog entry; the fixture's bare-pin arm now treats an SGR sequence as a boundary, with zero-parameter and multi-parameter probes; the self-check moved into a function so its probe is `local`, with a leak check; the interference sentence scoped to substring interference by a 4+ character needle. The fifth was already recorded and bounded, and is left as recorded.

The branch has been rebased repeatedly while main moved; this head carries no conflict. `verify.sh`'s `diag_output` is the one merge worth naming: main's bash-substring form (never `printf | grep -q`, #2582) for every needle, plus this branch's two, plus the CSI boundary.


go-to-k:
## Review — 3-axis + security add-on

Tier: `3-axis` (1602 LOC after excluding `docs/_generated/**`, 11 files) plus the
security add-on (secret redaction / sensitive-value persistence). The tier was
also up-biased on recency, which the size heuristic cannot see:
`src/deployment/secret-redaction.ts`'s last three commits on the base are all
`fix:` (#2562, #2511, #2492), and #2511 introduced the very arm this PR relaxes.

**Verdict: no blockers in the diff.** All four reviewers independently confirmed
the core argument — the four `redactByPath` recursion sites are all threaded, the
`effectiveProperties` exclusion is safe for every provider return shape
(`??` short-circuits on `{}`, a partial and same-identity alike), `singleSpanFrame`'s
`<=` is load-bearing and correct, and no marked object was found to outlive its
pass (state is re-parsed from S3 per `deploy()`, `observedCaptureTasks` is cleared
in `doDeploy`'s `finally`, and `attemptedResolvedProps` is now reset per deploy).

One thing to merge first, three minors, five nits.

### Must be resolved before merge (mechanical, not a review finding)

**M0 — the branch is `CONFLICTING` against `main`, and the rebase invalidates the
live-test evidence.** #2742 landed first and touched the same fixture, as this
PR's own body predicted. Conflicts: `docs/changelog-cdkd.md`,
`tests/integration/secrets-dynamic-ref/verify.sh`,
`tests/integration/secrets-dynamic-ref/lib/secrets-dynamic-ref-stack.ts`.

The consequence is not just the rebase. `verify.sh` was 2247 lines at the merge
base; #2742 took it to 2419 and this branch to 2543. The merged script is a
FOURTH form that neither PR's run exercised, so the `secrets-dynamic-ref` PASS
recorded in the ledger does not cover what will be on `main`. Since
`src/deployment/deploy-engine.ts` is in `integ-broad`'s scope, the broad-set run
has to be repeated on the rebased tree too. Flatten before rebasing (the
changelog clash is the standard trigger).

### Minor

**m1 — `src/deployment/deploy-engine.ts:1730` marks an object the engine did not
construct.** Three of the four reviewers converged here independently.
`markSameGenerationBag(observed)` marks whatever `provider.readCurrentState`
returned. The schema-upgrade auto-refresh path passes `resource.properties` — the
PREVIOUS generation's record bag — as that method's 4th argument (`:1961`), so a
provider returning its `properties` argument by identity would put a permanent
mark on a previous-generation object that is still installed on the record.
Combined with a resource the diff called UPDATE and the re-check then skipped
(`perResourceSecrets` populated at `:4501`, `perResourceTemplateProps` = today's
template), the persist walk would write today's token over a previous
generation's sub-floor leaf — exactly the fabrication the docstring forbids.

I grepped every `readCurrentState` under `src/provisioning/`; none does this
today, so it is unreachable now. But the contract is delegated to ~100 provider
implementations with no guard, in a file whose last three commits were all
`fix:`. The fix is one line and matches the pattern the PR already uses at its
two never-installed sites:

```ts
target.observedProperties = markSameGenerationBag({ ...observed });
```

**m2 — the safety premise behind marking EVERY drained readback is unpinned.**
`:1730` marks all captures, wider than the issue's "freshly created / updated
resource" wording. Reachability is closed by the fact that `perResourceSecrets`
is populated only in the create/update arms (`:4343`, `:4501`), so an unchanged
resource's auto-refresh walks an empty map and the arm cannot fire — correct, and
I verified it independently. But that fact lives only in a comment. A future
change to `redactStateForPersist`'s `secrets ?? new Map()` fallback (`:1605`)
would silently fabricate a token onto a previous-generation readback, and integ
Phase 1g cannot tell the two apart because it asserts the expression. Worth a
unit case that marks an auto-refresh readback and asserts the empty map keeps
the plaintext. Adopting m1 also narrows this.

**m3 — the replacement `propertiesToRecord` sites are untested.** `wasReplaced`
is never `true` in the new engine tests, so `:5197` and `:5984` are never
exercised. Low risk given marking is centralised in the `:6249` helper, but the
issue names all three call sites.

### Nits

- `positionByEmbeddedSpan`'s residual list and
  `.claude/rules/layout-deployment-secrets.md` do not name `cdkd scrub`, which is
  the documented repair tool for pre-GHSA state and cannot repair a sub-floor
  embedded leaf. Worth listing beside the `cdkd import` residual. (The next
  deploy of that resource does repair it, since the re-check compares unequal.)
- `verify.sh:249`: the bare-pin arm `(^|[^A-Za-z0-9])q7([^A-Za-z0-9]|$)` does not
  fire when an ANSI SGR sequence abuts the value (`\x1b[32mq7` — the terminating
  `m` is alphanumeric), so a colorized diagnostic could print the bare value.
  Strip ANSI first, or add `\x1b` to the boundary class. Only reachable if cdkd
  emits color into a captured non-TTY pipe, and `q7` is declared public fixture
  data.
- `verify.sh`, the `diag_output` self-check loop: `diag_probe` is not declared
  `local`, unlike every other helper variable in the script.
- The docstring's "the three interference cases all make the scan rewrite
  SOMETHING" holds only for needles of 4+ characters; a sub-floor interfering
  needle now falls through to the relaxed arm. The consequence is
  under-redaction of the OTHER secret (the scan's own pre-existing residual),
  never a fabricated expression, so the safety argument is unaffected — but the
  sentence is stated as a universal.
- `deploy-engine-short-embedded-secret.test.ts:352` records that the copy at
  `:1580` is not pinned, so a "mark `next.attemptedProperties` in place" mutant
  stays green. Bounded by the per-deploy reset, and the PR's table does not claim
  otherwise.

### Verified clean (so it is on the record)

- All four `redactByPath` recursion sites thread the flag (`:3220`, `:3229`,
  `:3249`, `:3264`); the three sibling fallbacks call `redactSecretsForState`
  with no source and never reach the span arm.
- Sink trace for the 2-character plaintext: `properties`, `observedProperties`,
  `outputs`, the exports index, the deploy summary, both journal bags and
  `cdkd diff` all improve to the token. `rollback --revert-failed` re-resolves
  the embedded token rather than shipping it literally. `scrub` / `import` keep
  the stated residual. `deployments/` carries no property bag. No sink regressed.
- No cross-generation mark leak found on any carrier.
- The negatives assert the exact feared shape (the plaintext literal, and the
  span negatives additionally assert equality with the value scan's own answer),
  and the engine half is pinned without going through the marking helper, so the
  suite is not self-referential. The four load-bearing mutation-table rows all
  hold; the unmarked-negative count is 9, not the 8 the table states.
- The fixture's five new jq programs were executed against a synthetic template
  and six mutants: non-vacuous and fail-closed in every arm.
- Unit suite re-run locally on this head: 3 files / 55 tests green, no type
  errors. CI is green on every check.

m1 is the only one I would ask for before merge; m2 and m3 are reasonable to
take as follow-ups if you would rather not grow the diff.

go-to-k:
Correcting the closing line of my review above — it named only `m1` as the ask
and left the branch state as a description. Stating the requests explicitly, with
the labels from that comment so the two line up.

**Requested before this PR can be merged, in order:**

1. **`M0` — rebase onto current `main` and force-push.** The branch is `CONFLICTING`:
   #2742 merged first and touched the same fixture, so
   `docs/changelog-cdkd.md`, `tests/integration/secrets-dynamic-ref/verify.sh`
   and `tests/integration/secrets-dynamic-ref/lib/secrets-dynamic-ref-stack.ts`
   all conflict. Flatten the branch first — the `changelog` clash is the
   standard trigger for that.

2. **`M0`, second half — re-run the integs on the rebased tree and update the ledger rows.**
   `verify.sh` was 2247 lines at the merge base, 2419 after #2742 and 2543 on
   this branch; the merged script is a fourth form that neither PR's run
   exercised, so the `secrets-dynamic-ref` PASS currently in
   `docs/_generated/integ-last-run.tsv` does not cover what lands on `main`.
   Because `src/deployment/deploy-engine.ts` is in `integ-broad`'s scope, the
   broad-set run has to be repeated as well. Both markers (`integ-destroy`,
   `integ-broad`) need to be set from those runs.

3. **`m1`** — mark a copy at `src/deployment/deploy-engine.ts:1730`
   (`markSameGenerationBag({ ...observed })`), so the mark never lands on an
   object a provider returned by identity.

`m2` and `m3` are fine as follow-up issues if you would rather not grow the
diff. The nits are yours to take or leave.

I have not touched the branch — head is still `c6d7b911`.

go-to-k:
## Review round 2 — `f74c740f`

Re-ran the `3-axis` code / test lanes and the security add-on against the new
head. **`M0`, `m1`, `m2`, `m3` and all five round-1 nits are addressed; no
blocker and no major.** The entire non-comment `src/` delta between the two
revisions is the one line I asked for:

```
- target.observedProperties = markSameGenerationBag(observed);
+ target.observedProperties = markSameGenerationBag({ ...observed });
```

Verified rather than assumed:

- The spread severs the `m1` channel — `markSameGenerationBag` has six call
  sites, all in `deploy-engine.ts`, and the object a provider returned is no
  longer among the marked. The nested-sharing residual is real but inert:
  `isSameGenerationBag` is read once at the top-level entry
  (`secret-redaction.ts:4696-4710`) and threaded down as a boolean, a nested bag
  walked on its own answers `false`, and both populations that could reach it
  (auto-refresh with an empty map, UPDATE-then-re-check-skip) refuse or converge.
- Nothing depended on `observedProperties` being identical to the provider's
  object: the only in-process reader (`isStatefulRecreateTargetForReplace`,
  `:4695`) reads values. `readCurrentState` is typed
  `Promise<Record<string, unknown> | undefined>`, so `{ ...arr }` is unreachable.
- The rebase composes: main's `deploy-engine.ts` delta is confined to
  `handleOutputResolutionFailure`, the `outputs` identity and its mark site are
  untouched, and #2739's null-prototype variables object never becomes a bag or
  a source leaf. `markSameGenerationBag` has no importer outside
  `deploy-engine.ts`, and the two private signature changes are compile-enforced.
- The exact-match arm really has no floor (`wholeValueExpr` at `:4786` runs at
  `:4897`, before `buildNeedleRegex`'s length filter at `:1623`), so the
  rewritten claim about sub-floor interference is sound.
- The three new unit cases discriminate. The `m1` case reds under a revert to
  `markSameGenerationBag(observed)`, and its rules choice is load-bearing —
  measured, `STATE_SOURCED_READBACK_RULES` yields the token even UNMARKED, so
  reusing the persist path's rules would have made the probe vacuous. `m3`
  reaches `:5216`, the property-driven replacement, so all three
  `propertiesToRecord` sites now have a case.
- The fixture self-checks were executed, not reasoned about: `local diag_probe`
  removed → the leak assertion fires; the containment arm removed → the mask
  probe fires; the SGR alternative removed → the diag probe fires; `[0-9;]*`
  narrowed either way reds. The leading-side-only argument is correct —
  `x=q7\033[0m` withholds because ESC is already in the plain class.
- Full CI green on this head, `MERGEABLE` / `CLEAN`, and the three ledger rows
  are dated after the rebase.

### Nits — none of these block

1. `src/deployment/secret-redaction.ts:2814-2816` — "a sub-floor needle matching
   the WHOLE leaf … is refused here as before" is wrong for the degenerate
   empty-frame leaf (leaf === middle, ≤3 chars). There `scanned` equals the
   survivor, so `:2802` matches and the leaf is ACCEPTED one line up, not
   refused at `:2820`. The answer is still this pass's own token, so the safety
   argument holds — only the word "refused" is false. Worth fixing precisely
   because this sentence replaced the round-1 nit about a universal claim.
2. `src/deployment/deploy-engine.ts:1731` — "like the two never-installed sites
   above": `:1580` is above, the no-change re-check at `:4579` is below.
3. `tests/unit/deployment/deploy-engine-short-embedded-secret.test.ts:637` —
   the comment cites `:1730`, which the round-2 comment expansion moved to
   `:1749`. Naming `drainObservedCaptures` instead would not go stale again.
4. `tests/integration/secrets-dynamic-ref/verify.sh:~250` — `EXPECTED_PIN` is
   interpolated unescaped into an ERE. Safe at `q7`; a future pin carrying a
   metacharacter would silently widen or break the match.
5. `verify.sh:291-292` — a CSI whose final byte is not `m` and abuts the value
   (`\033[2Kq7`, `\033[1Aq7`) is printed. Unreachable from cdkd, which emits
   only `\x1b[<n>m` and a bare `\r` — both already covered. Widening to
   `ESC\[[0-9;?]*[A-Za-z]` costs nothing.
6. The "What stays" residual list (`secret-redaction.ts:2700-2740`) does not
   name the masking channel: `maskSecretsInText`'s substring arm has the same
   four-character floor, so state now persists `port:{{resolve:...}}` while a
   warn line quoting an AWS message can still print `port:q7`. Pre-existing
   (the #2453 class) and not a regression, but the list is otherwise complete.
7. The replacement case does not bound the old resource's cleanup with
   `expect(mockProvider.delete).toHaveBeenCalled()`.

`m2`'s own comment already concedes that its mutant reds only at the
whole-value leaf and not at the embedded one — accurate, and the span arm's own
fence is the `m1` case, so I am not asking for more there.

nix-tkobayashi:
Done — `M0`, `m1`, and, rather than deferring them, `m2` and `m3` plus four of the five nits.

## M0

The branch is rebased onto `main` at 1c1286b3 and force-pushed. Three conflicts: `docs/changelog-cdkd.md` and the fixture's stack file are keep-both; `verify.sh`'s `diag_output` is a real merge, taking main's bash-substring form for every needle (never `printf | grep -q`, #2582) plus this branch's two.

Both integs re-run, and a third. The rows now in the ledger are the ones taken on the FINAL tree at 1c1286b3: `secrets-dynamic-ref` PASS 559 s and `lambda` PASS 137 s. That tree includes #2764, which merged between my first pair of runs and this one and changes `resolveSub` — a path this fixture exercises — so the earlier pair (611 s / 96 s, base 540debe3) does not attest to what will land, and I re-ran rather than reporting it. `ssm-secure-dynamic-ref` PASS 194 s is two commits earlier, at 5cc82973, which already carries #2764; nothing between it and 1c1286b3 touches `src/` or any fixture. I added that third fixture because `m1` changes the drained readback for every fixture, not just this lane's.

Clean destroy and zero surviving state versions on each; the bucket holds only `_index/` afterwards. `integ-destroy` and `integ-broad` are set from the final-tree runs.

## m1

`markSameGenerationBag({ ...observed })`, and it is pinned rather than asserted. After a deploy in which the provider returns its `properties` argument by identity, the test walks the object it handed back through `redactSecretsForState` under the DEFAULT rules, where the object's own mark is what decides the sub-floor middle; removing the spread reds that case. Worth recording, because it cost me two wrong probes first: `STATE_SOURCED_READBACK_RULES` does NOT discriminate, since `sourceIsSameGeneration: true` claims the generation itself, and an identity assertion through the persist path cannot either, because that path clones.

## m2 and m3

Taken here rather than as follow-ups. `m2` pins the premise on the docstring's own coincidence shape, and its comment states exactly which leaf the mutant moves: a fallback carrying this pass's pair reds it at the WHOLE-VALUE leaf through the floorless exact-match arm, not at the embedded leaf, whose fence is the mark `m1` pins.

`m3` needed a correction I would rather state than leave: `wasReplaced: true` reaches the ORDINARY update's `propertiesToRecord`, which existing cases already exercise. The case now drives `propertyChanges[].requiresReplacement`. Mutating the property-driven replacement site reds it; mutating the ordinary update site reds a different, pre-existing case instead, so the two are separated.

## Nits

Taken: `cdkd scrub` named as a residual in the docstring, the rules file and the changelog entry; the bare-pin arm's ANSI hole closed, with zero-parameter and multi-parameter SGR probes (a `[0-9;]+m` mutant reds the self-check); `diag_probe` made `local` by moving the self-check into a function, since `local` is a bash error at top level, plus a leak check that reds when it is dropped; and the interference sentence scoped to substring interference by a 4+ character needle, saying what the silent sub-floor case costs. I also added an interior-occurrence probe to the `mask` premise, which was pinning only its endpoints.

Left as recorded: the `:1580` copy note, which you had already bounded and which the table does not overclaim.

## Verification

Full unit suite run alone on the final tree: 942 files / 20470 tests. Codex: 2 more delta rounds (one found `m3`'s coverage claim false) and 3 more maintainer-checklist proxy rounds, 10 items in total, each acted on with a measurement rather than a reading — two of them corrected claims I had made in the code comments themselves.

go-to-k:
## `f4888e04` — blocker in the pending rebase

The only delta from `f74c740f` is three ledger rows (the integs re-taken on the
final tree at `1c1286b3` rather than the mid-rebase `540debe3`). Good catch on
your side; no code, test or fixture change, so my round-2 review still stands
for the diff itself.

But `main` moved again and the branch is `CONFLICTING`, and this time the
conflict is **not** mechanical. **#2788 rewrote `propertiesToRecord` — the exact
function this PR marks in — and resolving the conflict the obvious way silently
disables this fix for a subset of resources.**

### The blocker

`main` now has (`src/deployment/deploy-engine.ts`):

```ts
private propertiesToRecord(desiredProperties, result, resourceType, provisionedBy) {
  const effective = result.effectiveProperties ?? desiredProperties;
  if (provisionedBy !== 'sdk') return effective;
  return withoutSilentDropProperties(resourceType, effective);
}
```

and this branch has:

```ts
return result.effectiveProperties ?? markSameGenerationBag(desiredProperties);
```

The obvious merge — keep `markSameGenerationBag` where it sits and layer
#2788's narrowing on top — puts the mark on the wrong object, because
`withoutSilentDropProperties` (`src/provisioning/property-coverage.ts:141-150`)
returns a **fresh copy** whenever it removes anything:

```ts
const removable = removableSilentDrops(resourceType, properties);
if (removable.length === 0) return properties;   // same object — mark survives
const written = { ...properties };               // NEW object — mark LOST
for (const property of removable) delete written[property];
return written;
```

So on the SDK route, which is the default:

- no removable silent drop → the same object is returned, the mark survives,
  #2516 works;
- **any removable (non-create-only) silent drop present → the record holds a
  fresh, UNMARKED object, and an embedded 1-3 character secret persists in
  PLAINTEXT again** — precisely the defect this PR closes.

That is worse than a plain regression because it is conditional: it would pass
every existing test and every integ arm in this PR, and fail only for a
resource whose type has a removable silent drop in the bag. The generated
coverage table carries 134 types, so the population is not theoretical.

### What the resolution has to do

Mark the object the record actually HOLDS — i.e. after the narrowing, not
before — while keeping an `effectiveProperties` replacement unmarked. Something
of this shape, rather than wrapping `desiredProperties`:

```ts
private propertiesToRecord(desiredProperties, result, resourceType, provisionedBy) {
  // A provider's replacement is mixed provenance and is never marked.
  if (result.effectiveProperties) return result.effectiveProperties;
  if (provisionedBy !== 'sdk') return markSameGenerationBag(desiredProperties);
  // Still THIS pass's own resolution, minus keys the SDK route cannot write,
  // so the narrowed copy is inside the mark's contract.
  return markSameGenerationBag(withoutSilentDropProperties(resourceType, desiredProperties));
}
```

Please also add a case that would have caught it: an SDK-routed resource whose
type has a removable silent drop present in the bag, with the embedded
sub-floor secret, asserting the record holds the token. Without that, the same
merge could reintroduce the hole later and nothing would say so.

### The other conflicted site

The no-change re-check collides too, but composes once it is done deliberately:
this branch changed the LEFT operand
(`redactSecretsForState(markSameGenerationBag({ ...resolvedProps }), ...)`)
and #2788 changed the RIGHT one (`currentProps` → `currentPropsAsWritten`).
Take both.

While you are in there, please re-read that comparison once with #2788's
narrowing in mind and confirm it still settles: after #2788 the RECORD is
narrowed, while the left side is the full `resolvedProps`. If a removable
silent-drop key is present in `resolvedProps` but absent from both the record
and `currentPropsAsWritten`, the two sides differ on every deploy. I believe
#2788's desired-side twin (`withoutAcceptedSilentDropProperties`, used by the
diff) is what keeps that from firing, but the interaction is new and this PR is
the first thing to land on top of it.

### Round-2 nits

The seven nits from my previous comment are still open. None of them block, and
they can ride the same push as the conflict resolution.

Nothing else changed in my assessment: no blocker or major in the diff itself,
CI was green on `f74c740f`, and the ledger rows are now taken on the reviewed
tree.

nix-tkobayashi:
Your blocker and my rebase crossed — I hit the same `propertiesToRecord` collision while resolving it, reached the same conclusion, and the fix is on the branch at `94232915`. Confirming against your write-up rather than asserting it.

## The blocker

Resolved as you prescribe: the mark goes on the object the record actually HOLDS, after the narrowing, and an `effectiveProperties` replacement is never marked. One deliberate difference from your sketch — it keeps #2788's narrowing on the `effectiveProperties` arm too, which `main`'s `const effective = ... ?? desiredProperties` shape applied to both arms and an early return would drop:

```ts
if (result.effectiveProperties) {
  return provisionedBy === 'sdk'
    ? withoutSilentDropProperties(resourceType, result.effectiveProperties)
    : result.effectiveProperties;
}
const written =
  provisionedBy === 'sdk'
    ? withoutSilentDropProperties(resourceType, desiredProperties)
    : desiredProperties;
return markSameGenerationBag(written);
```

The case you asked for is there, and `AWS::Lambda::Function` already carries three removable silent drops, so it needed no second resource type: an SDK-routed CREATE whose bag carries `CapacityProviderConfig`, asserting both that the key is absent from the record and that the embedded sub-floor secret is still its token. MEASURED — moving the mark before the narrowing reds exactly that case and nothing else.

Its negative control is there too, because the route guard had none: a `cc-api` resource keeps the key, and narrowing that arm unconditionally reds that case alone.

## The other conflicted site

Both taken, as you said: the left operand is this branch's marked shallow copy, the right one is #2788's `currentPropsAsWritten`.

## The re-read you asked for

It settles, and your reading of why is right. `DiffCalculator` narrows BOTH sides (`withoutAcceptedSilentDropProperties` for the desired one, `withoutSilentDropProperties` for the current one), so a difference confined to a dropped key is never reported and the UPDATE arm is not entered.

What survives is narrower than the shape you were checking for, and I filed it rather than folding it in: the re-check's desired side is the FULL `resolvedProps` while its stored side is narrowed, so when the diff reports a change for one of the reasons the re-check exists to absorb AND the same bag carries a removable silent-drop key, the re-check cannot skip and the provider is called. The record stays correct — `propertiesToRecord` narrows what it writes — so it is wasted work, not a wrong result. #2809, with the case that fails today.

## Round-2 nits

All seven. The delta loop then found four more statements of my own that were over-broad: the whole-leaf sentence now splits on the frame (empty frame accepted one line up, nonempty refused), the CSI class covers any final byte plus colon and private parameters with a probe for each, the pin is escaped into the ERE with a note saying that guard is deliberately unpinnable by this fixture, and the changelog names the masking-channel residual. The replacement case now pins the DELETE's arguments rather than the call.

Two things the branch stopped carrying while this was in flight. It briefly re-derived `CORPUS_BYTES_MAX`, because `origin/main` was over the old ceiling and every lane was blocked; #2757 landed the same re-derivation while I was measuring, and #2807 has since replaced that ceiling with per-glob budgets, so the branch touches neither. And #2811 moved the changelog to one file per entry: the entry is now `changelog.d/entries/2026-09-09-2516-embedded-short-secret-token.md` and `_archive.md` is byte-identical to main.

## Verification

On the final branch tree, rebased onto `c3f2330c`, which carries #2777, #2807 and #2812: full unit suite 951 files / 20650 tests green, and both integs re-run — `secrets-dynamic-ref` PASS 550 s, `lambda` PASS 101 s, clean destroy and zero surviving versions on each, bucket back to `_index/`. `integ-destroy` and `integ-broad` are set from those runs.

One note for whoever rebases next: this worktree's `node_modules` predated #2812, and the stale `@aws-sdk/client-codebuild` made `HostKernel` an error type, so `vp check` failed with four errors in `src/provisioning/providers/codebuild-provider.ts`, a file the branch does not touch. `pnpm install --frozen-lockfile` cleared it. Worth knowing because the failure names a file the diff never opens.

Codex since your round 2: 5 delta rounds and 4 maintainer-checklist rounds, every item measured rather than reasoned about.

go-to-k:
## Review round 3 — `94232915`

Re-ran all three axes plus the security add-on. **The blocker is closed, and I
verified the resolution rather than taking the write-up for it. No blocker, no
major.** Three minors and six nits. `M2` and `M3` are requested before merge — both are
statements that are false as written, one of them in a rule file every future
session reads. `M1` belongs to #2809, not to this PR; the nits are
take-or-leave. The request is spelled out in the comment below.

Confirmed against the tree:

- `propertiesToRecord` (`deploy-engine.ts:6390-6399`) is right on all four
  combinations. Your correction to my sketch was the right call — `main`'s
  `const effective = … ?? desiredProperties` applied the narrowing to BOTH arms,
  and an early return would have dropped it from the `effectiveProperties` one.
- The `??` → truthiness change is behaviourally identical. `effectiveProperties`
  is `Record<string, unknown> | undefined`, so only `undefined` is falsy at type
  level; a runtime `null` from a JS double lands on the desired arm under both
  spellings; `0` / `''` / `false` are not type-permitted, and where `main` would
  have fed them to `withoutSilentDropProperties` the new code falls through to
  `desiredProperties`, which is strictly safer.
- `AWS::Lambda::Function` really does carry `CapacityProviderConfig` with an
  empty `createOnlyDrops`, so the case reaches the sdk arm with a non-empty
  `removableSilentDrops`. Under "mark before the narrowing" its exact-token
  assertion reds and its sibling stays green. The `cc-api` control
  discriminates, and its route is not merely labelled — dropping the guard reds
  it alone.
- All four (`effectiveProperties` × route) combinations have a case; two live in
  `deploy-engine-silent-drop-record.test.ts`, which this PR does not touch.
- The mark survives to the reader by identity (`redactStateForPersist:1610` →
  `scrubResourceRecord:5003` → `isSameGenerationBag` at
  `secret-redaction.ts:4722`), and the full route matrix lands on the intended
  answer, including the create-only-drop arm where the key is retained.
- The narrowing neither relocates nor fabricates: a dropped key is absent from
  the journal's `properties` too (it is the same object), events carry no
  property bag, and the exports index takes `outputs`.
- Contract not widened. `withoutSilentDropProperties` only `delete`s top-level
  keys and returns its input by reference when nothing drops, so every surviving
  leaf is byte-identical to this pass's resolution — an object-level claim over
  a SUBSET, which is strictly weaker than the one the issue sanctioned. That is
  a materially different thing from `effectiveProperties`, which the issue
  refuses because a provider carries previous-state values IN.
- The new fixture grammar was executed under bash 5.3.9 and 3.2.57: `\033[2K`,
  `\033[?25l`, `\033[1A`, `\033[38;2;255;0;0m`, `\033[38:5:1m`, `\033[m`,
  OSC-BEL, OSC-ST and a bare `\r` all withhold; the benign interior case still
  prints; the escaper is byte-identical for `q7` over ten metacharacter pins and
  the unescaped form provably over-matches. Every widening has its own
  discriminating probe.
- #2811 applied correctly — the only changelog path in the diff is
  `changelog.d/entries/2026-09-09-2516-embedded-short-secret-token.md` (1976
  bytes, under the cap); `docs/changelog-cdkd.md` and `_archive.md` are absent
  from `git diff c3f2330c..94232915 --name-only`. #2812 touches no
  masking/redaction surface.

### Minor

**M1 — #2809 is rated too low, and the case that shows it is a real
replacement.** Your issue says the residual is wasted work with a correct
record. The attribute-only branch (`deploy-engine.ts:4655`) sits INSIDE the
re-check's true arm, so for an SDK-routed resource whose bag carries an
allow-listed removable drop the re-check can never fire, and a
`DeletionPolicy` / `UpdateReplacePolicy`-only flip falls through to a real
`provider.update()` with identical property bags on both sides.
`AWS::SNS::Subscription` is the live case: `Region` is a non-create-only silent
drop, and `sns-subscription-provider.ts:229-245` implements `update()` as
unsubscribe-then-resubscribe. That is a resource replacement, not wasted work.
Inherited from #2788, not introduced here — please just re-scope #2809 rather
than fold anything into this PR.

**M2 — `verify.sh:295-297` states something false in a security-decision
comment.** It says cdkd emits only `\033[<n>m` and a bare `\r`, so the wider CSI
class is unreachable from it. `src/utils/live-renderer.ts:151,197,203` emit
`${ESC}1A${ESC}2K`, `${ESC}?25l` and `${ESC}?25h` — exactly the sequences round 3
widened the class for. The true reason is that they are TTY-gated
(`live-renderer.ts:72`), so a captured pipe never carries them. As written, a
later reader could narrow `csi` back to `[0-9;]*m` as dead grammar, and any
future non-TTY renderer write would put `\033[2Kq7` past the boundary class into
a CI log. Suggest the measured sentence instead.

**M3 — the `markSameGenerationBag` docstring no longer describes what is
marked.** `secret-redaction.ts:195-209` still says the mark goes on the record's
`properties` "only when that object is the resolved bag itself" and that
"**Two** calls mark a COPY … never installed". There are now three copy sites
and one of them (`deploy-engine.ts:6397`) IS installed. The same "when they are
the resolved bag" wording is in `secret-redaction.ts:2706-2709`, the
`positionByEmbeddedSpan` bullet in `.claude/rules/layout-deployment-secrets.md`,
and the changelog entry. No live fabrication path — the correct argument exists,
but only as a comment at one call site (`deploy-engine.ts:6385-6389`). Given
this PR's own theme of retiring over-broad statements, worth one pass.

### Nits

1. `deploy-engine-short-embedded-secret.test.ts:738` — the DELETE assertion pins
   the physical id (`'fn-phys'` vs `'fn-phys-2'` reds it) but arguments 4 and 5
   are `expect.anything()`, which only refuses `null` / `undefined`. A wrong
   properties bag would not red it, so "pins the DELETE's ARGUMENTS" holds for
   the identity, not the bag.
2. `verify.sh:307` — the "deliberately unpinnable by this fixture" note is fair
   about `diag_output` (every probe uses the alphanumeric pin, so escaped and
   raw are indistinguishable there) but over-broad as written: the escaper
   itself is pinnable without touching `diag_output`, by asserting its output
   for a metacharacter-carrying sample.
3. `verify.sh:308` — `/` is in the escaped class, producing `\/`, which is
   undefined in POSIX ERE (glibc and macOS both take it as a literal, measured).
   A sed-delimiter reflex, not an ERE metacharacter; drop it.
4. `deploy-engine-short-embedded-secret.test.ts:764` — the
   `getAllowedUnsupportedProperties` override is inert on this path
   (`withoutSilentDropProperties` takes no allow set; the registry's set is read
   only at `deploy-engine.ts:2335`, feeding a mocked `calculateDiff`), while the
   comment at `:201-204` reads as if the opt-in is load-bearing.
5. The changelog entry and the PR body never mention the #2788 interaction,
   which is the most surprising thing a future reader would want to find.
6. The PR body's rebase paragraph still names a `docs/changelog-cdkd.md`
   conflict this head no longer has.

Everything else from rounds 1-2 stayed fixed. Unit suite on the three changed
files: 3 files / 60 tests green, no type errors. Full CI green on this head,
`MERGEABLE` / `CLEAN`, and the two integ rows are dated on the final tree.

go-to-k:
**`M2` and `M3` are requested before merge.** They are one push and no code
change, and the reason is not tidiness:

- `M3` writes a FALSE INVARIANT into `.claude/rules/layout-deployment-secrets.md`
  — "the mark goes on the record's `properties` when they are the resolved bag
  itself". Every future session reads that file, and that invariant is exactly
  the one which, if relied on, reinstates the blocker this round caught: it is
  what makes "mark first, narrow after" look correct. The same wording is in
  `secret-redaction.ts:195-209` and `:2706-2709` and the changelog entry.
- `M2` puts a false premise under a security decision — `verify.sh:295-297`
  says cdkd emits only SGR, so a later reader has a written reason to narrow
  the CSI class back to `[0-9;]*m`.

`M1` stays out of this PR: re-scope #2809, nothing here. The six nits remain
take-or-leave.

Nothing else is outstanding — the diff itself is clean, CI is green on
`94232915`, and the integ rows are on the final tree. Push `M2` + `M3` and I
will merge.

go-to-k:
## Review round 4 — `f4ac65c9`

`M2`, `M3` and all six nits are done, and the rewrite is factually sound where I
checked it — the five-site docstring list, the `.claude/rules` index, the
changelog claims and the corrected TTY-gate reasoning all verify against the
code. The `ere_escape` self-check is a real pin, not circular: 23 mutations of
the escaper were run and all 23 are caught, on bash 3.2/BSD-sed, 5.3/BSD-sed and
5.2/GNU-sed-4.9. The whole premise block still passes on all three, and every
`diag_output` widening still reds when removed. You also caught the
`resolveOutputs` inaccuracy I had missed.

**Two requests before merge. Both are single sentences that are FALSE, and both
are in the class this round existed to retire.** After they land I will verify
those two lines and merge; I am not opening another wording round.

**`N1` — `verify.sh:378-379` names the one metacharacter that does NOT
disclose.** The comment says an unescaped `*` "makes the preceding character
optional, so the literal pin can stop matching and the secret is PRINTED".
Measured on the actual `pin_re`, with `value=<pin> tail` as the text and the pin
interpolated raw:

```
q*7    WITHHELD      <-- the named exemplar does not leak
q.7    WITHHELD      (the "benign" half is right)
q+7    PRINTED
q(7    PRINTED
q{7    PRINTED
q^7    PRINTED
q$7    PRINTED
qq*7   PRINTED
```

`q*7` withholds because the boundary alternation consumes the `*` and `q*`
matches zero occurrences. `*` only discloses when the preceding character
repeats (`qq*7`) or is itself a metacharacter. So the sentence's direction is
right and its example is the counter-example. Name `+` or `^` instead, or state
the mechanism without picking a character. The guard itself is unaffected — all
14 characters are probed either way — but this file's comments are the design
record, and round 3 already had to correct one wrong correction here.

**`N2` — "usually does not" points the wrong way.**
`secret-redaction.ts:230` and `.claude/rules/layout-deployment-secrets.md:168`
both say the `resolveOutputs` marked object is "usually not" the stored one.
`redactOutputs` returns its INPUT whenever `outputSecrets.size === 0`
(`deploy-engine.ts:1388`), which is the ordinary deploy, so the record usually
DOES hold the marked object. The exception is stated correctly in the very next
clause, so nothing downstream is misled — but "usually" is a frequency claim
pointing at the smaller population, which is the same shape as the universals
this round retired. Either drop the frequency word or scope it ("inside the
population where the arm can fire").

### Take or leave

- `secret-redaction.ts:240` — "the object the provider is handed is the unmarked
  original" is exact on the SDK route; on the `cc-api` route the provider gets
  `preparePropertiesForCcApi(...)`, a derived bag. Unmarked either way, so the
  safety claim holds; only "the original" is not literal.
- `ere_escape` does not catch a class WIDENED with a non-`/` non-metacharacter
  (`:`, `-`, `_`). Undefined POSIX ERE, literal on every platform measured,
  never a disclosure. The `a/b` probe already covers the dangerous widening —
  on glibc an over-escaped pin is an invalid regex (rc=2), which `[[ =~ ]]`
  reads as no-match and the pin PRINTS, so that probe is the highest-value one
  in the set.

Everything else is verified: no non-comment `src/` change this revision, the
three unit files are 3 / 60 green, CI is green on this head, and the diff is
`MERGEABLE` / `CLEAN`.

go-to-k:
`76b63b44` fixes the docstring's opening sentence — a good catch, and correct —
but it does not contain either of the two requests from my last comment. Both
are still on the head verbatim:

- **`N1`** — `tests/integration/secrets-dynamic-ref/verify.sh:378` still reads
  "while an unescaped `*` makes the preceding character optional, so the literal
  pin can stop matching and the secret is PRINTED". Measured on the actual
  `pin_re` with `value=<pin> tail`: `q*7` **withholds** (the boundary
  alternation consumes the `*` and `q*` matches zero), while `q+7`, `q(7`,
  `q{7`, `q^7` and `q$7` all print. The named exemplar is the one that does not
  leak. Name `+` or `^`, or state the mechanism without picking a character.
- **`N2`** — `src/deployment/secret-redaction.ts:235` still reads "and here it
  usually does not", and `.claude/rules/layout-deployment-secrets.md:168` still
  reads "the redaction INPUT and usually not the stored object".
  `redactOutputs` returns its input whenever `outputSecrets.size === 0`
  (`deploy-engine.ts:1388`) — the ordinary deploy — so the record usually DOES
  hold the marked object. Drop the frequency word or scope it.

Nothing else is outstanding. Push those three lines and I will check them and
merge.

nix-tkobayashi:
`M2` and `M3` are pushed at `76b63b44`, along with nits 1 through 5. Nothing in the round-3 follow-up changes behaviour — every edit of this round is a comment, a docstring, a rule sentence, a changelog sentence, a test assertion, or one new fixture premise check; the PR's own behaviour change is unchanged from the head you reviewed.

## M3

The contract is rewritten, and the thing that was wrong was not just the two counts: its conditions were stated GLOBALLY when they hold per site. "After the provider call" is false for the no-change re-check, which marks before a call it may skip entirely. "Every leaf this pass produced" is false for the auto-refresh readback of an UNCHANGED resource, where nothing was resolved and the empty secrets map, not the mark, is what keeps it safe. The docstring now says what the mark CLAIMS — the only property all five sites share — then names those two exceptions and hands the reader a per-site list.

Two further corrections the delta loop forced, both of which had survived my first rewrite. The outputs bag is the redaction INPUT rather than the stored object: `redactOutputs` returns a fresh bag and that is what is installed — except when this pass recorded no output secrets, where it returns its input and the marked object IS stored. And every drained readback is marked, not only a resource this pass wrote. The same three corrections are in the `positionByEmbeddedSpan` paragraph, `.claude/rules/layout-deployment-secrets.md` and the changelog entry, and the rules bullet now names the docstring as the authority rather than restating conditions it cannot keep current.

## M2

Replaced with the measured reason. `src/utils/live-renderer.ts` does emit `\033[1A\033[2K`, `\033[?25l` and `\033[?25h`; what keeps them out of a captured log is that the renderer refuses to start unless the stream is a TTY, and this fixture captures through a pipe. The comment says that, and says the class is reachable from cdkd's own code the moment anything writes those sequences ungated.

## Nits

1. The DELETE assertion pins the OLD generation's bag (`expect.objectContaining({ Handler: 'old.handler' })`), not just the physical id. Handing the delete `resolvedProps` reds it.

2 and 3. The escaper is now ONE `ere_escape` that both `diag_output` and its premise check call — the earlier check had its own copy, so mutating the real guard left it green and my "measured" result proved nothing. It walks every metacharacter in the class with three negatives each (substitution, omission, repetition) plus a mixed sample. Four mutants red it: `cat`, the class narrowed to `[.*]`, a `\&*` replacement, and a substitution missing its `g` flag. `/` is out of the class with a comment saying why, and its absence is asserted.

4. Both the mock and the case now say the `getAllowedUnsupportedProperties` override is inert for what the case asserts, and why it is set anyway.

5. The changelog names the #2788 interaction. It also regained two qualifications an earlier trim had dropped — the exact-match arm has no floor, and whole-leaf interference is scanned rather than left — which is the same class of error as `M3` and worth saying out loud.

6. The PR body's rebase paragraph is rewritten — this head carries none of those conflicts, so the paragraph now says only what is still true about the `diag_output` merge. The live-test durations in the body were stale against the ledger too and are corrected.

## M1

Taken as you asked — out of this PR. #2809 is re-scoped: its Impact section now leads with the replacement, names `AWS::SNS::Subscription` and why its `update()` makes an attribute-only flip a destroy-and-recreate, keeps the redundant-call reading for every other type, and its `Severity` line reads medium instead of low. Its verification section gained a second case so the replacement half is pinned rather than argued. That was the part I had wrong.

Also corrected while checking this reply: the contract's opening sentence still carried three global conditions — "produced by resolving today's template", "read back from a resource it just wrote", "about to install on a success path" — each of which is false at one site. That is the same defect `M3` names, one paragraph higher than where I fixed it.

## Verification

Rebased onto `f50f4cd0` at `76b63b44`, which carries #2829's import-path masking. Full suite 953 files / 20701 tests green; `check`, `docs`, `integ-destroy`, `integ-broad` and `verify-pr` all fresh at this head. Codex since your round 3: 2 delta rounds and 5 maintainer-checklist rounds, eleven items, every one measured rather than reasoned about — four of them corrections to statements I had just written, which is the honest shape of this round.

go-to-k:
Our comments crossed — yours landed 39 seconds after mine, so it answers round 3 (`M2` / `M3` / nits 1-5, all of which I have verified as done, and `M1` is correctly out of this PR; I fixed #2809's stale `severity:low` label to match its re-scoped body).

Still open and not in `76b63b44`: [`N1` and `N2`](https://github.com/go-to-k/cdkd/pull/2753#issuecomment-5595016678) — the `*` exemplar at `verify.sh:378`, and "usually" at `secret-redaction.ts:235` and `.claude/rules/layout-deployment-secrets.md:168`. Three lines; push them and I will merge.

go-to-k:
`N1` and `N2` are both correctly fixed at `f499d799` — the `*` exemplar is replaced with the measured set and the mechanism, and "usually" is now a conditional in both the docstring and the rules file. Thank you.

The close/reopen just above was me re-triggering CI, not a review action: **GitHub recorded your force-push at 03:45:52Z but created zero check suites for `f499d799`** (`check-suites` total 0, combined status `pending`/0, while the previous head `76b63b44` has 5). Not a config problem — no `paths` filter on `ci.yml`, Actions enabled, other PRs were dispatching normally at the same minute, and a fork-approval hold would have produced `action_required` runs rather than none. A dispatch miss on GitHub's side. Reopening fires `pull_request: reopened`, which the workflows accept, and it leaves your branch and its sha untouched — which matters here because the real-AWS markers are bound to that sha.

Nothing needed from you. I am running the two integs this merge needs from my own worktree (markers are per-worktree, so yours cannot carry over): `secrets-dynamic-ref` has passed — clean destroy, zero surviving state versions — and `lambda` is next. Once CI is green I will merge.

nix-tkobayashi:
I force-pushed `f499d799` to `66304976` at 04:24:28Z, about six minutes after your comment, without re-reading the thread. Please pause the `lambda` run if it is still underway. I am sorry — I changed the head while you were validating it, after you had said nothing was needed from me.

**What changed, measured.** Comparing each head's patch against its own merge base, the only difference is in `docs/_generated/integ-last-run.tsv`: the old patch's `lambda` row update disappears because the rebased head retains main's newer row. The patches for `src/`, `tests/`, `.claude/` and the changelog entry are unchanged.

Between the two merge bases, main changed comments in `src/deployment/retry.ts` and `src/deployment/retryable-errors.ts`; both paths are in the `integ-broad` activation scope. Those changes do not alter executable code.

**Your markers.** I cannot establish whether they remain valid in your worktree, and I do not want to assert it either way. `integ-destroy` is `hash: diff` against `origin/main`, and the scoped diff is unchanged in this comparison; `integ-broad` hashes its `.markgate-broad-integ-test` sentinel. The sha change alone does not establish that another AWS run is required — but you are the one who can see your own worktree, so I am flagging it rather than concluding.

**Validation on `66304976`.** The five local checks passed, and one full-suite run passed: 953 test files, 20,701 tests. I refreshed the `check`, `docs` and `verify-pr` markers and updated `.markgate-verify-pr-sha` to the new head. GitHub has created check runs for `66304976`.

**Why I did it.** My earlier polling showed `mergeable_state=dirty` and zero check runs. I interpreted that as a stale merge computation and rebased to trigger another evaluation. I should have re-read the thread before acting.

I will coordinate any further pushes with you while your validation is underway, and re-read the thread before changing the head.

go-to-k:
Merged as `c7dffa95`. Thank you — this was a hard one to get right, and the part that stood out was not the fix but the corrections: you caught the `resolveOutputs` identity claim and the docstring's three global conditions yourself, both of which I had read past, and you re-scoped #2809 after finding the `AWS::SNS::Subscription` replacement rather than defending the first reading. The habit of stating what a measurement does NOT establish is what made every round short.

No action needed on the crossed comments — that was mine, and #2840 is settled on the reviewer side.

Gate runs from my worktree, for the record: `secrets-dynamic-ref` and `lambda` both clean (0 errors, 0 orphans, 0 surviving state versions), full suite 953 files / 20701 tests, CI green. I will land those two ledger rows on `main` separately, since a fork branch has nowhere to commit them.

PR #2758, opened 2026-09-07
test(unit): carve the gitignored local settings path out of the scope checker as optional on disk

### LINKED ISSUE #2751: test(unit): check-scope-checker-inputs reports .claude/settings.local.json as an out-of-tree read target on any checkout that has one
## Summary

`tests/unit/scripts/check-scope-checker-inputs.test.ts` ("every existing out-of-tree read target is inside the check include") fails on any checkout that has a `.claude/settings.local.json`, since go-to-k/cdkd#2738 added `tests/unit/scripts/settings-bash-first-optout.test.ts`. That test never reads the file's contents: it passes the literal `LOCAL_SETTINGS = '.claude/settings.local.json'` to `git check-ignore -v` and `git ls-files`. The scope checker resolves bare literals as read candidates and checks the ones that exist on disk, so on a checkout where the (gitignored, per-developer) file exists it classifies that literal as an out-of-tree read target and reports it as missing from `.markgate.yml`'s `check.include`. CI and a checkout without the file never see it.

```
AssertionError: checker inputs outside the check include — add the path to .markgate.yml `check.include` (see the issue go-to-k/cdkd#2364 comment there), or record it as a repo-wide-scanner known limit: expected [ Array(1) ] to deeply equal []
+   ".claude/settings.local.json (read by tests/unit/scripts/settings-bash-first-optout.test.ts)",
```

Measured on a linked worktree at 8efbaf8e (origin/main c328db66 plus one commit that touches neither test nor `.markgate.yml`): with the file present, `mise exec -- vp test run tests/unit/scripts/check-scope-checker-inputs.test.ts` reports 1 failed / 18 passed; with the file moved aside, 19 passed.

`.markgate.yml`'s `.gitignore` entry (from the same PR) explains why `.gitignore` is a checker input and notes that the scope checker cannot see the `git check-ignore` read; it says nothing about the local settings path itself, which is the literal the checker does see.

## Which way to close it

The test's own message offers two doors. Adding the path to `check.include` would make a gitignored, per-developer file part of what stales the `check` marker, which does not fit the marker's meaning. The executable carve-out in the checker is `NON_READ_LITERALS` (the "KNOWN LIMITS" block in its header is documentation, not an exemption), and its fence asserts each entry's key, reason, on-disk existence and continued literal occurrence — so carving this path out means teaching that fence that an entry may be OPTIONAL on disk, or the clean checkouts fail instead.

## Classification

Session-fit: next (not this session) -- found while verifying an unrelated PR (go-to-k/cdkd#2742); the fix is in the scope checker and its fence, files that PR does not touch.
Severity: low -- CI is unaffected; a contributor with a local settings file gets one red critic and has to read its message to learn it is not theirs.
Effort: small (S) -- a `NON_READ_LITERALS` entry plus an optional-existence arm in its fence, or a decision to include the path.
Estimate: ~1 h -- the fence change and its own test; the decision between the two doors.
Dup-check: searched issues for `settings.local.json`, `check-scope-checker-inputs`, `checker inputs outside the check include`. Hits: go-to-k/cdkd#1455 / go-to-k/cdkd#1476 (closed hook-gate issues, unrelated); go-to-k/cdkd#2364 and go-to-k/cdkd#2381 (closed; genuine checker inputs missing from the include, and the hook / docs corpus coverage gaps); go-to-k/cdkd#2701 (open; five other tests' tracked-only `git ls-files` populations omitting untracked files, and it explicitly excludes this scope checker). None describes this one: a non-read literal classified as a read target only when an optional local file exists.

nix-tkobayashi:
Closes #2751

## What was wrong

`tests/unit/scripts/check-scope-checker-inputs.test.ts` ("every existing out-of-tree read target is inside the check include") failed on any checkout that has a `.claude/settings.local.json`. Since go-to-k/cdkd#2738, `settings-bash-first-optout.test.ts` spells that path as a literal (`LOCAL_SETTINGS`) it passes to `git check-ignore -v` and `git ls-files` — the file is never opened. The scope checker's bare-literal parser resolves every such literal and demands a `check.include` entry for each one that exists on disk, and the file is gitignored and per-developer, so the failure was invisible in CI and on a checkout without one, and red for a contributor who has one (measured: 1 failed / 18 passed with the file present, 19 passed with it moved aside).

## What changes (`tests/unit/scripts/check-scope-checker-inputs.test.ts`; no `src/` change)

- `.claude/settings.local.json` joins `NON_READ_LITERALS`, the carve-out for path-shaped literals that are arguments rather than reads, with its reason.
- A new `OPTIONAL_ON_DISK` set qualifies that entry. The carve-out fence's per-entry rules are now one pure function, `carveOutFenceProblems(entries, optional, exists, ignored)`: every non-optional entry must exist (a deleted file must not leave a dead exemption); an optional entry must be ignored by git whether or not it exists; every optional path must be a carve-out entry; every entry needs a reason. The live fence calls it with `existsSync` and `gitIgnores`. `gitIgnores` runs `git check-ignore -v` (deliberately WITHOUT `--no-index`: with the index consulted a TRACKED file is reported as not ignored even when a pattern matches it, so a tracked file cannot be listed as optional to dodge the existence check), counts only a `.gitignore:<line>:` source (a global excludes file or `.git/info/exclude` cannot make it true — so a verdict cannot differ between a developer machine with a global rule and CI), refuses a NEGATED pattern (`check-ignore -v` exits 0 and prints the `!pattern` line for a path a negation un-ignores, where `-q` exits 1), demands the TOP-LEVEL `.gitignore` specifically (a nested one wins and is named in the source column; refusing it is right because only the top-level file is inside the `check` include, so a rule moved down would stale nothing), refuses a path whose only ignore source is a global excludes file or `.git/info/exclude` (either would make the verdict differ between a developer machine and CI, which is the defect itself), passes `-c core.excludesFile=/dev/null` as belt-and-braces, and inspects the exit status (1 is "not ignored"; anything else — 128 for a non-repository or a bad pathspec — is rethrown, not read as a verdict). `OPTIONAL_ON_DISK`'s membership is pinned exactly beside the `NON_READ_LITERALS` key set.
- The direction-2 audit (an entry must still occur as a literal in `tests/unit`, or be dropped) covers the new entry unchanged.
- `.markgate.yml`: the `.gitignore` entry's comment now says the path that test hands to git is a literal the checker does see, and why it is deliberately not an include entry. `docs/changelog-cdkd.md`: one entry (1317 chars).

The other door the failure message offers — adding the path to `check.include` — would make a gitignored, per-developer file part of what stales the `check` marker, which does not fit what the marker attests.

## Tests

- `a per-developer literal is carved out of the bare-literal population`: the uncarved scan sees the literal and names `settings-bash-first-optout.test.ts` as its reader; the carved scan does not contain it; the JOIN idiom does not build it either. The case does not vary the file's presence and does not need to: a directory-shaped literal is never stat'ed, so presence cannot change the population.
- `the carve-out fence rules hold with the file present, absent, and under each invalid qualifier`: controlled inputs through `carveOutFenceProblems` — optional entry present and absent (no problem), a non-optional entry gone, an optional entry git does not ignore (refused even when it exists), an optional path that is not a carve-out entry, a too-short reason.
- `gitIgnores reads the index and the repo .gitignore only`: in a throwaway `git init` repo with `*.local.json` ignored, a tracked matching file reads as NOT ignored, an untracked matching one as ignored, a path ignored only through `.git/info/exclude` as NOT ignored, a path un-ignored by a `!` pattern as NOT ignored, a path ignored only by a nested `.gitignore` as NOT ignored, `.gitignore` itself as not, and a nonexistent working directory passed to `git -C` throws rather than reading as `false`; on this checkout the optional entry is ignored by the repo's own `.gitignore`. Built in a temporary repo because this checkout has no tracked file that also matches an ignore rule. The repo is created with `git -c init.templateDir= init` and `.git/info/` is made explicitly: that directory exists only because of the init template, so a machine setting `init.templateDir` (git-secrets' install instructions do) makes the `.git/info/exclude` write throw, which was green here and in CI and red on the maintainer's machine. The setup helper pipes stderr into any thrown error.
- The key-set assertion lists the three entries exactly; the live fence runs the same function against this checkout. The refusal message names both halves — `git does not ignore it through the top-level .gitignore (a tracked file is never ignored)` — because each half alone misdiagnoses one arm: naming only the top-level file blames it for a TRACKED file it does match, and naming only git blames git for a rule moved to a nested `.gitignore`. It is pinned on the tracked arm through the real predicate inside the throwaway repo, not only through an injected `() => false`, and keyed on `.claude/settings.local.json` — ignored by this repo, tracked by the throwaway one — so the root binding is load-bearing rather than incidental.

## Verification

- `vp check --fix` / `vp run check` / `vp run typecheck:test` / `vp run build`: 0 errors. The changed file passes with the local settings file present (22 cases) and absent (22 cases).
- Full `vp test run` at the default timeout on the tip WITH `.claude/settings.local.json` present — the configuration that failed before: 941 files, 20,426 passed (1 skipped), root-asserted in the PR worktree. Earlier revisions of this branch, each on the main it was rebased onto at the time: 939 / 20,378, 936 / 20,188, 934 / 20,034.
- Mutation probes, each restored byte-exact: the carve-out entry removed (3 cases fail with the file present); `gitIgnores` forced true (1); the carve-out bypassed for this path (2); `--no-index` added to `gitIgnores` (1, the throwaway-repo case); the optional branch disabled in the function (1); the `ignored` requirement dropped (1); the optional-subset check dropped (1); `OPTIONAL_ON_DISK` emptied with the file absent (1). Review-round probes: `OPTIONAL_ON_DISK` emptied with the file present (1 — the split verdict the first revision conceded is closed by the exact membership assertion); the `.gitignore:` source check dropped (1); the top-level anchor dropped so a nested `.gitignore` is accepted (1); the negation check dropped (1); the exit status left uninspected (1); `--no-index` added (1); the explicit `.git/info/` creation removed (1, and it fails here now that the template is disabled); the refusal message reverted to naming only the top-level file (2); the `root` binding dropped from the real-predicate assertion (1). Two survive and are reported as such: dropping the `-c core.excludesFile=/dev/null` override (equivalent under the source check — a global-file source is rejected by the prefix regardless; kept as belt-and-braces), and deleting the live fence call (equivalent on a healthy checkout, where every rule holds; the rules themselves are pinned by the controlled cases).
- Codex review loop on the first revision's delta to NO ISSUES (1 round) and the maintainer-checklist proxy to NO ISSUES (2 rounds — round 1: the fence rules were only testable against this checkout, and the tracked-file guarantee was asserted against a path no ignore rule matched); the review-round delta to NO ISSUES (2 rounds — round 1: `check-ignore -v` exits 0 for a path a negated pattern un-ignores, which the source check alone would have read as ignored); this body (7 rounds).


go-to-k:
Reviewed with two independent read-only reviewers (code + test adequacy) plus an inline pass over the diff. **No blockers.** The carve-out does not weaken the fence: the exact key-set assertion, the reason floor, the direction-2 occurrence audit and the new git-ignore qualifier all still apply to the new entry, and the JOIN parser is deliberately not carved, so a real read through that idiom is still caught.

Two premises checked by measurement rather than by reading the prose:

- `git check-ignore` without `--no-index` really does report a **tracked** file as not ignored (throwaway repo, `.gitignore` = `*.local.json`, `git add -f tracked.local.json` -> `check-ignore -q` rc=1; with `--no-index` rc=0; untracked matching rc=0; absent matching path rc=0). So `OPTIONAL_ON_DISK` cannot be used to dodge the existence rule for a tracked file.
- The carved literal is genuinely an argument, never a read: `.claude/settings.local.json` occurs once outside the fence, at `tests/unit/scripts/settings-bash-first-optout.test.ts:62`, reaching only `git check-ignore -v` and `git ls-files`. No `check.include` glob covers the path either, so the carve-out is the only door.

Findings, all minor or nit.

**1. `OPTIONAL_ON_DISK` membership is unpinned (minor).** The key-set assertion lists the three `NON_READ_LITERALS` keys exactly, but nothing lists `OPTIONAL_ON_DISK`'s. Emptying the set is **green on a checkout that has the file and red only on one that does not** -- a developer-vs-CI split verdict, which is the exact class this issue fixes. The PR body concedes this mutant survives; one line closes it in both disk states:

```ts
expect([...OPTIONAL_ON_DISK].sort()).toEqual(['.claude/settings.local.json']);
```

The dangerous direction (adding a tracked path) is already covered by the `ignored` rule -- this is the other one.

**2. `gitIgnores` asserts *that* git ignores the path, not *which* file says so (minor).** Same split, reached a different way. Measured on the reviewing machine: `~/.config/git/ignore:3` carries `**/.claude/settings.local.json`, and in a throwaway repo with **no `.gitignore` at all** `git check-ignore` still exits 0 from that global rule. So deleting `.gitignore:91` would leave the new assertion green here and red in CI. `settings-bash-first-optout.test.ts:305-315` hit this exact leak in #2738 and closed it with `-v` plus a `/^\.gitignore:/` match on stdout; the new helper regressed to `-q`. Today the repo rule alone does suffice -- `check-ignore -v` on this checkout reports `.gitignore:91` -- so this is about the fence's binding, not a live hole.

**3. `gitIgnores`'s bare `catch` conflates rc=1 with rc=128 (minor).** "not ignored", "not a git repository" / bad pathspec (measured rc=128) and git-missing all collapse to `false`. In the live fence that fails **closed**, which is the right polarity. In the throwaway-repo case, two of the three assertions expect `false`, so a broken invocation satisfies them and only the `true` one keeps the case honest. Inspecting `(err as { status?: number }).status` and rethrowing on anything other than 1 would remove the ambiguity.

**4. The throwaway repo is not config-isolated (nit).** No commit is made, so `user.*` / `init.defaultBranch` are irrelevant and the `try`/`finally` cleanup is sound -- but a contributor whose global excludes contain `*.local.json` would make the `untracked.local.json` assertion pass for the wrong reason. Worth flagging that the obvious fix is not sufficient: measured just now in a fresh repo, `GIT_CONFIG_GLOBAL=/dev/null` leaves the XDG excludes file active (rc=0), while `git -c core.excludesFile=/dev/null` actually isolates it (rc=1). `stdio: 'ignore'` on the setup `git()` helper also makes a setup failure opaque.

**5. The case title over-claims (nit).** "carved out whether or not the file exists" -- the case never varies presence. It rests on `extractBareTargets` calling `statSync` only for the root-file shape. Either say that in the comment, or drive `exists` through an injected predicate the way `carveOutFenceProblems` already allows.

**6. `.markgate.yml` comment (nit).** "carved out in the checker's `NON_READ_LITERALS` as optional on disk" -- the optional qualifier is a separate set, `OPTIONAL_ON_DISK`, not a property of `NON_READ_LITERALS`. Naming both would match the code.

**7. Pre-existing, not this PR's to fix: the measured-count comment is stale by ~80, not by 1.** Both reviewers independently proposed changing `Measured 97 on the shipping tree (99 with the carve-out disabled)` to 96, each reasoning by delta from the documented figure. Measured instead, by appending the probe that comment itself describes and writing the result to a file (the stream fence in `tests/setup.ts` swallows `console.error` on a passing test):

| tree | carved | uncarved (carve-out disabled) |
| --- | --- | --- |
| `origin/main` @ `333f3f64` | **178** | **180** |
| this branch @ `4d44ed14` | **175** | **178** |

Both deltas are internally consistent (2 carve-out entries on main, 3 here); the two trees' uncarved figures differ by 2 because this branch predates #2760 / #2763, which added literals. The comment was written in #2383, long before #2738 introduced this literal, so it was already ~80 off on `main` -- a separate issue rather than a change here. Worth noting alongside it that a floor of 75 was ~77% of the documented 97 and is ~42% of the real 178, so it fences considerably less than its own rationale claims.

Nothing above blocks the merge. 1 and 2 are the two worth taking, since both leave a verdict that differs between a developer checkout and CI.

nix-tkobayashi:
Thanks — 1 through 6 are taken in 614a8224 (one commit, rebased onto main; the fork tip was mine); 7 is yours as go-to-k/cdkd#2769 and is not touched here.

### 1. `OPTIONAL_ON_DISK` membership
Pinned exactly, in the same case as the `NON_READ_LITERALS` key set. Emptying the set now fails that case with the file present as well as absent — the split verdict the PR body had conceded is gone.

### 2. `gitIgnores` asserts which file says so
`check-ignore -v`, and only a `.gitignore:<line>:` source counts: a global excludes file or `.git/info/exclude` is rejected by the source, so deleting the repo's line reads red everywhere, your machine included. Pinned in the throwaway repo with a path ignored only through `.git/info/exclude` (reads as not ignored); dropping the source check fails that case. On this checkout the entry reports `.gitignore:` as its source, which is what CI sees. One thing `-v` changes that `-q` did not, found by the Codex pass on this round: `check-ignore -v` exits 0 for a path a NEGATED pattern un-ignores (it prints the `!pattern` line), where `-q` exits 1 — so the pattern column is read and a leading `!` reads as not ignored, pinned with a `!keep.local.json` line in the throwaway repo.

### 3. Bare `catch`
The thrown error's `status` is inspected: 1 is "not ignored", anything else is rethrown. The throwaway case asserts that a nonexistent directory passed to `git -C` throws rather than reads as `false`; collapsing every error to `false` fails it.

### 4. Config isolation
`-c core.excludesFile=/dev/null` is passed by the helper itself (your measurement that `GIT_CONFIG_GLOBAL=/dev/null` leaves the XDG file active is recorded in the comment), and the setup helper pipes stderr into the thrown error instead of discarding it. Reported as measured: with the source check of item 2 in place, dropping the `excludesFile` override survives every case — a global-file source is rejected by the `.gitignore:` prefix regardless — so it is belt-and-braces, not a fence of its own.

### 5. Case title
Retitled "carved out of the bare-literal population", and the comment says what it pins — the carve-out itself (seen uncarved, named to its reader, absent carved) — and why presence cannot matter (a directory-shaped literal is never stat'ed), rather than a presence it never varies.

### 6. `.markgate.yml` comment
Names both `NON_READ_LITERALS` and `OPTIONAL_ON_DISK`, and what the qualifier permits.

### Verification
- `vp check --fix` / `vp run check` / `vp run typecheck:test`: 0 errors; the file passes with `.claude/settings.local.json` present and absent (22 cases each); full `vp test run` at the default timeout on the rebased tree, with the file present: 939 files, 20,378 passed (1 skipped), root-asserted in the PR worktree.
- Mutation probes on the delta, each restored byte-exact: `OPTIONAL_ON_DISK` emptied with the file present (1 case fails); the source check dropped (1); the negation check dropped (1); the status left uninspected (1); `--no-index` added (1); the `excludesFile` override dropped (survives — equivalent under the source check, as above).
- Codex on the delta to NO ISSUES (2 rounds — round 1 found the negated-pattern case), on this reply (2 rounds).

Nothing further from my side — over to you.

go-to-k:
Round 1's six are all genuinely closed -- verified against the code rather than the description: `OPTIONAL_ON_DISK` is pinned exactly and presence-independently; the source check plus the `.git/info/exclude` arm bind the verdict to the repo's own rule; the status is inspected and a non-1 rethrown; the setup stderr reaches `error.message`; the case is retitled with a comment stating what it pins; `.markgate.yml` names both sets. The negated-pattern catch is a real one and worth recording: measured here, `check-ignore -q` exits 1 for a path a negation un-ignores while `-v` exits 0 and prints `.gitignore:2:!keep.local.json`, so switching to `-v` silently changes the exit-code contract and the `!` column read is load-bearing, not defensive.

One blocker in the fix round, though.

## Blocker -- the new throwaway-repo case fails outright on a machine that sets `init.templateDir`

```
FAIL  tests/unit/scripts/check-scope-checker-inputs.test.ts >
      gitIgnores reads the index and the repo .gitignore only: ...
Error: ENOENT: no such file or directory, open
  '/var/folders/.../cdkd-scope-fence-XBfZMC/.git/info/exclude'
  at check-scope-checker-inputs.test.ts:928
Tests  1 failed | 21 passed (22)
```

Measured on `614a8224`, git 2.53.0. `tests/unit/scripts/` as a whole: 1 failed / 3240 passed (86 files passed, 1 failed) -- this case is the only failure, so nothing else is affected.

`.git/info/` is created ONLY by the init template, and `init.templateDir` is an ordinary global config key:

| `git init` under | `.git/info/` |
| --- | --- |
| this machine's `init.templateDir` (`~/.git-templates/git-secrets`, which git-secrets' own install instructions set) | **absent** |
| `-c init.templateDir=` (template disabled) | **absent** |
| the default system template | present |

So `writeFileSync(join(root, '.git', 'info', 'exclude'), ...)` at line 928 throws before `git add` ever runs. CI has the default template and stays green.

That makes this **the developer-red / CI-green split this PR exists to remove, reintroduced inside the fix for item 2** -- and the case's own comment at line 911 asserts "The repo is config-isolated", which is not true of the setup: the isolation was applied inside `gitIgnores` (`-c core.excludesFile=/dev/null`) and never to `git init`, which also inherits the developer's template hooks.

This repo has already hit this exact trap and written down the remedy, at `.claude/hooks/stop-warn.test.sh:60-63`:

> `mkdir -p` is required, not defensive: this git does NOT create `.git/info/` on `init`, so the redirect below fails silently (no `set -e` here) and the exclude never exists. That is how the first attempt still saw `?? .claude/`.

Suggested fix: `mkdirSync(join(root, '.git', 'info'), { recursive: true })` before the write (adding `mkdirSync` to the `node:fs` import), and `git('-c', 'init.templateDir=', 'init', '-q')` so the setup is isolated in the way the comment claims -- note the second alone does not help, since disabling the template also leaves `.git/info/` absent. Worth correcting line 911 to say what is actually isolated.

## Minor -- the `^\.gitignore:` anchor misdiagnoses a nested `.gitignore`

`check-scope-checker-inputs.test.ts:236`. A nested ignore file outranks the root one and `check-ignore -v` names it as the source. Measured in a throwaway repo:

```
.claude/.gitignore:1:settings.local.json	.claude/settings.local.json
```

The anchor rejects that, so moving the rule into `.claude/.gitignore` would make the fence report `.claude/settings.local.json is listed as optional on disk but git does not ignore it` -- the wrong cause, on a repo that does ignore it. The direction is right (only the root `.gitignore` is in `.markgate.yml`'s `check` include, so a nested file would stale nothing), but neither the JSDoc nor the message says the constraint is "the top-level `.gitignore`" specifically. Wording the message that way, and listing the nested case in the JSDoc beside `.git/info/exclude` and the global file, would make the refusal self-explaining.

## Nit -- `-c core.excludesFile=/dev/null` is inert given the source check

`check-scope-checker-inputs.test.ts:226`. The global excludes file is the lowest-precedence source: it can never win over a `.gitignore` match, and when it does win, the source column names it and the anchor already rejects it. So the override changes no verdict the source check does not decide, which is consistent with the case's own note at line 914 that the arm cannot be exercised. Worth keeping as belt-and-braces -- just not described as a second defence.

## Also checked and clean

`status === 1` is unambiguous (`check-ignore` uses 0/1/128 only, and a spawn failure yields `status: undefined`, which is rethrown); the `!` check does not misfire on an escaped `\!foo` pattern (git prints the backslash verbatim); single-pathspec output is one line, so the missing `m` flag is harmless; the tmpdir is removed in `finally`; the changelog entry is 1318 bytes, under the cap.

Only the blocker needs to move before this is mergeable.

nix-tkobayashi:
Taken, with the reproduction. The blocker is exactly the shape you name. I missed it because I tested only with this machine's default template, under which `.git/info/` exists after `git init` and the case is green; disabling the template reproduces the absence here, on git 2.43.0, so it was discoverable on this machine and I did not look:

```console
$ d=$(mktemp -d); git -C "$d" init -q;                      ls -d "$d/.git/info"
/tmp/tmp.XXXX/.git/info
$ d=$(mktemp -d); git -C "$d" -c init.templateDir= init -q; ls -d "$d/.git/info"
ls: cannot access '/tmp/tmp.YYYY/.git/info': No such file or directory
```

Both halves of your suggested fix are in d391eacc: `git -c init.templateDir= init -q`, and an explicit `mkdirSync(root/.git/info, { recursive: true })` since disabling the template leaves the directory absent too. The comment no longer claims the repo is "config-isolated" on the strength of one axis; it names the two isolations separately and cites `.claude/hooks/stop-warn.test.sh:60-63`, which had already recorded this trap. With the template now disabled, removing the `mkdirSync` exposes the missing-directory failure on this machine too, so the case no longer depends on which template the runner has.

Two things I could have done and did not, since they are the same lesson the PR is about. My probes mutated the test helper and asked whether a case failed; the fixture's own setup was never varied, though varying it was available and this PR is nothing but test code. And `.git/info` was greppable prior art in this repo — `stop-warn.test.sh` records exactly this — which I did not search for before writing into a tool's internal directory.

### Minor — the nested `.gitignore`
Taken, and it reproduces here (`nested/.gitignore:1:nested-only.txt` in the source column, rc=0). The direction stays, for the reason you give: only the top-level file is inside `.markgate.yml`'s `check` include, so a rule moved down would stale nothing. What changed is that the JSDoc and the fence's message now say "top-level" specifically, so the refusal explains itself instead of reporting the wrong cause; the message is now `<path> is listed as optional on disk but the top-level .gitignore does not cover it`. A nested arm pins it, and dropping the `^` anchor fails that arm.

### Nit — `-c core.excludesFile=/dev/null`
Taken as worded: the JSDoc now says it changes no verdict the source column does not already decide, and keeps it as belt-and-braces rather than calling it a second defence.

### Verification
- `vp check --fix` / `vp run check` / `vp run typecheck:test` / `vp run build` / `gen:all-matrices` / `audit:coverage:check`: 0 errors on this machine. Full `vp test run` on the rebased tree with `.claude/settings.local.json` present: 940 files, 20,385 passed (1 skipped); the runner asserts the vitest root is this PR's worktree, so the count is that tree's and not another lane's.
- Probes on this round, each restored from a byte-exact copy: the top-level anchor dropped (1 case fails); the explicit `.git/info/` creation removed (1). Removing the `init.templateDir=` isolation survives on this machine with the `mkdirSync` present; the case does not independently enforce template isolation, and it is kept because `git init` also inherits the developer's template hooks.
- Codex on the delta to NO ISSUES (1 round), on this reply (2 rounds).

Nothing further from my side — over to you.

go-to-k:
The blocker is closed. Reproduced the fix on this machine at `d391eacc`: the file passes (22 cases), and the full suite is `940 files, 20,385 passed (1 skipped)` with the vitest root asserted to this worktree — byte-identical to your numbers. `vp check --fix` / `vp run check` / `vp run typecheck:test` / `vp run build` all clean, and `gen:all-matrices` + `audit:coverage:check` + `format` leave the tree unchanged.

Both halves verified independently: `git -c init.templateDir= init -q` leaves `.git` with only `HEAD config objects refs` — no `info/`, no `hooks/`, no `description` — so the `mkdirSync` is what does the work, and nothing else in the case needs template output. The nested arm genuinely discriminates: against `nested/.gitignore:1:nested-only.txt`, dropping the `^` or widening to `^[^\t]*\.gitignore:` both turn it red.

Three small things left, none of them blocking. The first is mine to own -- my round-2 wording suggestion introduced it.

## Minor -- the reworded message is false in the one arm the qualifier exists to catch

`tests/unit/scripts/check-scope-checker-inputs.test.ts:279`. `ignored()` returns false for a TRACKED file matching a top-level pattern, and there the top-level `.gitignore` **does** cover it -- the index is what refuses. Measured:

```console
$ git -c core.excludesFile=/dev/null check-ignore -v -- tracked.local.json
$ echo $?
1
$ git -c core.excludesFile=/dev/null check-ignore -v --no-index -- tracked.local.json
.gitignore:1:*.local.json	tracked.local.json
$ echo $?
0
```

That is exactly the abuse the design names ("a TRACKED file cannot be listed here to dodge the existence check"), so the most likely real trigger of this message now points the reader at the wrong file. Round 2's `but git does not ignore it` was accurate in all five arms; my suggestion traded that accuracy for the nested-`.gitignore` case's clarity, and both are gettable at once:

```
${rel} is listed as optional on disk but git does not ignore it through the top-level .gitignore (a tracked file is never ignored)
```

Worth noting no test pins the wording for that arm -- the controlled case drives `ignored` as `() => false`, so nothing would have caught the regression.

## Nit -- `:209` "Three things are deliberate:" heads four bullets

Pre-existing from round 2, but round 3 rewrote bullet 2 and left the count.

## Nit -- `:263` JSDoc drift

`carveOutFenceProblems`'s doc still says an optional entry "must be ignored by git"; the predicate now demands the top-level `.gitignore` specifically. Same sentence as the message above, so the two can move together.

## Docs -- the changelog entry has not moved since round 1

`docs/changelog-cdkd.md`, still 1318 chars and still describing the qualifier as:

> for an optional entry demands instead that **git ignores it** (`gitIgnores`, pinned against a tracked path and the ignored one)

Two rounds have since narrowed that to the repo's TOP-LEVEL `.gitignore` — a global excludes file, `.git/info/exclude` and a nested `.gitignore` are each refused by the source column, and the fence's own message now says so. A reader adding a second `OPTIONAL_ON_DISK` entry from this paragraph would get the rule wrong. The `-v` + source + negation read and the `init.templateDir` / `.git/info` isolation are the substance of two review rounds and appear nowhere in the entry either; the cap leaves room (1318 of 2000). The rest of the paragraph checks out — I confirmed the `existsSync` filter really is in the main assertion, so "demands an include entry for every one that exists on disk" is accurate.

## On the process note in your reply

Agreed, and the same lesson applies to me: I suggested the message rewording without asking which arms it had to stay true for, and the tracked-file arm is the one it broke. Varying the fixture's own setup, and grepping the repo for prior art before writing into a tool's internal directory, are both cheaper than the round they cost.

None of the above changes behaviour or needs a new test — one string, two comments, one changelog paragraph. After that I have nothing further.

nix-tkobayashi:
All four taken in 80747e4c, and the message one is the more interesting: your round-2 wording and round-1's were each true of a different arm, so the fix is to say both.

### Minor — the message
`<rel> is listed as optional on disk but git does not ignore it through the top-level .gitignore (a tracked file is never ignored)`. Naming only the top-level file blames it for a TRACKED file it does match, where the index is what refuses; naming only git blames git for a rule moved to a nested `.gitignore`, which git honours. The comment beside it records that each half was once wrong on its own, so the next reader does not simplify it back.

I also added an assertion that pins the full message for the tracked-file arm through the real predicate: inside the throwaway repo, `carveOutFenceProblems` is driven with the REAL predicate against the tracked file and its message asserted whole. Reverting to the round-2 wording now fails two cases, the controlled one and that one. It is the only behaviour-adjacent addition in this round; the rest is strings and comments.

### Nits
`Three things are deliberate` now reads `Four`, and `carveOutFenceProblems`' JSDoc says an optional entry must be ignored through the repo's top-level `.gitignore` rather than "by git" — the same sentence as the message, moved together as you suggested.

### Docs — the changelog entry
Rewritten, 1971 of 2000 chars. It now carries what the two rounds established: the `-v` read and the source column, why each rejected source is rejected (a global file or `.git/info/exclude` because either makes the verdict differ between a developer machine and CI, which is the defect itself; a nested `.gitignore` because it is outside the gate's `check` include, so a rule moved there would stale nothing), the negated-pattern refusal, that a nonzero exit other than 1 is rethrown, that a tracked file is never ignored, and the `init.templateDir` / `.git/info` isolation with the `stop-warn.test.sh` precedent. The Codex pass on it caught two of my own errors in that paragraph: I had written the nested case as another developer-versus-CI split, which it is not, and "a non-1 exit is rethrown", which wrongly includes the exit-0 success path.

### On the process note
Taken as mutual. Your rewording was a fair reading of the arm in front of it; what neither of us asked was which other arms the sentence had to stay true for, and the answer was one line of enumeration away.

### Verification
- `vp check --fix` / `vp run check` / `vp run typecheck:test` / `vp run build` / `gen:all-matrices` / `audit:coverage:check`: 0 errors on this machine. Full `vp test run` with `.claude/settings.local.json` present: 940 files, 20,385 passed (1 skipped), vitest root asserted to this worktree. Two earlier attempts failed on `tests/unit/local/*` and `gen-sdk-attr-coverage` 5-second timeouts while another lane's suite ran on the same host; each passes alone, and the run reported here had none.
- Probes, each restored from a byte-exact copy: the message reverted to the round-2 wording (2 cases fail).
- Codex on the delta to NO ISSUES (3 rounds — round 1 caught the two changelog errors above; round 2 reported them still present, because the rewrite had overrun the 2000-character cap and my edit aborted rather than truncating), on this reply (2 rounds).

Nothing further from my side — over to you.

go-to-k:
The message fix is right, and I verified it rather than taking it: reverting the producer's string to round 3's wording fails exactly 2 cases, the controlled one and the new real-predicate one. Full local run on `80747e4c` with the vitest root asserted to this worktree: `940 files, 20,385 passed (1 skipped)` -- identical to your numbers; `vp check --fix` / `vp run check` / `typecheck:test` / `build` / `gen:all-matrices` / `audit:coverage:check` all clean with no tree drift; the changelog entry is 1972 bytes and `changelog-entry-size.test.ts` passes. Every CI check on the PR is green.

**No code defects.** Everything below is prose accuracy, and I want to name the pattern before listing them, because it is now the third consecutive round in which the previous round's fix carried one.

## The pattern, before the items

Round 2's fix carried a blocker. Round 3's fix carried a message that was false in one arm. Round 4's fix carries the four below. Every one of them is an ENUMERATION or a COUNT in prose drifting from the code beside it -- "Three things" over four bullets, a rule list missing its first rule, a reason that holds for one path and is stated for all, a measured number nobody re-measures. Nothing in the suite watches any of them, so each round fixes the instances found and creates the next ones.

That is a rule-shaped problem, not a proofreading one, so I would rather not trade another round of round-trips for it. My suggestion: take items 1 and 2 (they are one line each), and treat 3 and 4 as optional -- I am happy to merge with them, and would rather see the class fenced later than see a sixth round of prose edits.

## 1. `carveOutFenceProblems`' JSDoc lists 3 of the 4 rules it enforces (minor)

`tests/unit/scripts/check-scope-checker-inputs.test.ts:255-267`. "Returns one message per violated rule:" then exist / ignored / optional-is-an-entry. The FIRST rule the loop applies -- `why.length <= 20` -> `needs a stated reason`, line 276 -- is not in the list, though the controlled case pins it. Exactly the drift round 3 fixed in the sibling JSDoc one function up, reintroduced one function down, which is what makes me want the class fenced rather than the instance fixed.

## 2. The changelog's nested-`.gitignore` reason is true of one path, stated of all (nit)

`docs/changelog-cdkd.md`: "a nested `.gitignore` because it is outside the gate's `check` include, so a rule moved there would stale nothing". True for `.claude/.gitignore`, which is how the JSDoc correctly scopes it. Not true generally -- `check.include` carries `src/**`, `tests/**`, `scripts/**`, `docs/**`, `.claude/skills/**`, `.claude/rules/**`, `.claude/agents/**` and `.claude/hooks/**`, so a nested `.gitignore` under any of those IS inside the include. Scoping the sentence the way the JSDoc does fixes it.

Same paragraph, smaller: "a `!`-negated pattern because `-v` exits 0 where `-q` exits 1" gives the reason the CHECK is needed, in a list whose other members give the reason the SOURCE is refused. A negation is refused because it means the path is not ignored; the exit-code difference is why `-v` cannot be trusted to say so by its status alone.

## 3. The new real-predicate assertion does not pin the arm its comment names (nit)

`check-scope-checker-inputs.test.ts:995-1006`. The binding is correct -- `(rel) => gitIgnores(rel, root)` -- but it is not load-bearing. Measured on this checkout:

```console
$ git -c core.excludesFile=/dev/null check-ignore -v -- tracked.local.json
$ echo $?
1
```

This repo's `.gitignore:92` is the literal `.claude/settings.local.json`, with no `*.local.json` glob, so `tracked.local.json` is unignored at `REPO_ROOT` too and the default root yields the same `false` and the same message. Dropping `root` leaves the assertion green. It genuinely pins the MESSAGE -- my mutation probe above confirms that -- but not the TRACKED arm the comment claims, so it reads stronger than it is. Keying the entry on `.claude/settings.local.json` (which the throwaway repo tracks) would make the root binding load-bearing: dropping it would then return `ignored=true`, `[]`, and a red.

## 4. The measured-count comment, for the record only

`check-scope-checker-inputs.test.ts:806`'s `Measured 97 on the shipping tree (99 with the carve-out disabled)` is stale, and this PR moves the gap from 2 to 3, so it is now internally inconsistent as well. It is go-to-k/cdkd#2769 and stays out of this PR.

One thing belongs here rather than in the issue, though. Re-measured on `80747e4c`: **178 carved, 181 uncarved** (`main` is 178 / 180). **Three independent reviewers have now proposed "96" for that comment, in three separate rounds of this PR**, each reasoning by delta from the documented figure instead of running the probe the comment describes. I have recorded that on the issue as evidence about the comment's shape: a stale absolute in prose invites arithmetic on itself, and the arithmetic is wrong every time because the base is. The fix there should be relational, not another number.

## Where this leaves the merge

Items 1 and 2 are one line each and I would take them. Items 3 and 4 need no action in this PR. Push 1 and 2 and I will merge on the green; or tell me you would rather ship as-is and I will merge and file 1 and 2 alongside go-to-k/cdkd#2769.

nix-tkobayashi:
Items 1 and 2 are in ed1ecb8d, and I took 3 as well — not to prolong the round, but because it is the same class you are naming: a comment claiming more than the code does. 4 stays with go-to-k/cdkd#2769.

On the pattern first, since you are right that it is rule-shaped. Three of the four are a count or a list in prose that no test reads. The carve-out keys and the optional-set membership both have exact assertions; the JSDoc rule list has no equivalent. I have not built that here.

### 1. The missing rule
Added, and described as the function actually branches: the reason floor per entry, then one of two mutually exclusive arms — ignored-through-the-top-level-`.gitignore` for an optional entry, exists-on-disk for any other — and then the optional-is-an-entry pass over the set.

### 2. The nested-`.gitignore` reason
Scoped to `.claude/.gitignore`, the way the JSDoc has it. You are right that the general form is false: `check.include` carries `src/**`, `tests/**`, `scripts/**`, `docs/**` and four `.claude/**` globs, so a nested file under any of those is inside the include. And the negation now gives the reason the SOURCE is refused — a negation means the path is not ignored — with the exit-code difference as what would otherwise hide it, rather than standing in as the reason.

### 3. The assertion that did not pin its arm
Taken, and your diagnosis was exact: this repo's `.gitignore` carries the literal `.claude/settings.local.json`, not a `*.local.json` glob, so `tracked.local.json` answered `false` at both roots and `root` was free to drop. The entry is now keyed on `.claude/settings.local.json`, which this repo ignores and the throwaway repo now tracks (`git add -f`), so the binding is load-bearing in the direction you named: dropping `root` answers `true` at `REPO_ROOT`, so the actual result becomes `[]` against an unchanged expectation and the case reds. Measured — dropping the `root` argument fails exactly that case.

### 4. The measured count
Left alone, and the re-measurement noted: 178 carved / 181 uncarved here, and this PR moves the gap to 3. I will not add a fourth "96" to the pile. Your reading of it is the one I would act on if I take go-to-k/cdkd#2769: the fix is relational, not another absolute — the floor should be expressed against the population the probe measures at run time, so nobody has to trust a number in a comment.

### Verification
- `vp check --fix` / `vp run check` / `vp run typecheck:test` / `vp run build` / `gen:all-matrices` / `audit:coverage:check`: 0 errors. Rebased onto main (the changelog conflicted, resolved keep-both with my entry first). Full `vp test run` with `.claude/settings.local.json` present: 941 files, 20,426 passed (1 skipped), vitest root asserted to this worktree.
- Probes, each restored from a byte-exact copy: the `root` binding dropped from the real-predicate assertion (1 case fails, where it did not before this round); the refusal message reverted to the round-3 wording (2).
- Codex on the delta to NO ISSUES (2 rounds — round 1 caught a false claim of loop order in the JSDoc I had just added, and two overstatements in this reply), on this reply (2 rounds).

Nothing further from my side.

go-to-k:
Merging. All four round-5 items are taken, and I verified the load-bearing one by mutation rather than by reading: dropping the `root` argument from the re-keyed real-predicate assertion now kills exactly one case, isolated against an unmutated-copy control (the control itself reds two unrelated cases as an artifact of being a copy; the mutation adds the third). Re-keying it to `.claude/settings.local.json` is what made that binding matter — with `tracked.local.json` the default root answered `false` too.

Measured on `7ef296e4`: `941 files, 20,426 passed (1 skipped)` with the vitest root asserted to this worktree; `vp check --fix` / `vp run check` / `typecheck:test` / `build` / `gen:all-matrices` / `audit:coverage:check` / `format` all clean with no tree drift; the changelog entry is 1993 bytes and both changelog fences pass. CI is `CLEAN`, every check green. One small thing for the record: the body's `939 files, 20,378 passed` predates the last rebase — not wrong for the tree it names, just no longer the tip.

## What I am NOT blocking on

Three prose items and one fragility are filed as go-to-k/cdkd#2786 rather than turned into a sixth round, with the four classification lines and a dup-check:

1. `carveOutFenceProblems`' new "in the order the loop applies them" is inaccurate — the `if/else` tests `optional.has` first, so the ignored bullet precedes the exists bullet, and the output is entry-major, so "one message per violated rule" is really one per (entry, rule) pair. Measured with an optional entry inserted first: `['b/opt.json … does not ignore it …', 'a/gone.txt no longer exists']`. The reason rule IS applied first for every entry, so that half holds.
2. The same JSDoc still says the top-level file "is the only one inside `.markgate.yml`'s `check` include" — the universal that round 5 correctly retreated from in the changelog. `check.include` carries `src/**`, `tests/**`, `scripts/**`, `docs/**` and four `.claude/*/**` globs, each of which would cover a nested `.gitignore` beneath it; the claim holds only because no include entry covers a bare `.claude/`.
3. `.markgate.yml:311`'s "The path **that test** passes to git" — nearest antecedent is `check-scope-checker-inputs.test.ts`, the previous sentence's subject, where the intended one is `settings-bash-first-optout.test.ts`. This PR makes the checker test also hand that path to git, so both readings are now true, which is what hides it.
4. The changelog entry sits at 1993 of 2000 bytes. Any word added trips the size fence; ~40 bytes of trim would leave room for the next correction.

One reported item I am recording as won't-do: the `// The three per-entry rules` comment is accurate as scoped — there are exactly three per-entry rules, and the fourth is set-scoped, which the sentence does not claim.

The stale `Measured 97` count stays go-to-k/cdkd#2769, now carrying a re-measurement on this head (178 carved / 181 uncarved; `main` is 178 / 180) and the suggestion to state the gap relationally instead of as another absolute.

## Why these are filed rather than fixed here

Five rounds, and every one of them found the same class in the previous round's fix: an enumeration or a count in prose drifting from the code beside it. Round 2's blocker, round 3's false message arm, round 4's four, round 5's three. Fixing the instances is what produced the next instances, and nothing in the suite watches any of them — which makes the durable remedy a fence, not a sixth proofreading pass. go-to-k/cdkd#2786 carries that as the real content.

## Thanks

This was a careful piece of work, and the parts that made it careful are worth naming: the negated-pattern behaviour of `check-ignore -v` was a genuine find that a `-q`→`-v` change would otherwise have shipped silently; the reproduction of my `init.templateDir` blocker on your own machine, by disabling the template rather than taking my word for it, is exactly the right response to a bug report you cannot see; and each round's mutation probes, restored byte-exact and reported honestly including the survivors, are why I could verify claims instead of re-deriving them. Two of my own review suggestions were wrong in ways your rounds exposed — the message rewording that broke the tracked arm, and the "in the order" phrasing I effectively prompted. Thank you.

nix-tkobayashi:
The stale count is fixed in the body rather than left for the squash message: it now reads the tip's `941 files, 20,426 passed (1 skipped)`, with the three earlier revisions listed as what they were, each against the main it was rebased onto at the time. The body update left the head at `ed1ecb8d`.

go-to-k/cdkd#2786 is the right call, and item 1 there is the sharpest version of the class: I wrote "in the order the loop applies them" while adding a bullet to that very list, which is the failure the issue is about happening inside the fix for the previous instance of it. That wording is gone at `ed1ecb8d` — the round-5 correction replaced the bullet list with prose describing the two mutually exclusive branches — but the naming was mine. Your measurement makes the entry-major behaviour explicit: the per-entry checks emit one message per (entry, rule) pair, then the optional-membership pass runs separately over the set.

Item 2 is the one I would not have caught by re-reading, and my round-5 sentence overstates a rationale that holds for `.claude/settings.local.json` specifically. The broad include globs were already there; the claim was never true in the universal form I gave it.

One possible follow-up, for whoever picks the issue up: an assertion for the entry-order behaviour, and a separate coverage assertion derived from `check.include` rather than restated beside it. The `NON_READ_LITERALS` key-set assertion is a useful precedent for the shape, though enumerating branches alone would not establish either property.

On the two suggestions of yours my rounds broke: both were reasonable readings of the arm in front of them, and in each case what was missing was the enumeration of the other arms the sentence had to survive. That is the same shortfall as items 1 and 2, from the reviewing side, which is a fair argument that the fence belongs in the file rather than in either of our habits.

Thanks for the depth of the reviews. The `init.templateDir` blocker in particular was invisible from here in the only way that matters — green on my machine and green in CI — and I would have shipped it.

go-to-k:
Merged as 2998f1b6. Thank you — this was a genuinely careful piece of work over six rounds.

Verified at `ed1ecb8d` before merging: `941 files, 20,426 passed (1 skipped)` with the vitest root asserted; `vp check --fix` / `vp run check` / `typecheck:test` / `build` / `gen:all-matrices` / `audit:coverage:check` / `format` clean with no tree drift; the changelog rebase resolved keep-both without duplicating the entry (uniqueness fence green, 1993 bytes); CI `CLEAN` on every check.

Three things stood out, and they are the reason I could verify rather than re-derive:

- The negated-pattern behaviour of `check-ignore -v` — that it exits 0 printing the `!pattern` line where `-q` exits 1 — was a real find. A `-q` to `-v` change would have shipped that silently.
- Reproducing my `init.templateDir` blocker on your own machine by disabling the template, rather than taking the report on trust, is the right response to a bug you cannot otherwise see.
- Every round's mutation probes, restored byte-exact and reported honestly including the survivors and the ones that survive for a good reason.

Two of my own review suggestions were wrong in ways your rounds exposed: the message rewording that broke the tracked arm, and the "in the order the loop applies them" phrasing I effectively prompted, which your Codex pass caught before I did.

Items 2, 3 and 4 from my last comment stay open as go-to-k/cdkd#2786 (the JSDoc's universal claim about the `check` include, the `.markgate.yml` pronoun, and the entry's 7 bytes of headroom), with item 1 recorded there as already fixed by `ed1ecb8d`. The measured count remains go-to-k/cdkd#2769. Neither needs anything from you — the real content in both is the fence, not the instances.

PR #2764, opened 2026-09-07
fix(resolver): resolve the Fn::Sub variable map into a fresh object instead of writing the caller's template in place

### LINKED ISSUE #2739: fix(resolver): resolveSub resolves the Fn::Sub variable map in place, leaving the resolved plaintext in the caller's template object
## What is wrong

`IntrinsicFunctionResolver.resolveSub` (`src/deployment/intrinsic-function-resolver.ts`) resolves the variable map of the two-argument `Fn::Sub` form IN PLACE:

```ts
[template, variables] = subArgs;
for (const [key, val] of Object.entries(variables)) {
  variables[key] = await this.resolveValue(val, context);
}
```

`variables` is the object inside the caller's template — `output.Value['Fn::Sub'][1]`, or a resource property's — so after one resolution the template object itself holds the RESOLVED value where it held an intrinsic or a `{{resolve:...}}` reference. When that value is a secret, the plaintext now sits in the in-memory template. The plain-string `Fn::Sub` form and `Fn::Join` do not mutate their input.

Measured (2026-09-07, while writing the go-to-k/cdkd#2728 unit cases): a test that shared one `Fn::Sub` literal across cases saw the second case never resolve — and never RECORD — the inner `{{resolve:secretsmanager:...:SecretString:password}}` reference, because the first case had replaced `variables.Pw` with the password; that case's resolver saw only the assembled outer reference, so its recording map lacked the needle and the masking under test could not fire. The PR builds the template per case; this issue is about the resolver.

## Why it matters

The consequence is a template object that lies about itself for the rest of the process, and two readers can be handed the lie:

- **A second resolution of the same object with a fresh recording map.** Every recording map is per pass (`buildResolverContext` creates it), so a later `resolve` of the mutated object returns the plaintext WITHOUT recording it — the `{{resolve:...}}` token is gone, so nothing in `resolveDynamicReferences` runs — and a redaction keyed on that pass's map then leaves the value unchanged. No such second resolution of one template object is known in the tree today (`resolveOutputs` runs once per deploy, on one of two mutually exclusive callers; a nested-stack child parses its own template; `cdkd diff` / `scrub` start from a fresh parse in their own process), which is why the measured instance is the test scaffold's. The invariant it breaks — "a template object is a description, not a cache" — is what a retry or a re-resolution added later would silently rely on.
- **The positioning source.** `DeployEngine.resolveOutputs` records `this.outputsTemplateSource[key] = output.Value`, the object `redactSecretsForState` positions a leaf by (issue go-to-k/cdkd#1910). An `Fn::Sub` source is an object, so the positioned pass takes `positionByCrossStackSource` / `positionByIntrinsicSkeleton` (`src/deployment/secret-redaction.ts`); the skeleton path needs the whole resolved leaf to be a recorded plaintext, so this spelling already positions an EMBEDDED secret by the value scan, mutation or not — and `intrinsicSkeletonPattern` derives the skeleton from the `Fn::Sub` template STRING and its placeholder positions, ignoring the variable map, so the mutation changes neither the skeleton nor the positioning result today. The stored positioning source simply carries plaintext where it carried the reference: a retained template description that is no longer true. Not a measured wrong result.

Bounded: the template object is never persisted (state holds resolved / redacted values), `cdkd export` re-synthesizes, and the per-pass map of the FIRST resolution does record the secret, so the deploy that mutates the object redacts correctly. The exposure is the in-process one above, and it is latent today.

## Proposed fix

Resolve into a fresh object — `const resolved: Record<string, unknown> = {}; for (...) resolved[key] = await this.resolveValue(val, context);` — and read `resolved` where `variables` is read below. A unit case in `tests/unit/deployment/` that resolves one `Fn::Sub` template object twice through the real resolver over a faked Secrets Manager client, with a FRESH recording map for the second resolution, and asserts (a) the template object is byte-identical after the first resolution (`JSON.stringify` before / after) and (b) the second map holds the inner reference's pair — served by the resolver's cache-hit arm, which re-records into the map it is handed (`resolveDynamicReferences`, the `cached.secret && cached.value` branch), so the fake client answering once is the expected count, not twice. Verification a fresh session can run: that case, then `/run-integ secrets-dynamic-ref` (its `OutputFailureLeak` output under `CDKD_TEST_OUTPUT_LEAK=true` is the live `Fn::Sub`-with-variables shape) and `/run-integ lambda` for the `integ-broad` marker.

## Classification

Dup-check: searched open and closed issues for `resolveSub variables in place mutates`, `Fn::Sub variable map mutated template`, `resolveSub mutation` — no hit; go-to-k/cdkd#2485 / go-to-k/cdkd#2516 are positioning residuals this could feed but do not name the cause.
Session-fit: next (not this session) — a resolver change on the cross-cutting `resolveSub` path with its own broad-integ cycle; found while fixing go-to-k/cdkd#2728, whose PR is scoped to the masking sites. The repro is the ten-line unit case above, so nothing is lost by deferring.
Severity: low — no plaintext is persisted or printed by this today; the second-resolution reader does not exist in the tree, and the first resolution's own map records the secret. It is an invariant break waiting for a caller.
Effort: small (S) — a three-line change plus one unit case, riding the existing `secrets-dynamic-ref` and `lambda` fixtures.
Estimate: ~1 h — the unit case's faked client and the two integ runs.

nix-tkobayashi:
Closes #2739

## What

`IntrinsicFunctionResolver.resolveSub` resolved the two-argument `Fn::Sub` form's variable map IN PLACE: each resolved value was written back into `subArgs[1]`, the object inside the caller's template (an Output's `Value['Fn::Sub'][1]`, a resource property's). After one resolution the template held a secret's plaintext where it had held the `{{resolve:...}}` reference or an intrinsic. Nothing persisted that object and the first pass's own map recorded the secret, but two readers could be handed the lie: a later resolution of the same object with a fresh per-pass recording map returned the plaintext WITHOUT recording it (no token left for `resolveDynamicReferences` to see), and the positioning source `DeployEngine.resolveOutputs` retains carried the secret. Measured while writing #2728's unit cases, where a shared `Fn::Sub` literal made the second case's masking unobservable.

- `src/deployment/intrinsic-function-resolver.ts`: the variables now resolve into a fresh **null-prototype** object; the caller's map is never written. Consequences, each stated in the comment and pinned: a `__proto__` variable (an own key after `JSON.parse`) still substitutes; a placeholder naming an `Object.prototype` member the map lacks (`${constructor}`) no longer renders that member's source text on either form and falls through to `Ref`; a `null` / primitive second element is an explicit `markNonRetryable` refusal (newly enforced: `null` and a non-empty string always threw, a number / boolean / empty string only once a placeholder reached the `in` test); an array second element still resolves by index, its non-enumerable `length` no longer a variable.
- `src/deployment/secret-redaction.ts` and `tests/unit/deployment/secret-redaction-cross-stack-source.test.ts`: the cross-stack reader's comments described the writer's `in` as a character-for-character mirror; they now say the reader refuses a conservative SUPERSET (an inherited binding falls through to `Ref` / `GetAtt` in the writer and is refused by the reader), and that a `hasOwn` reader would key the inherited case.
- `docs/changelog-cdkd.md`: the entry.

## Tests

`tests/unit/deployment/intrinsic-sub-variables-not-mutated.test.ts` (new, 7 cases), the real resolver over a faked Secrets Manager client: one `Fn::Sub` object resolved twice (byte-identical before and after; the fresh second map holds the entry via the cache-hit arm; the client answers once); a pseudo-parameter variable staying an intrinsic; `__proto__` substituting; `${constructor}` no longer rendering function source on either form; a resource named `constructor` reached through `Ref` on both forms; an inherited binding not substituted (the `Ref` arm answers); the refusal for `string` / number / `null` / `0` / `false` / `''` / `undefined` / the one-element form, with the whole message asserted exactly (the type named, the value never echoed) and the non-retryable marker, beside an array second element resolving by index and `${length}` no longer a variable.

Mutation probes against the final file (each run alone, sources restored byte-exact between): the in-place write restored (2 red); `{}` instead of `Object.create(null)` (3 red); `if (true)` at the bound-name test (3 red); a `for...in` copy (1 red); the refusal dropped, keyed on `${` in the template, truthiness-shaped, or presence-keyed (1 red each); `markNonRetryable` dropped (1 red); `String(variableMap)` or bare `typeof` in the diagnostic (1 red each).

## Live test

Run on this branch after its rebase onto `main` at d00f0067, which brought #2742's edit of the same `secrets-dynamic-ref` fixture (the later rebase onto 950a78b9 brought #2755's change to `evaluateConditions` and to `outputs-export-alias.ts`, neither of which touches `resolveSub`, and no fixture change; the `integ-destroy` / `integ-broad` markers digest this branch's delta against the merge base, so they stayed fresh across it, which is the trade CLAUDE.md records for those gates) (2026-09-07, `us-east-1`, ledger rows in `docs/_generated/integ-last-run.tsv`; an identical pair of runs before the rebase was also green):

| fixture | result | notes |
| --- | --- | --- |
| `secrets-dynamic-ref` | PASS, 559 s | every dynamic-reference / redaction / drift / rollback / refresh-observed / scrub arm green, including the `Fn::Sub` `Export.Name` (the plain-string form); clean destroy, 0 surviving versions |
| `lambda` (broad set) | PASS, 113 s | 9 deleted, 0 errors; events log swept after |

The changed arm was exercised live by the post-rebase `secrets-dynamic-ref` run: #2742's Phase 1b probe deploy declares `OutputFailureLeak`, a two-argument `Fn::Sub` whose `Pw` variable is a `{{resolve:secretsmanager:...:password}}` reference and whose body uses `${Pw}` as the JSON key of a second reference. The run's log shows the fixture's premise guard accepting that shape and the resolution failure reported masked (`key '***' not found in secret`), i.e. the two-argument arm still resolves its variable and substitutes it into the body after the change. That run exercises substitution, not the non-mutation itself: no fixture resolves one template object twice against a fresh map, so the non-mutation is pinned only by the unit file, through the real resolver. The plain-string `Fn::Sub` (the `Export.Name` arm) and every other intrinsic ran in the same fixture. `integ-destroy` and `integ-broad` markers set from these runs; the state bucket holds only `_index/` afterwards.

## Verification

- Codex delta review: 9 rounds (r1: `__proto__` through a plain `{}`; r3: the reader comment's stale mirror claim; r4: the historical-failure claim too broad and the array `length` change; r5: null / non-empty string always threw; r2, r6, r7, r8, r9: NO ISSUES). Maintainer-checklist proxy pass: 5 rounds, every item P2/P3 and fixed (the resource-backed `constructor` case, the explicit refusal and its marker, the falsy / undefined shapes, the inherited-binding case, the full diagnostic); stopped after round 5. PR-body review: 3 rounds (r1: stale mutation counts and a suite claim resting on a run under load; r2: a stale live-test paragraph written before #2742 landed; r3: NO ISSUES), then a maintainer-checklist pass over the body: 3 rounds (r1: four items, all fixed — the refusal's message asserted exactly, the live-test claim narrowed to substitution, the reader's `in` note reworded as a retained conservative refusal, the rules file's key sentence naming that over-refusal; r2: this `/check-docs` sentence; r3: NO ISSUES), with a Codex delta round (r9, NO ISSUES) over the three tree edits it produced.
- Reviewer agents on the working-tree diff: the security flow trace confirmed no reader relied on the mutation and no other intrinsic arm mutates its input, and raised the reader-comment mirror claim; the code / test review raised the both-forms prototype-member change and the redundant `clearRecordedSecretExpressions` call. All addressed.
- `/check`: 939 files / 20330 tests green on the final tree (rebased onto `main` at 950a78b9), run alone (runs taken while other sessions' suites held the machine at load 13-17 timed out four tree-scanning script tests at their 5 s bound, the #2553 class, none of them touched here; each passed on re-run alone); `/check-docs`: the one rules sentence touching this arm, `.claude/rules/layout-deployment-secrets.md`'s `crossStackSourceKey` bullet, now names the reader's over-refusal beside its by-construction claim (the rest of the satellites say nothing about `resolveSub`'s map); `/verify-pr`: all checks pass.


go-to-k:
Review: 3-axis (spec / code / test) plus the security add-on, dispatched because both touched `src/` files carry `fix:` in 3 of their last 3 commits and this is a secret-handling change.

**No blockers.** The fix does what #2739 prescribed, the flow trace confirms it records strictly more than before and never less, and all 11 mutation-probe claims in the PR body reproduce exactly. Four points below; only (1) is one I would like resolved before merge.

## 1. `${constructor}` now renders the literal string `undefined`, and the test comment names the wrong mechanism (medium)

`tests/unit/deployment/intrinsic-sub-variables-not-mutated.test.ts:128-131` says `resolveRef` "reads `context.resources[name]`, and an empty resources bag answers `undefined` for `constructor` through its own prototype". `({})['constructor']` is the `Object` **function**, which is truthy, so `resolveRef` (`src/deployment/intrinsic-function-resolver.ts:3065-3069`) takes the *resource* arm, not the not-found arm: `resolveRefValue(Object)` reaches `cfnRefValueFromPhysicalId(undefined, undefined, ...)`, which falls through every `resourceType` guard and returns `undefined` at `:585`; `resolveSub:4756` then `String()`s it.

Measured on this branch: `{"Fn::Sub": ["x-${constructor}", {"Lit": "y"}]}` renders `x-undefined` on both forms, where an ordinary unknown name renders `x-${A}` (placeholder retained). That path was unreachable from `Fn::Sub` before this PR — the plain-object `in` short-circuited it — so the change newly lets such a placeholder ship a literal `undefined` into a live property instead of the old function source.

The assertions at `:141-142` are negatives (`not.toContain('native code')`, `not.toContain('function')`) and both pass for `x-undefined`, so the actual new output is never stated. Please either correct the comment to name `cfnRefValueFromPhysicalId`'s pass-through and pin the rendered string with a positive `toBe('x-undefined')`, or close the prototype-chain read in `resolveRef` one layer down. Reached independently by the test and security lenses.

## 2. Element 0 is left unvalidated while element 1 gained a guard (minor)

`:4697` assigns `template = templateString` unchecked, so `[{"Ref": "X"}, {}]` still dies at `:4733` with `template.matchAll is not a function` — opaque and *unmarked*, right beside a neighbouring shape that now gets a named `markNonRetryable` refusal. `src/deployment/secret-redaction.ts:975` already asserts "a non-string template ... THROW[s] during resolution", so the reader depends on it throwing; folding `typeof templateString !== 'string'` into the same guard would make both shapes fail the same way.

## 3. The sibling object walk in the same file still drops `__proto__` (minor, pre-existing)

The new comment justifies `Object.create(null)` as "the same reason `redactByPath`'s object walk builds its output that way", but `resolveValue`'s generic property walk at `src/deployment/intrinsic-function-resolver.ts:2802` is still `const resolved: Record<string, unknown> = {}`. `Object.entries` yields `['__proto__', v]`, the assignment routes through the inherited setter, and the key is silently absent from the object deploy hands the provider — an arbitrary-key property bag (a custom resource's properties, ECS `DockerLabels`) loses the entry with no warning. `secret-redaction.ts` uses `Object.create(null)` at four comparable walks (`:3129` / `:4114` / `:4417` / `:4770`); this one is the odd man out. Not caused by this PR, so happy to file it separately if you would rather keep the diff scoped.

## 4. A third `in varMap` reader is now silently diverged (nit)

`src/analyzer/orphan-rewriter.ts:644` (`if (varMap && inner in varMap)`) tests the caller's plain object, so it treats an `Object.prototype` name as bound where `resolveSub` now falls through to `Ref`. Same conservative direction as the `secret-redaction.ts` reader this PR documents in two places; this third copy is in neither.

## Verified clean

- **Nothing depended on the mutation.** `variables` never escapes `resolveSub` (only `varNameStr in variables` and `String(variables[k])`), so the null prototype has no downstream `Object.keys` / spread / `JSON.stringify` exposure. Every reader of the resolved value is fed the same or better: `resolveDynamicReferences`, the cache-hit re-recording arm (`:7072-7079`), the cross-region sibling (`withoutProducerRegions` is a shallow spread, so `recordedSecretValues` is the same Map), `intrinsicSkeletonPattern` (reads `args[0]` only), `crossStackSourceKey` (key-tests `raw[1]`, never its values), the rollback journal, `deployments/` events (which carry no properties at all), and the logger (masked at `:4835`). `deploy-engine.ts:6901` / `:6968` retain `output.Value` by reference — the motivating leak the fix removes.
- **Re-resolution is bounded.** `cachedDynamicReferences` is instance-scoped and its hit arm re-records into the map it is handed, so a second pass costs zero extra AWS calls and still pins the needle. A cross-account resolver is a separate instance and takes the fresh-lookup recording path at `:7211`.
- **The new refusal is load-bearing, not merely defensive.** Without it `['x-${A}', 5]`, `[..., false]` and `[..., '']` would begin resolving *silently* — `Object.entries(5)` is `[]` and the `in` test now runs against the fresh object rather than the primitive — which is exactly the premise `crossStackSourceKey` documents at `secret-redaction.ts:972-977`. The only genuinely new refusal is a placeholder-free template with a primitive second element, a shape CloudFormation itself rejects. `markNonRetryable` is right: a plain `TypeError` was never retryable-classified either, so retry behaviour is unchanged.
- **The diagnostic cannot echo a value** on any input (only `typeof` and the literal `'null'`), and it is thrown before the placeholder loop, so `subPlaceholderWarning` and `rethrowStructuralSubFailure` are unreachable from it.
- **The "conservative SUPERSET" claim holds against the code**, not just the comment: the writer binds own-enumerable string keys via `Object.entries`, the reader tests `varName in (variables ?? {})` on the caller's plain object (own + inherited, enumerable + non-enumerable), so W is a subset of R and the reader never certifies a key the writer did not record. Rewriting that `in` to `Object.hasOwn` turns `secret-redaction-cross-stack-source.test.ts:1019` red, so the claim is pinned. Same wording in `.claude/rules/layout-deployment-secrets.md:167-170` is accurate.
- **Mutation probes.** All 11 rows in the PR body reproduce exactly (run in a git-free copy, working tree untouched). Two extras: killing the cache-hit recording arm at `:7072` turns case 1 red, so the fresh-map recording half is independently live rather than shadowed by the byte-identical assertion; and no existing test in the suite was passing on the old mutation (grep found no `${length}` / `${constructor}` / `${toString}` or primitive second element anywhere else, and the only production 2-arg construction at `orphan-rewriter.ts:577` always passes a defined `varMap`).
- Ledger rows at `docs/_generated/integ-last-run.tsv` match the PR body's claimed fixtures and durations; the changelog entry is 1990 chars, inside the 2000 cap.

go-to-k:
Filed items 3 and 4 as #2767, together with the `resolveRef` half of item 1 — they are the same read in three places and are all outside this PR's diff, so they should not widen it. Item 1's test-side half (the comment naming the wrong mechanism, and the negative assertions that never state `x-undefined`) still belongs here, since it is this PR that makes that path reachable from `Fn::Sub`.

Item 2 (the unvalidated element 0) is your call: it is one clause in the guard you already added, so folding it in keeps the two shapes failing the same way, but I am equally happy to see it go to #2767.

go-to-k:
Correction to item 1, before you act on it — my suggestion there and #2767 are in tension, and that is my error, not yours.

I asked you to pin the rendered string with a positive `toBe('x-undefined')`. #2767 fixes the `resolveRef` prototype-chain read that PRODUCES that `undefined`, so such an assertion would be invalidated within days. After #2767, `resolveRef` raises the ordinary not-found error, `rethrowStructuralSubFailure` declines it (the template does not declare a resource by that name), and `intrinsic-function-resolver.ts:4805-4806` warns and retains the placeholder — so `{"Fn::Sub": ["x-\${constructor}", {"Lit": "y"}]}` will render `x-\${constructor}`, the same as any other unknown name.

So **item 1 reduces to the comment correction only**: at `tests/unit/deployment/intrinsic-sub-variables-not-mutated.test.ts:128-131`, name `cfnRefValueFromPhysicalId`'s pass-through rather than "an empty resources bag answers `undefined` ... through its own prototype", and note that the literal `undefined` itself is tracked in #2767. Please do not add the positive assertion — #2767 will carry it, against the corrected behaviour.

On ordering: **#2764 goes first.** I will hold #2767 until this is merged and rebase it myself, so you carry no rebase for it — both PRs add a `docs/changelog-cdkd.md` entry and `docs/_generated/integ-last-run.tsv` rows, which would conflict for whoever lands second. Nothing else in the two diffs overlaps: #2767 touches `:2802` and `:3065`, roughly 1600 lines from `resolveSub`.

Item 2 (the unvalidated element 0) is unaffected by any of this and is still your call to fold in or leave to #2767.

go-to-k:
**Consolidated — this comment supersedes my three above. Read only this one.**

I split one review into three comments and the third retracted part of the second, which left the actionable set scattered. Sorry for the noise. Here is the current state.

## What this PR needs: one change

`tests/unit/deployment/intrinsic-sub-variables-not-mutated.test.ts:128-131` — the comment describes a mechanism that is not what happens. It says `resolveRef` "reads `context.resources[name]`, and an empty resources bag answers `undefined` for `constructor` through its own prototype". `({})['constructor']` is the `Object` **function**, which is truthy, so `resolveRef` (`src/deployment/intrinsic-function-resolver.ts:3065`) takes the *resource* arm, not the not-found arm; `resolveRefValue(Object)` reaches `cfnRefValueFromPhysicalId` with an undefined physical id, falls through every guard, and returns `undefined` at `:585`. Please reword it to name that pass-through, and add a pointer to #2767.

Leave the assertions at `:141-142` as they are. I earlier asked you to add a positive `toBe('x-undefined')` — **ignore that**, it was wrong of me: #2767 removes the `undefined`, so the pin belongs there, against the corrected behaviour.

That is the whole ask. Nothing else in the diff needs to change.

## What I took off your plate

Into #2767, so this PR stays scoped:
- the `resolveRef` prototype-chain read that produces the `undefined` (plus the `Fn::GetAtt` and parameter-arm twins of it, which I found while writing it up);
- `resolveValue`'s object walk at `:2802` still dropping a `__proto__` key;
- the `orphan-rewriter.ts:644` reader's undocumented divergence;
- **the unvalidated `subArgs[0]`** — I previously left this to you as "your call"; I have taken it, so you have exactly one thing to do.

## Ordering

**This PR merges first.** #2767 waits and I rebase it myself. Both PRs append to `docs/changelog-cdkd.md` and `docs/_generated/integ-last-run.tsv`, so whoever lands second carries that conflict, and it should not be you. #2767's own tests currently depend on this PR's change, which is a second reason it goes after.

Everything else in my first comment — the flow trace, the mutation-probe reproduction, the SUPERSET verification, the ledger and changelog checks — stands as written; there are no findings in it beyond the one above.

nix-tkobayashi:
Done — the one change is on the branch, plus a rebase current `main` forced.

## The comment (your `:128-131`)

It now names the pass-through instead of a not-found answer: `context.resources['constructor']` reads `Object.prototype.constructor`, the `Object` function is truthy so `resolveRef` takes the RESOURCE arm, that function carries no `physicalId`, and `cfnRefValueFromPhysicalId` falls through every recovery to return the `undefined` it was handed — so the placeholder renders `x-undefined`. It says that value is #2767's to pin, against the corrected behaviour, and links the issue.

I measured `x-undefined` rather than transcribing it from your comment: a temporary `toBe(...)` in that case, run, read off the failure, then the file restored byte-exact (`git diff --stat HEAD` empty). The measurement covers the two-argument form, which is where the loop failed; the comment attributes the rendering to `resolveRef`, which both forms reach identically once the placeholder is unbound.

The assertions at your `:141-142` are untouched, per your retraction — I agree with the reasoning. A positive pin here would have to be re-inverted by #2767, and the negative is the part this PR's change actually earns.

## Items 2, 3, 4 and the `resolveRef` read

Understood, and thank you for taking them into #2767 — including `subArgs[0]`, which I had carried as an unresolved "your call" residual rather than a decision. Nothing from them is in this diff.

## Rebase

The branch is now on `540debe3`. `docs/changelog-cdkd.md` conflicted and is resolved keep-both, my entry above the ones that landed; `docs/_generated/integ-last-run.tsv` merged cleanly and still carries the two rows the body cites. Nothing that landed since my old base touches `src/deployment/` at all (the only `src/` change is `src/provisioning/property-coverage.generated.ts`), and the one fixture added is `sdk-to-cc-autoroute`, not one this branch ran, so I did not re-run the integs. Both markers verify fresh, by different mechanisms: `integ-destroy` digests this branch's delta against the merge base, so a peer's merge cannot stale it, and `integ-broad` hashes the `.markgate-broad-integ-test` sentinel, which the rebase did not rewrite.

## Verification on the rebased tree

Full unit suite run alone: 940 files / 20382 tests green. One earlier run failed only `gen-handled-property-wiring.test.ts`'s `--check` spawn case at its declared 60 s bound while other sessions held this host at load ~13 (the #2553 class, nothing this branch touches); it passes alone and the run above is the one the marker is set from.

A Codex delta round and a maintainer-checklist proxy pass over the reworded comment both returned NO ISSUES.

Ordering understood: this one first, #2767 after, and you rebase it.

go-to-k:
Merged — thank you. The comment now names the real mechanism, and the delta review measured it independently rather than taking either of our words for it: both forms render `x-undefined`, the control `x-\${A}` keeps its placeholder, and the resource arm is the one that answers.

The rebase was checked the same way, since a keep-both resolution is where things quietly go missing: every blob this PR owns is byte-identical pre- and post-rebase, `docs/changelog-cdkd.md` diffs against `main` as one insertion and zero deletions with exactly one #2739 entry, and the ledger holds 299 rows with no duplicate test name. Your judgement that no integ re-run was warranted also holds — the only `src/` change in the rebase window is comment-only, and nothing on `main` gained a caller of what you changed.

One thing was mine to fix rather than yours: your comment points at #2767 for "that read and its `Fn::GetAtt` / parameter twins", and that was true of the WORK but not of the issue TEXT, which named only `resolveRef`. I have added the `Fn::GetAtt` read to that issue's body, so the cross-reference now resolves to something. The follow-up is [PR 2777](https://github.com/go-to-k/cdkd/pull/2777), which sweeps it along with several more sites the reviews kept surfacing one round at a time — the enumeration there ended up finding four that three review rounds had not, including a `SecretString:constructor` dynamic reference rendering a function's source as the resolved secret.

Gates were re-run against your head in a local checkout rather than inherited from another tree, since they are per-worktree: 940 files / 20382 tests green, and `lambda` PASS with a clean destroy and an empty account-wide state scan.

Nit for next time, not worth a push: the PR body's rebase chronology stops at `950a78b9` while the head sits on `540debe3`.

PR #2773, opened 2026-09-07
fix(diff): an Output the deploy skipped is previewed as absent, not as a phantom ADD, while its template inputs are unchanged

nix-tkobayashi:
Closes #2740

## What

An Output whose resolution fails INSIDE a secret lookup on every deploy (a `{{resolve:secretsmanager:...}}` naming a JSON key the secret does not hold) is skipped by the deploy under the default arm and never reaches `state.outputs`. `cdkd diff` resolves outputs with `skipDynamicReferences`, so the same reference assembles into its token there instead of failing, and `computeOutputsDiff` pushed a phantom `ADD` on every run of the unchanged stack — `cdkd diff --fail` stayed red.

The deploy now records what it skipped, and the diff trusts that record only while the template inputs it depends on are unchanged:

- **`StackState.skippedOutputs?: Record<string, string>`** (`src/types/state.ts`) — output key → sha256 over the output's own entry plus every top-level template section except `Resources` and the sibling `Outputs`. Informational, **no schema version bump**: absent on older records, the diff behaves as before, the next no-change deploy writes it. `skippedOutputsCarriedFrom` mirrors `exportNamesCarriedFrom`.
- **`src/analyzer/skipped-outputs.ts`** (new) — `skippedOutputDigest` (the ONE spelling of what the digest covers; canonical JSON with a null-prototype sort object), `collectSkippedOutputs` (every `undefined`-valued key of the resolved bag), `bindingSkippedOutputs` (the record's keys whose digest still matches a pristine template), `skippedOutputsEqual`.
- **`src/deployment/deploy-engine.ts`** — `doDeploy` snapshots the template (`Resources` dropped) at its TOP, before `resolveParameters` / `evaluateConditions`, and threads it into `resolveOutputs` / `executeDeployment`; the success-path save and the no-change-path save write this pass's record (omitted when empty); the no-change path gains `skippedOutputsChanged` as its own save trigger (a skipped output switches `outputsChanged` / `exportSetChanged` off, so nothing else would ever write it); the five partial saves (per-resource, pre-rollback, post-rollback and its fresh-ETag retry, `persistStateAfterOutputFailure`) carry the previous record.
[...]

Issue #1837, opened 2026-08-13
docs(changelog): the #1745 entry is duplicated on main, and the changelog conflicts on nearly every parallel-lane rebase


go-to-k:
Found incidentally while resolving changelog conflicts during the #1770 /
#1778 / #1786 lanes. Not caused by any of them — it is already on `origin/main`.

## What

`docs/changelog-cdkd.md` carries the `AWS::S3::Bucket` host-attributes entry
(issue #1745 / PR #1813) **twice**, as two byte-identical adjacent lines:

```console
$ git show origin/main:docs/changelog-cdkd.md | grep -c 'The four `AWS::S3::Bucket` host attributes derive their URL suffix'
2
```

Both land in the same bullet list, one immediately after the other.

## Why it happens, and why it will happen again

Every lane appends its entry to the TOP of the same list, so `docs/changelog-cdkd.md`
conflicts on essentially every rebase when more than one lane is open. The
natural resolution — "keep both sides" — is correct for two DIFFERENT lanes'
entries and wrong for two copies of the SAME lane's entry, which is what a
multi-commit branch produces after a rebase replays each commit's version of the
line. The #1778 lane hit exactly this shape during the same window and found
**three** stale copies of its own entry on its branch before deduplicating.

With the current rate of parallel lanes this is a recurring papercut rather than
a one-off.

## Suggested direction

Two separable pieces:

1. **The immediate fix** — delete the duplicate line.
2. **The durable one** — make the file stop being a conflict magnet, or make
   duplicates detectable. Options worth weighing: a CI check that fails on two
   identical bullet lines (cheap, catches every future instance, no workflow
   change); a per-entry unique marker (the issue number is already there, so a
   check for a repeated issue reference is nearly free); or a
   `docs/_generated`-style one-file-per-entry layout that is concatenated, which
   removes the conflict entirely at the cost of a build step.

Option (2)'s cheap form is a natural companion to the existing generated-doc
checks that already run in `check-build-test`.

Session-fit: next — no file overlap with the lanes that found it, and the
durable half wants its own decision about the changelog's layout. / Effort: S
for the duplicate line alone, M with a CI check.

## Sweep results (2026-09-02, terrapin lane on go-to-k/cdkd#1937 + go-to-k/cdkd#1938)

The sweep this issue asks for was run over the whole file. It found SEVEN duplicate clusters, fifteen lines, nine of them redundant -- this issue knew about one and go-to-k/cdkd#1937 about a second.

[...]

PR #2857, opened 2026-09-09
test(changelog): conserve the archive as a subsequence, so a fragment dated inside its span stops failing

### LINKED ISSUE #2813: changelog.d: a fragment dated inside the archive's span fails the verbatim-archive assertion, so no lane can date an entry 2026-05-31..2026-09-08
## Summary

A `changelog.d/` fragment whose date falls INSIDE the archive's date span fails `assemble-changelog.test.ts`'s verbatim-archive assertion. The archive spans 2026-05-31 to 2026-09-08, and its newest section is the migration day itself — so a lane writing an entry for the day go-to-k/cdkd#2811 merged cannot pass CI with a correctly dated fragment.

The assembler and that assertion disagree by construction, and both are doing what they were written to do.

## What happens

`assembleChangelog` splices a fragment whose date the archive already heads UNDERNEATH that heading, which `changelog.d/_header.md`'s "WHERE an entry goes" states as the intended behaviour and which exists to make the go-to-k/cdkd#1837 duplicate-heading defect structurally impossible. A fragment dated between two archive sections gets a new section inserted in descending position, i.e. into the archive's middle.

Both outcomes put text inside the archive's span, so the archive stops being a contiguous substring of the assembled document, and `tests/unit/scripts/assemble-changelog.test.ts` asserts exactly that it is one:

```
const archive = readFileSync(join(root, FRAGMENT_DIR, ARCHIVE_FILE), 'utf-8').replace(/\s+$/, '');
expect(doc.includes(archive), 'the archive is no longer reproduced verbatim ...').toBe(true);
```

A fragment dated strictly outside the span is fine: newer than the newest section opens a section above the whole archive, older than the oldest appends below it. Neither splits it.

## Measured

One probe fragment at a time, in an otherwise clean tree, running only that suite:

| fragment date | position relative to the archive | result |
| --- | --- | --- |
| 2026-09-09 | newer than the newest section | 16 passed |
| 2026-09-08 | the newest section itself | 1 failed / 15 passed |
| 2026-09-07 | inside the span | 1 failed / 15 passed |
| 2026-08-01 | inside the span | 1 failed / 15 passed |
| 2026-05-01 | older than the oldest section (2026-05-31) | 16 passed |
| 2026-01-01 | older than the oldest section | 16 passed |

The failing assertion is the verbatim one in every failing case.

## Why it matters

It is not calendar-dependent and does not heal on its own: a fragment dated 2026-09-08 fails on every later day too. What it costs is a lane that cannot record its entry anywhere the convention allows. The passing places are the archive, which the migration says no lane edits again, or a date that is not the entry's date. This lane hit it on go-to-k/cdkd#2773 and left the entry in `changelog.d/_archive.md`, which is the honest content in the wrong file rather than the reverse.

The window is wider than the migration day: any entry dated from 2026-05-31 through 2026-09-08 inclusive is affected, which includes a lane rebasing an older branch forward.

## Possible fixes

Author's judgment, but the shape of the disagreement suggests the ASSERTION is what is wrong rather than the splice, since the splice is the documented behaviour and the reason a duplicate heading is impossible.

1. Assert that the archive's ENTRIES survive rather than its byte span: every archived bullet still present, in order, under its own date heading. That is what "settled history is not reformatted" actually means, and it holds under a splice.
2. Assert verbatim-ness per archive SECTION rather than over the whole file, comparing a spliced section against its own text plus the fragments the assembler put into it.

A third shape, refusing a fragment dated at or inside the archive's span, would make the restriction explicit rather than emergent — but it is a behaviour change, not a fence fix: it refuses an entry for the migration day outright, and the suite's existing same-date and older-date acceptance cases would have to become rejection cases.

## Verification

`tests/unit/scripts/assemble-changelog.test.ts` should go green for a fragment at every date in the table above, and the check must still FAIL when the assembler genuinely reformats archived text rather than merely splicing into it — probe it by dropping an archived bullet and by re-wrapping an archived line, each on its own.

## Classification

Dup-check: searched open issues for `changelog`, `changelog.d`, `assemble`, `fragment` and `archive`; the migration is go-to-k/cdkd#2779 / go-to-k/cdkd#2811 and neither it nor any open issue names this interaction.
Session-fit: next (not this session) — it is the maintainer's own just-merged design decision which of the shapes above the fence should take, and the lane that found it has a review-round PR open on an unrelated file set.
Severity: medium — a correctly dated fragment cannot pass CI for any date inside the archive's span, with two workarounds that each break a different stated rule.
Effort: small (S) — one assertion and its fixtures, riding the unit suite.
Estimate: ~1 h — most of it deciding which shape the fence should have.

### LINKED ISSUE #2816: fix(scripts): assemble-changelog's same-date merge and its verbatim-archive fence cannot both hold
## What is wrong

`scripts/assemble-changelog.ts` (go-to-k/cdkd#2811) deliberately MERGES a fragment into an archive section that already heads its date:

> A date the archive already heads takes that section, newest entries first.

`tests/unit/scripts/assemble-changelog.test.ts` asserts the opposite property on the real tree:

> the ARCHIVE comes out verbatim -- the whole reason it is one file and not 574 is that settled history must not be reformatted

Those cannot both hold for a fragment whose date the archive already heads: merging inserts the fragment's text ABOVE that section's archived entries, so `doc.includes(archive)` is false. The archive's newest section is `2026-09-08`, so the first fragment dated `2026-09-08` fails.

## How it was found

Measured, as the first fragment written after the migration. `changelog.d/entries/` still held only `.gitkeep`, and the test's own comment anticipates exactly this:

> It is 0 today and that is the honest state: the archive holds every pre-migration entry and no fragment has been written yet, so this assertion is a TRIPWIRE for the first one rather than coverage now.

The tripwire fired on the first fragment. Renaming it `2026-09-09-...` makes the assembler insert a NEW section above the whole archive, the archive stays verbatim, and the suite is green -- which is what PR go-to-k/cdkd#2777 did to unblock itself. That is a WORKAROUND, not a fix: the next contributor who dates a fragment on the archive's newest day hits the same wall, with no note saying why.

## Why it matters

The failure is legible (the assertion names it), so nothing ships broken. What it costs is a contributor's afternoon and, worse, a nudge toward the wrong remedy -- the obvious reading of the message is "the assembler has a bug", when the assembler is doing what its own comment says it should.

## Proposed fix

One of three, and the choice is the maintainer's since it is a statement about what the archive IS:

1. **Narrow the fence to what the migration actually rests on.** Assert every archive SECTION appears verbatim, rather than the file as one string -- merging preserves each section's own text, so this keeps the real guarantee (settled history is not reformatted) and tolerates the supported merge.
2. **Refuse the collision at the entry-name check.** If a fragment may never share a date with an archive section, say so in `ENTRY_NAME`'s validation with that reason, and delete the merge branch as dead.
3. **Keep both and document the constraint** in `changelog.d/_header.md`, so the next author picks a free date deliberately rather than by bisecting a test failure.

(1) looks right from outside -- the merge branch was added by review for a measured defect (go-to-k/cdkd#1837's class arriving through the generator), so deleting it would re-open that -- but this is a judgement about the archive's contract, not a mechanical call.

## Classification

Dup-check: searched open and closed issues for `assemble-changelog verbatim`, `changelog fragment same date`, `archive reproduced verbatim` -- no hit. go-to-k/cdkd#2811 is the migration this is about and does not name the interaction; go-to-k/cdkd#1837 is the duplicate-entry class the merge branch exists to prevent.
Session-fit: next (not this session) -- the session that found it is landing go-to-k/cdkd#2777 and has an unblocking workaround in hand; the fix is a decision about the archive's contract, and options 1 and 2 point opposite ways. Next session verifies with `vp test --run tests/unit/scripts/assemble-changelog.test.ts` after re-dating a fragment onto `2026-09-08`, which is the reproduction.
Severity: low -- no user-visible defect and nothing ships broken; the cost is a contributor blocked by a message that points away from the cause.
Effort: small (S) -- one assertion or one validation branch, plus the case that reproduces the collision.
Estimate: ~1 h -- deciding which of the three is right, not writing it.

nix-tkobayashi:
Closes #2813
Closes #2816

## What

`assemble-changelog.test.ts` asserted that `changelog.d/_archive.md` is a
CONTIGUOUS SUBSTRING of the assembled document. The assembler SPLICES, so that
assertion refused a correct assembly.

A fragment whose date the archive already heads lands under that heading; one
dated between two archived sections opens a new section in descending position.
The first is what `changelog.d/_header.md` states under "WHERE an entry goes";
the second is documented in `assemble-changelog.ts` itself. Both are what makes
the go-to-k/cdkd#1837 duplicate-heading class structurally impossible. Both also put new text inside the archive's span, so the archive
stops being contiguous while nothing about settled history has changed.

The assertion is now that every archived line still appears, unmodified and in
its original order -- a SUBSEQUENCE. That is what "settled history is not
reformatted" actually claims, and a splice preserves it by construction.

## Why it mattered

No lane could date a fragment anywhere in **2026-05-31 .. 2026-09-08** -- the
whole archived span, which a lane rebasing an older branch forward reaches too.
The two available workarounds each break a stated rule: edit `_archive.md`,
which the migration says no lane edits again, or date the fragment something
other than its date. The lane that hit this on go-to-k/cdkd#2773 took the first.

## Measured

One probe fragment at a time on the REAL tree (51 archived sections, 2026-05-31
to 2026-09-08), running the changelog fences only.

| fragment date | position | before | after |
| --- | --- | --- | --- |
| 2026-09-09 | newer than the newest section | pass | pass |
| 2026-09-08 | the newest section itself | **FAIL** | pass |
| 2026-09-07 | inside the span | **FAIL** | pass |
| 2026-08-01 | inside the span | **FAIL** | pass |
| 2026-05-31 | the oldest section itself | **FAIL** | pass |
| 2026-05-01 | older than the oldest | pass | pass |
| 2026-01-01 | older than the oldest | pass | pass |

The issue's table named `assemble-changelog.test.ts` alone. The four other
fences that read the assembled document -- `changelog-entry-policy-sync`,
`changelog-entry-size`, `changelog-entry-uniqueness`, `changelog-ledger-citation`
-- were run with an inside-span fragment as well, and are green both before and
after (5 files, 50 tests).

## The check still FAILS on a real reformat

Proven against REAL code, not a synthetic fixture: each shape was introduced
into `scripts/assemble-changelog.ts` itself, measured against the real archive,
and the tree restored byte-exact between probes (`git diff --stat HEAD` receipt
taken each time).

Every row replaces the single expression `out.push(s.heading, ...s.body);` in
`assembleChangelog`, except the last, which replaces the two-line
`if (at === -1) ... else merged.splice(at, 0, fresh);`. The counts are of this
one test file's 30 cases and depend on exactly where the edit goes, so the
recipe is given rather than a description of it — run from the labels alone the
same idea lands elsewhere and reds a different number.

| edit | result |
| --- | --- |
| `out.push(s.heading, ...s.body.slice(1));` | **12 failed** |
| `out.push(s.heading, ...s.body.map((b) => b.replace(/  +/g, ' ')));` | **1 failed** |
| `{ const b = [...s.body]; if (b.length > 2) { const t = b[1]; b[1] = b[2]; b[2] = t; } out.push(s.heading, ...b); }` | **1 failed** |
| `out.push(s.heading, ...s.body.flatMap((b) => (b.startsWith('- ') ? [b, ''] : [b])));` | **5 failed** |
| `out.push(s.heading, '- FOREIGN LINE', ...s.body);` | **10 failed** |
| `if (at === -1 \|\| at === 0) merged.push(fresh); else merged[at - 1] = { ...merged[at - 1]!, body: [...es.map((e) => e.text), ...merged[at - 1]!.body] };` | **6 failed** |
| restored | 30 passed |

Two earlier probe passes were wrong and are recorded here rather than quietly
dropped. The first mutated `_archive.md` instead of the assembler, so the
expected value and the output moved together and all three "negatives" passed --
a circular probe that proved nothing. The second mutated the merge branch
(`{ ...s, body: [...] }`), which does not fire unless a fragment is dated at an
archived date, and was INERT: a probe that changes nothing is a probe that was
not run.

## Tests

16 -> 26 in the one file. Five date positions against a fixture archive
deliberately widened to two sections a week apart, because with the consecutive
days the other cases use, "strictly inside the span" is not a reachable date and
the interesting arm could not be written. Each asserts conservation AND that the
fragment landed exactly once -- conservation alone is satisfied by an assembler
that dropped the fragment -- plus heading uniqueness, descending order, and the
render check.

Five negative controls drive the check itself, including a **deleted blank
line**: an insertion-tolerant check is exactly the kind that could have let that
through, so it is pinned rather than assumed.

Full suite 953 files / 20,711 tests green; typecheck, `typecheck:test`, lint,
format and build clean. No `src/` change, so no changelog fragment -- nothing
about the shipped binary moves.

## Review round 1

**M0 — where the fragment lands is now pinned.** Conservation,
appears-exactly-once, heading uniqueness and descending order all pass if the
assembler merges the `2026-09-05` probe into the `2026-09-08` body instead of
opening its own section — the defect `assemble-changelog.ts` records review
catching once (`09-07 / 09-08 / 09-07`). Each of the five arms now asserts the
line directly under its own generated heading. The last row of the mutation
table above is a BROADER form of that defect — it also disturbs ordering, so it
fails six tests rather than isolating the landing assertions. I did not
reproduce the ordering-preserving form the review reported failing exactly two.

**M1 — the insertion budget.** A subsequence bounds presence and order and
nothing else, so 575 inserted blank lines and foreign text spliced mid-archive
both passed, and both had been rejected by `doc.includes(archive)`. Accepting
insertions is inherent to the form — the splice IS one — so
`assertArchiveConserved` now takes a `budget`, and every term of it is derived
from what the assembler can emit: `header + 1 + fragment lines + new headings +
seams + 1`. The measured delta is **183** against a bound of **185** — slack 2.

The seam term is the one that matters. The assembler inserts a blank before a
section only where the PREVIOUS body ends non-blank, which is 1 section of 51
here, so an earlier revision that allowed one per heading put 51 of its 52-line
slack in that term alone — and grew it by one for every dated section the
archive would ever gain. A merged fragment prepends and cannot change whether a
body ends blank; a new section can, at both edges, hence two per new heading.
The preamble contributes one only when it exists and ends non-blank, which it
does not here. (Earlier revisions of this description said 235 and then 234;
both were bounds this test no longer computes.) Residual, stated in the
comment rather than left implied: an insertion smaller than the seam allowance
is inside the budget, so the delta bounds a bulk reformat and not a few foreign
lines. Attributing every unconsumed line is the stronger form; this is the one
the review asked for and it closes the class that was measured open.

**Two arms the review named as untested**, both verified by hand there and
pinned here: the one legitimate variant heading
`**Recently Implemented** (2026-07-02, second batch):` — which this change is
what first makes a fragment of that date legal at all — and two fragments in one
run, one merging into an existing section and one opening a new one.

**Three corrections from the same review.** The two splice shapes are documented
in different files and the comment claimed both were in `_header.md`. "A deleted
blank line would still be caught" holds for THIS archive, whose body bullets are
unique, and not in general — a deleted line re-syncs against a later duplicate —
so the sentence is scoped to the corpus. And the failure message printed a
0-based index behind an ordinal phrasing, reporting "(0 of 51)" for the first
line.

Tests 26 -> 30. Full suite 953 files / 20,715 tests green.

## Review round 2

**M1b — the seam allowance is derived, not granted.** Covered above: 185 against
183, slack 2 rather than 51, and it no longer widens as the archive grows.

`HEADING` is now exported from `assemble-changelog.ts` and used by the test,
which had carried a copy missing the `[^)]*\):$` tail. Both return 51 on today's
archive, so a line that loose-matched and strict-failed would have inflated the
date set, under-counted new headings and left the budget one short of a
legitimate assembly, with nothing comparing the two.

Five notes, all taken: the variant arm's presence assertion could not fail
(conservation on the line above already requires it); `WIDE_BUDGET` is
`TWO_SECTION_BUDGET`, since it is derived for a two-section fixture and the
variant arm applies it to a different one; the call site no longer restates the
JSDoc; the doc comment's counter-example is executable again (both parameters
are strings and it was written with bare arrays); and the real-tree comment
still claimed no fragment had been written yet, which stopped being true before
this branch existed and is read by the budget's own `fragmentLines` term.

Every row of the mutation table above was re-measured against the tightened
budget. Full suite 953 files / 20,715 tests green.

## Live test

The behaviour under test is a fence, so the live test is the fence running
against the real corpus rather than a fixture: every row of both tables above
was executed on this tree, one mutation at a time with the assembler restored
byte-exact between probes, and the seven-date sweep was re-run after the final
commit.

### DIFF
--- a/scripts/assemble-changelog.ts
+++ b/scripts/assemble-changelog.ts
@@ -81,7 +81,13 @@ export function headingFor(date: string): string {
   return `**Recently Implemented** (${date}):`;
 }
 
-const HEADING = /^\*\*Recently Implemented\*\* \((\d{4}-\d{2}-\d{2})[^)]*\):$/;
+/**
+ * EXPORTED so a consumer cannot carry a looser copy: a pattern that matches a
+ * line this one rejects sees a section the assembler does not, and any count
+ * derived from it disagrees with the document. Both forms return 51 on today's
+ * archive, so the drift would be silent (go-to-k/cdkd#2813 review).
+ */
+export const HEADING = /^\*\*Recently Implemented\*\* \((\d{4}-\d{2}-\d{2})[^)]*\):$/;
 
 /**
  * "This line was MEANT to be a heading", looser than {@link HEADING} on the
--- a/tests/unit/scripts/assemble-changelog.test.ts
+++ b/tests/unit/scripts/assemble-changelog.test.ts
@@ -112,7 +112,8 @@ function assertEveryHeadingOpensItsBlock(d: string): void {
  * whole point (issue go-to-k/cdkd#2813). The assembler SPLICES: a fragment
  * whose date the archive already heads lands under that heading, and one dated
  * between two archived sections opens a new section in descending position --
- * both documented in `changelog.d/_header.md` and both the reason the
+ * the first documented in `changelog.d/_header.md` under "WHERE an entry goes",
+ * the second in `assemble-changelog.ts` itself, and both the reason the
  * go-to-k/cdkd#1837 duplicate-heading class is structurally impossible. Either
  * one puts new text INSIDE the archive's span, so the archive stops being a
  * contiguous substring while nothing about settled history has changed.
@@ -139,9 +140,26 @@ function assertEveryHeadingOpensItsBlock(d: string): void {
  *
  * Blank lines are compared like any other, so the seam blank the assembler may
  * INSERT before a heading is absorbed (an insertion is what a subsequence
- * tolerates) while one it DELETED would still be caught.
+ * tolerates). A DELETED one is caught for THIS archive, whose body bullets are
+ * unique, but not in general: a deleted line re-syncs against a later duplicate
+ * of itself, and `assertArchiveConserved(['a','b','','b'], ['a','','b'])`
+ * passes. The corpus is what makes the stronger reading true here, not the
+ * check.
+ *
+ * ## The insertion budget is part of the assertion, not a separate nicety
+ *
+ * Presence and order alone place NO bound on how much text lands BETWEEN
+ * archived lines, and accepting insertions is inherent to any subsequence form
+ * -- the splice IS an insertion. What the splice needs is a BOUNDED,
+ * ATTRIBUTABLE insertion. Measured on the real tree while reviewing this
+ * change: a blank line after every archived bullet (575 inserted lines) and
+ * arbitrary foreign text spliced into the middle of the archive were both
+ * ACCEPTED by presence-and-order, and both had been rejected by
+ * `doc.includes(archive)`. That is a real gap in the insertion class, and the
+ * `budget` closes it: the caller states how many lines it expects the assembler
+ * to add, so a splice passes and a reformat that pads settled history does not.
  */
-function assertArchiveConserved(doc: string, archive: string): void {
+function assertArchiveConserved(doc: string, archive: string, budget?: number): void {
   const want = archive.split('\n');
   const have = doc.split('\n');
   let i = 0;
@@ -149,8 +167,15 @@ function assertArchiveConserved(doc: string, archive: string): void {
   expect(
     i,
     `the archive is no longer reproduced in full -- the assembler is reformatting settled history. ` +
-      `First archived line not found in order (${i} of ${want.length}): ${JSON.stringify(want[i] ?? '')}`
+      `First archived line not found in order (${i + 1} of ${want.length}): ${JSON.stringify(want[i] ?? '')}`
   ).toBe(want.length);
+  if (budget !== undefined) {
+    expect(
+      have.length - want.length,
+      `the assembler inserted more than the ${budget} line(s) this assembly can account for -- ` +
+        `presence and order alone do not bound how much text lands between archived lines`
+    ).toBeLessThanOrEqual(budget);
+  }
 }
 
 
@@ -413,6 +438,16 @@ describe('assemble-changelog', () => {
       '- archived entry B',
     ].join('\n');
 
+    /**
+     * What ONE probe fragment may add to this fixture: the `# header` line and
+     * the blank after it, the probe's own bullet, at most one generated heading
+     * when the date opens a new section, and at most two seam blanks around it,
+     * plus the trailing newline. Stated as a number the caller can defend
+     * rather than a measured constant, so it fails if the assembler starts
+     * padding settled history.
+     */
+    const WIDE_BUDGET = 7;
+
     for (const [date, where] of [
       ['2026-09-09', 'newer than the newest archived section'],
       ['2026-09-08', 'the newest archived section itself'],
@@ -424,10 +459,23 @@ describe('assemble-changelog', () => {
         const doc = withTree(WIDE, { [`${date}-2813-probe.md`]: '- probe entry' }, (root) =>
           assembleChangelog(root)
         );
-        assertArchiveConserved(doc, WIDE);
+        // header line + the blank after it + the probe + at most a new heading
+        // and the seam blanks around it + the trailing newline.
+        assertArchiveConserved(doc, WIDE, WIDE_BUDGET);
         // Not vacuous: the fragment really did land, exactly once. Conservation
         // alone is satisfied by an assembler that dropped the fragment.
         expect(doc.split('- probe entry').length - 1, 'the fragment did not land exactly once').toBe(1);
+        // WHERE it landed, not merely that it did. Every other assertion in
+        // this arm passes if the assembler merges the 2026-09-05 probe into the
+        // 2026-09-08 body instead of opening its own section -- which is the
+        // defect `assemble-changelog.ts` records review catching once, where a
+        // fragment at the archive's SECOND heading yielded 09-07 / 09-08 /
+        // 09-07. This line is what would catch it coming back.
+        const lines = doc.split('\n');
+        expect(
+          lines[lines.indexOf(headingFor(date)) + 1],
+          'the fragment did not land directly under its own dated heading'
+        ).toBe('- probe entry');
         const h = headings(doc);
         expect(new Set(h).size, 'a date heading is duplicated').toBe(h.length);
         const dates = h.map((x) => /\((\d{4}-\d{2}-\d{2})/.exec(x)![1]!);
@@ -437,6 +485,56 @@ describe('assemble-changelog', () => {
         assertEveryHeadingOpensItsBlock(doc);
       });
     }
+
+    /**
+     * The one legitimate VARIANT heading in the real archive --
+     * `**Recently Implemented** (2026-07-02, second batch):`, a day
+     * deliberately split in two, which `headingFor` cannot reproduce and which
+     * `splitSections` keeps verbatim. This PR is what first makes a fragment
+     * dated 2026-07-02 legal at all, so nothing had exercised the pairing.
+     */
+    it('merges into a VARIANT heading and keeps it verbatim', () => {
+      const VARIANT = '**Recently Implemented** (2026-07-02, second batch):';
+      const archive = [VARIANT, '- archived entry V', '', headingFor('2026-07-01'), '- archived entry W'].join('\n');
+      const doc = withTree(archive, { '2026-07-02-2813-probe.md': '- probe entry' }, (root) =>
+        assembleChangelog(root)
+      );
+      assertArchiveConserved(doc, archive, WIDE_BUDGET);
+      const lines = doc.split('\n');
+      expect(lines.indexOf(VARIANT), 'the variant heading was rewritten rather than kept').toBeGreaterThan(-1);
+      expect(
+        lines[lines.indexOf(VARIANT) + 1],
+        'the fragment did not merge under the variant heading it shares a date with'
+      ).toBe('- probe entry');
+      expect(
+        headings(doc).filter((x) => x.includes('2026-07-02')),
+        'the variant heading was duplicated by a generated one'
+      ).toEqual([VARIANT]);
+    });
+
+    it('places TWO fragments in one run, one merging and one opening a section', () => {
+      const doc = withTree(
+        WIDE,
+        { '2026-09-08-2813-merge.md': '- merged entry', '2026-09-05-2813-fresh.md': '- fresh entry' },
+        (root) => assembleChangelog(root)
+      );
+      // Two probes, so two bullets plus the one new heading and its seams.
+      assertArchiveConserved(doc, WIDE, WIDE_BUDGET + 1);
+      const lines = doc.split('\n');
+      expect(lines[lines.indexOf(headingFor('2026-09-08')) + 1], 'the same-date fragment did not merge').toBe(
+        '- merged entry'
+      );
+      expect(lines[lines.indexOf(headingFor('2026-09-05')) + 1], 'the in-span fragment did not open a section').toBe(
+        '- fresh entry'
+      );
+      const dates = headings(doc).map((x) => /\((\d{4}-\d{2}-\d{2})/.exec(x)![1]!);
+      expect(dates, 'the two fragments did not leave the document newest-first').toEqual([
+        '2026-09-08',
+        '2026-09-05',
+        '2026-09-01',
+      ]);
+      assertEveryHeadingOpensItsBlock(doc);
+    });
   });
 
   /**
@@ -478,6 +576,25 @@ describe('assemble-changelog', () => {
         /no longer reproduced in full/
       );
     });
+
+    /**
+     * The budget arm. Presence and order accept ANY amount of inserted text --
+     * measured in review as 575 blank lines and as arbitrary foreign text
+     * spliced mid-archive, both of which `doc.includes(archive)` had rejected.
+     * These two pin that the count now refuses them while still passing the
+     * splice it exists to allow.
+     */
+    it('accepts a splice INSIDE its budget', () => {
+      expect(() => assertArchiveConserved(spliced, A, 1)).not.toThrow();
+    });
+
+    it('rejects padding that conserves every archived line but reformats around them', () => {
+      const padded = ['x', '', '- one', '', '- two', '', '', 'y'].join('\n');
+      // Every archived line is present and in order -- the subsequence arm
+      // passes on its own, which is the gap the budget closes.
+      expect(() => assertArchiveConserved(padded, A)).not.toThrow();
+      expect(() => assertArchiveConserved(padded, A, 1)).toThrow(/inserted more than the 1 line/);
+    });
   });
 
   it('assembles the REPO tree, and that tree is the shape the fences read', () => {
@@ -499,7 +616,34 @@ describe('assemble-changelog', () => {
     // and both live seam defects on this branch were found against the real
     // archive, not a fixture.
     const archive = readFileSync(join(root, FRAGMENT_DIR, ARCHIVE_FILE), 'utf-8').replace(/\s+$/, '');
-    assertArchiveConserved(doc, archive);
+    // The insertion budget, derived rather than measured-and-pinned: the header
+    // and the blank after it, every fragment's own lines, one generated heading
+    // per fragment date the archive does not already head, at most one seam
+    // blank per emitted heading, and the trailing newline.
+    //
+    // Residual, stated because the count alone does not establish it: an
+    // insertion SMALLER than the seam allowance is inside the budget, so the
+    // delta bounds a bulk reformat (the 575-blank padding measured in review is
+    // far outside it) without bounding a few foreign lines. Attributing every
+    // unconsumed line rather than counting it is the stronger form; this is the
+    // one the review asked for and it closes the class that was measured open.
+    const headerLines = readFileSync(join(root, FRAGMENT_DIR, HEADER_FILE), 'utf-8').replace(/\s+$/, '').split('\n')
+      .length;
+    const archiveDates = new Set(
+      archive
+        .split('\n')
+        .map((l) => /^\*\*Recently Implemented\*\* \((\d{4}-\d{2}-\d{2})/.exec(l)?.[1])
+        .filter((d): d is string => d !== undefined)
+    );
+    const fragments = readEntries(root);
+    const fragmentLines = fragments.reduce((n, e) => n + e.text.split('\n').length, 0);
+    const newHeadings = new Set(fragments.map((e) => e.date).filter((d) => !archiveDates.has(d))).size;
+    const emittedHeadings = archiveDates.size + newHeadings;
+    assertArchiveConserved(
+      doc,
+      archive,
+      headerLines + 1 + fragmentLines + newHeadings + emittedHeadings + 1
+    );
     assertEveryHeadingOpensItsBlock(doc);
     const onDisk = readdirSync(join(root, FRAGMENT_DIR, ENTRIES_DIR)).filter((n) => n !== '.gitkeep');
     expect(readEntries(root).length, 'a fragment on disk was not read into the assembly').toBe(onDisk.length);
--- a/tests/unit/scripts/assemble-changelog.test.ts
+++ b/tests/unit/scripts/assemble-changelog.test.ts
@@ -135,8 +135,10 @@ function assertEveryHeadingOpensItsBlock(d: string): void {
  * time with the tree restored between: dropping one body line per section
  * (`s.body.slice(1)`) failed at archived line 1, collapsing runs of spaces in
  * every body line failed at line 28, and swapping two body lines failed at
- * line 3. The substring form rejects those three too -- it is not weaker
- * against a real reformat, only wrong about a splice.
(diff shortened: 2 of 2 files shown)


go-to-k:
Reviewed at `6d0b5230`. The fix is at the right layer and the measurement in the description holds up — I re-ran it rather than taking it on trust, and it reproduces exactly, including the `2026-05-31` / `2026-05-30` boundary.

The splice preserves every archived line in order by construction (`assemble-changelog.ts:216-224` prepends into an existing section's body; a new section is spliced at section granularity), so a subsequence IS what "settled history is not reformatted" claims. Option 3 from go-to-k/cdkd#2813 would have made a legitimately-rebased lane unwritable. The greedy earliest-match loop is the standard complete subsequence test — no false negatives.

I also tried to break the new check, since replacing a fence is where a false pass gets introduced. Against the real 1440-line archive, with the header in the document so the whole-document match had somewhere to hide:

| mutation | probes | absorbed |
| --- | --- | --- |
| delete one archived line, at every index | 1440 | 0 |
| swap adjacent archived lines, at every index | 1439 | 1 — two identical blank lines, i.e. a no-op |
| collapse runs of spaces, every non-blank line | 1263 | 6 — all `---` lines, which contain no spaces, so the mutation itself was a no-op |
| a fragment whose text is byte-identical to an archived bullet, spliced above a dropped copy of that line | 1 | 0 |
| an entire archived section deleted | 1 | 0 |

Zero real reformats pass. The deletion claim in particular is sound: fragment text is prepended ABOVE a section's archived body, so it can never supply a matching line at a deleted line's position.

Two things to fix before this merges, then one separate issue and some take-or-leave notes.

---

### M0 — must fix before merge: the new arms never pin WHERE the fragment landed

`tests/unit/scripts/assemble-changelog.test.ts:423-438`. Conservation, appears-exactly-once, heading uniqueness and descending order all pass if the assembler merges the `2026-09-05` probe into the `2026-09-08` body without opening a new section. That is the exact defect class `assemble-changelog.ts:210-215` records review catching once ("a fragment dated at the archive's SECOND heading yielded 09-07 / 09-08 / 09-07"), and these five arms are where it would come back.

One line per arm closes it:

```ts
const lines = doc.split('\n');
expect(lines[lines.indexOf(headingFor(date)) + 1], 'the fragment did not land under its own heading').toBe('- probe entry');
```

### M1 — must fix before merge: the insertion budget is unbounded

`assertArchiveConserved` constrains presence and order but places no bound on how much text lands between archived lines. Measured on the real tree: inserting a blank line after every archived bullet — 575 inserted lines — is accepted, and so is arbitrary foreign text spliced into the middle of the archive. Both were rejected by `doc.includes(archive)`.

(The rendering effect is smaller than it first looks — the `<ul>` count is unchanged at 63, it converts tight lists to loose `<li><p>` — but it is still a reformat of settled history, which is the property the assertion is named for.)

Accepting insertions is inherent to any subsequence form, and that is fine: the splice is exactly an insertion. What the splice needs is a *bounded, attributable* insertion, not an arbitrary one. A line-count delta restores that without re-breaking the splice — in the WIDE loop the expected delta is the probe's one line plus at most one seam blank, and in the real-tree case it is `headerLines + 1 + Σ fragmentLines + seams + 1`.

Worth stating in `assertArchiveConserved`'s comment too: the current text says the check is not weaker "against a real reformat", and against the deletion / reorder / re-wrap classes that is measurably true, but the insertion class is a real gap until this is bounded.

### Separate issue — not this PR

go-to-k/cdkd#2859. Unlocking in-span dates widens the `changelog-entry-size` cap's evasion window: that fence caps only entries whose heading date is `>= 2026-09-05`, so a fragment named `2026-09-04-<n>-<slug>.md` now ships uncapped. The hole predates this PR (anything dated below `2026-05-31` was always both shippable and uncapped) and the fence's own header documents the limit deliberately, so this is not a defect in your change — but the window widens by ~97 days, and `readEntries` now makes the limit non-structural, which is what makes it fixable. Nothing to do here.

### Take-or-leave

- `:115` — "both documented in `changelog.d/_header.md`" is not quite right. `_header.md:143-149` documents the splice under an existing heading; the descending-insert of a NEW section is documented in `assemble-changelog.ts`, not there. Same wording in the PR description.
- `:142` — "one it DELETED would still be caught" overclaims in general: a deleted line re-syncs against a later duplicate (`assertArchiveConserved(['a','b','','b'].join('\n'), ['a','','b'].join('\n'))` passes). It holds for THIS archive because body bullets are unique, which is worth saying rather than dropping — scope the sentence to the corpus.
- `:152` — `(${i} of ${want.length})` prints a 0-based count where the phrasing ("First archived line not found in order") reads as an ordinal, so the first line reports "(0 of 51)". `${i + 1}` reads correctly.
- The WIDE fixture has no variant heading, so `**Recently Implemented** (2026-07-02, second batch):` (`_archive.md:1054`) is untested — and this PR is what first makes a fragment dated `2026-07-02` legal at all. I checked it by hand and it works (the heading is kept verbatim and conservation holds); a sixth arm would pin it.
- Two fragments in one run — one merging into an existing section, one opening a new one — is untested. Also verified working by hand (`09-08 / 09-05 / 09-01`).
- This also fixes go-to-k/cdkd#2816, which is the same defect filed narrower (the same-date case only). Worth adding to the `Closes` line so it does not stay open.

nix-tkobayashi:
Addressed M0, M1 and the review notes in `eecac660`. CI is green.

**M0.** Added the suggested assertion in all five date-position cases: the probe must appear directly under its own dated heading.

**M1.** Added a line-count budget to `assertArchiveConserved`. On this tree the measured delta is 183 lines against a calculated limit of 234. The comment documents the remaining slack: an insertion smaller than the seam allowance still fits inside the budget.

The seam term counts heading LINES rather than distinct dates. Those differ on this archive — 51 against 50, since 2026-07-02 carries the variant heading beside the plain one — and my first cut counted dates, one line tighter than its own stated derivation. The PR description also said 235, which came from an ad-hoc probe taken before the derivation was written into the test and never re-measured against it; it now says 234.

I did not, as I first believed, correct the sentence you flagged. "It is not weaker against a real reformat, only wrong about a splice" was still there, with the new budget section added below it. It now claims only the deletion, reorder and re-wrap classes and points at the budget for the insertion class.

**Notes.** Corrected the documentation references in the comment and the PR description, scoped the deletion claim to this corpus — your counter-example is exact, a deleted line re-syncs against a later duplicate — and fixed the diagnostic's ordinal. Added the variant-heading case and the two-fragment merge/insert case. Added the closing reference for go-to-k/cdkd#2816 to the PR body.

Mutation results are in the PR description. Assembler tests 26 -> 30; local full suite 953 files, 20,715 tests passed and one skipped.

go-to-k:
Reviewed `eecac660`. M0 and M1 are both answered, and the self-corrections in your reply — the seam term counting heading LINES rather than dates, the stale 235, and withdrawing the claim that you had already reworded the sentence — are the three I would otherwise have had to raise. Thank you for stating the third rather than quietly fixing it.

I re-ran the parts a reviewer should not take on trust.

**M0 is load-bearing, and I checked it the hard way.** Rather than confirm the assertion exists, I introduced the historic defect shape into `assemble-changelog.ts` itself: fold a new-section fragment into the NEWER section above it, emitting no heading, so heading uniqueness and descending order both stay intact. That reds **exactly 2 of 30** — the two arms carrying your new landing assertion. 28 pass. Your claim that every other assertion in the arm survives that defect is exact.

**M1's budget bites on the class that was measured open.** Padding the assembler with a blank after every archived bullet reds 5 tests including the real-tree one. Tree restored byte-exact after each probe.

**The arithmetic reproduces**: 234 against a delta of 183, heading lines 51 against 50 dates.

---

### M1b — must fix before merge: the seam term is ~52x the need, and it grows

This is the last item; everything else below is take-or-leave and I will merge on M1b alone.

`emittedHeadings` budgets one seam per emitted heading, but the assembler inserts a seam only when the previous section's body ends non-blank. Measured on the real archive:

```
sections: 51
sections whose body ends NON-blank (the only ones that cause a seam):  1
budgeted seam allowance:                                              52
```

So 51 of the 51-line slack is the seam term alone, and it is not a fixed allowance — it grows **+1 for every dated section the archive ever gains**. A year from now the window for unattributed foreign text inside settled history is a few hundred lines. The residual note calls it a seam allowance, which reads as a constant; as written it is a monotonically widening acceptance window.

Deriving it instead makes the bound tight, and the derivation is short because a merge PREPENDS to a section's body and so cannot change whether that body ends blank:

```ts
// A seam is inserted before a section only when the PREVIOUS section's body
// ends non-blank. A merged fragment prepends, so it cannot change that; a NEW
// section's body is the fragment text, which can. Plus one for a non-empty
// preamble.
const bodyEndsNonBlank = /* archive sections whose last body line is non-blank */;
const seams = bodyEndsNonBlank + newHeadings + 1;
```

Measured, that gives `176 + 1 + 3 + 1 + 3 + 1 = 185` against the same delta of 183 — slack 2 instead of 51, and it stops growing. I checked the emitted-seam rule against `assemble-changelog.ts`'s loop directly: `out` is empty before the first section on this archive (the first line IS a heading, so `preamble` is empty), and thereafter the guard reads the previous body's last line.

### Take-or-leave

- `archiveHeadingLines`'s regex drops the `[^)]*\):$` tail that `HEADING` carries, so it is strictly looser than the assembler's own. Loose and strict both return 51 today — I measured — so this is inert, but a line that loose-matches and strict-fails would inflate `archiveDates`, under-count `newHeadings`, and leave the budget one short of a legitimate assembly. Exporting `HEADING` and using it removes the whole question. (Also: the name says LINES, the value holds dates.)
- `expect(lines.indexOf(VARIANT)).toBeGreaterThan(-1)` in the variant case cannot fail — `assertArchiveConserved` on the line above already requires VARIANT present and in order.
- `WIDE_BUDGET`'s JSDoc derives it for the `WIDE` fixture but the variant case applies it to a different archive; widening `WIDE` later would silently loosen that case.
- The comment above `assertArchiveConserved(doc, WIDE, WIDE_BUDGET)` restates the JSDoc, so there are two copies to keep in step.
- In the doc comment, `assertArchiveConserved(['a','b','','b'], ['a','','b'])` would `TypeError` as written — the parameters are strings. The claim is correct in the `.join('\n')` form (I ran it); restoring the joins keeps it executable.
- Pre-existing, now sitting next to your new code: the real-tree comment says no fragment has been written yet and the count is 0, while `changelog.d/entries/` holds three and your new `fragmentLines` term reads them.

Push the M1b tightening and I will merge.

nix-tkobayashi:
Addressed M1b and all six optional notes in `03ef39d5`. CI is green.

The seam budget is now `sectionEndsNonBlank + 2 * newHeadings + preambleSeam`. On this archive those terms are `1 + 2 * 1 + 0` — the preamble is empty because the first line is a heading. The full budget is `176 + 1 + 3 + 1 + 3 + 1 = 185` against an assembled-line delta of 183: slack 2. It no longer adds an allowance for every archived heading.

The budget now uses `HEADING` exported from the assembler. Removed the redundant variant assertion and the call-site comment, renamed `WIDE_BUDGET` to `TWO_SECTION_BUDGET`, restored `.join('\n')` in the counter-example, and corrected the stale zero-fragment comment.

Two corrections to what the PR description claimed, both found by having the mutation results re-run rather than re-read.

It called my fold-upward mutation "exactly that shape" as yours. It is not: mine also disturbs ordering, which is why it fails six tests where yours isolates the two landing assertions. I did not reproduce your ordering-preserving form, and the description now says so.

And the table labelled each mutation in prose. Run from those labels the same idea lands in a different place and reds a different number — measured, on two of the six rows. The table now carries the exact expression each row substitutes, so the counts are reproducible rather than merely reported.

go-to-k:
Merged as `2adb5942`. Thank you — this closed a fence that was blocking every lane from dating an entry anywhere in a three-month window, and it closed go-to-k/cdkd#2816 with it.

I verified the seam derivation rather than reading it: 3 foreign lines spliced mid-archive now red five tests including the real-tree one, where under the previous 234-line budget they passed silently. Budget 185 against a delta of 183, slack 2 and no longer growing with the archive.

Three things stood out across the rounds, and they are the reason this took two rounds instead of five.

You kept finding your own errors before I did. The seam term counting heading LINES rather than dates, the stale 235 from a probe taken before the derivation existed, and the mutation table labelling its rows in prose — that last one is the sharpest, because "run from those labels the same idea lands in a different place and reds a different number, measured on two of six rows" is exactly the failure mode a table of mutation results is supposed to prevent and usually doesn't. The table now carrying the substituted expression per row is strictly better than what I asked for.

You withdrew a claim instead of quietly fixing it. "I did not, as I first believed, correct the sentence you flagged" cost you nothing to leave unsaid, and saying it is what let me trust the rest of the reply without re-deriving it.

And you corrected my own framing where it was wrong: my fold-upward probe and yours are not the same shape — yours disturbs ordering and reds six, mine preserves it and isolates the two landing assertions. You could have accepted the flattering version.

The residual is recorded honestly too: an insertion inside the remaining 2-line slack still fits, and attributing every unconsumed line is the stronger form nobody has written. That is the right place to leave it.

PR #2797, opened 2026-09-08
fix(resolver): drain every concurrent part before a rejection surfaces, so a sibling's secret is recorded first

nix-tkobayashi:
Closes #2563

## What was wrong

`IntrinsicFunctionResolver` resolved a join's parts (`resolveJoin`) and a list's elements (`resolveValue`'s array arm) with `Promise.all`, which surfaces a rejection AT ONCE. Resolving a secret dynamic reference records `plaintext -> expression` into `context.recordedSecretValues` just before its promise settles, so a caller's `catch` / `finally` ran while a sibling part was still in flight.

`DeployEngine`'s `Export.Name` block copies its private `nameSecrets` map into the pass map in exactly such a `finally`. A recording that landed after that copy died with the local, leaving a plaintext the outputs pass did not know as a needle — for the masking of later export-name refusal messages and for `redactOutputs` at persist. `cdkd scrub` had the same shape and go-to-k/cdkd#2562 replaced it there with a view of the pass map; the engine still had the copy.

## What changes (`src/deployment/intrinsic-function-resolver.ts`)

`allSettledKeepingFirstRejection` wraps both concurrent sites. It returns `Promise.all`'s result and throws `Promise.all`'s choice of error — the EARLIEST REJECTION IN TIME, via a per-promise `catch` whose first assignment wins, PER HELPER INVOCATION — while awaiting every input, so unless the cap below has expired, every recording the resolution triggered has happened by the time a caller is reached.

- **Both sites, not just the join.** A list nested inside a join part rejects through the array arm's own `Promise.all`, so draining `resolveJoin` alone would still let a late recording escape.
- **In the resolver, not per consumer.** go-to-k/cdkd#2531's shared-map view lets a late write land whenever it happens; it cannot make it land before the next consumer runs.
- **The engine's copy shape is left alone.** The drain is what makes the existing end-of-block copy safe as it stands, and three comments in `src/deployment/deploy-engine.ts` that described this residual as live are corrected. The engine is touched for a different reason: both `resolveOutputs` call sites wrap the pass in one drain budget, below.

[...]

PR #2896, opened 2026-09-10
docs(diff): ground the resolveBestEffort clone in the ownership contract, not the retired resolver write-back

### LINKED ISSUE #2832: diff-calculator: three comments still explain the clone by the resolver write-back that #2764 retired
## Summary

Three comments still explain `DiffCalculator`'s clone by a mechanism go-to-k/cdkd#2764 retired: "the intrinsic resolver mutates its input in place". `resolveSub`'s write-back into the caller's `Fn::Sub` variable map was that mechanism, and #2764 removed it.

- `src/analyzer/diff-calculator.ts:787` — the comment above the `resolveBestEffort` call, whose parenthetical names the retired write-back specifically ("it rewrites an `Fn::Sub` variable map's `Fn::GetAtt` to the resolved current-state value").
- `tests/unit/analyzer/diff-calculator.test.ts:1514` — the same claim as the rationale of `does NOT mutate the desired template (resolveBestEffort resolves a clone)`.
- `tests/unit/analyzer/diff-calculator.test.ts:1541` — "(as the real intrinsic resolver does)", attached to the deliberately mutating resolver that case injects. This one does not share the other two's wording, so a grep for the sentence misses it.

A fourth copy in `src/analyzer/outputs-diff.ts` was re-grounded in go-to-k/cdkd#2773; the maintainer flagged these as informational and out of that PR's scope, which is why they are filed rather than folded in.

## Why it matters

The clone itself is correct and must stay. What is wrong is the reason given for it, and that is the direction that gets a guard deleted: a reader who checks the claim finds no in-place write anywhere in the real resolver and concludes the clone is dead weight.

The clone's actual job is DEFENSIVE, and the test says so by construction rather than by its comment. `resolveBestEffort` collects into a fresh object, so with today's resolver nothing writes through to the shared desired template and removing the clone changes no outcome. What the clone protects against is an `IntrinsicResolveFn` that mutates its argument — which is exactly the resolver `diff-calculator.test.ts:1542` injects, and why that case still discriminates. Stating it as "the resolver mutates" makes a defensive guard read as a response to live behaviour, and then the retired behaviour makes the guard read as obsolete.

## Suggested fix

Re-ground all three on the contract rather than on any resolver's behaviour: `resolveBestEffort` is handed a leaf of the SHARED desired template and must not let a resolver write through it, whatever that resolver does. Keep the retired write-back as labelled history so the next reader does not re-derive it as live, which is the shape `src/analyzer/outputs-diff.ts` uses after go-to-k/cdkd#2773. The injected mutating resolver keeps its comment, minus the claim that the real one behaves that way.

## Verification

`grep -rniE "resolver (mutates|rewrites)[^.]*in place|as the real intrinsic resolver does" src/ tests/` returns nothing, and `does NOT mutate the desired template (resolveBestEffort resolves a clone)` still passes and still reds when the `structuredClone` is removed — the injected resolver is what keeps that probe meaningful, so it must stay mutating.

## Classification

Dup-check: searched open issues for `mutates`, `in place`, `resolveSub`, `clone` and `diff-calculator`; go-to-k/cdkd#2764 is the merged PR that retired the mechanism and no open issue tracks the remaining copies.
Session-fit: next (not this session) — documentation-only in a file neither this lane nor its review touches, and explicitly scoped out of go-to-k/cdkd#2773.
Severity: low — wrong text that does not execute, but it argues for deleting a guard that is still needed.
Effort: small (S) — three comments, riding the unit suite.
Estimate: ~20 min — most of it stating the ownership contract correctly.

nix-tkobayashi:
## Summary

Four comments in `src/analyzer/diff-calculator.ts` and its test explained the `structuredClone` in `resolveBestEffort` by a mechanism that no longer exists — "the intrinsic resolver mutates its input in place". That was `resolveSub`'s write-back into the caller's `Fn::Sub` variable map, retired by go-to-k/cdkd#2764.

The clone is correct and stays. Only its stated reason was wrong, and in the direction that gets a guard deleted: a reader who checks the claim finds no in-place write in the resolver and concludes the clone is dead weight. Each site is re-grounded on the OWNERSHIP contract — `resolveBestEffort` is handed a leaf of the SHARED desired template, so no `IntrinsicResolveFn` may write through it, whatever that function does. Shared is checked, not assumed: `DeployEngine.deploy` hands `calculateDiff` the same `effectiveTemplate` it then provisions from. The retired write-back is kept as labelled history, which is the shape `src/analyzer/outputs-diff.ts` took in go-to-k/cdkd#2773.

**The sweep found a fourth site the issue did not name**, in the same file and added by the same commit: the `rawGetAttRefs` snapshot note, which said the comparison loop "mutates in place" the desired property intrinsics. `git blame` puts that note and the `resolveBestEffort` one in the same commit, 772dd03f (go-to-k/cdkd#939), which is also where the clone arrived — `git show 772dd03f^:src/analyzer/diff-calculator.ts` has the bare `resolved[key] = await resolveFn(value)`. So the loop never mutated the template: it collects into a fresh object, and nothing in the file assigns to `.Properties` at all (`grep -nE '\.Properties\s*=' src/analyzer/diff-calculator.ts` is empty). The raw intrinsics survive the loop, which is what made the old note's ordering rationale wrong.

The second commit narrows that same note. Its first version also called the old text's "a GetAtt can no longer be detected" false, which over-claims — that IS true of the resolved bag, as the paragraph above it says — so only the unqualified half is kept.

Comments only — **0 non-comment changed lines**, measured on the diff rather than asserted:

```
git diff origin/main -U0 -- src tests | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' \
  | sed 's/^[+-]//;s/^[[:space:]]*//' | grep -vE '^//|^$' | wc -l
0
```

## Test plan

No new test: the behaviour under discussion is already fenced by `does NOT mutate the desired template (resolveBestEffort resolves a clone)`, and this PR's job is to stop that fence reading as dead weight. The verification the issue prescribes, plus a probe that the fence still discriminates:

- **Verification grep** — `grep -rniE 'resolver (mutates|rewrites)[^.]*in place|as the real intrinsic resolver does' src/ tests/` returns nothing (before, it returned the three lines the issue names).
- **Mutation probe on the committed tree** — `resolved[key] = await resolveFn(structuredClone(value))` -> `resolveFn(value)` (receipt: bytes 51079 -> 51062, anchor 1 -> 0), then `vp test run tests/unit/analyzer/diff-calculator.test.ts`: `1 failed | 34 passed (35)`, the single failure being `does NOT mutate the desired template (resolveBestEffort resolves a clone)`. Restored from a byte-exact copy; `git diff` clean afterwards. The injected mutating resolver therefore stays mutating — it is what keeps the case discriminating.
- **Full suite** — 968 files, 21109 passed | 1 skipped, no type errors, rc=0, RUN root asserted as this worktree, re-run after the second commit. (One earlier run of the same tree reported the same 968/21109 with rc=1 and a single `Worker exited unexpectedly`; a peer session's suite was on the host. rc=0 on a quiet one.)
- `vp check --fix` + `vp run check` + `vp run typecheck:test` + `vp run build` + `vp run gen:all-matrices` / `audit:coverage:check` — all clean, nothing regenerated dirty.
- **Docs / rules**: nothing to update. `grep -rniE 'resolver (mutates|rewrites)|mutates? (its|the) (input|caller|argument)' .claude/rules/ docs/ CLAUDE.md README.md changelog.d/` returns zero hits, and neither `.claude/rules/analyzer.md` nor `.claude/rules/layout-analyzer.md` mentions `diff-calculator.ts`, `resolveBestEffort` or the clone.

**Review round 1** (go-to-k/cdkd#2896 review of 2026-09-10, commit `9563177c`) — five items, all taken:

- **M0** (blocker): the previous text said the loop "never" mutated the desired property intrinsics. False, and in this PR's own defect class: at `772dd03f^` `resolveBestEffort` called `resolveFn(value)` with no clone, `DeployEngine` handed it the real resolver, and `resolveSub` bound the caller's `Fn::Sub` variable map by destructuring and wrote into it. The note was TRUE when written and go-to-k/cdkd#939 falsified it in the same commit. The sentence is temporal now.
- **m1**: the clone note names both fences — `tests/unit/deployment/intrinsic-sub-variables-not-mutated.test.ts` for the resolver side, and `does NOT mutate the desired template (resolveBestEffort resolves a clone)` for the clone side. Probed rather than cited: reinstating the write-back in `resolveSub` reds exactly the two cases the comment names (`2 failed | 5 passed (7)`), restored byte-exact.
- **m2**: the ownership rationale covers both `calculateDiff` callers; `cdkd diff` never reaches a deploy phase. (It shares the template with the Outputs resolution, but not with the skipped-output digests — `outputsDigestSource` is a `structuredClone`.)
- **m3**: the `rawGetAttRefs` snapshot is named a precomputation, with no cost claim attached (none was derivable).
- **m4**: the restating clause is gone, and the test comment describes the injected resolver's actual walk.

Verification for the round: full suite 968 files / 21109 passed / 1 skipped, rc=0, root asserted; the greps above still return nothing; still 0 non-comment changed lines. Codex on the delta: 2 rounds (r1 found the walk description overstated) plus a maintainer-checklist proxy pass, NO ISSUES.

`main` advanced by go-to-k/cdkd#2797 and go-to-k/cdkd#2898 during the round; neither touches `diff-calculator.ts` or its test, so no rebase. The three cross-module facts these comments assert were re-derived against `origin/main` as it stands after that merge.

**Review round 2** (commit `4a592670`) — two required items, both taken:

- **M1**: "the raw template survives the loop (`resolveBestEffort` clones, and collects into a fresh bag)" did not cover the `catch` arm, where `resolved[key] = value` aliases the raw leaf by identity. The sentence now names the real reason — the `try` arm clones, the `catch` arm only aliases into a bag no consumer mutates — after re-walking those consumers (`withoutAcceptedSilentDropProperties` and all four live `canonicalizeDesiredProperties` implementations either return their input unchanged or build a fresh object).
- **M2**: "stands for ANY `IntrinsicResolveFn`" was unearned — the injected resolver rebuilds arrays, so array-ELEMENT mutation is outside what the case exercises. The finding is taken; the maintainer's proposed wording ("does not stand in for the narrow write-back") is not, because it overshoots the other way: that resolver DOES perform the retired write-back on the `Fn::Sub` variable map. Probed — a resolver doing ONLY the narrow write-back passes with the clone (`35 passed`) and reds the case without it (`1 failed | 34 passed`), so the narrow shape alone is what discriminates. The text now says the walk covers that write-back, goes wider along the OBJECT axis, and names the axis it does not model.

`diff-calculator.ts`'s sentence about `cdkd diff` sharing the template is left as written, per the maintainer's own ruling on that item.

Verification for the round: full suite 968 files / 21109 passed / 1 skipped, rc=0, root asserted (re-run on the amended tree); every gate command rc=0; still 0 non-comment changed lines. Codex on the delta: NO ISSUES; the maintainer-checklist proxy pass caught the M2 overshoot, then NO ISSUES. Three Codex rounds on the reply itself caught a "below" that should read "above" (fixed in the comment too) and a claim that `outputs-diff.ts` reads `Outputs` only (it walks the whole template, read-only).

No integ: the diff touches no `integ-destroy` / `integ-broad` / `integ-local` / `integ-schema-migration` scope, and changes no executable line.

Closes #2832

### DIFF
--- a/src/analyzer/diff-calculator.ts
+++ b/src/analyzer/diff-calculator.ts
@@ -159,14 +159,18 @@ export class DiffCalculator {
 
     // Snapshot each resource's `Fn::GetAtt` / `Fn::Sub`-`${X.Attr}` references
     // from the RAW template, in one pass, for promoteInPlaceAttributeDependents
-    // below: it needs the references a property declared, while the comparison
-    // loop reads the RESOLVED values, where a GetAtt to an
+    // below: it needs the references a property DECLARED, while the comparison
+    // loop works on RESOLVED values, where a GetAtt to an
     // in-place-referenceable resource has already been replaced by its resolved
-    // current value.
-    // (This said the loop "mutates in place" the desired property intrinsics.
-    // It never has: go-to-k/cdkd#939 added the clone in the same change that
-    // wrote the note, and the loop collects into a fresh bag, so the RAW
-    // template keeps its intrinsics and this snapshot reads them either way.)
+    // current value. A PRECOMPUTATION rather than a rescue — that function also
+    // receives `desiredTemplate` and could re-extract from it, since the raw
+    // template survives the loop (`resolveBestEffort` clones, and collects into
+    // a fresh bag).
+    // (This note used to say the loop "mutates in place" the desired property
+    // intrinsics. It has not since go-to-k/cdkd#939 added the clone in the same
+    // change that wrote the note; BEFORE that commit the loop ran the real
+    // resolver over the template leaf itself, and `resolveSub` wrote its
+    // resolved values back into the caller's `Fn::Sub` variable map.)
     const rawGetAttRefs = new Map<string, Map<string, Map<string, Set<string>>>>();
     for (const [logicalId, desiredResource] of Object.entries(desiredResources)) {
       if (desiredResource.Type === 'AWS::CDK::Metadata') continue;
@@ -789,18 +793,24 @@ export class DiffCalculator {
     for (const [key, value] of Object.entries(properties)) {
       try {
         // Resolve a CLONE. The reason is the OWNERSHIP contract, not any
-        // resolver's behaviour: this value is a leaf of the SHARED desired
-        // template, which the deploy phase re-resolves later against the
-        // in-flight state, so nothing here may let a resolver write through it.
-        // A resolved current-state value baked in here would still be read as a
+        // resolver's behaviour: this value is a leaf of a template its CALLER
+        // still shares with other consumers, so nothing here may let a resolver
+        // write through it. Both call sites share it — `cdkd diff` hands the
+        // same object on to the Outputs resolution that follows, and `cdkd
+        // deploy` re-resolves it against the in-flight state and provisions
+        // from it. The deploy one is the consequence with teeth: a
+        // resolved current-state value baked in here would still be read as a
         // literal then, and a genuinely-changed dependent of an in-place-updated
         // upstream would be skipped. Cloning makes that impossible whatever the
         // resolver does.
-        // (It is not true that the resolver mutates its input today: `resolveSub`
-        // wrote back into the caller's `Fn::Sub` variable map until
-        // go-to-k/cdkd#2764 retired it, and this comment named that write-back.
-        // Kept as history so the clone is not deleted as dead weight, which is
-        // exactly what the contract forbids.)
+        // (The resolver does not mutate its input today: `resolveSub` wrote back
+        // into the caller's `Fn::Sub` variable map until go-to-k/cdkd#2764
+        // retired it, and this comment named that write-back. Kept as history so
+        // the clone is not read as dead weight. Both directions are fenced:
+        // `tests/unit/deployment/intrinsic-sub-variables-not-mutated.test.ts`
+        // holds the resolver to it, and deleting this `structuredClone` reds
+        // `does NOT mutate the desired template (resolveBestEffort resolves a
+        // clone)` in `tests/unit/analyzer/diff-calculator.test.ts`.)
         resolved[key] = await resolveFn(structuredClone(value));
       } catch {
         resolved[key] = value;
--- a/src/analyzer/diff-calculator.ts
+++ b/src/analyzer/diff-calculator.ts
@@ -158,11 +158,16 @@ export class DiffCalculator {
     const processedLogicalIds = new Set<string>();
 
     // Snapshot each resource's `Fn::GetAtt` / `Fn::Sub`-`${X.Attr}` references
-    // from the RAW template, BEFORE the comparison loop below resolves (and
-    // mutates in place) the desired property intrinsics. Used by
-    // promoteInPlaceAttributeDependents — once `resolveBestEffort` runs, a
-    // GetAtt to an in-place-referenceable resource has been replaced by its
-    // resolved current value and can no longer be detected.
+    // from the RAW template, in one pass, for promoteInPlaceAttributeDependents
+    // below: it needs the references a property declared, while the comparison
+    // loop reads the RESOLVED values, where a GetAtt to an
+    // in-place-referenceable resource has already been replaced by its resolved
+    // current value.
+    // (This said the loop "mutates in place" the desired property intrinsics
+    // and that a GetAtt "can no longer be detected" once `resolveBestEffort`
+    // runs. Both were false from the commit that wrote them — go-to-k/cdkd#939
+    // added the clone in the same change, so the template keeps its raw
+    // intrinsics.)
     const rawGetAttRefs = new Map<string, Map<string, Map<string, Set<string>>>>();
     for (const [logicalId, desiredResource] of Object.entries(desiredResources)) {
       if (desiredResource.Type === 'AWS::CDK::Metadata') continue;
@@ -784,14 +789,19 @@ export class DiffCalculator {
     const resolved: Record<string, unknown> = {};
     for (const [key, value] of Object.entries(properties)) {
       try {
-        // Resolve a CLONE: the intrinsic resolver mutates its input in place
-        // (e.g. it rewrites an `Fn::Sub` variable map's `Fn::GetAtt` to the
-        // resolved current-state value). Mutating the shared desired template
-        // here would bake the OLD value into the template, so the deploy phase's
-        // later re-resolution (against the in-flight state, where an in-place-
-        // updated upstream now holds its NEW value) would still see the stale
-        // literal and skip a genuinely-changed dependent. Cloning keeps the raw
-        // intrinsics intact for the deploy phase.
+        // Resolve a CLONE. The reason is the OWNERSHIP contract, not any
+        // resolver's behaviour: this value is a leaf of the SHARED desired
+        // template, which the deploy phase re-resolves later against the
+        // in-flight state, so nothing here may let a resolver write through it.
+        // A resolved current-state value baked in here would still be read as a
+        // literal then, and a genuinely-changed dependent of an in-place-updated
+        // upstream would be skipped. Cloning makes that impossible whatever the
+        // resolver does.
+        // (It is not true that the resolver mutates its input today: `resolveSub`
+        // wrote back into the caller's `Fn::Sub` variable map until
+        // go-to-k/cdkd#2764 retired it, and this comment named that write-back.
+        // Kept as history so the clone is not deleted as dead weight, which is
+        // exactly what the contract forbids.)
         resolved[key] = await resolveFn(structuredClone(value));
       } catch {
         resolved[key] = value;
--- a/src/analyzer/diff-calculator.ts
+++ b/src/analyzer/diff-calculator.ts
@@ -164,8 +164,8 @@ export class DiffCalculator {
     // in-place-referenceable resource has already been replaced by its resolved
     // current value. A PRECOMPUTATION rather than a rescue — that function also
     // receives `desiredTemplate` and could re-extract from it, since the raw
-    // template survives the loop (`resolveBestEffort` clones, and collects into
-    // a fresh bag).
+    // template survives the loop: the `try` arm clones, and the `catch` arm
+    // only aliases the leaf into a bag no consumer mutates.
     // (This note used to say the loop "mutates in place" the desired property
     // intrinsics. It has not since go-to-k/cdkd#939 added the clone in the same
     // change that wrote the note; BEFORE that commit the loop ran the real
--- a/src/analyzer/diff-calculator.ts
+++ b/src/analyzer/diff-calculator.ts
@@ -163,11 +163,10 @@ export class DiffCalculator {
     // loop reads the RESOLVED values, where a GetAtt to an
     // in-place-referenceable resource has already been replaced by its resolved
     // current value.
-    // (This said the loop "mutates in place" the desired property intrinsics
-    // and that a GetAtt "can no longer be detected" once `resolveBestEffort`
-    // runs. Both were false from the commit that wrote them — go-to-k/cdkd#939
-    // added the clone in the same change, so the template keeps its raw
-    // intrinsics.)
+    // (This said the loop "mutates in place" the desired property intrinsics.
+    // It never has: go-to-k/cdkd#939 added the clone in the same change that
+    // wrote the note, and the loop collects into a fresh bag, so the RAW
+    // template keeps its intrinsics and this snapshot reads them either way.)
     const rawGetAttRefs = new Map<string, Map<string, Map<string, Set<string>>>>();
     for (const [logicalId, desiredResource] of Object.entries(desiredResources)) {
       if (desiredResource.Type === 'AWS::CDK::Metadata') continue;
--- a/tests/unit/analyzer/diff-calculator.test.ts
+++ b/tests/unit/analyzer/diff-calculator.test.ts
@@ -1541,10 +1541,13 @@ describe('DiffCalculator - replacement propagation to dependents (issue #807)', 
       },
     };
 
-    // A MUTATING resolver: it rewrites the Fn::Sub variable map's GetAtt in
-    // place. NOT what the real resolver does — go-to-k/cdkd#2764 retired that
-    // write-back — and injected here precisely because the clone must hold
-    // against ANY resolver. It is what keeps this case discriminating.
+    // A MUTATING resolver, and deliberately a BLUNTER one than any real
+    // resolver: every object it descends into has each key written back in
+    // place (a node carrying `Fn::GetAtt` is terminal — replaced whole, its
+    // members never walked; arrays it rebuilds). So it stands for ANY
+    // `IntrinsicResolveFn`, not for the narrow `resolveSub` write-back
+    // go-to-k/cdkd#2764 retired — that generality is the comment's own
+    // argument, and it is what keeps this case discriminating.
     const mutatingResolver = async (value: unknown): Promise<unknown> => {
       const walk = (v: unknown): unknown => {
         if (v === null || typeof v !== 'object') return v;
--- a/tests/unit/analyzer/diff-calculator.test.ts
+++ b/tests/unit/analyzer/diff-calculator.test.ts
@@ -1541,13 +1541,13 @@ describe('DiffCalculator - replacement propagation to dependents (issue #807)', 
       },
     };
 
-    // A MUTATING resolver, and deliberately a BLUNTER one than any real
-    // resolver: every object it descends into has each key written back in
-    // place (a node carrying `Fn::GetAtt` is terminal — replaced whole, its
-    // members never walked; arrays it rebuilds). So it stands for ANY
-    // `IntrinsicResolveFn`, not for the narrow `resolveSub` write-back
-    // go-to-k/cdkd#2764 retired — that generality is the comment's own
-    // argument, and it is what keeps this case discriminating.
+    // A MUTATING resolver, and deliberately not a faithful one: every object it
+    // descends into has each key written back in place (a node carrying
+    // `Fn::GetAtt` is terminal — replaced whole, its members never walked;
+    // arrays it rebuilds). That COVERS the narrow `resolveSub` write-back
+    // go-to-k/cdkd#2764 retired — the `Fn::Sub` variable map above is one of
+    // the objects it rewrites — and goes wider along the OBJECT axis. What it
+    // does NOT model is a resolver mutating array ELEMENTS in place.
     const mutatingResolver = async (value: unknown): Promise<unknown> => {
       const walk = (v: unknown): unknown => {
         if (v === null || typeof v !== 'object') return v;
--- a/tests/unit/analyzer/diff-calculator.test.ts
+++ b/tests/unit/analyzer/diff-calculator.test.ts
@@ -1511,9 +1511,13 @@ describe('DiffCalculator - replacement propagation to dependents (issue #807)', 
   });
 
   it('does NOT mutate the desired template (resolveBestEffort resolves a clone)', async () => {
-    // The intrinsic resolver mutates its input in place; resolveBestEffort must
-    // clone first so the raw Fn::GetAtt survives for the deploy phase to
-    // re-resolve against the in-flight (new) upstream value.
+    // resolveBestEffort is handed a leaf of the SHARED desired template, so it
+    // must clone before passing it to an `IntrinsicResolveFn` — whatever that
+    // function does with what it is given — and the raw Fn::GetAtt must survive
+    // for the deploy phase to re-resolve against the in-flight (new) upstream
+    // value. The resolver this case injects mutates in place; the real one no
+    // longer does (go-to-k/cdkd#2764 retired `resolveSub`'s write-back), which
+    // is why the mutation is injected rather than relied upon.
     const state = baseState();
     state.resources['Base'] = {
       physicalId: 'base',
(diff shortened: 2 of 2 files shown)


go-to-k:
Reviewed at `6d0d4bc6` against merge-base `27b5869c`. The direction is right and the mechanical claims in the body hold — I re-derived them rather than taking them:

- 0 non-comment changed lines (three-dot against the merge base).
- `#2764` (`f7937c6e`) really did retire the write-back: `resolveSub` now resolves into a fresh `Object.create(null)`, and the only surviving member assignment in `intrinsic-function-resolver.ts` (`:4873`) targets that fresh object. Both production `IntrinsicResolveFn`s are `resolver.resolve`.
- `#939` (`772dd03f`) added the `structuredClone` and the `rawGetAttRefs` note in one commit; its parent has the bare `resolveFn(value)`.
- Nothing in `diff-calculator.ts` assigns to `.Properties`, and the sharing is real — `deploy-engine.ts:2365` hands `calculateDiff` the same `effectiveTemplate` it re-resolves at `:2466` and provisions from at `:2701`.
- The verification grep and the docs/rules greps both return nothing; the change is outside every integ gate's scope.

One blocker, then four smaller items.

## M0 — `src/analyzer/diff-calculator.ts:167`: "It never has" is false, and it is this PR's own defect class (required before merge)

> (This said the loop "mutates in place" the desired property intrinsics. **It never has**: go-to-k/cdkd#939 added the clone in the same change that wrote the note, ...)

The loop did mutate the shared desired template, right up until the commit that wrote the note. Three facts, all at `772dd03f^`:

- `src/analyzer/diff-calculator.ts:347` — `resolved[key] = await resolveFn(value);`, no clone.
- `src/deployment/deploy-engine.ts:964` — `const diffResolveFn = (value) => this.resolver.resolve(value, diffResolverContext);`, so the loop ran the real resolver.
- `src/deployment/intrinsic-function-resolver.ts:1524-1528` — `[template, variables] = subArgs;` binds `variables` to the caller's map by destructuring, then `variables[key] = await this.resolveValue(val, context);` writes into it.

So a desired property holding `{"Fn::Sub": ["...", {V: {"Fn::GetAtt": [...]}}]}` had its variable map rewritten in place by the comparison loop. The old note was TRUE when it was written; `#939` falsified it in the same commit by adding the clone. The current text replaces "true-then, false-now" with "never", which is a new false claim — and it fails in the same way the PR is fixing: a reader who checks the history finds it wrong and stops trusting the rest of the paragraph.

Please make it temporal. Something like:

> It has not since go-to-k/cdkd#939 added the clone in the same change that wrote this note; before that commit the loop ran the real resolver over the template leaf itself, and `resolveSub` wrote its resolved values back into the caller's `Fn::Sub` variable map.

That also rescues the rest of the sentence, which currently leans on "#939 added the clone" as if it were the proof of "never" — it is the proof of the cutoff, not of the absence.

## m1 — `diff-calculator.ts:799`: fence the cross-module claim (recommended, same edit round)

> (It is not true that the resolver mutates its input today: ...)

This is prose asserting how `intrinsic-function-resolver.ts` behaves, from a file that cannot see it — structurally the same shape as the claim being retired here, and the reason it went stale in the first place. There is already a fence for exactly this: `tests/unit/deployment/intrinsic-sub-variables-not-mutated.test.ts` ("keeps the template object byte-identical across a resolution", "does not write a resolved intrinsic back into the variable map either"). Naming it makes the claim checkable instead of re-derivable, and gives the next `#2764` a test to red rather than a comment to silently outlive.

While there: the `src` comment never names the fence that reds if the clone is deleted (`does NOT mutate the desired template (resolveBestEffort resolves a clone)`). The PR's stated goal is to stop the clone reading as dead weight, and a named test that goes red is the strongest form of that. `src/analyzer/skipped-outputs.ts:145-151` — the sibling of the `outputs-diff.ts` precedent this PR cites — does exactly that, in both directions.

## m2 — `diff-calculator.ts:791-794`: the contract is declared caller-independent, then defined by one caller (optional)

> The reason is the OWNERSHIP contract, not any resolver's behaviour: this value is a leaf of the SHARED desired template, **which the deploy phase re-resolves later against the in-flight state** ...

`calculateDiff` has two callers. `diff-recursive.ts:590` (`cdkd diff`) passes the same `effectiveTemplate` object, shares it with the Outputs diff resolved through the same resolver, and with the skipped-outputs digests — and never reaches a deploy phase at all. A reader on that path finds the stated reason inapplicable and is back to asking whether the clone is load-bearing, one level up from where this PR is fixing it. Naming the sharing generically (both diff callers hand over a template other consumers still read) and keeping the deploy-phase re-resolution as *one* consequence would close it.

## m3 — `diff-calculator.ts:161-169`: the corrected premise leaves the snapshot unexplained (optional)

The rewritten note justifies `rawGetAttRefs` by "the comparison loop reads the RESOLVED values". But `promoteInPlaceAttributeDependents` already receives `desiredTemplate` (`:428`) and reads it (`:641`), so it could extract the refs itself from the raw properties. Under the old, false premise the snapshot was *necessary*; once the premise is corrected it is a precomputation — which is fine, and the "in one pass" phrasing hints at it, but the sentence that follows argues necessity. Saying plainly that the raw template is available either way and this is a single-pass precomputation would stop the next reader repeating this PR's exercise.

## m4 — two nits (optional)

- `:802-803` — "Kept as history so the clone is not deleted as dead weight, which is exactly what the contract forbids" restates `:791-798`, and "which" has no clear antecedent (the deletion? the write-through?).
- `tests/unit/analyzer/diff-calculator.test.ts:1542` — "it rewrites the Fn::Sub variable map's GetAtt in place" understates the injected resolver: it deep-mutates every object it walks, which is what makes it stand for ANY resolver rather than for the retired one. Worth saying, since that generality is the comment's own argument.

---

M0 is required before merge. m1 is worth folding into the same round while the file is open; m2-m4 are take-or-leave. I have not merged — ping me when M0 is pushed and I will re-review.

nix-tkobayashi:
Thanks — M0 is right, and I checked it before taking it rather than after. At `772dd03f^`: `diff-calculator.ts:347` is the bare `resolved[key] = await resolveFn(value)`, `deploy-engine.ts:964` passes `this.resolver.resolve`, and `resolveSub` at `intrinsic-function-resolver.ts:1524-1528` binds `variables` to the caller's map by destructuring and writes into it. So the note was true when written and go-to-k/cdkd#939 falsified it in the same commit. Worse for my side: "never" was the SECOND version of that sentence — the first said both halves were false, I narrowed it to one half in `bdb9bd39`, and narrowing is what turned "true-then, false-now" into "never". Two revisions of a history claim in one PR whose subject is a stale history claim.

Pushed as `9563177c`. All five items taken.

**M0** — the sentence is temporal now and says what the pre-#939 loop did:

> (This note used to say the loop "mutates in place" the desired property intrinsics. It has not since go-to-k/cdkd#939 added the clone in the same change that wrote the note; BEFORE that commit the loop ran the real resolver over the template leaf itself, and `resolveSub` wrote its resolved values back into the caller's `Fn::Sub` variable map.)

**m1** — both fences are named, and I probed the one I was about to assert rather than citing it: reinstating the write-back in `resolveSub` (resolve into `variableMap` instead of the fresh `variables`) reds exactly the two cases the comment names — `keeps the template object byte-identical across a resolution` and `does not write a resolved intrinsic back into the variable map either`, `2 failed | 5 passed (7)`. Restored byte-exact, `git diff` clean. The clone side keeps its own probe from the PR body (`1 failed | 34 passed (35)`). And I re-derived both against `origin/main` as it stands after go-to-k/cdkd#2797 merged, since that PR moved `intrinsic-function-resolver.ts`: `resolveSub` still resolves into the fresh map, and both cases are still there.

**m2** — the rationale no longer belongs to the deploy caller. It now says both call sites share the object, names what `cdkd diff` hands it to next, and keeps the deploy re-resolution as the consequence with teeth. One correction to your description while I was there: `diff-recursive.ts` shares the template with the Outputs resolution (`resolveTemplateOutputs(effectiveTemplate, ...)`), but NOT with the skipped-output digests — `outputsDigestSource` is a `structuredClone` (`:404`). I dropped that half rather than repeat it.

**m3** — the snapshot is called a precomputation, with the reason: `promoteInPlaceAttributeDependents` also receives `desiredTemplate` and could re-extract, because the raw template survives the loop. I left out any claim about it being cheaper — I could not derive one (both shapes are a single pass over the same resources), and this PR is a poor place to ship an unmeasured clause.

**m4** — the restating clause is gone. The test comment now describes the injected resolver as it actually walks: each key of every object it descends into is written back in place, a node carrying `Fn::GetAtt` is terminal and replaced whole, arrays are rebuilt. That last detail came out of the Codex round on this delta, which called my first draft ("writes every OBJECT member back in place") an overstatement — correctly, since a `Fn::GetAtt` node's members are never walked.

Verification for the round: full suite 968 files / 21109 passed / 1 skipped, rc=0, run root asserted; `vp check --fix` + `vp run check` + `typecheck:test` + `build` + `gen:all-matrices` + `audit:coverage:check` clean; the issue's verification grep and the docs/rules greps still return nothing; still 0 non-comment changed lines. Codex on the delta: 2 rounds plus a maintainer-checklist proxy pass, NO ISSUES.

`main` moved by two commits (go-to-k/cdkd#2797, go-to-k/cdkd#2898) while this sat. I did not rebase: neither touches `diff-calculator.ts` or its test, GitHub still reports the PR mergeable, and the three cross-module facts the comments assert I re-checked against `origin/main` rather than my merge base. Say the word if you would rather have it rebased anyway.

go-to-k:
Round 2 on `9563177c`. **M0 is fixed and I verified the fix at both ends of the boundary, not just at HEAD** — `772dd03f^` has the bare `resolveFn(value)`, `deploy-engine.ts:964` passed the real resolver over `effectiveTemplate`, and `resolveSub` at `intrinsic-function-resolver.ts:1524-1529` wrote into the caller's map by identity. The note is authored by `772dd03f`, not moved into it, and the clone was never removed and re-added, so "has not since #939" is exact. m1-m4 all landed; the two named fences exist and pin what they are cited for, and deleting the `structuredClone` does red `does NOT mutate the desired template (resolveBestEffort resolves a clone)`.

Your correction on m2 is right and mine was wrong: `outputsDigestSource` at `diff-recursive.ts:404` is a `structuredClone`, so the skipped-output digests never shared the object. Dropping that half rather than repeating it was the correct call.

Two items left. Both are one-line, both are this PR's own defect class — a sentence that does not survive a reader checking it — which is why I am asking rather than filing them.

## M1 — `src/analyzer/diff-calculator.ts:167-168`: the parenthetical does not cover the `catch` arm (required before merge)

> since the raw template survives the loop (`resolveBestEffort` clones, **and collects into a fresh bag**)

The conclusion is true, and I confirmed it. The two cited mechanisms are not why it is true for every key. Eleven lines below, `:816` is `resolved[key] = value` — on the `catch` path the raw template leaf goes into the "fresh bag" **by identity**, so for those keys there is no clone and the bag is not fresh in the sense the word is doing work here. Survival there rests on a third condition the sentence does not name: nothing downstream mutates the bag — `withoutAcceptedSilentDropProperties`, `canonicalizeProperties`, `compareProperties`. That condition holds today (I walked all four live `canonicalizeDesiredProperties` implementations and `property-coverage.ts:183`; every one builds a fresh object), but it is an unfenced cross-module assumption of exactly the kind this same delta added fences for at `:809`.

A reader who checks this parenthetical finds a case it does not explain, which is the failure mode the PR exists to close. Naming the actual reason fixes it:

> since the raw template survives the loop — the `try` arm clones, and the `catch` arm only aliases the leaf into a bag no consumer mutates.

## M2 — `tests/unit/analyzer/diff-calculator.test.ts:1547-1548`: "stands for ANY" does not follow from the premises listed (required before merge)

> (a node carrying `Fn::GetAtt` is terminal — replaced whole, its members never walked; **arrays it rebuilds**). **So it stands for ANY `IntrinsicResolveFn`**

The description itself is line-accurate — I checked it against `:1552-1560` clause by clause. The inference is not. The parenthetical lists two shapes the injected resolver does *not* model, and then concludes it covers all of them. `if (Array.isArray(v)) return v.map(walk);` rebuilds, so a resolver that mutates array elements in place (`v[i] = ...`) is outside what this case exercises — and `Fn::Sub`'s variable map arrives as `subArgs[1]`, an element of an array, so that is not a hypothetical axis.

This is the same shape as M0: an unqualified universal standing where a bounded claim is what the evidence supports. The bounded version is also the stronger argument for keeping the resolver as-is:

> …arrays it rebuilds). Blunter than `resolveSub` along the OBJECT axis, which is what makes this case discriminating — it does not stand in for the narrow write-back go-to-k/cdkd#2764 retired.

## Not asking for a change — `diff-calculator.ts:798-799`

For the record, so it is not re-raised later. The reviewer on this round flagged that `cdkd diff` "hands the same object on to the Outputs resolution" is sharing without a reachable write-through: `resolveTemplateOutputs` reads `template.Outputs`, and the resolver touches `context.template.Resources` only to test key presence (`intrinsic-function-resolver.ts:4762`), never a resource's `Properties`. That is accurate, but the sentence is stating what THIS function may not do with an object its caller still holds, not forecasting what that particular consumer would do with a write — and the text already ranks the two ("the deploy one is the consequence with teeth"). It came out of my m2, so I am closing it myself rather than leaving you to guess: leave it as written.

---

M1 and M2 are required. Nothing else is outstanding — CI is green, the full suite passes here (968 files / 21109 passed / 1 skipped, rc=0), `gen:all-matrices` and `audit:coverage:check` are clean, and the diff is outside every integ gate's scope. Push those two and I will merge.

nix-tkobayashi:
Both taken, pushed as `4a592670`. M1 verbatim; M2 with one deliberate deviation, and the deviation is the interesting part.

**M1** — the parenthetical now names the real reason:

> since the raw template survives the loop: the `try` arm clones, and the `catch` arm only aliases the leaf into a bag no consumer mutates.

I re-walked the third condition rather than inheriting it: `withoutAcceptedSilentDropProperties` (`property-coverage.ts`) returns its input untouched when nothing is removable and otherwise spreads into a new object, and all four live `canonicalizeDesiredProperties` implementations (cognito / ec2 / kinesis / s3-bucket) either return `properties` unchanged or build a fresh one. Agreed that it is an unfenced cross-module assumption; the sentence now says what it rests on instead of implying the clone covers it.

**M2 — your finding is right and your replacement overshoots, so I took the finding and not the wording.** "So it stands for ANY `IntrinsicResolveFn`" was unearned, for exactly the reason you give. But `it does not stand in for the narrow write-back go-to-k/cdkd#2764 retired` is false in the other direction: the injected resolver DOES perform that write-back. `derivedValue` is `{'Fn::Sub': ['hello-${V}', {V: {'Fn::GetAtt': [...]}}]}`, and the walk reaches `{V: ...}` as an object and rewrites `V` in place — the retired shape exactly. What it adds on top is wider object mutation, not a replacement for it.

Probed rather than argued. I swapped the injected resolver for one doing ONLY the narrow write-back (resolve the `Fn::Sub` variable map's members in place, return the substituted string, touch nothing else):

| probe | clone | result |
| --- | --- | --- |
| narrow write-back only | present | `35 passed (35)` |
| narrow write-back only | removed | `1 failed \| 34 passed (35)`, the failure being `does NOT mutate the desired template` |

So the narrow shape alone is what discriminates, and the broader mutation is not carrying the case. Both files restored byte-exact afterwards. The text now reads:

> arrays it rebuilds). That COVERS the narrow `resolveSub` write-back go-to-k/cdkd#2764 retired — the `Fn::Sub` variable map above is one of the objects it rewrites — and goes wider along the OBJECT axis. What it does NOT model is a resolver mutating array ELEMENTS in place.

That keeps your bound (the array axis is named as uncovered) without the claim the probe contradicts. Say the word if you would rather have your wording as written and I will take it.

**`diff-calculator.ts:798-799`** — left as written, per your ruling. I checked your reviewer's fact rather than just accepting the closure, and would narrow it by one step: the Outputs path does reach past `Outputs` — `containsSecretDynamicReference(template)` at `outputs-diff.ts:384` walks the WHOLE template, resource properties included. Every touch on that path is a read, though (`:261-265` is `Array.some` / `Object.values`), the resolver's `context.template.Resources` access is an `Object.hasOwn` presence test, and there are zero writes through `context.template` anywhere in the resolver. So the no-reachable-write-through conclusion holds, by read-only-ness rather than by narrow reach. So the observation holds, and it is also why the sentence should stay — it states what this function may not do with an object its caller still holds, and the reviewer's fact is a snapshot of what one consumer does today, which is the kind of thing this PR exists to stop comments from resting on.

Verification for the round: full suite 968 files / 21109 passed / 1 skipped, rc=0, run root asserted; `vp check --fix` + `vp run check` + `typecheck:test` + `build` + `gen:all-matrices` + `audit:coverage:check` all rc=0; still 0 non-comment changed lines; the issue's verification grep and the docs/rules greps still return nothing. Codex on the delta: NO ISSUES. The maintainer-checklist proxy pass is what caught the M2 overshoot above — it returned NO ISSUES after the correction.

(The sha moved from `f7b00c83` to `4a592670` — a Codex round on this reply caught the new comment saying the variable map is "below" when `derivedValue` is defined above it. Amended rather than stacked, since you had not started reading yet; gates and full suite re-run on the amended tree.)

`main` has moved four commits now (go-to-k/cdkd#2797, go-to-k/cdkd#2898, go-to-k/cdkd#2892, go-to-k/cdkd#2882); none touches `diff-calculator.ts` or its test, and GitHub still reports the PR mergeable, so I have not rebased.

go-to-k:
Merged. Thanks — three rounds and every one of them made the text more checkable than the round before.

Two things stood out. On M2 you did not take my suggested wording, and you were right not to: I had written "it does not stand in for the narrow write-back #2764 retired", which is false — the injected resolver rewrites the `Fn::Sub` variable map as one of its objects, so it covers that write-back and goes wider along the object axis. I replayed the walk over the fixture's graph to confirm before merging: the variable map keeps its identity and comes back `{"V":"world"}`, while arrays come back as new objects. Your "COVERS ... and goes wider ... what it does NOT model is array ELEMENTS" is the accurate version of the bound I was asking for, and it says more than what I asked for would have. Catching `below` -> `above` on the amend was the same instinct applied to your own sentence.

And the whole PR turned out to be a good demonstration of its own thesis. M0 landed because a claim about the past was checked only at HEAD; every subsequent round was a sentence that read fine until someone followed it to the code. That is the failure the clone's comment existed to prevent, reproduced four times in the process of fixing it — which is the strongest argument the file could have for why the rationale, not just the guard, has to survive review.

Verified before merge, at `4a592670` in a worktree at that sha: full suite 968 files / 21109 passed / 1 skipped, rc=0; typecheck, lint, build clean; `gen:all-matrices` + `audit:coverage:check` clean with nothing regenerated dirty; 0 non-comment changed lines against the merge base; the issue's verification grep and the docs/rules greps both empty; CI green; outside every integ gate's scope.

Issue #2904, opened 2026-09-10
test(local): the websocket drain's completion arm is never executed, and the leaked-count assertion cannot separate its two positions


nix-tkobayashi:
Found while reviewing go-to-k/cdkd#2795's fix. Two coverage gaps in the same describe block, filed together because both need the same missing fixture capability. Neither is a production defect: `src/local/websocket-server.ts`'s drain behaves correctly today.

## 1. The leaked-handler COUNT assertion cannot separate its two positions

`tests/unit/local/websocket-server.test.ts`'s bounded-drain case asserts the warn contains `1/1 $disconnect handlers`. The message is built as `${inFlightDisconnects.size}/${drainStartCount}`, and the fixture opens ONE connection, so numerator and denominator are both 1. Swapping the two interpolations still passes, and so does hardcoding `1/1`. What the assertion pins is the magnitude, not the positions.

Producing `1/2` needs TWO handlers pending at the moment `drainStartCount` is snapshotted — which is after `await Promise.all(closes)` — and then one of them settling before the ceiling fires. They need not be dispatched BY that call: the socket close listener adds every `$disconnect` promise to `inFlightDisconnects`, so a client-initiated disconnect dispatched earlier and still pending counts too. What matters is that both are pending at the same snapshot.

## 2. The drain's COMPLETION arm is executed incidentally and never asserted

Nothing asserts that a well-behaved `$disconnect` lets `close()` return through the drain's COMPLETION arm.

**Correction, and it matters for the plan below.** An earlier version of this issue said the arm is never EXECUTED, on the strength of dropping `drainComplete.then(() => 'complete' as const)` from the `Promise.race` and seeing the file stay green. That is not a reliable signal, and the underlying reason it stayed green was wrong.

`inFlightDisconnects.add(disconnectPromise)` runs in the connection-time `close` listener, which is installed on ADMISSION. So it fires for every close of an admitted connection regardless of whether a `$disconnect` route is configured — not for a denied `$connect`, and not for a client that closes during the `$connect` await, both of which the suite covers separately and neither of which registers anything. For an admitted connection the promise may therefore still be pending when `close()` reaches `if (inFlightDisconnects.size > 0)`, and whether it is depends purely on scheduling. Measured both ways on the same tip with only that arm removed:

- idle host: 18/18 green, the other 17 cases at 2-166 ms and only the drain case at 5003 ms; still green at load average 25;
[...]

Issue #527, opened 2026-05-23
WebSocket #524: MAJOR robustness follow-ups (serialization / docker portability / shutdown drain)


go-to-k:
## WebSocket #524 — MAJOR robustness follow-ups

Surfaced by `/review-pr` 3-axis review on [#524](https://github.com/go-to-k/cdkd/pull/524) — see [the review comment](https://github.com/go-to-k/cdkd/pull/524#issuecomment-4523980689). Companion to #526 (BLOCKERS bundle).

These are robustness / portability MAJORs — significant production-shape gaps but NOT pre-merge gates. Can ship #524 with #526 fixed; address these as fast-followups.

### M1 — Per-connection message dispatch is not serialized

[src/local/websocket-server.ts:199-225](https://github.com/go-to-k/cdkd/blob/feat/local-start-api-websocket/src/local/websocket-server.ts#L199). `'message'` listener fires `void dispatchMessage(...)` per frame WITHOUT chaining. Client sends 3 messages back-to-back → 3 concurrent `pool.acquire(lambdaLogicalId)` calls race the warm-pool entry. AWS WebSocket semantics SERIALIZE handler invocations per connection; cdkd does not.

Silent ordering bug — invisible until the user hits "why is my chat-app conversation history out of order locally but in order on AWS?"

**Fix**: per-connection promise chain
```ts
const chainByConnection: Map<string, Promise<void>> = new Map();
// inside 'message' listener:
const prev = chainByConnection.get(connectionId) ?? Promise.resolve();
const next = prev.then(() => dispatchMessage(...)).catch(reportFailure);
chainByConnection.set(connectionId, next.finally(() => {
  if (chainByConnection.get(connectionId) === next) chainByConnection.delete(connectionId);
}));
```

Across-connection dispatch stays parallel.

### M2 — `host.docker.internal` hardcoded for the mgmt endpoint

[src/cli/commands/local-start-api.ts:765-792](https://github.com/go-to-k/cdkd/blob/feat/local-start-api-websocket/src/cli/commands/local-start-api.ts#L765). `AWS_ENDPOINT_URL_APIGATEWAYMANAGEMENTAPI` env injected as `http://host.docker.internal:<port>`, with `--add-host=host.docker.internal:host-gateway` mapping. Three sub-issues:

[...]

PR #2905, opened 2026-09-10
test(local): pin the websocket drain against its own mechanism, not the wall clock

### LINKED ISSUE #2795: test(local): the websocket bounded-drain case asserts a wall-clock ceiling that fails under concurrent suites
## Summary

`tests/unit/local/websocket-server.test.ts`'s `graceful shutdown bounded drain (M3) > logs a warn naming the leaking $disconnect count when drain times out` asserts a WALL-CLOCK window around the 5000 ms drain ceiling:

```ts
const elapsed = Date.now() - t0;
// Drain ceiling = 5000ms; some scheduler jitter accepted.
expect(elapsed).toBeGreaterThanOrEqual(4_500);
expect(elapsed).toBeLessThan(7_500);
```

The upper bound is what failed, twice in one session, in two separate full-suite runs on the same tree while another lane's suite ran on the same host:

| run | elapsed | bound |
| --- | --- | --- |
| 1 | 7751 ms | 7500 ms |
| 2 | 7815 ms | 7500 ms |

Both times the file passed on its own immediately afterwards: three consecutive `vp test run tests/unit/local/websocket-server.test.ts`, 18/18 each. The tree under test touched neither `src/local/**` nor this file.

## What it costs

The margin is 2500 ms over a 5000 ms ceiling. Under concurrent suites on this host that margin was not enough twice, and each failure costs a full-suite re-run, which in this repo is 6 to 11 minutes. The failure also reads as a `$disconnect` drain defect when what actually happened is that the elapsed measurement overran; the two assertions the case exists for -- exactly one `graceful shutdown drained for` warn, carrying `still in flight` -- are never reached, because the elapsed assertions precede them.

Those two assertions do not reference `elapsed`, but they are not free of timing either: the warn is emitted only because the drain ceiling wins against the mocked handler's 10,000 ms delay.

## Possible directions, for the owner to weigh

- Loosen or drop the upper bound. The case carries a 10,000 ms per-case timeout, which would still fail an asynchronous shutdown that never resolves, though it does not preserve the 7500 ms limit as a behavioural statement and cannot interrupt a blocked event loop.
- Drive the clock with `vi.useFakeTimers()`. The drain's timer is a reachable global `setTimeout`, but this case also does real HTTP / WebSocket I/O and waits on socket-close events before that timer is registered, so this needs careful synchronisation rather than a plain advance.

Either way the question is whether the upper bound is meant as a behavioural limit or as a hang guard, which is the owner's call.

## Classification

Session-fit: next (not this session) -- found while verifying an unrelated resolver change (go-to-k/cdkd#2563); the fix is in a `tests/unit/local` file that branch does not touch. Start with: `/work-issues <this issue number>`.
Severity: low -- no user-facing defect; the cost is a false red and a full-suite re-run when another suite shares the host.
Effort: small (S) -- one assertion, or one case rewritten onto fake timers; verified by `vp test run tests/unit/local/websocket-server.test.ts` plus a full `vp test run`, with no integ.
Estimate: ~30-60 min -- deciding whether the bound is a behavioural limit or a hang guard, and, if fake timers are chosen, sequencing them against the socket-close waits.
Dup-check: searched issues for `websocket drain flake`, `websocket-server` and open issues mentioning `flake`; go-to-k/cdkd#1843 is about `local start-api` region chains and unrelated. Nothing covers this case.

nix-tkobayashi:
Closes #2795

## What was wrong

The bounded-drain case asserted a wall-clock window around the server's 5 s `SHUTDOWN_DRAIN_MS`:

```ts
expect(elapsed).toBeGreaterThanOrEqual(4_500);
expect(elapsed).toBeLessThan(7_500);
```

The upper bound went red twice under a concurrent suite with no defect present, at 7751 ms and 7815 ms, each costing a full-suite re-run of 6 to 11 minutes. It also misreports: the two assertions the case exists for sit AFTER it and never run, so the failure reads as a `$disconnect` drain defect when what overran was the measurement.

## The decision the issue left open

Whether that bound is a behavioural limit or a hang guard. **It is a behavioural limit, and a wall-clock constant is the wrong instrument for one.** A host stall inflates `elapsed` while the constant stays put, so the comparison moves on one side only. Three assertions replace it, each reading the property off the mechanism instead.

- **The drain gave up before the handler.** The hung `$disconnect` records when it settles; the drain must return first. `close()` races `Promise.allSettled` of the in-flight handlers against the ceiling timer, so "the drain gave up" IS "that `allSettled` had not settled".
- **The ceiling CONSTANT is still 5 s.** Pinned off the warn, which interpolates `SHUTDOWN_DRAIN_MS`, so moving it to 8 s rewrites the string where the settled flag sees nothing.
- **The leaked-handler COUNT the case is named for.** It was named and not pinned, so a warn reporting the wrong number satisfied every other line. New coverage rather than a restatement. It pins the MAGNITUDE and not the two positions: with one connection the numerator and denominator are both 1, so a swap would still pass. Separating them needs a second handler pending at the drain's count snapshot, released on a signal rather than a sleep, so it is filed as go-to-k/cdkd#2904 rather than built here on a timing window.

`elapsed >= 4_500` stays. A lower bound cannot fail from a slow host, only from one running faster than real time, so it was never part of this defect.

The hung sleep and the per-case timeout become named constants, and the timeout goes `10_000` to `20_000` so the ordering is timeout > handler > ceiling. A drain that lost its ceiling is then reported by an assertion rather than killed by the runner at a deadline the handler was still racing.

## Verification

Every mutant restored from a byte-exact copy. Each is the SOLE failing assertion for its row.

| Mutation | Red on | Previously caught by |
|---|---|---|
| warns at the ceiling, then still blocks on the handler | the settled flag | the removed bound |
| ceiling constant `5_000` to `8_000` | `drained for 5000ms` | the removed bound |
| leaked-handler count reported wrong | `1/1 $disconnect handlers` | nothing |

Deleting the settled-flag assertion lets its own mutant pass 18/18, so it is load-bearing rather than decorative.

**The reported failure, reproduced and compared side by side.** Under a deliberate 16-way CPU load, one run carrying BOTH the old bound and the new assertions red the OLD bound at `elapsed=7907` while the new ones passed in the same run, on the same host, in the same process. The fixed case then went 5/5 at load average 35. That is the closest thing to a controlled comparison this defect allows, and it is also the live test of the changed behaviour: the behaviour here IS the case's response to load.

Full `vp test run`, vitest root asserted to this worktree: 969 files / 21,135 passed / 1 skipped, clean. `vp check --fix` / `run check` / `typecheck` / `typecheck:test` / `build` / `gen:all-matrices` / `audit:coverage:check`: 0 errors, nothing regenerated dirty.

No integ, and the gates agree: the diff is `tests/unit/local/websocket-server.test.ts` alone with zero `src/` changes, so the cross-cutting, deletion and local-execution detections in `/verify-pr` step 6 all report not-touched. `tests/unit/local/**` is not in the `integ-local` scope, which is `src/local/**`, `src/cli/commands/local-*.ts` and `tests/integration/local-*/**`. No changelog entry, under the rule that only a user-visible behaviour delta writes one.

## What review changed, since it was not cosmetic

- The first cut asserted only the settled flag. That WEAKENED the case: a ceiling regressed from 5 s to 8 s left the flag false, the lower bound passing and the warn assertions indifferent, where the removed bound caught it. Hence the constant is pinned separately.
- The second cut claimed pinning the constant recovered "the one thing" the old bound caught. It does not. A drain scheduling a delay other than the constant it logs waits 8 s and passes everything.
- The residual then named only that one shape. The honest statement is the class: **nothing here bounds how long `close()` takes any more**, so any added latency under the handler's 10 s is invisible. The removed bound did not see all of that either, only the part pushing the total past its own 7500, so an extra 1 s passed it too.
- The same-loop argument is recorded as MEASURED resilience, not as an invariant. The two timers are armed at different moments, the handler's when the close listener dispatches `$disconnect` and the drain's after the socket closes settle, so a stall between those moments does move their deadlines relative to each other. What makes it hold is the ceiling-to-handler margin, `HUNG_DISCONNECT_MS - (socket-close phase + SHUTDOWN_DRAIN_MS)`, where the removed bound had 2.5 s against a fixed constant. (Corrected after merge: this line still carried the pre-resize `10 s` figure. It never reached `main` — the commit message uses the derivation form — and the maintainer had already noted it and merged over it rather than spend a round on it.)

All four are written into the case rather than left here, so the next reader of that file gets them without finding this PR.

An independent test-adequacy reviewer then found four more, and its measurement overturned one of my own claims. The residual said an extra await after the race was invisible; awaiting `drainComplete` is in fact CAUGHT, because the mock sets its flag before returning, so the flag is already true when `allSettled` resolves. It also showed the residual contradicted itself (the settled flag IS a duration bound, at the handler's 10 s rather than at 7500), that the slack is about 5 s and not 10, and that the comment describing `ws.close()` as firing `$disconnect` was wrong — the dispatch happens inside `close()`'s own socket-close loop, one loopback round-trip later, which makes the margin argument stronger than I had sold it. All corrected in the case.

## Filed, not fixed here

go-to-k/cdkd#2904 covers two coverage gaps the same reviewer found, bundled because both need one missing fixture mechanism: the degenerate `1/1` count above, and the drain's COMPLETION arm, which is never asserted.

**Corrected after review.** I first published that the completion arm "is never executed", because dropping `drainComplete.then(() => 'complete')` from the race left the file green here. That was a claim about one host stated as a property of the suite. The close listener that registers a disconnect promise is installed on ADMISSION, so it fires for every close of an admitted connection whether or not a `$disconnect` route is configured, and such a promise may still be pending when `close()` reaches the drain. Whether it is depends purely on scheduling: idle here the mutant is green, with the other 17 cases at 2-166 ms and only the drain case at 5003 ms, and still green at load average 25; on the maintainer's run it was `7 failed | 11 passed`, all `Test timed out in 5000ms`. The arm is therefore reached unpredictably and simply never asserted, and that mutant's colour is not a usable baseline. The commit message and go-to-k/cdkd#2904 both carry the corrected premise, the issue's verification plan no longer rests on it, and the fix is required to red BY NAME rather than by a runner timeout.

Pre-existing, no user-visible defect, `severity:low` / `effort:medium`.

### DIFF
--- a/tests/unit/local/websocket-server.test.ts
+++ b/tests/unit/local/websocket-server.test.ts
@@ -782,21 +782,52 @@ describe('per-connection dispatch serialization (M1)', () => {
 
 // Issue #527 M3: graceful shutdown drains in-flight $disconnect
 // dispatches with a bounded ceiling. A hung handler pre-fix would leak
-// its rieTimeoutMs (60s+) past close()'s 5s socket-close timeout.
+// its rieTimeoutMs (60s+) past close()'s 5s socket-close timeout. Note that
+// `close()` has TWO 5s timers and they are unrelated: the socket-close
+// path's defensive per-socket timeout, and `SHUTDOWN_DRAIN_MS`, the drain
+// ceiling this block is about. Everything below turns on the second.
 describe('graceful shutdown bounded drain (M3)', () => {
+  // The hung `$disconnect` outlasts `SHUTDOWN_DRAIN_MS`, and the per-case
+  // timeout outlasts BOTH. That ordering is what lets an assertion do the
+  // reporting: a drain that lost its ceiling finishes when this handler
+  // does, and is named rather than killed by the runner at a deadline the
+  // handler was still racing.
+  //
+  // The socket-close phase runs BEFORE the drain and carries its own 5s
+  // timer, so the effective margin is
+  // `HUNG_DISCONNECT_MS - (that phase + SHUTDOWN_DRAIN_MS)` -- about 5s at
+  // these values once a slow phase is allowed for, against the 2.5s the
+  // removed wall-clock bound had.
+  //
+  // What it is bought with, scoped to the shape that pays it: a mutant that
+  // leaves the drain BLOCKED on the hung handler cannot report until that
+  // handler settles, so its failure arrives at roughly `HUNG_DISCONNECT_MS`.
+  // Widening this constant delays that one report by the same amount. The
+  // other mutants probed against this case fail at their own ceilings
+  // instead, and no cost lands on the passing path at all, which never
+  // awaits this handler.
+  const HUNG_DISCONNECT_MS = 15_000;
+  const CASE_TIMEOUT_MS = 25_000;
+
   beforeEach(() => {
     rieModule.__resetQueue();
     rieModule.invokeRie.mockClear();
   });
   it('logs a warn naming the leaking $disconnect count when drain times out', async () => {
     rieModule.__resetQueue();
-    rieModule.__queueInvokeResult({ statusCode: 200 }); // $connect for the live conn
+    // Flips only when the hung `$disconnect` finally returns — the thing the
+    // drain must NOT wait for. `close()` races `Promise.allSettled` of the
+    // in-flight handlers against a `SHUTDOWN_DRAIN_MS` timer, so "the drain
+    // gave up" IS "that allSettled had not settled", which this observes
+    // directly.
+    let disconnectSettled = false;
     // $disconnect hangs longer than the 5s drain window.
     rieModule.invokeRie.mockImplementation(async (_id: string, _meta: unknown, event: any) => {
       const routeKey = event?.requestContext?.routeKey;
       if (routeKey === '$connect') return { payload: { statusCode: 200 }, raw: '{}' };
       // Sleep past the drain window — exact value doesn't matter past 5s.
-      await new Promise((r) => setTimeout(r, 10_000).unref?.());
+      await new Promise((r) => setTimeout(r, HUNG_DISCONNECT_MS).unref?.());
+      disconnectSettled = true;
       return { payload: {}, raw: '{}' };
     });
 
@@ -821,22 +852,109 @@ describe('graceful shutdown bounded drain (M3)', () => {
     try {
       const ws = await openWebSocket(port, '/prod');
       await waitFor(() => attached.registry.size() === 1, 1000);
-      // Closing the WebSocket fires the close listener, which kicks off
-      // $disconnect. Then immediately call attached.close() which awaits
-      // the drain. The drain hits the 5s ceiling and the warn fires.
-      // We do NOT await ws's close event here because attached.close
-      // itself drives the close to completion.
+      // `ws.close()` only starts the handshake -- the close frame is an
+      // async write, so the server-side listener that dispatches
+      // `$disconnect` runs INSIDE `attached.close()`'s socket-close loop,
+      // one loopback round-trip later. We do not await ws's close event
+      // here because `attached.close()` drives it to completion. The drain
+      // then hits the ceiling and the warn fires.
+      //
+      // Calling `attached.close()` in the SAME tick is load-bearing twice
+      // over: `close()` snapshots `wss.clients` synchronously, and calling
+      // it synchronously is what GUARANTEES that snapshot is non-empty --
+      // sufficient rather than necessary, since an intervening await need
+      // not let the socket actually close; and the `$disconnect` promise is added to
+      // `inFlightDisconnects` synchronously inside the connection-time
+      // `close` listener, which was registered BEFORE `close()`'s own
+      // `ws.once('close', ...)`, so registration always wins the race
+      // against `Promise.all(closes)` rather than winning it by timing.
       ws.close();
       const t0 = Date.now();
       await attached.close();
       const elapsed = Date.now() - t0;
-      // Drain ceiling = 5000ms; some scheduler jitter accepted.
+      // The drain really waited rather than returning at once. Kept as a
+      // wall-clock bound because a LOWER one cannot fail from a slow host,
+      // only from one running faster than real time.
+      //
+      // It is also what stops the next assertion passing VACUOUSLY. If no
+      // `$disconnect` ever registered, `inFlightDisconnects` would be empty,
+      // `close()` would skip the drain block entirely and return in ~5 ms
+      // with `disconnectSettled` still false -- which reads identical to a
+      // correct give-up. This line is the guard that fails there, so the
+      // two are a pair rather than one bound plus a leftover.
       expect(elapsed).toBeGreaterThanOrEqual(4_500);
-      expect(elapsed).toBeLessThan(7_500);
+      // ...and it gave up before the handler. This and the `drained for
+      // 5000ms` assertion below together replace an `elapsed < 7_500` upper
+      // bound that went red twice under a concurrent suite with no defect
+      // present (issue #2795): a host stall inflates `elapsed` against a
+      // fixed constant, where this compares two timers that a stall delays
+      // together. Measured, not assumed to be absolute, across separate
+      // runs: the old bound red at 7751 / 7815 ms when the issue was filed,
+      // and again at 7907 ms in a run carrying BOTH bounds where the new
+      // assertions passed; the fixed case then went 5/5 at load average 35
+      // on 16 cores. It is not an INVARIANT -- the two timers are armed at different
+      // moments (the handler's when the close listener dispatches
+      // `$disconnect`, the drain's after the socket closes settle), so a
+      // stall between those two moments does move their deadlines relative
+      // to each other. What makes it hold in practice is the margin between
+      // the ceiling and the handler, where the removed bound had 2.5s
+      // against a fixed constant.
+      //
+      // This half alone would NOT: it only says the drain quit somewhere
+      // short of `HUNG_DISCONNECT_MS`, so a ceiling raised to 8s still
+      // passes it -- and the wider that constant is set, the wider this
+      // undisclosed window gets, which is the cost side of the sizing above.
+      // The removed bound did catch that, which is why the ceiling's VALUE
+      // is pinned below rather than left to this.
+      //
+      // What the trio does NOT recover: `close()` is still bounded, but at
+      // the handler's settle point rather than at 7500, so latency added between
+      // those two figures is invisible where the removed bound saw it. That
+      // is UNRELATED latency -- awaiting `drainComplete` itself is caught
+      // (measured), because the mock sets the flag before returning, so it
+      // is already true when `allSettled` resolves. A drain scheduling a
+      // delay other than the constant it logs is the reachable case. The
+      // removed bound did not see all of that class either, only the part
+      // pushing the total past its own 7500, so an extra 1s passed it too;
+      // what it had was 2.5s of slack where this has the ceiling-to-handler
+      // margin. Narrowing
+      // that needs a duration bound, the instrument this case drops on
+      // purpose, having measured it red twice with nothing wrong. A
+      // `setTimeout` spy is not a way back: the socket-close path above
+      // schedules its own 5s timer, so the delays are not separable by
+      // value, and separating them by count or order would pin the socket
+      // loop's shape instead of the drain's.
+      // The message says only what the flag OBSERVES. `disconnectSettled`
+      // being true means the handler settled before `close()` RETURNED, and
+      // `close()` still awaits `wss.close()` after the drain, so a long
+      // enough stall there flips it without the drain having waited at all.
+      // Blaming the drain in the message would be the same misreporting
+      // this case was rewritten to stop doing.
+      expect(
+        disconnectSettled,
+        'the in-flight $disconnect settled before close() returned'
+      ).toBe(false);
       const drainWarns = warnSpy.mock.calls.filter(
         (c) => typeof c[0] === 'string' && c[0].includes('graceful shutdown drained for')
       );
       expect(drainWarns).toHaveLength(1);
+      // The ceiling CONSTANT, pinned off the message rather than the clock:
+      // the warn interpolates `SHUTDOWN_DRAIN_MS`, so moving it to 8s
+      // rewrites this string where `disconnectSettled` sees nothing. It pins
+      // the constant, NOT the delay actually scheduled -- see the residual
+      // noted above.
+      expect(drainWarns[0]![0]).toContain('drained for 5000ms');
+      // The COUNT this case is named for. It was named and not asserted, so
+      // a warn reporting the wrong number of leaked handlers satisfied every
+      // other line here. What it pins is the MAGNITUDE and not the two
+      // POSITIONS: one connection makes numerator and denominator both 1, so
+      // swapping them in the message would still pass. Separating them needs
+      // a second connection whose `$disconnect` settles after the
+      // socket-close loop but before the drain snapshots its count -- and
+      // the width of that window is the load-dependent quantity this case is
+      // being rewritten to stop depending on. Left as the magnitude check
+      // deliberately; the stronger fixture is filed instead.
+      expect(drainWarns[0]![0]).toContain('1/1 $disconnect handlers');
       expect(drainWarns[0]![0]).toContain('still in flight');
     } finally {
       warnSpy.mockRestore();
@@ -846,5 +964,5 @@ describe('graceful shutdown bounded drain (M3)', () => {
         server.closeAllConnections?.();
       });
     }
-  }, 10_000);
+  }, CASE_TIMEOUT_MS);
 });


go-to-k:
Reviewed. The change itself is sound and I am not merging it yet — one item below has to be fixed first, and it is in the published prose rather than in the diff.

The verification table reproduces. Re-run independently, each of the three mutations is the SOLE failing assertion for its row, and mutant A (`await drainComplete;` after the warn) reports at **10009 ms** — past the old `10_000` per-case timeout, so on that mutant the old case really would have been runner-killed rather than reporting. That corroborates the misreporting argument instead of just asserting it, and it is also what makes the `20_000` raise the right call rather than a loosening.

Two things I checked separately that are worth recording, because neither is stated in the case:

- The trio is not vacuous, and the thing fencing it is the assertion that stayed. If `$disconnect` never registers, `inFlightDisconnects.size === 0`, `close()` skips the drain block entirely, and the surviving `elapsed >= 4_500` fails at ~5 ms. So `disconnectSettled === false` cannot pass for the wrong reason — but only because the lower bound is still there. Its comment sells it as "the drain really waited"; it is also the anti-vacuity guard.
- The `ws.close()` comment is correct and load-bearing: `close()` snapshots `wss.clients` synchronously, so calling `attached.close()` in the same tick is what guarantees a non-empty snapshot. Also, the `$disconnect` promise is added to `inFlightDisconnects` synchronously inside the connection-time `close` listener, which is registered before `close()`'s own `ws.once('close', ...)`, so registration always wins the race against `Promise.all(closes)`.

Dropping the dead `__queueInvokeResult({ statusCode: 200 })` primer is right — the queue is read only by the default `invokeRie` impl, which this case replaces wholesale.

---

## M0 — required before merge: a measured claim in the PR body, the commit message and go-to-k/cdkd#2904 is false

The PR body and the commit message both say:

> dropping `drainComplete.then(() => 'complete')` from the race leaves the whole file green, because the other cases close their clients before server cleanup and bypass the drain

and go-to-k/cdkd#2904 is built on the same premise.

I re-ran it: PR tip applied, that one arm removed from `src/local/websocket-server.ts`, nothing else changed, tree restored byte-exact afterwards.

```
Tests  7 failed | 11 passed (18)
```

All seven are `Test timed out in 5000ms`. So `inFlightDisconnects.size > 0` IS true at `close()` in those cases, the race DOES run, and the completion arm IS executed — it is just unasserted, and which cases cross the default 5 s timeout varies between runs. The honest statement is "executed incidentally, never asserted", not "never executed at all".

This matters in two places and both need the edit:

1. **The commit message**, because a squash writes it into `main` permanently.
2. **go-to-k/cdkd#2904's body**, more importantly: its verification plan starts from "measured GREEN before it". Whoever picks that issue up will drop the arm, see a pre-existing red, and read it as coverage their own change created. Please correct the premise there before this merges, otherwise the issue ships a trap.

The PR body can be edited or left as-is once the other two are right — it is the one of the three nobody re-reads later.

## m1 — optional, take-or-leave: the new assertion message can name the wrong mechanism

```ts
expect(
  disconnectSettled,
  'the drain waited for the in-flight $disconnect instead of giving up at its ceiling'
).toBe(false);
```

`disconnectSettled === true` at the read only means the handler settled before `close()` RETURNED. `close()` still does `await new Promise(resolve => wss.close(resolve))` after the drain, so latency there under a long enough stall flips the flag too, and the message then blames the drain for something the drain did not do. That is the same misreporting class go-to-k/cdkd#2795 was filed about, in the assertion replacing it. Narrow the message to what is observed, or leave it — the case fails correctly either way, only the label is wrong.

## m2 — optional, take-or-leave: the constants sever the M3 doc comment

`HUNG_DISCONNECT_MS` / `CASE_TIMEOUT_MS` land between the `// Issue #527 M3: ...` block and the `describe` it documents, so that block now reads as documentation for the two constants. Both are used only inside the describe; moving them in restores the adjacency. The same block also puts two different 5 s values next to each other — `close()`'s socket-close timeout and `SHUTDOWN_DRAIN_MS` — with nothing distinguishing them, which is worth one clause given the rest of the case turns on exactly that distinction.

## m3 — optional, take-or-leave: two measurement wordings

- "5/5 green at load average 35, in runs where the old bound red at 7907 ms" fuses two separate runs into one sentence. The 7907 was the dual-bound comparison run; the 5/5 was a different set; the original reds were 7751 / 7815.
- "about 5s" of slack is the nominal figure. The effective margin is `10_000 - (socket-close phase + 5_000)`, and the socket loop carries its own 5 s defensive timer, so a run shaped like the reported `elapsed=7907` leaves closer to ~2 s if the excess was loop time. If you want the margin the prose claims, `HUNG_DISCONNECT_MS = 15_000` / `CASE_TIMEOUT_MS = 25_000` keeps the ordering and costs nothing on a green run.

---

Everything under `m` is yours to take or leave and none of it blocks. Push the M0 correction — the commit message and go-to-k/cdkd#2904's premise — and I will merge.

nix-tkobayashi:
M0 is fixed on `c82d655b`, and all three of m1, m2 and m3 are taken.

## M0 — you are right, and the reason is worse than the wording

I could not reproduce your seven timeouts, which is what made me go looking properly rather than just adopting your sentence. Same tip, that arm alone removed, tree restored byte-exact: **18/18 green here**, the other seventeen cases at 2-166 ms and only the drain case at 5003 ms, and still green under a deliberate 16-way load at average 25.

So both measurements stand, and the thing that reconciles them is the mechanism, not the run. `inFlightDisconnects.add(disconnectPromise)` sits in the connection-time `close` listener, which is installed on ADMISSION — so it fires for every close of an admitted connection whether or not a `$disconnect` route is configured. Not for a denied `$connect` and not for a client closing during the `$connect` await, which the suite covers separately and which register nothing. For an admitted connection the promise can still be pending when `close()` reaches `size > 0`, and whether it is depends purely on scheduling. Idle, those promises settle first and the block is skipped; on your run seven of them did not, entered the drain, and waited the full ceiling into vitest's default 5 s per-case timeout.

Which makes my published claim wrong twice over: "never executed" was false, and "the other cases bypass the drain" was a property of one host stated as a property of the suite. That is the same failure you named at the end of go-to-k/cdkd#2797 — reporting what was looked at as though it were the whole set — and I did it again while writing up someone else's finding rather than measuring it myself. The reviewer's original note said "leaves THIS CASE green"; the widening to the file was mine.

Corrected in all three places, in the order you ranked them:

- **go-to-k/cdkd#2904's body**, patched first. Its plan no longer rests on that baseline at all: the gap-2 mutant's colour before the fix is documented as host-dependent and explicitly not usable, and what the fix must show instead is that it reds BY NAME — the new case's own assertion, not `Test timed out in Nms` — plus proof the case entered the drain rather than skipping the `size > 0` guard. The gap-1 mutant keeps "measured GREEN before it", which is sound there because both interpolations are 1 with one connection, so the swap is a string identity rather than a timing fact.
- **The commit message**, since the squash writes it into `main`.
- **The PR body**, corrected too rather than left.

## The two things you checked and found unstated

Both are now in the case, because you were right that neither was.

The anti-vacuity point I had not seen at all, and it changes how the lower bound reads: it is not a leftover beside the new assertions, it is what stops `disconnectSettled === false` passing for the wrong reason. Probed it directly — dropping the `inFlightDisconnects.add` line so nothing ever registers gives `expected 1 to be greater than or equal to 4500`. Its comment now says the two are a pair.

The same-tick and registration-order facts are in as well: that `close()` snapshots `wss.clients` synchronously, so calling it in the same tick is what GUARANTEES a non-empty snapshot — sufficient rather than necessary, since an intervening await need not let the socket actually close — and that the connection-time listener is registered before `close()`'s own `ws.once('close', ...)`, so registration wins that race structurally rather than by timing.

## m1, m2, m3 — all taken

**m1.** You are right that the message named a mechanism the flag does not isolate, and doing that inside the assertion replacing a misreporting bound would have been a poor joke. It now reads `the in-flight $disconnect settled before close() returned` and stops there. My first attempt at this fix kept a second clause saying the drain did not give up ahead of it, which re-committed the exact error one sentence later; a review pass caught it. A comment carries the reasoning instead: `close()` still awaits `wss.close()` after the drain, so a long enough stall there flips the flag without the drain having waited.

**m2.** The constants moved inside the describe, so the `Issue #527 M3` block documents the describe again. That block now also distinguishes the two 5 s timers by name, since the rest of the case turns on exactly that.

**m3.** Both wordings fixed. The fused sentence now separates the runs: 7751 / 7815 when the issue was filed, 7907 in the dual-bound comparison run where the new assertions passed, and 5/5 at load average 35 afterwards. And I took your sizing rather than softening the prose — `HUNG_DISCONNECT_MS` 15_000 and `CASE_TIMEOUT_MS` 25_000, which keeps the ordering, costs nothing on a green run since the handler is never awaited on the passing path, and makes the margin the comment claims actually exist.

## Verification

Kill set re-run after every edit above, each still the SOLE failing assertion: warn-then-still-block on the settled flag, ceiling 5_000 to 8_000 on `drained for 5000ms`, wrong count on `1/1 $disconnect handlers`, and the new anti-vacuity probe on the lower bound. Full `vp test run` root-asserted to this worktree: 969 files / 21,135 passed / 1 skipped. `vp check --fix` / `run check` / `typecheck` / `typecheck:test` / `build`: 0 errors. Codex to NO ISSUES on the corrected go-to-k/cdkd#2904 body in 2 rounds — the FIRST caught me writing "every socket close" where the listener only exists post-admission, which is the same over-broad reflex M0 is about, and the second was clean. A further pass over this reply caught the two code errors named above.

go-to-k:
Re-reviewed at `c82d655b`. M0 is properly closed and m1-m3 are all taken. One new item, below, caused by the m3 resize I asked for — one line, then I merge.

## Verified

- **Baseline** at the new tip: 18/18, 7.75 s.
- **Mutant A** (`await drainComplete;` after the warn) with the resized constants: sole failure, reported **BY NAME** at **15007 ms** — `AssertionError: the in-flight $disconnect settled before close() returned: expected true to be false` — inside the 25 s timeout. So the ordering still holds after the resize, and the m1 rewording is right: the message now states what the flag observes and nothing else.
- Your mechanism explanation for M0 checks out. `ws.on('close', ...)` is installed after the admission branch, and `onDisconnect` returns early when no `$disconnect` route is configured, so the promise is registered for every admitted connection's close regardless — which is exactly what makes the mutant's colour a scheduling fact rather than a property of the suite. Both our measurements are real and neither generalizes, which is what #2904 now says.

The M0 correction is better than what I asked for. I asked you to fix a sentence; you removed the mutant's colour from #2904's verification plan entirely and replaced it with "reds BY NAME, and prove the case entered the drain". That is the version that survives a host it was not written on. Keeping "measured GREEN before it" only for the gap-1 swap, on the grounds that a string identity is not a timing fact, is the right line to draw.

## M1 — required before merge: the resize left a stale figure in the residual paragraph

`HUNG_DISCONNECT_MS` went 10_000 to 15_000, and two references did not follow:

- **line 893**: "it only says the drain quit somewhere short of 10s, so a ceiling raised to 8s still passes it" — now 15s. This one matters because of where it sits: the residual paragraph is the disclosure of what the trio does NOT catch, and the undisclosed window widened from 10s to 15s with the resize. The conclusion does not invert (a raised ceiling still passes, more easily), but the paragraph now understates its own residual by 50%, which is the failure mode this PR exists to remove.
- **line 794**: "Sized so the slack is the ~10s the prose below claims" — the slack figure is right (15s minus the 5s ceiling, less the socket-close phase), but the prose below no longer claims it; you correctly genericized those figures to "the margin between the ceiling and the handler". The self-reference now points at nothing. Either restate the figure at the one place that computes it and drop the pointer, or point at the constants.

My fault for suggesting a bare resize without saying which figures it invalidated.

## Not a request, worth recording

The resize moves the failure path from 10009 ms to 15007 ms — mutant A now takes 5 s longer to report. That is the correct trade (it only costs on a red, and the passing path never awaits the handler), but it is the one thing the wider margin is actually bought with, and neither the case nor the PR body says so. Take it or leave it.

Push the M1 line and I will merge.

nix-tkobayashi:
M1 is fixed on `e98f1eb8`, and I took the optional note too.

## M1 — both figures

**The residual's "short of 10s"** now reads "short of `HUNG_DISCONNECT_MS`", and says the thing your framing made obvious: widening that constant widens this undisclosed window by the same amount, so the sizing has a cost side and this is it. You were right that the conclusion did not invert but the disclosure did — a paragraph whose whole job is to state what the trio misses, understating its own miss by 50%, is the failure this PR was opened to remove.

**The dangling self-reference** is gone. The figure now lives at the one place that computes it, with no pointer: the socket-close phase runs before the drain and carries its own 5s timer, so the effective margin is `HUNG_DISCONNECT_MS - (that phase + SHUTDOWN_DRAIN_MS)`, about 5s at these values once a slow phase is allowed for, against the 2.5s the removed bound had.

No apology needed for the bare resize. You flagged the sizing; keeping the figures that depended on it in step was my job, and I had just finished writing a paragraph about not letting a number outlive what it describes.

## The optional note, taken — and then narrowed twice

I wrote it in and a proxy pass caught two problems with my own wording, both the habits you have now named on three consecutive rounds.

First I wrote that the resize delays "the FAILING path". It does not: only a mutant that leaves the drain BLOCKED on the hung handler waits for it. The other three probes red at roughly 8s, 5s and 6 ms — their own ceilings, not the handler's. Scoped to the shape that pays it.

Then I had put the measured `15007 / 10009 ms` pair in the comment. That embeds the resize history in the case and adds a second maintenance-sensitive figure to a file I had just cleaned of one. Both numbers deleted; the tradeoff is expressed in `HUNG_DISCONNECT_MS` alone, and the measurements live here instead. A third pass then caught "every other mutant fails at its own ceiling" generalising past the probes actually run, and a leftover "the figure lives here rather than being referred to from the prose below", which is editing history wearing a comment's clothes.

For the record, since it belongs here rather than in the case: mutant A reds by name at **15009 ms** on my tree in the final run, against your 15007, and the pre-resize figure was 10009.

## Verification

Kill set re-run against the tree as it now stands, after the last of the edits above, each still the SOLE failing assertion:

| Mutant | Assertion | Reports at |
|---|---|---|
| drain warns, then still blocks | the settled flag | 15009 ms |
| ceiling `5_000` to `8_000` | `drained for 5000ms` | 8010 ms |
| leaked count wrong | `1/1 $disconnect handlers` | 5010 ms |
| nothing ever registers | the retained lower bound | 6 ms |

Baseline 18/18. `vp check --fix` / `run check` / `typecheck` / `typecheck:test` / `build`: 0 errors. Codex to NO ISSUES on the delta, plus three maintainer-proxy rounds, the last clean.

**A suite-level finding, reported with what it does and does not establish.** Three of six full-suite runs on this box exited non-zero while reporting `Test Files 969 passed` and `Tests 21135 passed | 1 skipped`. Every failure is the same shape: `[vitest-pool]: Worker forks emitted error` / `Worker exited unexpectedly`, no test named, no file attributed.

I ran the discriminating experiment rather than guessing. Same worktree, same box, my change REVERTED to `origin/main`'s copy of the file, nothing else touched: the same worker-exit signature reproduced. That the signature occurs without the change is the whole of what it establishes — it does not prove the two failures share a cause, and it does not exclude the change contributing to the ones that carried it. The cause remains unknown.

| run | tree | result |
|---|---|---|
| 1 | this PR | SUITE_OK |
| 2 | this PR | worker exit |
| 3 | this PR | SUITE_OK |
| 4 | this PR | worker exit |
| 5 | **main's copy of the file** | **worker exit** |
| 6 | this PR | SUITE_OK |

I did not capture memory or process state at the moment of a crash, only afterwards, and I had been running deliberate 16-way CPU load on this box earlier in the session for the flake reproduction. It may well be local to this host or worktree. I am not filing an issue on the strength of one machine, but if you or CI have seen the same signature it is probably worth one, and I can file it with these numbers.

The marker is set on run 6, a clean run of the tree as pushed.

The branch is 6 commits behind `main` but conflict-free and GitHub reports it mergeable at one changed file. Say the word if you would rather I rebase before you merge.

go-to-k:
Merged. Thank you — three rounds and every one of them came back tighter than what I asked for.

What stood out, since it is worth naming rather than just merging:

**You did not adopt my sentence.** I handed you "7 failed, your claim is wrong" and the easy move was to take it. You could not reproduce it, went and found the mechanism instead, and came back with the answer that reconciles both runs: the close listener is installed on ADMISSION, so the promise registers for every admitted connection's close, and whether it is still pending when `close()` reaches the drain is pure scheduling. My 7-red and your 18-green are both real and neither generalizes. That is a better outcome than either of us had alone, and it only exists because you treated a maintainer's measurement as evidence rather than as a verdict.

**And the correction went past its own blast radius.** I asked you to fix a sentence. You removed the mutant's colour from go-to-k/cdkd#2904's verification plan entirely and replaced it with "must red BY NAME, and prove the case entered the drain" — a criterion that holds on a host neither of us has run on. Keeping "measured GREEN before it" only for the gap-1 swap, on the grounds that a string identity is not a timing fact, is exactly the right place to draw that line.

**The self-narrowing on the optional note** is the part I would not have caught. You wrote the tradeoff in, then cut "the FAILING path" to the one mutant shape that actually pays it, then deleted the `15007 / 10009` pair rather than plant a second maintenance-sensitive figure in a file you had just cleaned of one. Declining my suggestion's literal form to keep its intent is the correct read.

My verification, for the record: 969 files / 21135 passed / 1 skipped, and mutant A reds by name at 15007 ms against your 15009. No stale figures left in the case — everything now names `HUNG_DISCONNECT_MS` or states its derivation, so the resize cannot strand a number again.

Two things recorded, neither needing action:

- The PR body still says "the 5 s against 10 s margin" in one line. It does not reach `main` — the commit message uses the derivation form — so I merged over it rather than spend a fourth round on the artifact I had already ranked lowest. Noting it only so it is not mistaken for something I missed.
- On the worker-exit signature: I cannot corroborate it. One full run on my box was clean, and no `main` CI run in the recent history carries it. That is weak evidence against a repo-level cause and no evidence about your host, so I agree with your call not to file on one machine. If it recurs on a second box, file it with your six-run table — the reverted-file control is what would make that issue worth reading.

go-to-k/cdkd#2904 stands as filed.

Issue #2881, opened 2026-09-09
fix(drift): the --revert mask refusal names a NoEcho custom resource as the cause, which is now only one of two


go-to-k:
## The defect

`cdkd drift --revert` refuses a resource whose baseline holds `SECRET_MASK` at a
position AWS reports nothing for, and the refusal message NAMES a cause:

```
the recorded baseline holds only the redaction mask there (a NoEcho
custom-resource value), and AWS reports nothing to preserve ... Force that
custom resource to update (change one of its properties, e.g. a nonce) and
re-deploy, so its handler runs again and supplies the value
```

`src/cli/commands/drift.ts:4323`. Since the readback position walk started
FAILING CLOSED (issue (#2852)), a mask in `observedProperties` has a SECOND
cause: a position the walk could not certify — a reshaped container, an
identity key AWS normalised, an unkeyed array it reordered, or a record whose
`properties` hold the raw `Fn::Join` object `cdkd import`'s warn path writes.
For that cause the message names the wrong mechanism and prescribes a remedy
that cannot work: there is no custom resource to force, and forcing an
unrelated one changes nothing. The correct remedy is a `cdkd deploy` of the
resource, after which the baseline is captured against a template cdkd can
position against.

A second, smaller claim in the same file is now weaker rather than false:
`preserveLiveValuesAtMaskedLeaves`' doc comment
(`src/cli/commands/drift.ts:3741`) reasons that "THE MASKED POSITION IS THE
PROOF THE VALUE IS SECRET, since nothing but this module's own redaction puts a
mask at a leaf". Still true that only cdkd's redaction writes one — but a
fail-closed mask deliberately over-masks a value AWS normalised, so the live
value it registers as a mask-only needle may be an ordinary literal
(`US-EAST-1`). The consequence is bounded to over-masking in log output, which
is the safe direction, but the comment should say so rather than assert
secrecy.

## Not fixed in the PR for (#2852) / (#2846) / (#2869)

That lane's file scope is `src/deployment/secret-redaction.ts` +
`src/cli/commands/state.ts` and their tests; `src/cli/commands/drift.ts` was
held by another lane in the same run. The two sibling messages that carry the
same NoEcho attribution — `src/cli/commands/export.ts:3899` and
`src/deployment/rollback-executor.ts:1198` — were checked and are NOT affected:
both read `properties`, and the fail-closed mask only ever lands in
`observedProperties`.

## Remedy directions

[...]

Issue #2852, opened 2026-09-09
fix(secret-redaction): the readback pass FAILS OPEN at every position it cannot certify, persisting a decrypted secret (reshaped containers, non-round-tripping identity keys, unpairable arrays)


go-to-k:
## The class

`redactSecretsForState`'s readback pass (`refuseUncertifiedReadbackPositions`)
substitutes the source expression only at positions it can CERTIFY. At every
branch it cannot certify it FAILS OPEN -- `return bag` -- so the decrypted
readback is persisted verbatim.

This was filed as an array issue. It is not. Measured 2026-09-09 by driving the
module with the capture's exact arguments (`source` -> `readback`, EMPTY map,
`STATE_SOURCED_READBACK_RULES`), the class is **position REACHABILITY in
general**: arrays are one instance, not the boundary. Three independent axes
leak, and two of them involve no array at all:

- **a reshaped container** -- the readback changes a container's shape relative
  to the source at that position or ANY ancestor (object -> array, an added
  wrapper level, a scalar promoted to a container);
- **an identity key that does not round-trip BYTE-IDENTICALLY** -- an AWS case
  or ARN normalisation drops a keyed-array element to `partner === undefined`
  and it is returned unchanged;
- **a readback key the source does not carry** -- the object walk hands the raw
  value on.

Arrays add a fourth: an unkeyed array that anchor corroboration cannot certify.
`unkeyedArrayPairsByAnchors` requires same length AND every non-reference
position deep-equal AND a distinguishing anchor per reference-bearing element
AND pairwise-distinct anchor signatures. **Same length is the FIRST of four**,
so "an unkeyed array of unchanged length" is NOT safe -- an earlier revision of
this issue said it was, and that claim was false.

## The measurement, verbatim

Refutation-oriented boundary probes. `LEAK` means the decrypted needle survived
into the persisted bag. Pasted as produced rather than summarised -- a summary
is an unbacked claim, and an earlier revision's count of these rows disagreed
with the rows themselves:

```
safe | whole-token @depth | plain nested object, structure preserved | {"A":{"B":"{{resolve:secretsmanager:s:SecretString:pw::}}"}}
LEAK | whole-token @depth | ANCESTOR container reshaped (obj -> array) | {"A":["PLAINTEXT-NEEDLE-1"]}
LEAK | whole-token @depth | ANCESTOR gains a wrapper level | {"A":{"B":{"C":"PLAINTEXT-NEEDLE-1"}}}
[...]

Issue #2759, opened 2026-09-07
fix(secret-redaction): a secret passed through Fn::Base64 is persisted to state.json and logged, because the needle set matches the plaintext literally


go-to-k:
## What is wrong

`redactSecretsForState` and `maskSecretsInText` match the recorded plaintext
LITERALLY. `Fn::Base64` over a dynamic reference produces a value that is the
secret in a trivially reversible encoding, and no needle matches it — so the
encoded secret is persisted to `state.json` and printed to the log.

**Measured, not read.** Against the real `src/deployment/secret-redaction.ts`
with a bag holding one recorded plaintext:

```
b64                       cHJvYmUtcHctbmV2ZXItbG9nZ2Vk
decodes_back_to           probe-pw-never-logged
persisted_value           cHJvYmUtcHctbmV2ZXItbG9nZ2Vk
b64_survived_redaction    true
raw_plaintext_control     {{resolve:secretsmanager:s:SecretString:password}}
masked_log_line           Resolved Fn::Base64: *** -> cHJvYmUtcHctbmV2ZXItbG9nZ2Vk
```

The control is the point: the RAW plaintext in the same bag is correctly
replaced by its expression, so the needle set works. Only the encoded form
passes through.

The log line is the sharper illustration —
`src/deployment/intrinsic-function-resolver.ts:6601` masks BOTH sides:

```ts
`Resolved Fn::Base64: ${this.maskSecretsForLog(resolvedValue, context)} -> ${this.maskSecretsForLog(result, context)}`
```

The left side becomes `***` because the resolved value IS a needle; the right
side is handed through untouched because `base64(secret)` is not. The line masks
the secret and then prints it in the same breath.

## Why it reaches state, not just the log

`resolveBase64` (`:6591`) RETURNS `result`, so it becomes a resolved property
value and flows to the save choke point, where `redactSecretsForState` scans with
the same literal needles. The probe above measures exactly that call.

`Fn::Base64` over a secret is not contrived: it is the ordinary CloudFormation
spelling for EC2 `UserData`, and putting a credential there is a common (if
unwise) pattern. `tests/integration/intrinsics-torture/` and
`intrinsics-torture-2/` already exercise `Fn::Base64`, though not over a
dynamic reference.

## Scope — this is the ENCODING class, not the length one

Distinct from the two residuals already tracked:
- go-to-k/cdkd#2516 / go-to-k/cdkd#2745 are the needle-LENGTH floor (a 1-3
[...]

Issue #1784, opened 2026-08-13
feat(drift): a readback reverse-map inside an ARRAY has no getDriftUnknownPaths companion, so existing observedProperties baselines strand


go-to-k:
Recorded while fixing #1767 — the residual its `getDriftUnknownPaths` companion cannot reach, and the second instance of the same wall #1742 already records for `AWS::DynamoDB::GlobalTable`.

## The shape

`feedback: dropping a key from a readback manufactures drift for every EXISTING baseline` (the #1760 lesson) says: when `readCurrentState` stops emitting something, the ALREADY-WRITTEN `observedProperties` records still carry it, so the first `cdkd drift` after the upgrade reports a one-sided phantom until each stack redeploys. The remedy is a `getDriftUnknownPaths` companion that removes the path from the COMPARISON as well.

That remedy works when the affected value is a TOP-LEVEL key (#1760's `WarmThroughput`). It does not exist when the value is a MEMBER OF AN ARRAY ELEMENT: `calculateResourceDrift`'s `isIgnoredPath` is never asked about a path that crosses an array, because `diffAt` compares arrays wholesale via `deepEqual` and never descends into elements. So there is no per-member ignore path, and the only expressible suppression is the WHOLE array.

#1767 therefore ships:

- the reverse-map (drops `IndexStatus` / `Backfilling` / `ItemCount` / `IndexSizeBytes` / `IndexArn` / the on-demand `{0, 0}` placeholder / the computed per-index `WarmThroughput` from `GlobalSecondaryIndexes[]` and `LocalSecondaryIndexes[]`);
- a companion that ignores each list when the template declares NO index list (covers that population's transition AND steady state, the #1760 shape);
- and CARRIES the residual for the population that DOES declare indexes: their pre-#1767 observed baseline reports a one-sided `GlobalSecondaryIndexes` difference until the next deploy re-captures it, or `cdkd drift --accept` does.

Suppressing that residual with the only available mechanism would mean ignoring the whole array for every table that HAS indexes — i.e. never detecting an out-of-band index add / remove / capacity change again, permanently, to remove a one-time report. That trade was refused.

## Why it is worth a mechanism rather than a per-provider workaround

The class recurs on every reverse-map that touches an array-shaped property, and cdkd is acquiring them steadily (#1742 GlobalTable per-index `WarmThroughput` — still open for exactly this reason, #1767 Table indexes). Each one currently has to choose between shipping a transitional phantom and permanently blinding an array.

[...]

PR #2912, opened 2026-09-10
fix(drift): refuse a raw-intrinsic revert baseline, pair token-position arrays by identity, and name both mask-refusal causes

### LINKED ISSUE #2855: fix(drift): --revert can push an unresolved intrinsic OBJECT to a live property for a resource refused an observedProperties baseline
## The hazard

A resource refused an `observedProperties` baseline by (#2842)'s THROW arm ends
up in a state no previous record reached: `properties` are the RAW intrinsic
shape AND `observedProperties` is `undefined`. `cdkd drift --revert` then makes
the raw bag its revert baseline and can push an unresolved intrinsic OBJECT to
a live AWS property.

`drift.ts` computes `revertBaseline = observedProperties ?? properties`, so a
refused resource falls to `properties`. `resolveStateSecretExpressions` re-
resolves only `{{resolve:` STRINGS, so an `{Fn::Join: [...]}` OBJECT walks
through untouched into `provider.update`.

The neighbouring comment in `drift.ts` already knows that pushing an unresolved
TOKEN corrupts a live property. The OBJECT form sits outside that guard.

## Why it is new

On `main` these resources had a readback baseline, so `--revert` was a
near-no-op for them. (#2842)'s refusal is what removes the baseline and routes
them to the raw bag. The refusal is still the right trade against a plaintext
disclosure -- this issue is the cost, not an argument to undo it.

Note the refusal is deliberately coarse: it fires for EVERY throwing resource,
including ones carrying no dynamic reference at all, so the affected population
is wider than "resources with secrets".

## UNMEASURED, and it decides the severity

Whether `provider.update` FAILS LOUDLY or CORRUPTS SILENTLY when handed an
intrinsic object where it expects a scalar has not been measured. It plausibly
differs per provider -- an SDK provider may reject on type, while Cloud Control
may serialize the object into the payload. Measure this before choosing a
remedy; the issue is filed at high severity on the pessimistic reading.

## Remedy directions

The fix belongs on the REVERT side, with the raw-shape family in (#2846),
rather than in the import refusal:

- refuse to revert a resource whose baseline leaf is a non-scalar intrinsic
  object, naming the resource -- the same shape `drift.ts` already uses for the
  unresolved-token case;
- or teach `resolveStateSecretExpressions` to walk intrinsic OBJECTS, which is
  a larger change and shares a boundary with the resolver.

## Classification

Session-fit: next (not this session) -- the remedy is in `cdkd drift`'s revert
path, a different command from the one (#2842) touches, and the severity turns
on a measurement not yet taken.
Severity: high -- on the pessimistic reading a `--revert` writes a corrupted
value to a live AWS resource, which is worse than a stale baseline. Downgrade
to medium if `provider.update` is measured to fail loudly for every route.
Effort: small (S) -- a guard on the revert side plus unit cases, once the
measurement is in.
Estimate: ~1-2 h, most of it measuring the `provider.update` behaviour across
an SDK route and a Cloud Control route.
Dup-check: searched open and closed issues on 2026-09-09 for `drift revert raw
intrinsic`, `revert baseline object`, and `resolveStateSecretExpressions`.
(#2846) is the closest -- the same raw-shape record family, but its subject is
`state refresh-observed` persisting a plaintext, not the revert-side write.
(#2850) / (#2852) / (#2854) are the import refusal's own residual classes. No
duplicate.

### LINKED ISSUE #2884: fix(drift): --revert copies live array values POSITIONALLY into masked leaves, so a reordered list sends the wrong value
## The defect

`preserveLiveValuesAtMaskedLeaves` (`src/cli/commands/drift.ts`) walks the send
bag against AWS's reported properties and, at every leaf holding
`SECRET_MASK`, copies AWS's live value in so `cdkd drift --revert` sends a real
value rather than the literal `***`. Its array descent is **POSITIONAL**,
guarded only by a length check:

```ts
const liveItems = Array.isArray(live) && live.length === value.length ? live : undefined;
return value.map((item, i) => walk(item, liveItems?.[i], `${path}[${i}]`));
```

So element *i* of the baseline takes element *i* of AWS's current list. When the
two lists describe the same elements in a different ORDER, the value copied in —
and pushed to the live resource — belongs to a different element. It is also
registered as a mask-only needle by `recordMaskOnlyValue`, so the wrong value
becomes a redaction needle for the rest of the run.

## Why it matters more now

Before the fail-closed change of (#2852), a mask in a persisted bag came from
one place: a `NoEcho` custom-resource `Data` value ((#2274)). Those sit at
scalar leaves of properties the template names, so the positional descent was
rarely exercised against a reordered list.

(#2852) added a second source, and it lands in exactly the population the
positional descent is weakest on: a leaf is masked *because* the walk could not
pair the baseline against the readback — an unkeyed array AWS reordered, or a
keyed element whose identity AWS normalised. A resource that produced a mask is
by construction a resource whose list AWS does not return in a stable,
pairable form.

Reproduction shape (not yet run against real AWS):

- record properties `Env: [{Name:'DB', Value:'{{resolve:secretsmanager:...}}'}, {Name:'REGION', Value:'us-east-1'}]`
- AWS returns the same two entries with `Name: 'db'` (case-normalised), so the
  element carrying the reference finds no partner and the whole element is
  masked into `observedProperties`
- a later `cdkd drift --revert` runs while AWS reports the list in the other
  order; `preserveLiveValuesAtMaskedLeaves` copies index 0's live value into
  index 0's mask, which is now the OTHER entry

## Remedy directions

- Pair by the same identity rule the redaction walk uses (`identityKeyFor` /
  `ARRAY_IDENTITY_KEYS`) before descending, and report the path as
  UNPRESERVABLE when no pairing is available — the caller already refuses the
  whole resource on an unpreservable path, which is the correct conservative
  outcome here.
- Or refuse positionally-descended arrays outright when any element carries a
  mask. Stricter, and it costs a revert on lists that would have been fine.

Either way the sibling `preserveLiveValuesAtUnresolvedTokens` shares the
positional descent and should be settled at the same time — it is the same
question asked about a different marker.

## Not fixed in the PR for (#2852)

`src/cli/commands/drift.ts` is outside that lane's file scope. Found by the
security review of go-to-k/cdkd#2882.

## Classification

Session-fit: next (not this session) — the edit is in `src/cli/commands/drift.ts`,
held by a different lane of the same `/work-issues` run. It is also not a
regression that PR introduces: the positional descent predates it, and the
pre-(#2852) behaviour at those same positions was to push a stale decrypted
secret, which is not better.
Severity: medium — wrong values written to a live resource, but only on a
resource whose list AWS reorders between the baseline capture and the revert,
and only for the elements that carry a mask. No disclosure.
Effort: small (S) — reuse `identityKeyFor` at one call site plus unit cases; it
drags the unit cycle, and a `drift-revert-arrays` integ run to confirm.
Estimate: ~2-3 h — most of it deciding whether an unpairable list should refuse
the resource or only the path, and the integ round trip.
Dup-check: searched open and closed issues on 2026-09-10 for
`preserveLiveValuesAtMaskedLeaves`, `positional array revert` and
`drift revert array order`, and grepped `src/cli/commands/drift.ts` for the
sibling walk. (#2274) is the mask's origin and (#2881) is the refusal MESSAGE
that misattributes its cause; neither names the positional descent. (#1783) /
(#1784) are the array seams `drift-revert-arrays` fences, on the comparison
side rather than the preserve side. No duplicate.

### LINKED ISSUE #2893: fix(drift): preserveLiveValuesAtUnresolvedTokens pairs array elements POSITIONALLY, so a reordered readback copies the wrong live value into a token position
## The defect

`preserveLiveValuesAtUnresolvedTokens` (`src/cli/commands/drift.ts`) walks the
bag `cdkd drift --revert` is about to send against AWS's reported properties
and, at every leaf whose WHOLE value is a surviving `{{resolve:...}}` token,
copies AWS's live value in — so the revert leaves that position as AWS has it
instead of writing the literal token. Its array descent is POSITIONAL, guarded
only by a length check:

```ts
const liveItems = Array.isArray(live) && live.length === value.length ? live : undefined;
return value.map((item, i) => walk(item, liveItems?.[i]));
```

When AWS returns the same elements in a different ORDER, element *i* of the
send bag takes element *i* of AWS's current list, so the value copied into the
token's position — and then pushed to the live resource by `provider.update` —
belongs to a different element.

## Why this is filed separately from (#2884)

(#2884)'s body states two requirements. The masked-leaf half is fixed in
go-to-k/cdkd#2882: `preserveLiveValuesAtMaskedLeaves` now pairs by
`identityKeyFor`, and with no identity field requires every non-mask LEAF to
corroborate the order before it will align positionally. Its closing sentence —
"the sibling `preserveLiveValuesAtUnresolvedTokens` shares the positional
descent and should be settled at the same time" — is this issue.

It was NOT settled at the same time because the corroboration rule does not
transfer, which is a design difference rather than a scheduling one:

- The mask arm's wildcard is `SECRET_MASK`; here it would be the token. Sharing
  the helper literally would compare the token leaf against AWS's resolved
  value, they differ, and every array holding a token would refuse — the
  opposite of what the function exists to do.
- Parameterising the wildcard is not enough either, and the reason is the
  RESIDUAL, not the send bag's shape. (An earlier revision of this bullet said
  the token arm differs because its send bag is `buildRevertNewProperties`'
  output whose drifted subtrees "differ from AWS by construction" — but
  `preserveLiveValuesAtMaskedLeaves` receives the SAME bag, so the mask arm
  over-refuses on drifted arrays too, and that is its accepted behaviour;
  corrected in round 4 of go-to-k/cdkd#2882.) What actually differs is what a
  refusal COSTS: a mask refusal DROPS the whole resource — nothing is written,
  the refusal is reported — while a token-arm refusal falls back to KEEPING
  the token, which `provider.update` then SHIPS to AWS (the token walk returns
  the literal on a missing live value). A corroboration rule that refuses
  often is a safe posture for the mask arm and an actively harmful one here,
  so the sibling needs a rule tuned to its own residual.

So the sibling needs its own corroboration rule (identity-field pairing is the
part that plausibly transfers; the unkeyed fallback is not), its own fixtures,
and its own real-AWS confirmation.

## Remedy directions

- Pair by `identityKeyFor` / `ARRAY_IDENTITY_KEYS` first, as the masked-leaf
  arm now does. For its population this is strictly better than index
  alignment: an `Environment` / `Tags` list AWS reordered becomes preservable,
  and an element whose identity AWS normalised gets no live value, which falls
  back to KEEPING the token — the residual this function already documents as
  safe (`cdkd deploy` sends the literal token too).
- Decide what the unkeyed fallback should be. Options: keep index alignment
  (status quo, wrong under a reorder), refuse whenever the array holds a token
  (safe, costs preservation on lists that were fine), or corroborate on
  something that survives a revert's own deltas — e.g. only the positions the
  drift comparator did NOT report as drifted.
- Either way the residual on refusal is the same one the function already
  takes: keep the token.

## Verification

`vp test run tests/unit/cli/drift-masked-leaf-preserve.test.ts
tests/unit/cli/drift-secret-redaction.test.ts` for the unit half, plus a
`/run-integ drift-revert-arrays` round — that fixture is the one that has
caught this module's array-ordering defects before (issues (#1783) / (#1784)).

## Classification

Session-fit: next (not this session) — the corroboration rule for tokens is a
different decision from the one go-to-k/cdkd#2882 makes, and settling it needs
a `drift-revert-arrays` real-AWS run, which that lane cannot take (integ runs
are serialized on the orchestrator).
Severity: medium — a value belonging to another array element is written to a
live AWS resource. No disclosure: since issue (#2482) every CloudFormation
`{{resolve:...}}` spelling resolves, so a SURVIVING token is text that merely
looks like a reference and the live value at its position is ordinary data.
Effort: small (S) — one call site plus unit cases, mirroring the shape
go-to-k/cdkd#2882 already landed one function over.
Estimate: ~2-3 h — most of it deciding the unkeyed fallback and the integ round
trip.
Dup-check: searched open and closed issues on 2026-09-10 for
`preserveLiveValuesAtUnresolvedTokens` (3 hits: (#2102), (#2884), closed (#1958)) and
for `revert positional array token` (0 hits), and grepped
`src/cli/commands/drift.ts` for both preserve walks. (#2884) is the MASKED-leaf
twin and its masked half is fixed; (#2102) is mask-by-SPAN — the whole-leaf
gate on which leaves are preservable at all, not which live element pairs with
which. (#2274) is the mask's origin and (#2881) is the refusal MESSAGE.
No duplicate.

go-to-k:
## Summary

Three fixes on the same `cdkd drift --revert` write-back path in `src/cli/commands/drift.ts`, shipped together because one decision — what an unpairable or unresolvable position does on the write path — decides all three.

### Issue (#2855) — refuse an unresolved intrinsic OBJECT baseline

A record refused an `observedProperties` baseline by issue (#2842)'s throw arm reverts from raw `properties` holding intrinsic OBJECTS (`{Fn::Join: ...}`) written by `cdkd import`'s warn path. `resolveStateSecretExpressions` re-resolves only `{{resolve:...}}` STRINGS, so the object survived into the send bag and `provider.update` shipped it.

**The issue's open measurement, taken with mocked SDK clients (both routes):**

- `SSMParameterProvider.update` — no throw; `PutParameterCommand.input.Value` carried the raw object `{"Fn::Join":["",[...]]}`.
- `CloudControlProvider.update` (ordinary type) — `UpdateResourceCommand` issued with the object verbatim as a JSON-patch value; AWS's schema decides.
- `CloudControlProvider.update` (`JSON_STRING_PROPERTIES` type, `AWS::Events::Rule.EventPattern`) — the object is `JSON.stringify`ed into a schema-VALID string: **AWS cannot reject it on type, i.e. guaranteed silent corruption.**

No route fails loudly on cdkd's side, so the issue's pessimistic reading (Severity: high) is confirmed and the remedy is a revert-side refusal (`collectUnresolvedIntrinsicObjectPaths`; counted `totalUnresolvable`, exit 2, remedy `cdkd deploy`), with three scopings from the review rounds:

- the scan reads the **BASELINE** (`desiredProperties`), never the merged send bag — `mergeUntemplatedValue`'s key-merge can FUSE a single-key intrinsic with a plain-record live value into a multi-key object the single-key predicate misses;
- it runs only at the **drifted top-level keys**, the only ones the baseline sources into the send bag;
- it is **gated on `observedProperties === undefined`** — the raw-`properties` fallback is the only provenance that can hold a genuine intrinsic, while a readback-derived observed baseline holding a `Ref`-shaped map is a real AWS value a refusal would pin unrevertable forever.

The refusal also runs BEFORE the token warnings, so a refused resource is never first promised "left UNCHANGED by this revert".

### Issue (#2893) — pair token-position arrays by identity, not index

`preserveLiveValuesAtUnresolvedTokens` paired array elements positionally behind a length test, so a readback AWS reordered copied ANOTHER element's live value into a token position and `provider.update` shipped it. The descent now goes through the same `pairedLiveItems` rule as the masked-leaf walk — identity field first, else the array's own literal frame corroborated per leaf — with two per-caller parameters, both derived from what a refusal COSTS:

- the **wildcard** (whole tokens + masks for this pass; the mask alone for the sibling);
- **`acceptForcedSingleton`**: a 1-vs-1 array is a FORCED position (no other candidate to mis-pair with; the copy is position-neutral), and the token walk's refusal residual WRITES the token over the live value — so the token walk accepts there (pre-change parity for every 1-vs-1 array), while the mask walk keeps refusing (its refusal residual is a safe drop). Length ≥ 2 keeps every bound: identity pairing, per-leaf corroboration, the one-wildcard-slot rule and the `compared > 0` floor.

On any pairing refusal the token is KEPT — this pass's documented residual — never the mask walk's drop-the-resource refusal. The mask walk's `corroborationSource` (PRE-token bag) is unchanged and still load-bearing.

This closes the token half of issue (#2884); PR (#2882) closed the masked half.

### Issue (#2881) — the mask refusal names both causes (PARTIAL — Refs, not Closes)

The refusal message asserted "(a NoEcho custom-resource value)" as THE cause and prescribed a nonce bump, which does nothing for the now-common (#2852) fail-closed cause. It now states both causes and both remedies, the token warnings state the pairing condition, and `preserveLiveValuesAtMaskedLeaves`' doc no longer claims a masked position proves secrecy. **This PR deliberately does NOT close (#2881)**: its checklist's remaining items (the `Fn::Base64` third source and the `deploy-engine.ts` fourth message — both premised on the still-open (#2759), whose `resolveBase64` calls no `recordMaskOnlyValue` at this tree — plus pointer comments in `export.ts` / `rollback-executor.ts` / `deploy-engine.ts`) live in files outside this lane's scope; the disposition is recorded on the issue.

## Test plan

- 971 unit files / 21,263 tests green (`vp test run`), plus `vp check`, `vp run typecheck:test`, `vp run build`, `vp run gen:all-matrices`.
- Shape-enumerated tests written before the predicates: S1–S17 in `tests/unit/cli/drift-masked-leaf-preserve.test.ts` (keyed reorder / AWS-added element / normalised identity / token identity echo / corroborated unkeyed / reordered unkeyed / two-token / 1-vs-1 forced-position preserve + length-2 all-token refuse / length mismatch / nested keyed / mask sibling / non-plain member / forced-singleton keyed + contradicting-sibling / nested-array slots / mixed-leaf conjunct), H1–H8 for the intrinsic scan, and 7 command-level cases in `drift-secret-redaction.test.ts`.
- Mutation probes, each a single mutation with byte-exact `cp` restore (`git diff` empty after each): positional-descent revert **10/57 red**; wildcard args dropped **1/57** (S5); carries-arg mask-only **2/57** (S7, S11); refusal wiring disabled **2/92**; `Fn::` arm dropped **6/149**; drifted-key scoping removed **2/149**; old single-cause message **1/92**; overlaid-scan (fuse) **1/94**; unmasked paths **1/94**; wildcard-identity guard removed **2/59**; singleton wiring off **3/64** (S8, S14, S15); singleton bound removed **14/64**; deep-array slot arm neutered **1/64** (S16); mixed-leaf conjunct dropped **1/64** (S17); provenance gate removed **1/95**; old summary text restored **1/95**.

## Real-AWS coverage needed before merge

This lane ran no real-AWS integ (serialized on the orchestrator). Before merging run:

- `/run-integ drift-revert`
- `/run-integ drift-revert-arrays` (the fixture that has caught this module's array-ordering defects — (#1783) / (#1784) — and the one issues (#2893) / (#2884) name for verification; it is what decides whether `Closes #2884` stands)

## Deferred with classification

- Issue (#2919) — the MASK walk's identity arm pairs on a corr identity that is a whole token (flagged by the round-1 security review).
- Issue (#2920) — the token preserve arm copies a NON-STRING live value into a string-typed token position (flagged by the round-2 test review; pre-existing at length >= 2, widened by the singleton acceptance).

Closes #2855
Closes #2893
Closes #2884
Refs #2881

Issue (#2884) is closed by this PR because both of its stated requirements are met: the masked-leaf half landed in PR (#2882) and the token half is this PR's (#2893) fix; its real-AWS confirmation is the `drift-revert-arrays` run listed above.

https://claude.ai/code/session_01FiqeKgpkYzA3eqFNESUmEC


go-to-k:
Review-round disposition record (rounds 1-2, code + security reviewers, both blocker-free):

- Fixed in 42ea5510: the round-1 MAJOR (merge fuses a single-key intrinsic into a multi-key bag — the scan now reads the baseline), security M1 (path lists through `maskSecretsInText`, applied to all three messages on this path), security N1 (explicit wildcard-identity refusal in `pairedLiveItems`), and the `Object.hasOwn` nit.
- Won't-do (recorded here): the code reviewer's display-ambiguity nit — a refused path whose property KEY itself contains `.` or `[` renders ambiguously in the refusal message. Display-only: no pairing, scan, or write decision consumes the rendered path, and the repo-wide path rendering shares the convention, so a local escape would diverge from every sibling message. Shipping as-is is deliberate.
- Round 2 re-review of the fix delta: both reviewers reported no new findings (BLOCKERS: none).

Issue #2907, opened 2026-09-10
test(deployment): the intrinsic-resolver wrapped-budget assertion is a wall-clock RATIO and flakes under full-suite load


go-to-k:
## What happened

`tests/unit/deployment/intrinsic-resolver-concurrent-drain.test.ts:502` failed once during a full-suite run on 2026-09-10:

```
× a LOOP of resolutions wrapped in one budget is bounded once, not per iteration
AssertionError: the wrapped loop paid it once: expected 96 to be less than 94
```

The assertion is `expect(wrapped, 'the wrapped loop paid it once').toBeLessThan(bare / 2)`. Both operands are measured wall-clock durations, so the verdict is a RATIO of two timings taken under whatever load the machine is carrying. At the failure the margin was 2 ms on a ~94 ms budget — about 2%.

## It is a flake, not a regression

- The run that failed was concurrent with several other agent sessions on the same machine.
- The same file passes **26/26 three times in a row** when run alone, immediately afterwards.
- The full suite passes 970/970 on a re-run under normal load.
- The branch that observed it (go-to-k/cdkd#2903) touches only `tests/unit/scripts/claude-md-size.test.ts`, `CLAUDE.md` and three `SKILL.md` files — nothing this test reads.

## Why it is worth fixing rather than re-running

It fails in the direction that accuses an unrelated change: a contributor whose PR happens to run while the machine is loaded gets a red on a deployment test they never touched, and the natural next move is to look for a cause in their own diff. The repo already carries three siblings of exactly this family — go-to-k/cdkd#2741 (a floor with no declared timeout), go-to-k/cdkd#2444 (a wall-clock chunk race), go-to-k/cdkd#2120 (SSM read lag) — so the remedy shape is probably shared: assert the INVARIANT (the budget was paid once) by counting the payments rather than by comparing two durations, or give the ratio enough margin that ordinary scheduling noise cannot cross it.

Line 350 (`toBeLessThan(flat * 2)`) and line 416 (`toBeLessThan(firstCap / 2)`) in the same file are the same shape and should be assessed together.

Dup-check: searched open issues for `concurrent-drain`, `intrinsic-resolver`, `budget`, `flake` and `wall-clock` across titles and bodies. Hits: go-to-k/cdkd#2741, go-to-k/cdkd#2444 and go-to-k/cdkd#2120 -- all the same FAMILY (a timing assertion flaking under load) but each names a different file and a different assertion; none covers this one.

[...]

PR #2915, opened 2026-09-10
test(resolver): make the drain-budget fence per-CALL rather than per-loop, and fence the cross-stack pre-pass by its call sites

### LINKED ISSUE #2894: test(resolver): the owner-keyed drain-budget fence misses two wrapper spellings, and its false-positive message names a resolver call that is not there
Follow-up from the review of go-to-k/cdkd#2797 (maintainer item m18), filed rather than fixed there because the PR was otherwise ready to merge.

## What is wrong

`tests/unit/deployment/intrinsic-resolver-concurrent-drain.test.ts` carries an OWNER-KEYED fence over `cdkd scrub`'s inline drain budget: it walks `scrubStack`'s subtree, asserts every `.resolve(` call in it sits inside a `withSharedDrainBudget(` argument, under ONE shared wrapper, with no iteration statement between that wrapper and the owner. Two wrapper spellings pass it while restoring the per-iteration budget the hoist removed, and a third behaviour is deliberate but misdescribed in the failure message.

1. **Callback iteration is not an iteration statement.** The loop clause walks ancestors with `ts.isIterationStatement(n, false)`, which matches `for`, `for..of`, `for..in`, `while` and `do`. It does not match a callback. So

   ```ts
   await Promise.all([1].map(async () => withSharedDrainBudget(async () => { /* the loops */ })));
   ```

   passes, although it opens one budget per element exactly as a `for` would. Stated precisely, because the comment above that clause is easy to over-read: what the fence asserts is ONE shared wrapper by SYNTACTIC position, and a single syntactic wrapper invoked repeatedly satisfies that while being per-iteration at runtime. The comment's "a wrapper inside a loop is per-iteration by another spelling" is about the shapes it does check; it does not promise callback coverage, and a reader can take it to.

2. **The wrapper is matched by identifier TEXT.** The ancestor walk accepts any `ts.isIdentifier` call whose text is `withSharedDrainBudget`. A local shadow that opens nothing therefore satisfies the fence:

   ```ts
   const withSharedDrainBudget = async (fn: () => Promise<void>) => await fn();
   ```

   Contrived, and worth stating exactly: the CALL side is not "already closed" by declaration resolution either. It catches aliased receivers by matching any receiver's `resolve` property, deliberately and without symbol resolution. Closing this would introduce symbol resolution as a NEW mechanism in this fence, not extend an existing one.

3. **A plain `Promise.resolve()` placed OUTSIDE the wrapper in `scrubStack` reds the fence.** Inside the wrapper it passes. The call matcher takes any property access named `resolve` in the owner's subtree, deliberately, and the comment says so: "a false positive here is a loud test rather than a silent hole." The FAILURE MESSAGE does not — it reads

   > a resolver.resolve in scrubStack outside withSharedDrainBudget — that loop then spends one cap per iteration with the scrub lock held and `saveState` downstream

   which sends a reader looking for a resolver call that is not there.

## Why it is bounded

1 and 2 require someone to write an unusual wrapper spelling while removing a working one, so neither is reachable by accident. 3 IS reachable by accident — an ordinary `Promise.resolve()` added outside the wrapper trips it — but it fails loudly and misleads rather than passing something broken. The fence catches the shapes that actually occurred during review: a deleted hoist, per-call wrappers, a budget opened inside the first loop, an unwrapped nested arrow, an aliased receiver, an element-access call, a nested `function` declaration, a nested class method, a conditional receiver, and `while` / `for..of` between wrapper and owner.

## Proposed fix

- **Keep the iteration-statement check and ADD a function-boundary check**: the wrapper's enclosing function must BE the owner. The boundary check alone is not a replacement — a wrapper inside a `for` written directly in `scrubStack` has `scrubStack` as its enclosing function and would go green, turning today's loop mutants green with it. The two clauses catch different shapes and both are needed.
- Resolve the wrapper identifier to its declaration and require it to be the import from `src/deployment/drain-budget.ts`. This is a new mechanism for this fence; weigh it against the contrivance of the shape it closes.
- Reword the failure message so it names what was matched (`a call to .resolve(...) in scrubStack`) rather than asserting it is the resolver's.

## Verification plan

Each clause gets a mutation probe, since a fence that survives the mutant of the clause it claims to pin is asserted rather than pinned:

- baseline green;
- the `.map(async () => withSharedDrainBudget(...))` shape must RED after the fix, and is measured GREEN before it;
- the local-shadow shape must RED after, and is measured GREEN before;
- a wrapper inside a `for` written directly in `scrubStack` must stay RED after the fix — this is the probe that catches the boundary check being taken as a replacement for the iteration check;
- a plain `Promise.resolve()` outside the wrapper must still RED, with the new message naming the match rather than the resolver, and one inside the wrapper must stay green;
- the ten shapes the fence already catches must stay RED, so the rewrite adds coverage rather than trading it.

Dup-check: searched open and closed issues for "drain budget fence scrub owner-keyed", "withSharedDrainBudget fence evasion" and "resolveCrossStackReads" — no match. The nearest neighbours are go-to-k/cdkd#2805 and go-to-k/cdkd#2814, which are runtime residuals of the same PR rather than fence coverage.
Session-fit: next (not this session) -- the fence is correct for every shape the code actually contains today; widening it is test-only work with no user-visible change, so it does not need to ride go-to-k/cdkd#2797's merge. Start with: `/work-issues 2894`; verified by `vp test --run tests/unit/deployment/intrinsic-resolver-concurrent-drain.test.ts`
Severity: low -- nothing user-facing is wrong; the cost is that two contrived wrapper spellings could reintroduce a per-iteration budget without the fence saying so, and the false-positive message names a call that is not there
Effort: small (S) -- an AST predicate widening plus its probes, riding unit verification that is already paid
Estimate: ~1-2 h -- writing a probe per missed spelling that is measured GREEN before the change and RED after is the work; the predicate edit itself is minutes

### LINKED ISSUE #2895: test(resolver): the owner-keyed drain-budget fence excludes resolveCrossStackReads, which carries the per-leaf multiplier
Follow-up from the review of go-to-k/cdkd#2797 (maintainer item m19), filed rather than fixed there because the PR was otherwise ready to merge.

## What is wrong

`cdkd scrub`'s drain budget is hoisted over three resolve loops in `scrubStack`, and the OWNER-KEYED fence in `tests/unit/deployment/intrinsic-resolver-concurrent-drain.test.ts` walks `scrubStack`'s subtree to hold that hoist in place. `resolveCrossStackReads` is outside that subtree: it is built by `makeCrossStackPrePass` at module scope and returned as a closure, so its own per-leaf `resolver.resolve` is invisible to the fence.

That call carries the per-leaf multiplier. It is a per-leaf resolve loop invoked from ALL THREE wrapped loops in `src/cli/commands/scrub.ts` (the resources loop and both output loops), which is why the note above the wrap states the aggregate drain WAIT as at most `(#resources + 2 x #outputs) x (1 + #cross-stack-leaves) x` the cap rather than `(#resources + 2 x #outputs) x` it. Whether the leaves actually dominate depends on the input and on which resolutions drain at all — the expression is an upper bound, not a measurement.

## Why it is correct TODAY

`withSharedDrainBudget` inherits: a caller already inside a budget keeps it rather than opening a fresh one, and the budget lives in an `AsyncLocalStorage` store. All three of `resolveCrossStackReads`' call sites are inside `scrubStack`'s wrap, so at runtime its drains spend the hoisted budget and nothing leaks. The fence's silence costs nothing while that stays true.

## What is un-fenced

A future caller reaching `resolveCrossStackReads` from OUTSIDE a budget — a fourth site in `scrub.ts` added above the wrap is the reachable one today, since `makeCrossStackPrePass` is NOT exported and no other module can import it yet — makes each leaf open a budget of its OWN. Precisely: the unwrapped path still goes through the public `resolver.resolve`, which opens a budget via `withSharedDrainBudget`, so this is a separate budget per leaf rather than the drain's no-store fallback. Each is individually correct; the aggregate is one cap per cross-stack leaf with the scrub lock held and `saveState` downstream. Nothing reds.

The maintainer's own count on go-to-k/cdkd#2797 (item m15) confirmed the shape: `resolveCrossStackReads` really is a per-leaf resolve loop reached through `walk` -> `readOne` -> `resolver.resolve`, invoked at three sites, all currently inside the wrap.

## Proposed fix

Fence the CALL SITES rather than the owner's body: assert that every call to `resolveCrossStackReads` is lexically inside a `withSharedDrainBudget(` argument, the way the existing callee-keyed table does for the other loop entry points. That fits the property wanted, which is about one function's call sites and not about one function's body.

Two limits to state in the fix rather than discover after it:

- A name-keyed table catches calls spelled `resolveCrossStackReads`. A differently named binding for the factory's return value escapes it. Either narrow what the fence claims, or track the binding and add mutation probes for the aliased spelling.
- Having `makeCrossStackPrePass` open a budget itself does NOT remove the need for the fence. `AsyncLocalStorage` does not capture a store when a closure is created, so a budget opened in the factory does not reach the returned function's later invocations. Wrapping the returned function's EXECUTION would work per invocation and would inherit an existing budget, but separate unwrapped invocations would still get separate budgets — which is the very thing being fenced. It also takes the choice away from the caller, the reason `evaluateConditions` deliberately does not wrap its loop.

## Verification plan

- baseline green;
- MUTANT: add a fourth `resolveCrossStackReads` call in `scrub.ts` ABOVE the `withSharedDrainBudget` wrap. It must RED after the fix, and is measured GREEN before it — that measurement is the point of this issue;
- MUTANT: move one of the three existing call sites out of the wrap. Must RED;
- MUTANT: call it through an aliased binding. Must RED if the fix claims binding coverage, and the issue's wording must be narrowed if it does not;
- the existing owner-keyed and callee-keyed cases must stay green, so the new fence adds coverage rather than restating theirs.

Dup-check: searched open and closed issues for "resolveCrossStackReads", "drain budget fence scrub owner-keyed" and "withSharedDrainBudget fence evasion" — no match. go-to-k/cdkd#2805 and go-to-k/cdkd#2814 are runtime residuals of the same PR, not fence coverage.
Session-fit: next (not this session) -- every call site of `resolveCrossStackReads` is inside the wrap today, so the gap is a missing FENCE rather than a live hold; it is test-only and does not need to ride go-to-k/cdkd#2797's merge. Start with: `/work-issues 2895`; verified by `vp test --run tests/unit/deployment/intrinsic-resolver-concurrent-drain.test.ts`
Severity: low -- no user-visible defect today; what is unprotected is the per-leaf multiplier, so a future caller reaching `resolveCrossStackReads` from outside a budget would restore a per-leaf hold with nothing reporting it
Effort: small (S) -- one more owner row in the existing table plus its probes, on unit verification already paid
Estimate: ~1-2 h -- deciding whether the row keys on the helper's own body or on its callers is the work, since the helper inherits at runtime rather than lexically; the row itself is minutes

nix-tkobayashi:
## Summary

Two follow-ups from go-to-k/cdkd#2797's review (m18 / m19), filed as go-to-k/cdkd#2894 and go-to-k/cdkd#2895 and taken together because both land in the same fence in one file. Tests only — no `src/**` change.

**go-to-k/cdkd#2894 — the owner-keyed case counted the wrong barrier.** It asserted "one shared wrapper, and no ITERATION STATEMENT between it and the owner", and `ts.isIterationStatement` does not match a callback. So

```ts
await Promise.all([1].map(async () => withSharedDrainBudget(async () => { /* the loops */ })));
```

passed while opening one budget per element — exactly the per-iteration budget the hoist removed. The barrier is now any function boundary as well: in the shape the case exists to hold, the wrapper sits directly in the owner's own body, so a function-like node between the two means the budget is opened once per call of THAT function rather than once per call of the owner. The walk stops at the owner's FUNCTION node rather than its declaration, so a `const` or class-field owner is not counted as its own barrier.

Its false-positive message is corrected too. The walk matches ANY receiver's `resolve` by design, so a plain `Promise.resolve()` in `scrubStack` reds it — deliberate, but previously reported as "a resolver.resolve … outside withSharedDrainBudget", naming a call that need not be there.

**go-to-k/cdkd#2895 — the cross-stack pre-pass is fenced by its call sites.** `resolveCrossStackReads` carries the per-leaf multiplier and is invisible to the owner-keyed walk: it is built by `makeCrossStackPrePass` at module scope and returned as a closure, so its own `resolver.resolve` sits outside `scrubStack`'s subtree. Its CALL SITES are inside `scrubStack`, which is the property worth holding, so it becomes a row in the callee-keyed table — the shape that already fences `resolveOutputs`, `resolveImportedProperties` and `buildResolvedParametersPerStack`.

**One part of go-to-k/cdkd#2894 is deliberately NOT closed**, and is recorded as a KNOWN LIMIT at the wrapper match: the wrapper is matched by identifier TEXT, so a local `const withSharedDrainBudget = async (fn) => await fn()` shadowing the import satisfies the walk while opening nothing. Closing it means introducing symbol resolution — a mechanism this fence uses nowhere else — for a shape nobody writes by accident. Chasing one spelling per round is what made the parent PR need nine of them; the limit is named instead.

## Test plan

The walk is extracted into `analyseSharedBudget(sourceText, owner)` so the two NEW clauses are pinned by synthetic sources, not only by the real tree — which carries neither shape, and so leaves both mutants green. Every number below is measured, of 30 cases in this file:

| mutation | reds |
| --- | --- |
| `scrubStack`'s wrap rewritten as `Promise.all([1].map(async () => withSharedDrainBudget(...)))` | the owner-keyed real-tree case, `['ArrowFunction']` — and the PRE-change fence stayed green on the same mutation (`26 passed`), which is the hole go-to-k/cdkd#2894 reports |
| a fourth `resolveCrossStackReads` call added above the wrap | the new callee row, naming the added line |
| the callee renamed at its three sites | the row's floor (`expected 0 to be >= 3`) |
| `ts.isFunctionLike` deleted from the barrier | the callback case alone (`1 failed \| 29 passed`) |
| the barrier walk stopped at the owner's DECLARATION | both arrow-owner cases (`2 failed \| 28 passed`, reproduced three times) |
| `ts.isPropertyDeclaration` dropped from the normalisation | the class-field case alone (`1 failed \| 29 passed`) |

The real-tree case survives all three of the last group — which is why the synthetic ones exist. Production files were restored byte-exact after every probe (`git diff` clean).

Full suite 971 files / 21246 passed / 1 skipped, rc=0, run root asserted; `vp check --fix` + `vp run check` + `typecheck:test` + `build` + `gen:all-matrices` + `audit:coverage:check` all rc=0.

**No docs or changelog entry**, deliberately: CLAUDE.md writes an entry only for a user-visible behavior delta, and this changes no shipped behavior. `.claude/rules/layout-deployment.md` describes which src loops wrap in `withSharedDrainBudget`, which this PR does not change.

Codex: 3 delta rounds and 4 maintainer-checklist proxy rounds. The proxy pass is what found both defects worth having — a comment claiming "nothing reddens" for a mutation this PR's own new row catches, and the two new clauses being asserted rather than pinned.

**Review round 1** (commit `4e338e7d`) — all five required items and all six optional ones. The blocker was mine: with the function boundary added, `ts.isIterationStatement` became unreachable in every shape under test, so deleting it left all 30 cases green. go-to-k/cdkd#2894's Proposed fix warned about exactly that ("the boundary check alone is not a replacement ... both are needed") in a part of the issue body I had truncated away; two more required items were bullets in the same tail.

The fence now has 40 cases and every clause has a mutant that reds it and nothing else:

| mutation | reds |
| --- | --- |
| `ts.isIterationStatement(n, false) \|\|` deleted | loop case + source-order case |
| `ts.isFunctionLike(n)` deleted | callback case |
| barrier walk stopped at the DECLARATION | both arrow-owner cases |
| `ts.isPropertyDeclaration` dropped | class-field case |
| body test replaced by a `declare`-only rejection | overload case |
| `ts.isIdentifier(node.name)` dropped | quoted-method case |
| `node.name !== undefined` dropped | anonymous-declaration case |
| `distinctWrappers` counting `undefined` | no-wrapper case |
| barrier walk driven by `owned[0]` | source-order case |
| receiver match narrowed to `resolver.resolve` | Promise.resolve case + element-access case |
| a fourth `resolveCrossStackReads` above the wrap / the callee renamed | the new row / its floor |

Also in the round: the `scrub.ts` wrapper is inside a `try`, so "sits directly in the owner's body" was wrong; the function-boundary clause is an OVER-approximation (a helper called once reds) and is now labelled `KNOWN LIMIT` in both the comment and the failure message; the go-to-k/cdkd#2895 row's claim is narrowed to calls SPELLED `resolveCrossStackReads`, since an aliased binding passes.

One bullet of go-to-k/cdkd#2894's verification plan is deliberately unmet: the local-shadow mutant still passes. That is the KNOWN LIMIT documented at the wrapper match — closing it needs symbol resolution, which the issue itself offers as a branch to weigh rather than a requirement.

Full suite 971 files / 21256 passed / 1 skipped, rc=0, root asserted. Codex: 2 delta rounds and 3 maintainer-checklist proxy rounds on the round's delta, plus 3 rounds on the reply itself.

Closes #2894
Closes #2895

### DIFF
--- a/tests/unit/deployment/intrinsic-resolver-concurrent-drain.test.ts
+++ b/tests/unit/deployment/intrinsic-resolver-concurrent-drain.test.ts
@@ -1236,6 +1236,7 @@ describe('the callers that must wrap a resolve LOOP do (issue #2563)', () => {
     ['src/deployment/deploy-engine.ts', 'resolveOutputs', 2],
     ['src/cli/commands/import.ts', 'resolveImportedProperties', 2],
     ['src/cli/commands/export.ts', 'buildResolvedParametersPerStack', 1],
+    ['src/cli/commands/scrub.ts', 'resolveCrossStackReads', 3],
   ])('every %s call to %s is inside a shared drain budget', (file, callee, atLeast) => {
     // The wraps are this round's whole deliverable and nothing held them:
     // deleting either engine one left `tests/unit/deployment` and then the
@@ -1248,11 +1249,23 @@ describe('the callers that must wrap a resolve LOOP do (issue #2563)', () => {
     // does not -- its justification is the lock plus the CFn import
     // parameters it builds.
     //
-    // `scrub.ts` wraps its loops with ONE INLINE budget rather than at a
-    // callee, so it does not fit this callee-keyed table -- it gets the
-    // OWNER-KEYED case below instead. Calling it "not expressible" here was
-    // wrong, and that sentence is what let a third, unwrapped loop sit in
-    // that file for a round while both narratives said it was done.
+    // `scrub.ts` wraps its three resolve LOOPS with ONE INLINE budget rather
+    // than at a callee, so THOSE do not fit this callee-keyed table -- they
+    // get the OWNER-KEYED case below instead. Calling it "not expressible"
+    // here was wrong, and that sentence is what let a third, unwrapped loop
+    // sit in that file for a round while both narratives said it was done.
+    //
+    // Its cross-stack pre-pass DOES fit, and is the last row (issue
+    // go-to-k/cdkd#2895). `resolveCrossStackReads` is built by
+    // `makeCrossStackPrePass` at module scope and returned as a closure, so
+    // its own per-leaf `resolver.resolve` sits OUTSIDE `scrubStack`'s subtree
+    // and the owner-keyed case below cannot see it -- while its CALL SITES
+    // are inside `scrubStack` and are the property worth holding. Correct
+    // today because `withSharedDrainBudget` INHERITS, so a call already
+    // inside the hoisted budget spends that one. Before this row a fourth
+    // call site added above the wrap would have opened a budget per leaf with
+    // nothing reddening; that mutation is what this row now catches (measured:
+    // it reports the added line).
     //
     // Syntactic, like the fences above, and with the same limits: it asks
     // whether the call sits lexically inside a `withSharedDrainBudget(...)`
@@ -1317,6 +1330,147 @@ describe('the callers that must wrap a resolve LOOP do (issue #2563)', () => {
  * Uses the REAL resolver — `deploy-engine-outputs-export-name-collision.test.ts`
  * replaces it wholesale and so pins the copy's SEQUENTIAL shape only.
  */
+/**
+ * The owner-keyed walk, extracted so its CLAUSES can be pinned against
+ * synthetic sources as well as against the real `scrub.ts`.
+ *
+ * Both clauses issue go-to-k/cdkd#2894 added are invisible to a real-tree-only
+ * case, which is the same shape as the hole they close: deleting
+ * `ts.isFunctionLike` from the barrier, or stopping the barrier walk at the
+ * owner's DECLARATION rather than at its function, left the whole repository
+ * green while the cases below did not exist, because today's `scrubStack` is a
+ * function declaration whose wrapper sits directly in its body. The three
+ * cases below carry the shapes that tell the mutants apart, and the real-tree
+ * case stays GREEN under every one of them — which is the point.
+ */
+function analyseSharedBudget(
+  sourceText: string,
+  owner: string
+): {
+  resolveCount: number;
+  unwrappedLines: number[];
+  distinctWrappers: number;
+  perCallBarriers: string[];
+  ownerFound: boolean;
+} {
+  const sf = ts.createSourceFile('probe.ts', sourceText, ts.ScriptTarget.Latest, true);
+
+  // Find the OWNER's declaration and walk its whole subtree, rather than
+  // asking each call which function encloses it: the innermost-enclosing
+  // reading assigns a nested arrow's calls to the arrow, so an unwrapped
+  // `const fourth = async () => { await resolver.resolve(...) }` inside
+  // `scrubStack` walked straight past the first cut of this case
+  // (measured). `makeCrossStackPrePass` is a separate top-level function,
+  // so it stays out for free.
+  let ownerBody: ts.Node | undefined;
+  const findOwner = (node: ts.Node): void => {
+    if (ownerBody !== undefined) return;
+    const named =
+      (ts.isFunctionDeclaration(node) && node.name?.text === owner) ||
+      ((ts.isVariableDeclaration(node) || ts.isPropertyDeclaration(node)) &&
+        ts.isIdentifier(node.name) &&
+        node.name.text === owner &&
+        node.initializer !== undefined &&
+        (ts.isArrowFunction(node.initializer) || ts.isFunctionExpression(node.initializer)));
+    if (named) {
+      ownerBody = node;
+      return;
+    }
+    ts.forEachChild(node, findOwner);
+  };
+  findOwner(sf);
+  if (ownerBody === undefined) {
+    return {
+      resolveCount: 0,
+      unwrappedLines: [],
+      distinctWrappers: 0,
+      perCallBarriers: [],
+      ownerFound: false,
+    };
+  }
+
+  // The barrier walk below stops at the owner's FUNCTION node, not at its
+  // declaration: for `const scrubStack = async () => {...}` the arrow sits
+  // BETWEEN the wrapper and the VariableDeclaration, so stopping at the
+  // declaration would count the owner itself as a callback barrier and red a
+  // correct tree. Today `scrubStack` is a function declaration and the two
+  // coincide; the `const` arm of the const-arrow case below is what keeps this
+  // honest, because the real tree cannot exercise it.
+  const ownerDecl: ts.Node = ownerBody;
+  const ownerFn: ts.Node =
+    (ts.isVariableDeclaration(ownerDecl) || ts.isPropertyDeclaration(ownerDecl)) &&
+    ownerDecl.initializer !== undefined
+      ? ownerDecl.initializer
+      : ownerDecl;
+
+  const owned: { line: number; wrapped: boolean; wrapper: ts.Node | undefined }[] = [];
+  const walk = (node: ts.Node): void => {
+    // Any receiver spelling, deliberately: `const r = resolver; r.resolve()`
+    // and `resolver['resolve']()` both escaped a receiver-name filter
+    // (measured). Inside this subtree a `.resolve(` call is the resolver's
+    // until shown otherwise, and a false positive here is a loud test
+    // rather than a silent hole.
+    if (
+      ts.isCallExpression(node) &&
+      ((ts.isPropertyAccessExpression(node.expression) &&
+        node.expression.name.getText() === 'resolve') ||
+        (ts.isElementAccessExpression(node.expression) &&
+          ts.isStringLiteralLike(node.expression.argumentExpression) &&
+          node.expression.argumentExpression.text === 'resolve'))
+    ) {
+      // KNOWN LIMIT, and deliberate (issue go-to-k/cdkd#2894): the wrapper
+      // is matched by identifier TEXT, so a local
+      // `const withSharedDrainBudget = async (fn) => await fn()` shadowing
+      // the import would satisfy this walk while opening nothing. Closing it
+      // means introducing symbol resolution, which this fence does not use
+      // anywhere -- a new instrument for a shape nobody writes by accident.
+      // Named here rather than chased.
+      let wrapper: ts.Node | undefined;
+      for (let n: ts.Node | undefined = node.parent; n; n = n.parent) {
+        if (
+          ts.isCallExpression(n) &&
+          ts.isIdentifier(n.expression) &&
+          n.expression.text === 'withSharedDrainBudget'
+        ) {
+          wrapper = n;
+          break;
+        }
+      }
+      owned.push({
+        line: sf.getLineAndCharacterOfPosition(node.getStart(sf)).line + 1,
+        wrapped: wrapper !== undefined,
+        wrapper,
+      });
+    }
+    ts.forEachChild(node, walk);
+  };
+  walk(ownerBody);
+
+  // An iteration statement is only ONE way to lose "opened once per call of
+  // the owner". A wrapper inside a CALLBACK --
+  // `await Promise.all(xs.map(async () => withSharedDrainBudget(...)))` --
+  // opens one budget per element while matching no `ts.isIterationStatement`,
+  // and the earlier version of this clause passed it (issue
+  // go-to-k/cdkd#2894). So the barrier is any function boundary too: in the
+  // shape this fence exists to hold, the wrapper sits directly in the owner's
+  // own body, and a function-like node between the two means the budget is
+  // opened once per call of THAT function rather than of the owner.
+  const perCallBarriers: string[] = [];
+  for (let n: ts.Node | undefined = owned[0]?.wrapper?.parent; n && n !== ownerFn; n = n.parent) {
+    if (ts.isIterationStatement(n, false) || ts.isFunctionLike(n)) {
+      perCallBarriers.push(ts.SyntaxKind[n.kind]);
+    }
+  }
+
+  return {
+    resolveCount: owned.length,
+    unwrappedLines: owned.filter((c) => !c.wrapped).map((c) => c.line),
+    distinctWrappers: new Set(owned.map((c) => c.wrapper?.getStart(sf))).size,
+    perCallBarriers,
+    ownerFound: true,
+  };
+}
+
 describe('an INLINE budget is fenced by its OWNER, not by a callee (issue #2563)', () => {
   it('every `resolver.resolve` inside `scrubStack` is under a shared drain budget', () => {
     // `cdkd scrub` hoists ONE budget over its three resolve loops rather than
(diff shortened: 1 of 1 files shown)


go-to-k:
Reviewed at `eb70d23`. **Please push a fix round — I am not merging this yet.** Five items are required before merge (M1-M5), six are take-or-leave (m1-m6).

Every number in the PR body's mutation table reproduces exactly. I re-ran each mutant against a byte-exact copy of the PR head, restoring the production file after every probe:

| mutation | measured |
| --- | --- |
| `ts.isFunctionLike` deleted from the barrier | `1 failed \| 29 passed`, the callback case alone |
| barrier walk stopped at the owner's DECLARATION | `2 failed \| 28 passed`, both arrow-owner cases |
| `ts.isPropertyDeclaration` dropped from the normalisation | `1 failed \| 29 passed`, the class-field case alone |
| a fourth `resolveCrossStackReads` call added above the wrap | `1 failed \| 29 passed`, naming line 3790 |
| the hoisted wrapper in `scrub.ts` renamed (unwrap-equivalent) | `2 failed \| 28 passed`, `[3838, 3981, 4123]` and `[3840, 3985, 4127]` — the extraction did not weaken the real-tree case |

The new callee row is right: exactly three `resolveCrossStackReads` call sites (`src/cli/commands/scrub.ts:3838,3981,4123`), all lexically inside the wrap at `scrub.ts:3790`; the `const` binding at `:3710` and the named function expression at `:3072` are correctly not counted. CI is green on every check.

## Required before merge

**M1 — `ts.isIterationStatement(n, false)` is now completely unpinned.** `tests/unit/deployment/intrinsic-resolver-concurrent-drain.test.ts:1460`.

Measured: delete `ts.isIterationStatement(n, false) ||` from that line, leaving only `ts.isFunctionLike(n)`, and **all 30 cases stay green**. In every one of the four consumer shapes the loop sits *inside* the wrapper, so it is never on the upward path the barrier walks. `analyseSharedBudget` is module-local, so nothing else in the repository can reach the clause either.

This is verbatim the hole go-to-k/cdkd#2894 warned about in its Proposed fix — "The boundary check alone is not a replacement — a wrapper inside a `for` written directly in `scrubStack` has `scrubStack` as its enclosing function and would go green, turning today's loop mutants green with it. The two clauses catch different shapes and both are needed." — and its third verification bullet asked for exactly this probe.

One synthetic case closes it. Measured with it added: baseline `31 passed`, and the same mutant then gives `1 failed | 30 passed`.

```ts
const source = `
  async function scrubStack(ids: string[]) {
    for (const id of ids) {
      await withSharedDrainBudget(async () => {
        await resolver.resolve(id);
      });
    }
  }
`;
const found = analyseSharedBudget(source, 'scrubStack');
expect(found.perCallBarriers).toContain('ForOfStatement');
```

**M2 — go-to-k/cdkd#2894's verification item 4 has no test, and the reworded message now asserts the behaviour in prose.** `:1415-1419` (the matcher) and `:1499-1503` (the message).

The message now states "This walk matches ANY receiver by design, so an unrelated `.resolve(` in this function — a plain `Promise.resolve()`, say — reds here too". Nothing pins either arm. Measured: narrowing the matcher to `resolver.resolve` only — which also removes the whole `ts.isElementAccessExpression` arm — leaves **all 30 cases green**, because every source under test spells it `resolver.resolve(...)`.

The gap in the matcher itself predates this PR (the "escaped a receiver-name filter (measured)" comment is unchanged from `main`), but go-to-k/cdkd#2894 item 3 is one of the things this PR closes, and its verification plan asks for both arms: a `Promise.resolve()` outside the wrapper must red with the new message, one inside must stay green. Post-extraction that is two synthetic sources.

**M3 — the go-to-k/cdkd#2895 row's comment overclaims.** `:1239`, comment at `:1265-1268`.

"Before this row a fourth call site added above the wrap would have opened a budget per leaf with nothing reddening; that mutation is what this row now catches". Measured on the real `scrub.ts`:

- a fourth **direct** call above the wrap → RED, `unwrappedLines=[3790]` ✅
- the binding renamed at all three sites → RED via the floor, `count=0 < 3` ✅
- `const aliasedPrePass = resolveCrossStackReads; await aliasedPrePass(...)` added above the wrap → **`30 passed`, fully green** ❌

So the floor self-fences a *rename*, not an *added aliased site* — and an added site above the wrap is precisely the shape go-to-k/cdkd#2895 names. That issue asked the fix to "narrow what the fence claims" if it does not track the binding. Please narrow the sentence to calls *spelled* `resolveCrossStackReads`, or add the aliased probe.

**M4 — `:1342` and `:1455-1456` state something false about `scrub.ts`.** Both say the wrapper "sits directly in [the owner's] body". It does not: `scrub.ts:3790` sits inside the `try` opened at `scrub.ts:3573`. Measured ancestor chain — `AwaitExpression → ExpressionStatement → Block@3573 → TryStatement@3573 → Block@3553 → FunctionDeclaration@3532`. The conclusion survives (a `try` is neither an iteration statement nor function-like), but the stated reason is wrong. "with no iteration or function boundary between the two" is the accurate phrasing.

**M5 — `:1456-1457`'s semantic claim is false as written.** "a function-like node between the two means the budget is opened once per call of THAT function rather than of the owner." Measured counter-example, correct code that the clause reds:

```ts
async function scrubStack(ids) {
  const run = async () => {
    await withSharedDrainBudget(async () => {
      for (const id of ids) await resolver.resolve(id);
    });
  };
  await run();
}
```

→ `perCallBarriers: ["ArrowFunction"]`, though the budget is opened exactly once per `scrubStack` call. Same for an IIFE. The clause is a deliberate over-approximation and a good one; the file labels its other over-approximations `KNOWN LIMIT` (`:1421`) and this one is unlabelled. Restate as "the fence can no longer tell how many times it is opened, so it reds", plus a `KNOWN LIMIT` note.

## Take-or-leave

**m1 — `:1369`: a bodyless `FunctionDeclaration` is accepted as the owner.** Measured: `declare function scrubStack(...)` (or an overload signature) gives `{resolveCount: 0, ownerFound: true}` — `findOwner` stops at the signature and walks an empty subtree. The `>= 3` floor still reds, so it is not a hole, but `ownerFound` is a lie. `&& node.body !== undefined` fixes it.

**m2 — `:1468`: `distinctWrappers` counts `undefined` as a member.** Measured: an owner with one unwrapped resolve and no wrapper anywhere returns `distinctWrappers: 1`, which reads as "one shared wrapper". Masked today by `unwrappedLines` asserting first. Filter `undefined` before the `Set`.

**m3 — `:1459`: the barrier walk is driven by `owned[0]` alone.** If the first `.resolve(` in source order is unwrapped, the start point is `undefined` and the barrier check silently becomes vacuous — safe only because `unwrappedLines` and `distinctWrappers` red first. Driving it from the single wrapper that `distinctWrappers === 1` already implies removes the ordering dependency.

**m4 — `:1399`: leftover from the extraction.** `ownerDecl` is a pure alias of `ownerBody`, and `ownerBody` is a misnomer — it holds the *declaration*, and `walk(ownerBody)` at `:1447` walks the declaration subtree. One correctly-named binding would do.

**m5 — `:1315-1332`: the "engine `Export.Name` copy" JSDoc now documents `analyseSharedBudget`.** The misplacement is pre-existing on `main` (it heads the INLINE-budget describe while describing the one two blocks down), but inserting the helper into that gap stacks two JSDoc blocks on one function. Same file, no behaviour risk — worth moving down to `describe('the engine Export.Name copy sees a concurrent sibling record ...')` while you are here.

**m6 — `:1370`: `MethodDeclaration` is an asymmetric omission.** Now that a class-field arrow owner is accepted, `class Scrubber { async scrubStack() {...} }` still reports `ownerFound: false`. It reds loudly on the first assertion so there is no hole, but the finder's coverage now looks arbitrary. Also: `distinctWrappers` in the two negative-control cases (`:1571`, `:1597`) is implied by `resolveCount === 1 && unwrappedLines === []` and is decorative there.

Nothing else — the extraction preserved every real-tree assertion one for one, the `KNOWN LIMIT` for the wrapper-shadow is documented where a reader hits it and nothing claims it closed, and no false negative reachable by an ordinary refactor survives the three assertions in combination.

Please push M1-M5 (M1 is the one that changes coverage; M2-M5 are a case pair and three sentences), take or leave m1-m6, and ping me — I will re-review and merge.

nix-tkobayashi:
All eleven taken, pushed as `4e338e7d`. 40 cases now, baseline green.

**M1 is yours and the miss is mine, in a way worth stating plainly.** The clause was unreachable in every shape under test because in all of them the loop sits INSIDE the wrapper, never on the upward path — your probe reproduces here. What makes it worse than an oversight: go-to-k/cdkd#2894's Proposed fix says "The boundary check alone is not a replacement ... The two clauses catch different shapes and both are needed", and its Verification plan asked for exactly this probe. I never read either. I pulled the issue body through `.body | split("\n") | .[0:22]` to save context, and the tail I cut held the Proposed fix and the whole Verification plan. M2 was a bullet in that same tail; M3's alias requirement is in go-to-k/cdkd#2895's plan, which I had also read only in part. That is now a standing rule for me: an issue, PR or review body I am about to act on gets written to a file and read whole, never sampled.

Working from both plans in full this time, the round covers every bullet in them EXCEPT one, deliberately: go-to-k/cdkd#2894 asks for the local-shadow mutant to RED, and it still passes. That is the KNOWN LIMIT you accepted — closing it needs symbol resolution, which the same issue offers as a branch to weigh rather than a requirement. Everything else in both plans is now measured:

**M1** — a `for..of` written directly in the owner, where only the iteration clause can report. Measured: before this case, deleting `ts.isIterationStatement(n, false) ||` left all 30 green; now that mutant reds this case and the source-order one below it, and nothing else.

**M2** — three cases, not two: `Promise.resolve()` outside the wrapper reds, one inside does not, and `resolver['resolve']()` outside reds. The element-access arm needed its own, because the narrowing mutant you describe drops that arm wholesale.

**M3** — narrowed, not widened. The comment now says the row holds every call SPELLED `resolveCrossStackReads`, names the aliased shape as passing, and says why tracking the binding is declined: it is the same symbol-resolution instrument the wrapper-shadow KNOWN LIMIT already declines. go-to-k/cdkd#2895 offered exactly this branch.

**M4** — reworded to "no iteration or function boundary between the two". Your ancestor chain reproduces; the `try` at `scrub.ts:3573` is what the old wording ignored.

**M5** — labelled `KNOWN LIMIT` and restated, and the FAILURE MESSAGE with it: it now says the fence can no longer read how many times the budget is opened, so it refuses both spellings, rather than asserting per-iteration. Your once-called-helper counter-example is what the old message was false about.

**m1-m6, all taken**, and m1 grew a second half under probing: rejecting only `declare` passes every case while an OVERLOAD signature captures the owner from its implementation and reds a correct tree, so the test is written against the invariant — the owner is the declaration that HAS a body — with an overload-plus-implementation case. While pinning the name test I also split it: `ts.isIdentifier` is pinned by a quoted-method case, and the separate `node.name !== undefined` guard by an anonymous `export default function () {}` control, since `FunctionDeclaration.name` is optional and an unguarded `ts.isIdentifier(undefined)` throws before the real owner is reached.

Every clause now has a mutant that reds it and nothing else:

| mutation | reds |
| --- | --- |
| `ts.isIterationStatement(n, false) \|\|` deleted | loop case + source-order case |
| `ts.isFunctionLike(n)` deleted | callback case |
| barrier walk stopped at the DECLARATION | both arrow-owner cases |
| `ts.isPropertyDeclaration` dropped from the normalisation | class-field case |
| body test replaced by a `declare`-only rejection | overload case |
| `ts.isIdentifier(node.name)` dropped | quoted-method case |
| `node.name !== undefined` dropped | anonymous-declaration case |
| `distinctWrappers` counting `undefined` again | no-wrapper case |
| barrier walk driven by `owned[0]` again | source-order case |
| receiver match narrowed to `resolver.resolve` | Promise.resolve case + element-access case |
| a fourth `resolveCrossStackReads` above the wrap / the callee renamed | the new row / its floor |

Two notes on measurement hygiene, since both bit me this round. Every "measured" sentence in the file is now temporal — a tally written before the next case exists goes false the moment it lands, which happened twice; the delta review caught both, not me. And this file's RUNTIME cases fail intermittently under host load: in two separate probe batches one unrelated runtime case failed once and did not reproduce on re-run (`the cap timer is cleared on the way out`, and `nested drains waiting CONCURRENTLY charge the budget once`). Neither is tied to a mutation — each failed under a mutant that cannot reach it, and passed on the next run of the same tree. I have treated both as load artifacts rather than discrimination; say the word if you would rather have that filed as its own issue.

Verification: full suite 971 files / 21256 passed / 1 skipped, rc=0, run root asserted — one earlier run of the same tree reported the SAME counts with rc=1 and a single `Worker exited unexpectedly`. A peer session was running its own suite from another worktree throughout, and still was during the clean re-run (two of its vitest processes recorded immediately before it), so the artifact is intermittent rather than a function of the peer being present. `vp check --fix` + `vp run check` + `typecheck:test` + `build` + `gen:all-matrices` + `audit:coverage:check` all rc=0. Codex: 2 delta rounds and 3 proxy rounds. The delta review found the two stale tallies; the proxy pass then found six more — four unpinned clauses, the M5 wording surviving in the failure message after the comment was fixed, and the m4 alias that had been renamed rather than removed.

`main` has moved six commits since the review; none touches this file, and GitHub still reports the PR mergeable, so I have not rebased.

go-to-k:
Merged as `51e7af7c`, both issues auto-closed. Thank you — all thirteen mutants in your table reproduce here exactly, each reddening the cases you name and nothing else, and the real-tree case still reds when the hoisted wrapper is renamed away. 40 cases, baseline green, full suite 971 / 21256 / 1 skipped rc=0 from my own tree.

Two things this round did that are worth naming.

**You extended the fix past what the finding asked for, in the direction the finding implied.** M1 was one missing case; you added the source-order case beside it, so the iteration clause is now pinned twice from different angles. m1 was "reject a bodyless declaration"; you found under probing that a `declare`-only rejection passes every case while an OVERLOAD signature reds a correct tree, and wrote the test against the invariant instead — the owner is the declaration that HAS a body. Splitting the name guard into the quoted-method case and the anonymous-declaration control is the same move: `ts.isIdentifier(node.name)` and `node.name !== undefined` are two clauses and now have two mutants, and the second one throws rather than returning false, which a single combined case would have hidden.

**The prose was corrected everywhere it appeared, not just where I quoted it.** M5's over-approximation is labelled `KNOWN LIMIT` in the comment AND restated in the failure message — that second half is the one that usually survives a round, since nobody re-reads a message after fixing the comment above it.

On the flakes: no new issue — go-to-k/cdkd#2907 already owns this file and the same family, and I have [added both cases there](https://github.com/go-to-k/cdkd/issues/2907#issuecomment-5620882395). They are a sub-shape it did not name: an ABSOLUTE wall-clock ceiling (`toBeLessThan(100)`, `toBeLessThan(420)`) rather than a ratio of two measured durations, which is if anything the weaker form — a ratio at least normalises against the machine. The remedy that issue proposes covers them unchanged.

The truncation rule you drew for yourself is the right one, and generalises to the reviewing side too: an issue that proposes a fix usually also says how to prove it, so its Verification plan is the cheapest checklist to run against a PR that closes it. All three required items this round came from bullets that were already written down.

Issue #2914, opened 2026-09-10
fix(rollback): decide whether a rollback should re-adopt a DeletionPolicy: Retain resource into state


go-to-k:
## Summary

Split out of go-to-k/cdkd#2902, whose reporter offered two remedies. That issue
is being fixed with the FIRST one — on a redeploy collision, say that cdkd
itself orphaned the resource via an earlier `Retain` rollback and name
`cdkd import` with the physical name cdkd already computed. This issue tracks
the SECOND: having a rollback **re-adopt** a `DeletionPolicy: Retain` resource
into state instead of only removing it, so the next deploy sees an existing,
unchanged resource rather than attempting a colliding CREATE.

They are split because this half is a design decision about how far cdkd
departs from CloudFormation, and because it is **not implementable
symmetrically today**.

## Why it is not just "stop deleting the record"

There are TWO orphan arms in `src/deployment/rollback-executor.ts`, and only
one of them still knows what it orphaned.

- `case 'orphan-retain'` — a COMPLETED create being rolled back. `:1877` binds
  the record, `:1880` deletes it, and `:1906-1912` publishes `physicalId` plus
  a `reason` into `ROLLBACK_RESOURCE_SUCCEEDED`. The in-code comment at
  `:1893-1895` states that this event is the ONLY place the retained
  resource's id survives.
- `case 'orphan-failed-create-retain'` — a FAILED in-flight create under the
  same policy. Classified at `:868`, arm at `:3034-3065`, record deleted at
  `:3050`, and its event at `:3056-3063` carries **no `physicalId` and no
  `reason`**. The orphan is untraceable even in the event stream.

The rollback journal does not close the gap either:
`settleJournalAfterCleanRollback` (`src/deployment/deploy-engine.ts:3614-3644`)
pops the segment after a clean rollback and either deletes the journal
(`:3619-3627`) or re-records it with `operations: []` (`:3638-3644`),
discarding the completed CREATE that held the id.

So re-adoption needs one of: a documented read of the `deployments/` event
stream, or a new durable orphan record — plus a fix for the second arm, which
records nothing to read.

## Why this is a design decision, not a bug fix

CloudFormation also leaves a `Retain` resource outside the stack, so removing
the record matches CFn semantics and cdkd is template-compatible by intent.
What differs is the OTHER half: CFn's auto-generated physical names carry a
random suffix, so a retained orphan never collides with a later deploy, while
[...]

Issue #2901, opened 2026-09-10
fix(deployment): Cognito UserPool create races its just-created SmsRole's trust-relationship propagation — no IAM_PROPAGATION_ERROR_MESSAGE_PATTERNS entry catches it


yama3133:
## Summary

Deploying a CDK stack containing a Cognito UserPool with an MFA/SMS
configuration (the default `UserPool` L2 auto-creates an SNS-publish IAM role
— `smsRole`) via `cdkd deploy` fails when the UserPool CREATE races the
just-created SMS role's IAM trust-relationship propagation. The AWS error is
`InvalidSmsRoleTrustRelationshipException`, and none of the existing
`IAM_PROPAGATION_ERROR_MESSAGE_PATTERNS` entries in
`src/deployment/retryable-errors.ts` match its wording, so the create fails
outright instead of being retried on the dense IAM-propagation schedule.

## Reproduction steps

1. Deploy a CDK stack that includes `aws-cdk-lib/aws-cognito`'s `UserPool` L2
   (with its default-generated SMS role) alongside other resources, via
   `cdkd deploy <stack> --yes`.
2. cdkd creates the SMS role (`UserIdentityUserPoolsmsRole...`,
   `AWS::IAM::Role`) and, per the structured event log, issues the Cognito
   `CreateUserPool` call only 336ms later — before the role's trust policy has
   propagated to Cognito's assume-role check.
3. The create fails.

## Actual error

Raw CLI output:

```
Failed to create UserIdentityUserPoolA179CDAC: Failed to create Cognito User Pool UserIdentityUserPoolA179CDAC: Role does not have a trust relationship allowing Cognito to assume the role
```

Structured event (`cdkd events <stack> --run <runId> --format json`):

```json
{
  "eventType": "RESOURCE_FAILED",
  "logicalId": "UserIdentityUserPoolA179CDAC",
  "resourceType": "AWS::Cognito::UserPool",
  "provisionedBy": "sdk",
  "durationMs": 336,
  "error": {
    "name": "ProvisioningError",
    "awsErrorCode": "InvalidSmsRoleTrustRelationshipException",
    "message": "Failed to create Cognito User Pool UserIdentityUserPoolA179CDAC: Role does not have a trust relationship allowing Cognito to assume the role"
  }
}
```

Reproduced twice: the first `cdkd deploy` attempt, and a retry 60 seconds
later (see the companion rollback issue — the second and third retries after
that failed for a different, downstream reason before this API call was
[...]

Issue #2610, opened 2026-09-05
fix: 14 user-facing remedy messages name a flag or a re-run whose precondition the code never checks (sibling sweep of go-to-k/cdkd#2579)


go-to-k:
`LogsLogGroupProvider`'s `LogGroupClass` refusal advised `--replace --force-stateful-recreation` without checking whether the log group carried `DeletionProtectionEnabled`, so the advised command could not succeed. That was fixed for that one site (issue go-to-k/cdkd#2579). A sweep for the SHAPE rather than the name -- a user-facing message naming a concrete remedy whose precondition the emitting code never checks -- found **14 more sites** in two families.

Ground facts, verified against the tree rather than remembered:

- `cdkd deploy` does NOT accept `--remove-protection`. `deployOptions` (`src/cli/options.ts`) registers no such Option. The three commands that do register it are `cdkd destroy` (via `destroyOptions`), `cdkd state destroy` (its own `.option()`), and `cdkd orphan` (inherited, and inert -- `orphan` never calls `provider.delete()`).
- `src/deployment/**` contains ZERO occurrences of `removeProtection`. Every `DeleteContext` the deploy engine builds sets only `expectedRegion` / `forceDataDelete` / `finalSnapshotIdentifier`. So every deploy-side replacement DELETE runs with `removeProtection` unset, which is what makes family (a) real rather than theoretical.

## Family (a): the advised replacement is blocked by a protection flag the code never reads

Each of these emits replacement advice while the same provider (or the engine) holds a bag that could answer whether the resource is protected.

| # | Site | Unchecked precondition | Gate that would clear it |
| --- | --- | --- | --- |
| 1 | `src/provisioning/providers/elbv2-provider.ts:773` | `LoadBalancerAttributes` carrying `deletion_protection.enabled=true` (already in `handledKeys` in the same function) | `elbv2-provider.ts:1074` |
| 2, 3 | `src/provisioning/providers/emr-cluster-provider.ts:517`, `:532` | `Instances.TerminationProtected` -- read one line above the throw | `emr-cluster-provider.ts:754` |
| 4 | `src/provisioning/providers/cognito-provider.ts:2286` | `DeletionProtection === 'ACTIVE'` | `cognito-provider.ts:2336`, whose JSDoc already states the dead-end verbatim |
[...]

Issue #2038, opened 2026-08-19
fix(deployment): the rollback replay threads an UNMASKED logger into withRetry, so an exhausted retry can print a resolved secret


go-to-k:
## Summary

`src/deployment/rollback-executor.ts` resolves state secrets back to PLAINTEXT for a replay and then threads the **raw** `Logger` into `withRetry`, so the give-up summary's verbatim AWS message can print an unmasked secret at `warn` level — the exact hole `src/cli/commands/drift.ts` closes for the drift-revert path.

Found by a security review during https://github.com/go-to-k/cdkd/issues/2026's PR. **Pre-existing, not introduced there** — that PR adds `name` / `httpStatusCode` / `requestId` to the same line, and all three were traced and confirmed non-sensitive. The exposure is the AWS **message**, which has been interpolated into this line since https://github.com/go-to-k/cdkd/issues/2018.

## Mechanism

- `rollback-executor.ts:1548-1549` re-resolves the secret bag to plaintext for the replay.
- `rollback-executor.ts:1568` constructs the `withRetry` options with `ctx.logger` directly.
- `retry.ts` emits `opts.logger?.warn?.(summary())`, and `summary()` interpolates the raw AWS `message`.
- AWS commonly quotes the offending value back: `Value 'hunter2-real' at 'password' failed to satisfy constraint ...`.

So an IAM-propagation (or now 5xx) sequence that exhausts during a rollback replay prints the secret to stdout at DEFAULT verbosity.

`src/deployment/deploy-engine.ts:4631` threads `this.logger` on the same shape and deserves the same audit, though the deploy path does not do the plaintext re-resolution that makes the rollback path acute.

## The established remedy

`drift.ts:2768-2771` already shows it — thread a masking logger, masking the CONCATENATED string:

```ts
{
  debug: (m) => logger.debug(maskSecretsInText(m, secrets)),
  warn:  (m) => logger.warn(maskSecretsInText(m, secrets)),
}
```

Note `RetryLogger.warn` is optional precisely so this masking is possible; a required `warn` would have been silently satisfied by an unmasked one (GHSA-tied, https://github.com/go-to-k/cdkd/issues/1914).

## Acceptance

1. Every `withRetry` caller that has a resolved secret bag in scope threads a masking `warn` AND `debug`.
[...]

PR #2916, opened 2026-09-10
fix(deploy): name the orphan a colliding CREATE hit, and the command that recovers it

### LINKED ISSUE #2902: fix(rollback): an automatic rollback's DeletionPolicy: Retain resources are orphaned from state under a deterministic physical name, permanently blocking redeploy of the same stack
## Summary

When `cdkd deploy` fails and triggers an automatic rollback, resources
carrying `DeletionPolicy: Retain` are (correctly, per CloudFormation
semantics) left in AWS and removed from cdkd's tracked state. But because
cdkd generates deterministic physical names (`<StackName>-<LogicalId>`,
per `docs/troubleshooting.md`), the *next* `cdkd deploy` of the same stack
tries to create a new resource under the exact same physical name and fails
with `EntityAlreadyExists`. That failure triggers another automatic rollback,
which reproduces the same orphaned-Retain situation, so the stack cannot be
redeployed at all until an operator manually finds and deletes the orphaned
AWS resources outside of cdkd. No error message at any point in the loop
identifies the cause or points at a remedy (e.g. `cdkd import`).

## Reproduction steps

1. Deploy an 88-resource CDK stack containing several `RemovalPolicy.RETAIN`
   resources (CDK L2 defaults for KMS keys, and two IAM roles —
   `ApiGatewayAccountCloudWatchRole`, `UserIdentityUserPoolsmsRole`) via
   `cdkd deploy <stack> --yes`, and have the deploy fail partway through so
   that an automatic rollback runs. (In our case the trigger was a separate
   filed issue — the Cognito SmsRole/UserPool timing race — but any
   mid-deploy failure that unwinds past a `Retain` resource reproduces this.)

2. The automatic rollback logs (raw CLI output):

   ```
   Rollback: Leaving ApiLogsKeyEFD45BEB (AWS::KMS::Key) in AWS (DeletionPolicy: Retain) — removed from state
   Rollback: Leaving UserIdentityUserPoolsmsRole6CF6EC5B (AWS::IAM::Role) in AWS (DeletionPolicy: Retain) — removed from state
   Rollback: Leaving AgentServiceAgentSessionBucket428B1999 (AWS::S3::Bucket) in AWS (DeletionPolicy: Retain) — removed from state
   Rollback: Leaving AgentServiceAgentSessionKey5BE85384 (AWS::KMS::Key) in AWS (DeletionPolicy: Retain) — removed from state
   Rollback: Leaving WebsiteWebsiteKeyC399DC31 (AWS::KMS::Key) in AWS (DeletionPolicy: Retain) — removed from state
   Rollback: Leaving ApiGatewayAccountCloudWatchRoleFD29A2DA (AWS::IAM::Role) in AWS (DeletionPolicy: Retain) — removed from state
   ```

3. Run `cdkd deploy <stack> --yes` again, with no manual AWS changes in
   between. It fails immediately on the two IAM roles, whose physical names
   are deterministic and therefore already taken:

   ```
   Failed to create ApiGatewayAccountCloudWatchRoleFD29A2DA: Failed to create IAM role ApiGatewayAccountCloudWatchRoleFD29A2DA: Role with name nx-cdkd-race-infra-sandbox-Application-ApiGatewayAccoun-19184149 already exists.
   Failed to create UserIdentityUserPoolsmsRole6CF6EC5B: Failed to create IAM role UserIdentityUserPoolsmsRole6CF6EC5B: Role with name nx-cdkd-race-infra-sandbox-Application-UserIdentityUser-f4ecf9cd already exists.
   ```

   Structured event:

   ```json
   {
     "eventType": "RESOURCE_FAILED",
     "stackName": "nx-cdkd-race-infra-sandbox-Application",
     "operation": "CREATE",
     "logicalId": "ApiGatewayAccountCloudWatchRoleFD29A2DA",
     "resourceType": "AWS::IAM::Role",
     "provisionedBy": "sdk",
     "durationMs": 680,
     "error": {
       "name": "ProvisioningError",
       "message": "Failed to create IAM role ApiGatewayAccountCloudWatchRoleFD29A2DA: Role with name nx-cdkd-race-infra-sandbox-Application-ApiGatewayAccoun-19184149 already exists.",
       "awsErrorCode": "EntityAlreadyExists",
       "requestId": "d115692e-e9f0-4399-92fa-71643f575f17"
     }
   }
   ```

4. That failure triggers another automatic rollback (no new Retain resources
   this time, since the create itself never succeeded). A third
   `cdkd deploy` attempt, 60 seconds later, reproduces the identical
   `EntityAlreadyExists` failure. We stopped at 3 attempts; the loop does not
   self-resolve with time or retries alone.

5. The stack became redeployable again only after manually detaching
   policies from and deleting the two orphaned IAM roles via the AWS IAM API
   directly (and separately scheduling deletion of the orphaned KMS keys /
   deleting the orphaned S3 bucket). `cdkd deploy`, `cdkd destroy`, and
   `cdkd state list` were not enough to unblock this on their own: `cdkd
   destroy` only acts on what is in state, which no longer includes these
   resources after the rollback removed them, and nothing in the CLI output
   at any point named the collision or suggested `cdkd import` as the fix.

## What I checked before filing

`docs/troubleshooting.md` ("Recovering from Orphaned Resources") documents
that cdkd's generated physical names are deterministic
(`<StackName>-<LogicalId>`) by design, for state-recovery/import purposes,
and warns that redeploying after a *manually deleted* `state.json` hits
`EntityAlreadyExists` for most resource types. That is a different trigger
than what we hit: here, state was never deleted or corrupted — the
automatic rollback itself (a normal, working feature) is what orphans the
`Retain` resource from state, on every deploy attempt that fails past it, and
the resulting collision on redeploy is not covered by that guidance or
surfaced in the CLI's own error/remedy messages.

For comparison, a plain `cdk deploy` (standard CDK CLI) against the same
template did not hit this: `cdk destroy --all` also leaves `Retain`
resources behind (confirmed: same KMS keys / IAM roles / S3 bucket / Cognito
UserPool), but standard CloudFormation's auto-generated physical names
include a random suffix for most of these resource types, so a
Retain-orphaned resource from a prior stack never collides with a later
deploy's new resource of the same logical ID.

## Expected behavior

Not prescribing a specific implementation, since we don't know cdkd's
intended design here, but one of:

- The `EntityAlreadyExists` (or equivalent) failure on redeploy should
  identify that the collision is with a resource cdkd itself orphaned via an
  earlier `Retain` rollback, and suggest a concrete remedy (e.g. `cdkd
  import` with the physical name cdkd already computed), instead of
  surfacing the bare AWS SDK error with no further context.
- And/or: an automatic rollback could re-adopt (import) a
  `DeletionPolicy: Retain` resource back into state instead of only removing
  it, so the next deploy sees it as an existing, unchanged resource rather
  than attempting a colliding CREATE.

## Environment

- `@go-to-k/cdkd`: 0.288.3
- `aws-cdk-lib`: 2.268.0
- `aws-cdk` (CLI; used only for the `cdk deploy` comparison, not `cdkd`): 2.1139.0
- Node.js: v24.15.0
- Region: ap-northeast-1
- Stack: 88 resources (Cognito UserPool, API Gateway, Lambda, CloudFront, WAF, KMS, AgentCore Runtime, among others)
- Auth: IAM user via `AWS_PROFILE` (not a CDK-bootstrapped deploy role — cdkd does not go through CDK bootstrap's deploy-role/lookup-role)

## Related issues found while searching (not exact duplicates)

- [#1407](https://github.com/go-to-k/cdkd/issues/1407) (closed/released) — a
  similar `EntityAlreadyExists` collision on a `Retain`
  `AWS::ApiGateway::Account` CloudWatch role, but caused by re-running the
  *same integ fixture* back-to-back (`cdkd destroy` correctly leaving a
  `Retain` resource, then a fresh `cdkd deploy` of the same template) — not
  by an automatic rollback. The fix there was to change the test fixture
  (`cloudWatchRole: false`), not cdkd itself, so it doesn't cover the
  rollback-triggered path reported here.
- [#2598](https://github.com/go-to-k/cdkd/issues/2598),
  [#2603](https://github.com/go-to-k/cdkd/issues/2603),
  [#2886](https://github.com/go-to-k/cdkd/issues/2886) — several
  open/closed issues about rollback's handling of `Retain` /
  `UpdateReplacePolicy: Retain` for *replacement* rollbacks specifically;
  none address the deterministic-physical-name redeploy collision on a
  plain CREATE rollback reported here.
- [#2018](https://github.com/go-to-k/cdkd/issues/2018) — the report that
  surfaced the same "rollback leaves a `RETAIN` + deletion-protected Cognito
  UserPool behind" symptom we saw, in a different (Lambda-triggered)
  reproduction; it treats the retained resource as expected behavior in
  isolation and doesn't address the redeploy-blocking consequence.
- #2901 — the Cognito SmsRole propagation race that triggered this rollback
  in our reproduction.

Dup-check: searched open issues for EntityAlreadyExists / orphan / Retain / deterministic-physical-name concepts -- #1710 (a throw AFTER a successful create orphans the resource) and #2631 (a replacement cleanup delete leaving an unrecorded survivor) are adjacent orphan classes with different triggers, and #2018 reports the same retained-UserPool symptom without the redeploy-blocking consequence. None covers this root cause.

go-to-k:
## What

A deploy that collides with a resource cdkd itself left behind now says so, and prints the command that recovers it.

Reported by an external user, as the follow-on from the deploy failure in go-to-k/cdkd#2901: their stack became permanently un-redeployable, and the only way out they found was hand-deleting resources through the AWS API.

## The loop

A resource carrying `DeletionPolicy: Retain` stays in AWS when a deploy rolls back and its state record is dropped. That is CloudFormation's behaviour too, and it is what `Retain` is for. What differs is the other half: cdkd's generated physical names carry no random component (`generateResourceName` is sha256-derived, nothing random), so the next `cdkd deploy` asks AWS for a name the orphan still holds, fails `EntityAlreadyExists`, rolls back again, and repeats. Re-running never resolves it.

CloudFormation never shows this — its generated names carry a random suffix, so a retained orphan cannot collide with a later deploy. **The breakage is that combination, not either half**, which is why this PR ships a diagnosis rather than a behaviour change.

## Scope: the first of the reporter's two remedies

The issue offered two. This is the first — name the cause and the remedy. The second, having a rollback **re-adopt** the retained resource into state, is split out as go-to-k/cdkd#2914, for two reasons:

- it is a design decision about how far cdkd departs from CloudFormation semantics, and it should be made against the measured behaviour of *this* remedy rather than ahead of it;
- it is **not implementable symmetrically today**. The completed-CREATE orphan arm publishes the retained `physicalId` into its `ROLLBACK_RESOURCE_SUCCEEDED` event (`rollback-executor.ts:1906-1912`), but the failed-in-flight-CREATE arm publishes neither an id nor a reason (`:3056-3063`), and `settleJournalAfterCleanRollback` discards the record that held it. Re-adoption would need a durable orphan record first.

## The refusals are most of the change

`orphanedNameCollisionAdvice` returns `undefined`, or falls back to a delete-only arm, for every case it cannot vouch for:

- **A name the TEMPLATE supplied gets nothing.** That resource may be someone else's entirely, and `cdkd import` would be advice to adopt it.
- **A type whose provider implements no `import` is told to delete instead.** `runImportForResource` SKIPS such a type (`skipped-no-impl`) rather than failing, so a blanket recommendation would route the user through a command reporting "provider does not implement import" and leave them where they started — issue go-to-k/cdkd#2610's class. The precondition is checked before the command is named, using the same `getProvider` call `cdkd import` itself makes.
- **A NESTED-STACK child is told to delete too.** Its stack name is `<parent>~<child>`, and CDK's own stack-name rule bars `~`, so no Cloud Assembly stack can ever carry it and the printed import could never resolve — the same go-to-k/cdkd#2610 class one level down.

### A derived name is not proof the resource is yours

The first revision asserted the orphan outright. Review showed that is wrong in two reachable ways, and the message now names the orphan as the *likely* case and asks the reader to confirm ownership first:

- a **globally-namespaced** type can collide with **another account's** resource, and because cdkd's derivation is predictable that name can be pre-registered by someone else. For `AWS::S3::Bucket` this is not hypothetical: the own-account collision resolves elsewhere, so the path that reaches this advice *is* the cross-account one — precisely where "an earlier cdkd run left it behind" is false;
- the same stack name deployed in **two regions** derives the same name for a global type, so the collision is with a live resource another state file already owns. Importing it would give two stacks one resource, and either `cdkd destroy` would then delete it out from under the other.

### The command is quoted, because it is meant to be pasted

`looksLikeCdkdGeneratedName` compares only the ALPHANUMERIC skeleton — it strips everything else before comparing — so a name whose extra characters are all shell metacharacters passes the guard. The command now takes this repo's established sanitize / quote / suppress treatment (`replacement-protection-advice.ts` is the precedent). The two halves do different jobs: `shellQuote` neutralises metacharacters by single-quoting outside a conservative safe set, and `displaySafe(asciiOnly)` + SUPPRESS covers what quoting cannot make safe to print at all.

## Verified live, end to end

The part a unit test cannot reach is whether the printed remedy actually *works*. Against real AWS:

```
deploy                → role created under a cdkd-derived name
state orphan          → record dropped, role still live in AWS
redeploy              → FAILS already-exists, AND prints the diagnosis
<the command the message printed>  → 1 imported, 0 failed
redeploy              → rc=0, "Unchanged: 1"      ← the loop is broken
```

That sequence is now the fixture `tests/integration/retain-orphan-redeploy`. It **parses the import argument out of the message** rather than hard-coding one, so the assertion is about cdkd's advice and not about a command the fixture author believed in.

Two premises the fixture pins, because both are load-bearing and neither is obvious:

- the role must have **no `roleName`** — otherwise the name is template-supplied and the diagnosis correctly refuses to fire, and the fixture would assert nothing;
- it must be an **IAM role**. `AWS::S3::Bucket`, `AWS::Logs::LogGroup` and `AWS::SNS::Topic` silently ADOPT an existing resource instead of refusing, so a fixture built on one of those would redeploy green.

Probe-verified: with the message removed from the built bundle the fixture exits 1 and reports that the AWS error is present while the diagnosis is not — distinguishing a reworded diagnosis from an absent condition, rather than going quietly green.

## Unit fences, and what is honestly NOT one

Most of the unit cases are refusals. Every branch the diff adds was probed, and where a probe did not discriminate the code now says so instead of reading as a fence:

- the **`physicalId` guard** is subsumed at runtime by `looksLikeCdkdGeneratedName`'s own falsy check (deleting either alone stays green; deleting both reds). It stays for type narrowing.
- the **`getProvider` catch** is unreachable for the type that got there — `provisionResource` already called `getProvider` with that same type before the create, to read its timeout floor, so a throwing lookup fails the deploy earlier. A test for it was written and then **dropped rather than shipped**, with the reason recorded at the catch.

Two corrections review forced, both of the same class:

- the **CREATE guard** was described as unreachable defence-in-depth. It is not: a replacement collision *does* reach this catch (the `NAMED_REPLACEMENT_COLLISION` throws happen inside `provisionResourceBody`, called inside the same `try`, and this PR's own test asserts their line was logged). What refuses them is the `ProvisioningError` check, since they throw `CdkdError` — and that makes the CREATE guard live, for a non-CREATE `ProvisioningError` carrying `already exists`. Three of the four reviewers found this independently.
- the **quoting test** originally asserted the command was SUPPRESSED for a `$()` payload. Wrong measurement: `shellQuote` quotes it, which is the stronger outcome, and that is what is asserted now.

The replacement case was itself rewritten after the first version passed with the CREATE guard deleted — it had been driving no replacement at all (empty state map). One probe was VOID rather than non-discriminating (it targeted the wrong file); re-run against the right one, it reds.

## Docs

`docs/troubleshooting.md` gains the rollback-orphan case — the section previously covered only a *manually deleted* `state.json`, which is a different trigger from the one users actually meet.

Its physical-name description is also corrected in both places it appeared. It stated `<StackName>-<LogicalId>` flatly; that form holds only when it fits the type's length limit, otherwise cdkd truncates and appends `-` plus 8 hex characters. The reporter's own observed name (`...-ApiGatewayAccoun-19184149`) is the truncated form — so a user following the documented format for a long stack name would construct a name that does not exist and conclude the import was impossible.

## Review

Four axes (spec / code / test / security) in parallel, then three delta rounds. No blocker at any point.

The security lens was dispatched **additively** — no path here is on the security-surface list, so the tier did not require it. It was worth running: the finding that a cdkd-derived name is not proof of ownership came from it, and it traced the new log line to every reader (terminal only; it reaches neither the events store nor the thrown error, and its masking is byte-identical to the go-to-k/cdkd#2038-masked sibling beside it).

**Each delta round found a false or incomplete rationale inside the previous round's fix**, three for three:

| round | what it found |
|---|---|
| 1 | "a replacement collision never reaches this catch" — it does; the `ProvisioningError` check is what refuses it. Three of four reviewers found this independently. |
| 2 | "`displaySafe`+SUPPRESS makes printing safe" — the prose still printed the raw value in both branches. The suppression's real job is deciding whether the command names the RIGHT RESOURCE. |
| 3 | that comparison covered the physical id and the stack name, but the command carries a **third** value — the logical id — admitted by the same guard for the same reason. |

That is the shape worth recording: the text written while agreeing with a reviewer is the text least likely to have been checked against the code. Every claim that survived is now either probe-verified or stated as *not* a fence.

### Nits shipped as-is, with the reason

- **The `getProvider` catch is untested.** It cannot be driven: `provisionResource` already calls `getProvider` with that same type before the create, so a throwing lookup fails the deploy earlier. A test was written and dropped rather than shipped, and the catch says so at the call site instead of implying a checked precondition.
- **The fixture verifies only the `--resource` half of the emitted command**, supplying the stack name itself. The unit suite pins the full string including the stack name; duplicating that in a real-AWS run buys nothing.
- **The quoting test's negative assertion cannot fail independently of its positive one**, because `=` is outside `shellQuote`'s safe set and so the argument is quoted unconditionally. That is a property of `shellQuote`, not of this code; a test constructed to separate them would be asserting the quoting rule twice.

## Verification

- unit: 972 files / 21256 tests green
- `tests/integration/retain-orphan-redeploy` PASS 72s — 1 deleted, 0 errors, 0 orphans; re-run after every round that touched `deploy-engine.ts`, since `integ-destroy` digests the branch delta
- `lambda` PASS 81s for `integ-broad`, **re-run for this lane** rather than inherited: that marker is sentinel-bound and reads fresh for code its run never saw
- post-run orphan scan clean (state, IAM roles, Lambda functions)
- every branch the diff adds was probed; two probes were VOID rather than non-discriminating (one targeted the wrong file, one had a broken regex) and were re-run correctly

Closes #2902


go-to-k:
Follow-up: a claim in the message this PR shipped is wrong, and none of the four review axes or four delta rounds caught it. The maintainer did, by asking whether CloudFormation really behaves the same.

The emitted diagnosis said `... drops it from state (CloudFormation does the same), and the next deploy then asks AWS for the name it still holds.` The parenthetical is true of the clause it is attached to and false of the one right after it — and in a message announcing a stuck deploy it reads as "cdk deploy would stick too". For a resource the template does not name, CloudFormation generates a fresh random-suffixed name and the redeploy SUCCEEDS; the retained resource is orphaned but blocks nothing. Both resources in the originating report are that shape, so the sentence was wrong for precisely the case it prints on.

(cdkd sticks in BOTH columns because its derivation is deterministic — `resource-name.ts:238-258`, sha256, no random component. CloudFormation sticks only for an explicitly-named resource, which is why the corrected text says "for an unnamed resource" rather than claiming CFn never collides.)

`docs/troubleshooting.md` had this right — it said "The difference is that cdkds generated names carry no random suffix". The emitted string is what drifted, and nothing fenced it, so the same reword could have happened again. The follow-up adds the fence with the wording, probe-verified red against the old text.

Worth noting for the record because of what it says about this PRs review: the four rounds each found a false RATIONALE — a comment explaining the code. This one is a false claim in the PRODUCT text, about an external systems behaviour, which no amount of probing the code would surface. Different class, and the reviewers were not asked to check it.

go-to-k:
Correction PR: go-to-k/cdkd#2921.

PR #2928, opened 2026-09-10
test(local): pin the websocket drain's completion arm, and separate the two positions of its leaked-count message

### LINKED ISSUE #2904: test(local): the websocket drain's completion arm is never executed, and the leaked-count assertion cannot separate its two positions

nix-tkobayashi:
## Summary

Closes the two coverage gaps go-to-k/cdkd#2904 filed against the bounded-drain case in `tests/unit/local/websocket-server.test.ts`, with the one fixture both of them needed: a `$disconnect` that is **released on a signal** rather than after a sleep. Tests only — no `src/**` change; `src/local/websocket-server.ts`'s drain is correct today and is untouched.

**The signal is `Promise.allSettled`.** It is called exactly once in the production file, inside the drain and one line AFTER `drainStartCount` is snapshotted (the `Promise.all(closes)` above it is a different method), so a spy on it says *the drain has started and the count it will report is already taken*. That is an ORDERING fact, not a timing one: the release can never land before the snapshot. A sleep tuned between the socket-close loop and the ceiling could not promise that — and the issue is explicit that tuning one there is how go-to-k/cdkd#2795's flake class comes back in a new place. The pre-existing case rejects a `setTimeout` spy for this job because the socket-close loop arms a 5s timer of its own; `allSettled` has no such twin.

**Gap 2 — the COMPLETION arm.** One connection, held pending until the drain is running, then released; `close()` must return through completion. The discriminator is deliberately **not elapsed time**: the ceiling is armed several microtasks before any clock the case can read, so a stall spends part of it uncounted and the mutant slips under a `< 5000` bound. That was the first draft, and the review measured it — an await-the-ceiling-only path reported `501ms` after a 4.5s pause. It is now a count of the 5s `setTimeout` callbacks that actually FIRE, read at the moment `close()` returns: zero on the completion path, non-zero on the mutant, which cannot return until the ceiling callback has run. ("No warn" alone cannot kill that mutant — the warn also needs handlers still in flight, and this one finished.)

**Gap 1 — the two positions of the message.** `1/1` from one connection passes under a swap and under a hardcoded literal. Two connections are now pending at the snapshot, one released after it and one never, so the warn reads `1/2`. The count's INPUT is asserted separately, off the array the spy recorded, so `1/2` cannot be reached by a one-connection race that registered twice.

Both cases **race** their signal against `close()` returning rather than awaiting it bare: if the drain is skipped through the `size > 0` guard the signal never settles, and a bare await is killed at the case timeout naming nothing. The pre-existing `1/1` paragraph is corrected while being superseded — it said the second connection must settle BEFORE the snapshot, which produces `1/1` again.

## Test plan

Every bullet of the issue's verification plan, measured against the final tree (baseline: 20 cases in the file, full suite 974 files / 21417 passed / 1 skipped, rc=0):

| mutation | measured |
| --- | --- |
| drop `drainComplete.then(...)` from the race | the COMPLETION case reds **by name** — `close() returned before the drain ceiling fired: expected 2 to be +0` — not a runner deadline |
| swap `${inFlightDisconnects.size}` and `${drainStartCount}` | the `1/2` case reds; the same mutant against `origin/main`'s test file is **GREEN** (18 passed), which is the measurement the issue asks for |
| hardcode `1/1` in the message | the `1/2` case reds |
| `if (inFlightDisconnects.size > 0)` -> `if (false)` | all three drain cases red **in milliseconds, by name** — the vacuity guard the plan's fifth bullet asks for |
| remove the ceiling from the race | the pre-existing case and the `1/2` case red by name (the latter at its own reporting bound, below the case timeout) |
| move `SHUTDOWN_DRAIN_MS` to 8s | only the pre-existing case reds — the new coverage adds rather than trades |

Production files were restored byte-exact after every probe (`git diff` clean).

`vp check --fix` + `vp run check` + `typecheck:test` + `build` + `gen:all-matrices` + `audit:coverage:check` all rc=0. No integ: the diff is outside every integ gate's scope and changes no shipped behaviour, so no changelog entry either.

Codex: 3 delta rounds and 2 maintainer-checklist proxy rounds. The delta review caught the elapsed bound; the proxy caught two comments describing mechanisms that were not in the code.

**Review round 1** (commit `2f2c8703`) — four items, all taken.

The blocker was a justification standing on the wrong guarantee: the header explained the `allSettled` signal by "called exactly once in `websocket-server.ts`", which is the ORDERING argument (the call sits one line after the snapshot) but is not what makes the signal trustworthy, since the spy patches the GLOBAL. `drain.calls() === 1` is what closes that — a foreign call makes it 2 — and the comment and the assertion message now say so. Both spy installs also moved to immediately before `attached.close()` and inside the `try`, which narrows the window to that one call and removes the path where a throw between two global patches leaves one in place for the rest of the file (`vite.config.ts` sets neither `restoreMocks` nor `clearMocks`). The never-released gate is released in `finally` so its handler stops pending for the rest of the run, and the reporting timer is cleared when `close()` wins.

Coverage is unchanged: every mutant above was re-measured after the round and reds exactly the same case or cases.

Closes #2904

### DIFF
--- a/tests/unit/local/websocket-server.test.ts
+++ b/tests/unit/local/websocket-server.test.ts
@@ -808,6 +808,11 @@ describe('graceful shutdown bounded drain (M3)', () => {
   // awaits this handler.
   const HUNG_DISCONNECT_MS = 15_000;
   const CASE_TIMEOUT_MS = 25_000;
+  // How long the `1/2` case waits for `close()` before reporting that the
+  // drain never gave up. Above the 5s ceiling by the same margin
+  // `HUNG_DISCONNECT_MS` carries, and below `CASE_TIMEOUT_MS` so the
+  // assertion reports rather than the runner.
+  const CEILING_REPORT_MS = 15_000;
 
   beforeEach(() => {
     rieModule.__resetQueue();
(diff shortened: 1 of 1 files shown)


go-to-k:
Reviewed with two parallel read-only reviewers (code + test adequacy) plus my own pass over the production drain. **The substance holds up.** Both claimed gaps are genuinely closed, and I re-ran the mutants rather than taking the PR body's word for them:

| mutant | measured here |
| --- | --- |
| drop `drainComplete.then(...)` from the race (`src/local/websocket-server.ts:537`) | new completion case reds **by name** — `close() returned before the drain ceiling fired: expected 2 to be +0` |
| swap `${inFlightDisconnects.size}` / `${drainStartCount}` (`:545`) | **exactly one** failure, the new `1/2` case |

File runs 20 passed / 12.26s; the completion case is 2ms, the `1/2` case 5003ms (the real ceiling, unavoidable). Every factual claim in the new comments checks out against source: `Promise.allSettled` occurs once (`src/local/websocket-server.ts:535`), one line after the snapshot at `:534`, `Promise.all(closes)` at `:529` is a different method, the socket-close timers at `:523-526` are never cleared (zero `clearTimeout` in the file), and the "measured: 2" figure reproduces under the first mutant. No vacuous assertion found, and no `finally` masks a failure.

Four items below. **All four are requested for this PR** — they are one round, all inside the block this PR already edits. None of them says the coverage is wrong.

### m1 — `tests/unit/local/websocket-server.test.ts:981`: the comment's justification is not the guarantee that is actually load-bearing

> It is called exactly once in `websocket-server.ts`, inside the drain and one line AFTER `drainStartCount` is snapshotted

Both halves are true, and the ordering half is exactly right. But `watchDrainStart` patches the **global** `Promise.allSettled` (`:1023`), not the binding that file imports — so "once in `websocket-server.ts`" does not by itself make `started` trustworthy. Any other caller in the worker that reaches `allSettled` inside the spy window resolves `started` with a foreign array first.

What actually closes that is `expect(drain.calls()).toBe(1)` (`:1132`, `:1240`): a foreign call makes it 2 and reds. That assertion is currently commented as the drain-skip guard only, and it is carrying a second job silently.

The residual after that guard is narrow — a false GREEN needs *exactly one* foreign call *and* a drain that then skips — but the reason the comment gives is not the reason the case is safe, and this repo has been bitten repeatedly by prose that describes a mechanism it is not actually standing on. Please either narrow the spy window to the `attached.close()` call, or restate `:981` to name `drain.calls() === 1` as the guarantee and keep the file-scoped count as the ordering argument it really is.

### m2 — `:1108-1109`: the `watchCeilingFire` window is wider than the question it asks

The spy is installed before `openWebSocket` / `waitFor` / `ws.close()`, so anything arming a 5s timer in that window is counted too. The `ms === 5_000` discriminator cannot separate the ceiling (`src/local/websocket-server.ts:538-539`) from the per-socket timers (`:523-526`) — which the helper's JSDoc says, and the false-GREEN direction is genuinely impossible (on the mutant path `close()` cannot return until the ceiling callback has run). So this is a false-RED residual only. Still, moving both spy installs to immediately before `attached.close()` closes the window and makes m1's wording problem go away at the same time.

### m3 — `:1108-1109`: restore is not exception-safe across the two helper calls

`watchDrainStart()` runs outside the `try`. If `watchCeilingFire()` threw, the global `Promise.allSettled` patch would survive for the rest of the file — `vite.config.ts` sets neither `restoreMocks` nor `clearMocks`, so nothing else would put it back. Nothing between those two lines can realistically throw, so this is defence-in-depth; an `afterEach(() => vi.restoreAllMocks())` on the describe block, or moving both calls inside the `try`, settles it.

### m4 — `:1224` / `:1235`: case 2's two danglers

`gates[1]` is never released, so that `$disconnect` pends forever and is retained by `inFlightDisconnects` (`src/local/websocket-server.ts:177`) and by `dispatchChainsByConnection`, whose `.finally` cleanup at `:164-168` therefore never runs. No hang and no unhandled rejection — production attaches a `.catch` at `:388-392` — but `gates[1]!.release()` as the first line of the `finally` drains both. Separately, the `setTimeout(… 'hung', CEILING_REPORT_MS)` at `:1235` is not cleared when `closing` wins; hoisting the handle and clearing it in `finally` costs one line.

---

Please push m1-m4 and I will re-review. m1 is the one I would not merge without: the other three are cheap and in the same block, which is why they are in this round rather than a follow-up.

nix-tkobayashi:
All four taken, pushed as `2f2c8703`.

**m1 is right, and the distinction is the one worth having.** The file-scoped count is the ORDERING argument — `allSettled` sits one line after the snapshot, so the release cannot land early — and it is not what makes `started` trustworthy, because the spy patches the global. I had written one as if it were the other. The comment now separates them and names `drain.calls() === 1` as the guarantee, and the assertion says both jobs in its own message rather than carrying the second one silently:

> That file-scoped count is NOT what makes the signal trustworthy … the spy replaces the GLOBAL `Promise.allSettled`, so any other caller reaching it inside the spy window would resolve `started` first, with a foreign array. Two things close that, and both are deliberate — the spy is installed immediately before `attached.close()` and removed in `finally`, so the window is that one call; and `drain.calls()` is asserted to be exactly 1, which a foreign call turns into 2.

**m2 and m3 together** — both installs moved to immediately before `attached.close()` and inside the `try`. That is the same edit twice: it narrows the window to the one call, and it removes the path where a throw between two bare global patches leaves one in place for the rest of the file. I checked the premise rather than taking it: `vite.config.ts` sets neither `restoreMocks` nor `clearMocks`, so nothing else would have put it back. Restores are optional-chained now. Nothing between the install and `close()` can arm the signal — the installs are the last two statements before it, in the same tick as `ws.close()`, which the pre-existing case's comment explains is load-bearing for a different reason.

**m4** — `gates[1]!.release()` is the first line of the `finally`, after every assertion has run, so nothing it could have affected was still being read; and the `CEILING_REPORT_MS` handle is kept and cleared when `close()` wins.

**Coverage is unchanged, and the mutants say so rather than my saying it.** Re-measured after the round, against the final tree:

| mutant | measured |
| --- | --- |
| drop `drainComplete.then(...)` | completion case alone, by name |
| swap the two interpolations | `1/2` case alone |
| hardcode `1/1` | `1/2` case alone |
| `if (inFlightDisconnects.size > 0)` -> `if (false)` | all three drain cases, in milliseconds, by name |
| `SHUTDOWN_DRAIN_MS` -> 8s | the pre-existing case alone |

Baseline 20 passed in the file; full suite 974 files / 21417 passed / 1 skipped, rc=0, run root asserted. `vp check --fix` + `vp run check` + `typecheck:test` + `build` + `gen:all-matrices` + `audit:coverage:check` all rc=0. Codex on the delta and a maintainer-checklist proxy pass over it: NO ISSUES from both, which after this round I read as "nothing NEW", not as a second opinion that the round was needed — your m1 is the finding neither of them made.

`main` gained one commit (go-to-k/cdkd#2926) while this sat; it does not touch this file and GitHub still reports the PR mergeable, so no rebase.

go-to-k:
Merged as `e19751df`. Thank you — this was a good round.

I re-measured the delta rather than taking the summary, in a throwaway worktree at `2f2c8703`:

- File 20/20, `typecheck:test` rc=0 under `strict: true`, full suite 974 files / 21417 passed / 1 skipped, rc=0 — matching your figures exactly.
- **The mutant kills survive the narrowed spy windows**, which was the round's real risk. Completion arm dropped → the completion case alone, by name, and `ceiling.fired()` is still exactly **2**, the number the helper's doc comment claims. Interpolations swapped → the `1/2` case alone. `size > 0` → `if (false)` → all three drain cases in 6ms / 3ms / 2ms.
- The `drain.calls()` message rename cost nothing diagnostically: under the skip mutant the new cases fail earlier, on the race assertion whose message already says "without entering the drain" verbatim, so the renamed line is never the one that speaks.
- The `| undefined` narrowing is real rather than incidental — deleting the assignment produces `TS18048` at both read sites, so the check is live.
- Every factual claim added by the commit checks out at file:line, including the two I could most easily have been wrong about: the spy installs really are the last statements before `attached.close()` with no `await` between, and `vite.config.ts` really sets neither `restoreMocks` nor `clearMocks`.

On m4 specifically: releasing `gates[1]` does let the handler's continuation resume after `warnSpy.mockRestore()`, which I had flagged as a theoretical bleed. It cannot fire — `mockReset()` does not touch an already-entered async body, so `invokeRie` resolves normally and the production `.catch` has nothing to catch. Confirmed empirically with the stream fence off: no stray stderr after case 2.

**What I liked.** m1 was the finding your two automated passes did not make, and the fix did not just reword around it — separating the ordering argument from the trust guarantee, then narrowing the window so the guarantee is structural rather than asserted, is a strictly better answer than the one I asked for. The note that the assertion was "carrying the second job silently" is now in the assertion's own message, which is where it can be read by whoever hits it.

**One nit, deliberately not fixed:** `'exactly one allSettled — the drain\'s'` is not the form oxfmt emits (it would write `"…the drain's"`). Shipping as-is is right — `vite.config.ts`'s `sourceOnlyIgnorePatterns` puts `tests/**` outside both `vp fmt` and `vp lint`, so no gate can flag it, and there are already 134 escaped-apostrophe occurrences under `tests/`. Changing it would make this file inconsistent with its neighbours rather than with the formatter.

Separately, reviewing this surfaced a gap on my side, not yours: `/verify-pr` step 1 prescribes `vp run typecheck`, which excludes `**/*.test.ts`, so the local step that should have covered a tests-only PR does not. Filed as go-to-k/cdkd#2929; CI's `typecheck:test` was the backstop the whole time.

Issue #3000, opened 2026-09-11
`--allow-unsupported-properties` reads as the opposite of what it does, and rhymes with a sibling flag that runs the other way


go-to-k:
## What is wrong

`--allow-unsupported-properties` reads as the opposite of what it does, and it
misleads by RHYMING with a sibling flag that runs the other way.

| Flag | What it actually does |
| --- | --- |
| `--allow-unsupported-types` | ATTEMPTS the resource via Cloud Control — an escape hatch that ENABLES a route (`provider-registry.ts` rule 7) |
| `--allow-unsupported-properties` | ACCEPTS the property being dropped, keeping the resource on the SDK path — it DISABLES the Cloud Control auto-route (rule 3-5, via `findActionableSilentDrops`) |

Same `--allow-unsupported-*` shape, inverse meanings: one says "let cdkd try",
the other says "let cdkd give up". A reader who has met the first flag reads the
second as "let me use properties cdkd does not support, and cdkd will find a
way" — which is backwards. Without the flag, cdkd already finds a way: an
unwired property auto-routes the whole resource through Cloud Control, which
forwards the full property map (issue
go-to-k/cdkd#614). The flag is what turns that
OFF, in exchange for staying on the SDK fast path and out of the sticky
`provisionedBy: "cc-api"` route.

**Measured, not supposed.** Two independent readers got the direction backwards
in one session: the maintainer, and an agent that had just written the name into
44 public issue bodies while describing its behaviour — the bodies claimed the
properties were "dropped silently" and that this flag was what reached Cloud
Control. Both claims were wrong and both were corrected in
go-to-k/cdkd#2998's pull request.

**The help text is not the problem.** It already says the right thing —
"Comma-separated `<ResourceType>:<PropertyName>` tokens to **accept as silently
dropped**" — and the word `accept` carries the meaning. What misleads is the
flag NAME, which is what a reader sees first and what they remember.

## What to do

Rename to something whose verb matches the semantic and whose shape does not
rhyme with the type-level flag. `--accept-dropped-properties` is the suggestion:

- `allow` -> `accept`: the user is not granting a permission, they are accepting
  an outcome.
- `unsupported` -> `dropped`: names what happens to the value rather than a
  property of cdkd.
- it no longer pattern-matches `--allow-unsupported-types`, which is the half
  that actually causes the confusion.

This is user-facing CLI surface, so it needs the deprecation shape rather than a
bare rename: add the new spelling, keep the old one working as a hidden alias
[...]

Issue #2750, opened 2026-09-07
fix(state): a property dropped under --allow-unsupported-properties is still RECORDED, so the later Cloud Control auto-route diffs it as unchanged and never sends it


go-to-k:
## What happens

`--allow-unsupported-properties <Type>:<Prop>` keeps a resource on the SDK
provider and accepts that the property is not written to AWS. But the state
record it writes carries the **template** bag, including that property. The
record then claims a value AWS does not have.

On a later deploy without the flag, the resource is correctly auto-routed to
Cloud Control — `cdkd diff` even announces `[via CC API: <Prop>]` — but
`CloudControlProvider.update` builds a JSON Patch from `previousProperties`
(the state record) to the desired properties
(`src/provisioning/cloud-control-provider.ts:586-625`). The property is
identical on both sides, so the patch omits it, Cloud Control sends nothing for
it, and **the property never reaches AWS**. There is no error and no warning;
the deploy reports success.

The auto-route exists to close the silent-drop class (go-to-k/cdkd#614). This
reopens it through the state record.

## Measured, live

`tests/integration/sdk-to-cc-autoroute/` (added by go-to-k/cdkd#2749) on
`AWS::CloudWatch::Alarm` / `EvaluationWindow`, us-east-1, same binary, same
template, same flag-less deploy — only the history differs:

| Sequence | `EvaluationWindow` on AWS after the flag-less deploy |
| --- | --- |
| base (no property) → add property, no flag | **present** — auto-route applied it in place |
| base → add property WITH `--allow-unsupported-properties` → same property, no flag | **absent** |

The state record after the opt-out deploy, dumped from
`s3://.../state.json` mid-run:

```
AlarmName,ComparisonOperator,EvaluationPeriods,EvaluationWindow,MetricName,Namespace,Period,Statistic,Threshold
```

`EvaluationWindow` is recorded. `DescribeAlarms` at the same moment does not
report it.

## Why this matters more than it looks

The properties that land in `silentDrop` are not cosmetic. The
`--allow-unsupported-properties` documentation itself lists the class it warns
about — `KmsKeyArn`, `MonitoringRoleArn`, `MasterUserSecret`, IAM policy
attachments, resource-policy fields, encryption settings, TLS configuration —
and says a silent drop there "is a real-world incident". A user who opts into
the drop once, then removes the flag expecting the auto-route to close it, gets
a deploy that reports success and changes nothing.

It is also silent in both directions: `cdkd drift` compares against the same
[...]

Issue #2718, opened 2026-09-06
No mechanism watches CFn schema fixture freshness, so a property AWS adds after the snapshot is silently dropped instead of auto-routing to Cloud Control


go-to-k:
## Lane and session plan (authoritative)

This block is duplicated verbatim on [issue 2718](https://github.com/go-to-k/cdkd/issues/2718), [issue 2719](https://github.com/go-to-k/cdkd/issues/2719) and [issue 2720](https://github.com/go-to-k/cdkd/issues/2720), so picking up any ONE of them shows the whole shape without reading the other two. **It supersedes anything about ordering or session grouping in the comment threads below** — those were written incrementally, and one of them was corrected.

**A session ends at MERGED, not at "PR opened".** A session that leaves its own PR open is NOT CLOSEABLE, and the `/work-issues` contract carries each issue through merge, pull, rebuild of the linked binary, and worktree cleanup.

### Session A — issue 2718, alone

Decide the direction with Fable, record it on the issue, implement, `/check` (plus `/check-docs` if docs are touched), `/verify-pr`, PR, CI green, reviewers at the tier the size heuristic selects, merge, clean up. No real-AWS gate is in scope for this lane, so there is no long integ wait.

Optional and cheap: take issue 2719's DECISION step here too — Fable settling its four sub-questions and recording them on that issue, with NO implementation. It lets this lane's PR body state correctly that the `cc-api` flip is currently one-way.

### Session B — issue 2720 then issue 2719, in that order, in ONE session

Two sequential PRs:

1. **[issue 2720](https://github.com/go-to-k/cdkd/issues/2720)** (`chore(hooks)`): confirm the blast radius by reading the routing call sites, then move every copy of the scope list together. Touches `CLAUDE.md` and `.claude/skills/**`, so it needs BOTH the `check` and `docs` markers. Merge it FIRST.
2. **[issue 2719](https://github.com/go-to-k/cdkd/issues/2719)**: the routing change, landing UNDER the gate that PR 1 just created — plus its own binary-swap integ arm (deploy on the CC route, swap the binary, redeploy, assert the physical id is unchanged and the state record flipped) and whatever broad-set integ the newly-applicable gate demands.

Why ONE session for these two and not two: there is no file overlap between them, but there IS a strict order, and an order that has to survive a session boundary is an order that gets lost. Land the routing change before its own gate exists and the gate protects nothing. PR 1 is minutes of offline work, so it does not dilute the lane.

[...]

PR #3015, opened 2026-09-11
fix(provisioning): thread --pin-cc-api into the silent-drop reporting path

### LINKED ISSUE #3009: `--pin-cc-api` reproduces both silent-drop reporting defects through the pin: a false "will be silently dropped", and no inert-preference warning
## What is wrong

`--pin-cc-api` reproduces, through the PIN, both defects go-to-k/cdkd#3000 fixed
for the sticky STATE RECORD. `reportSilentDropDecisions` never sees the pin, so
it reasons about a route that is not the one the deploy will take.

A resource pinned for this deploy routes its update through Cloud Control
regardless of what its state record says (`deploy-engine.ts` passes the pin into
the routing decision). With `provisionedBy: 'sdk'` and every silent drop covered
by `--prefer-sdk-route`:

- `stickyCc` is false and `autoRouted` is empty, so the inert-preference warning
  go-to-k/cdkd#3000 added does NOT fire — yet the preference IS inert, because
  Cloud Control forwards the full property map and writes the values anyway.
- worse, the `overridden` branch fires and says the values **"will be silently
  dropped"**. That is false for exactly the same reason, and it is the sentence
  go-to-k/cdkd#2750 retired for the other two routes.

So on a pinned resource cdkd states the opposite of what happens, and stays
silent about the thing the user would want to know. Found by review on
go-to-k/cdkd#3007 (`provider-registry.ts`, the `named` / `overridden` branches).

## What to do

Thread the pin into the reporting path the way the state record already is.
`stickyCc` is computed from `provisionedBy` plus the exemption table; the pin is
a third input to the same question — "will this resource reach Cloud Control
whatever its properties say" — and all three answers belong in one predicate
rather than two of them being consulted and one ignored.

Then the existing branches behave: the false "will be silently dropped" line
stops firing for a pinned resource, and the inert-preference warning fires with
the PIN named as the cause. The remedy differs again — for a pin the user drops
the flag, which is neither of the other two remedies — so it needs its own
sentence, and the module's standing rule applies: state the outcome, prescribe a
command only where it is unconditionally right.

## The adjacent fence worth adding

Review also noted that today's exemption table makes the guard correct **by
accident**: no `'sdk-coverage'`-exempt type currently carries `silentDrop`
entries. A future exemption on a type that does would reopen this same shape,
because `!STICKY_CC_MIGRATION_EXEMPT.has(resourceType)` does not approximate
`wouldReturnToSdkProvider`'s `forceCcApi` and absent-`previousProperties` gates.
A unit fence asserting the two predicates agree over the live table would catch
that at the moment the exemption is added rather than in a user's log.

## Classification

Dup-check: searched open and closed issues for `pin-cc-api`, `silently
dropped`, `inert` — go-to-k/cdkd#2750 fixed this class for the auto-route and
the state record, go-to-k/cdkd#3000 added the inert warning for those two, and
neither covers the pin. No existing report.

Session-fit: next (not this session) — it needs the pin threaded from
`deploy-engine.ts` into `ProviderRegistry`'s reporting path, which widens a pull
request whose subject is a flag rename, and the wrong-direction message it fixes
has been shipping since the pin existed rather than being introduced by that
change.
Severity: medium — cdkd tells a user the opposite of what happens ("will be
silently dropped" while Cloud Control writes the value), which is the class
go-to-k/cdkd#2750 exists to prevent. Condition-specific: it needs a pinned
resource whose silent drops are all covered by `--prefer-sdk-route`.
Effort: small (S) — one predicate and its call sites, plus the unit cases; no
new integration fixture, since `cc-api-fallback-transitions` already exercises
the pin.
Estimate: ~1-2 h — most of it in deciding whether the three route-forcing inputs
collapse into one predicate or stay separate with a shared helper.

go-to-k:
## Summary

Closes go-to-k/cdkd#3009.

THREE inputs force a resource onto Cloud Control regardless of its properties
-- a `provisionedBy: 'cc-api'` record, an un-allowed silent drop, and
`--pin-cc-api` -- and the reporting path saw only two. The pin BEATS the
record: a resource recorded `sdk` still routes its update through Cloud
Control while pinned, so every branch reasoned about a route the deploy was
not going to take.

Both consequences are the defects their own issues had just removed, reached
through the pin instead of the record:

- go-to-k/cdkd#2750's false "will be silently dropped" came back -- printed
  over values Cloud Control was about to write.
- go-to-k/cdkd#3000's inert-preference warning stayed SILENT in the case a
  user is most likely to report: pinned, every drop covered, values written
  anyway, cdkd says nothing at default verbosity.

The three scattered tests are now ONE predicate, `forcedCcRoute`, returning
the CAUSE rather than a boolean. That shape is the fix for the RECURRENCE, not
just this instance: a caller that has to explain itself to a user cannot get
the explanation from a `true`, which is how two readers came to be updated
and the third not, twice.

The pin's remedy is its own and differs from both others -- dropping
`--pin-cc-api <LogicalId>` is plain, complete and non-destructive, so it is
the only one of the three that names a command. The sticky case still hands
off to docs/cli-deploy-safety.md, for the reasons recorded there.

Six probes, each applied alone with the tree restored byte-exact between them.
Five red immediately: the pin arm deleted from the predicate, the pin
cause/remedy falling through to sticky, the third reader reverting to the
record-only test, and both engine-side wiring mutations. The SIXTH survived
and is the finding -- deleting the engine's `pinnedToCcApi:` line left every
registry-level case green, because they all hand the flag in directly. A
probed callee says nothing about its WIRING. Fenced separately, with the
fence's own limit stated: it is source-shape, so it catches deletion and a
detached sibling, not a wrong value -- the pre-flight call sits past a synth,
a state read and a lock, and the sibling wiring suite drives the per-resource
dispatch instead and cannot reach it.

991 files / 21,820 tests green, asserted rooted in this worktree (the session
hit the cwd race once and a verification run silently targeted main).

REAL AWS, us-east-1: `lambda` clean -- 9 resources deleted, 0 errors, 0
orphans, state 404. Required because deploy-engine.ts is in integ-destroy's
scope; it also refreshes integ-broad.

Local `vp run check` reports one no-irregular-whitespace error in
src/cli/yaml-cfn.ts, a file untouched since May and not in this diff. Not
chased: main's CI is green at its tip, the two local worktrees disagree with
each other about the count, and CI is the authority.


go-to-k:
Closing unmerged: **the premise is false, and this change would make things worse.**

`--pin-cc-api` does NOT force an `sdk`-recorded resource onto Cloud Control. `forceCcApi` is read only inside `if (provisionedBy === 'cc-api')` (`provider-registry.ts:488-494`), and the flag's own help says so outright: *"Has no effect on a resource already on the SDK path."* What it actually does is DECLINE the automatic return to the SDK provider for a resource already recorded `cc-api` — and only for an `sdk-coverage` exemption, since `cc-broken` ignores it.

So `forcedCcRoute` returning `'pin'` for an SDK-path resource is wrong in the harmful direction: it suppresses a CORRECT "will be silently dropped" warn, emits a false cause with a remedy that changes nothing, and re-opens go-to-k/cdkd#2718's silent-drop hole for unrecognized properties. Both reviewers found it independently; I then verified it against the source rather than taking either on faith.

And the residual is EMPTY. The one case where the pin changes routing is a `cc-api`-recorded resource of a `sdk-coverage`-exempt type — and both exempt types (`AWS::Scheduler::Schedule`, `AWS::SNS::Topic`) are absent from `PROPERTY_COVERAGE_BY_TYPE` entirely, so `drops` is always empty for them and every branch this PR touches is unreachable. There is nothing to fix.

The bad premise came from a review finding on go-to-k/cdkd#3007 that I encoded into code and into go-to-k/cdkd#3009's body without checking it against the source. Recorded on the issue, which I am closing as invalid.

Issue #3019, opened 2026-09-11
docs(deploy): cli-deploy-safety.md still describes the fixed record-side silent-drop defect as open


nix-tkobayashi:
`docs/cli-deploy-safety.md`, section "When to use it", still describes go-to-k/cdkd#2750 as an open defect. The paragraph at lines 288-300 (at `76168d40`) opens "**One sequence defeats this, and it is a bug, not a reason to reach for the flag.**" It says:

- an earlier `--prefer-sdk-route` deploy "still RECORDED the property in cdkd state without writing it to AWS";
- the later flag-less deploy therefore sends nothing;
- "until then" a user must remove and re-add the property across two deploys, or recreate the resource.

The section's last paragraph repeats the workaround ("the two-deploy workaround named there is also reasoned rather than measured", line 322).

That no longer describes cdkd for a removable drop. go-to-k/cdkd#2750 is closed: the SDK route leaves removable silent-drop properties out of the record (`withoutSilentDropProperties`), while a create-only one stays recorded even when unwritten (go-to-k/cdkd#2790). `tests/integration/sdk-to-cc-autoroute/verify.sh` asserts the fixed behavior on a live alarm:

- phase 4: the opt-out deploy leaves `EvaluationWindow` out of the record;
- phase 5: the flag-less re-route delivers it (the alarm reads back `Timezone == UTC`).

`docs/provisioning-layers.md` step 3 already says the auto-route delivers the property after an `--prefer-sdk-route` deploy, except for a create-only one, so the two public pages contradict each other. The paragraph asked whoever fixed the defect to update this page; the fixture's phase 5 was rewritten to assert the fix, but the page was not updated.

The fix is docs-only, in the same section:

- rewrite the lines 288-300 paragraph to describe current behavior, keeping the create-only exception as the one sequence that still does not deliver the property;
- drop the two-deploy workaround sentence from the closing paragraph;
- revise lines 317-320, which say both exception cases "fail loudly". The create-only residual in go-to-k/cdkd#2790 can report success without applying the property.

Dup-check: searched open issues (titles and bodies) for cli-deploy-safety, "One sequence defeats", "two-deploy workaround", "RECORDED the property" -- none covers this; go-to-k/cdkd#2790 is the create-only residual in the code, a different root cause from this page's wording
Session-fit: next (not this session) -- a public-docs rewrite with its own review; the lane that found it is fixing go-to-k/cdkd#2809 in a different file, and this page is not a forced parallel change of that fix
[...]

Issue #2651, opened 2026-09-05
fix(deploy): a --recreate-via-* target whose diff is NO_CHANGE is silently ignored


go-to-k:
`cdkd deploy MyStack --recreate-via-cc-api MyTable` on a template whose diff
for `MyTable` is NO_CHANGE exits 0, migrates nothing, logs nothing, and leaves
`provisionedBy` where it was. The user has passed an explicit destroy-and-
recreate flag, answered its `[y/N]` consent prompt (which prints a per-target
plan naming `MyTable`), and gets a clean run that did not do it.

Mechanism: the engine reads the target set only inside `case 'UPDATE'` in
`provisionResourceBody`, and only AFTER the post-intrinsic-resolution
`JSON.stringify` equality short-circuit (`src/deployment/deploy-engine.ts`
around the `Skipping ${logicalId}: no actual changes` debug line). A resource
whose properties did not change never reaches the recreate branch. The
pre-flight cannot see this either: it validates the id against the template and
the state record, which is a different question from whether this deploy
produces a diff for it.

Evidence that the constraint is already known but unwritten:
`tests/integration/loggroup-never-expire-guard/verify.sh` documents its phase 4
as unable to assert the recreate executed, "because the deploy engine reads
`recreateTargets` only inside its `case 'UPDATE'`"; and the fixture added in
go-to-k/cdkd#2649 had to give BOTH colliding resources a real property change
to avoid measuring nothing. `docs/cli-deploy-safety.md` says none of it, and
"the template did not change" is exactly the state a user is in when they
reach for this flag — the whole point of `--recreate-via-cc-api` is to move a
resource between routing layers, which is not a template edit.

Suggested fix: after the DAG completes, reconcile the validated target set
against the ids an UPDATE arm actually consumed, and warn per unconsumed id
(naming the reason: no change in this deploy). A refusal would be wrong — a
multi-target run where one target legitimately has no diff should still deploy
— so this is a warn plus, arguably, a non-zero-summary line rather than an
error. Document the constraint in `docs/cli-deploy-safety.md` in the same
change.

Found by the security-review round on PR go-to-k/cdkd#2649.

Dup-check: searched open issues for "recreate target no change silently
ignored" and "recreate-via NO_CHANGE" — no hits; go-to-k/cdkd#2567 (the
nested-stack scope of the same flags, closed by the PR above) is adjacent but
is about WHICH STACK a target matches in, not about a target that matches
nothing.
[...]

Issue #2668, opened 2026-09-05
fix(deployment): a resource Type change routes the OLD resource's delete at the NEW type's provider, so changing a Type to AWS::CloudFormation::Stack destroys a child stack


go-to-k:
Changing a resource's `Type` to `AWS::CloudFormation::Stack` in a template
routes the OLD resource's delete through `NestedStackProvider`, which ignores
the physical id it is handed and destroys the child stack cdkd derives from the
logical id instead. No flag is needed — a plain `cdkd deploy` reaches it.

The chain, each link read rather than inferred:

- `src/analyzer/diff-calculator.ts:185-204` — a Type change emits
  `changeType: 'UPDATE'` with `resourceType: desiredResource.Type` (the NEW
  type) and `propertyChanges: [{ path: 'Type', requiresReplacement: true }]`.
- `src/deployment/deploy-engine.ts:4054` — `const resourceType =
  change.resourceType`, so the template's type is what the UPDATE branch uses.
- `src/deployment/deploy-engine.ts:4335` / `:4367` — `requiresReplacement` sets
  `propertyDrivenReplacement`, so the replacement path runs with no
  `--recreate-via-*` involved.
- `src/deployment/deploy-engine.ts:4493` — `oldDeleteProvider =
  getProviderFor({ resourceType, provisionedBy: currentResource.provisionedBy })`.
  `src/provisioning/provider-registry.ts:205-244` selects on that `resourceType`
  unless `provisionedBy` is `'cc-api'`, and
  `src/provisioning/register-providers.ts:357` maps
  `AWS::CloudFormation::Stack` to `NestedStackProvider`.
- `src/provisioning/providers/nested-stack-provider.ts:296-323` — `delete`
  IGNORES `_physicalId` and `_resourceType` and derives
  `<parent>~<logicalId>`, then destroys that stack.

So for a row the state records as (say) `AWS::SNS::Topic` whose template now
declares it an `AWS::CloudFormation::Stack`, the delete does not delete the
topic. If cdkd holds state for a child at `<parent>~<logicalId>` — which it
does whenever a nested stack of that logical id was ever deployed there — that
CHILD STACK and every resource it owns is destroyed. Where no such state
exists, the delete is an idempotent no-op and the old topic is silently leaked
instead.

Neither outcome is what the template asked for, and the property-driven
stateful guard does not stop either: `AWS::CloudFormation::Stack` is not in
`STATEFUL_TYPES`, so a stateful old resource is not protected by it here.

[...]

Issue #1943, opened 2026-08-18
fix(state): redactOutputs asserts same-generation at the persist choke point, and a __proto__ state key is silently dropped


go-to-k:
## Context

Bundled residuals from the GHSA-p5qg-v9gv-hc7w follow-up run (#1914 / #1915 /
#1917 / #1919). Both were raised by security reviewers on that run, neither is a
regression of it, and neither belonged in a lane already past its final review
gate. Filed together because both are small and both live in the redaction path.

## 1. `redactOutputs` keeps `descendArrays: true` at the persist choke point

`src/deployment/deploy-engine.ts:806` calls `redactSecretsForState` for
`state.outputs` with the default `TEMPLATE_DERIVED_RULES`, whose
`descendArrays` is `true` — i.e. it asserts the bag and the source are the same
generation and that positional descent into a list is sound.

At that choke point `state.outputs` **can be a previous generation**: the bag is
whatever the last successful resolve produced, while the source is today's
template. That is precisely the reasoning the #1917 lane used to move
`src/cli/commands/scrub.ts`'s outputs call off the same default onto a
`TEMPLATE_SOURCED` constant — the deploy-side twin was left alone because that
file was owned by a parallel lane at the time.

Consequence is narrow: outputs are rarely list-valued (though `StackState.outputs`
is explicitly NOT string-coerced, so a list-valued `Fn::GetAtt` does persist a
JSON array), and it needs a colliding pair to matter. But the two calls should
agree, and today they disagree for a reason that no longer applies.

Decide together with the sibling below whether the right fix is a constant swap
or a generation check.

## 2. A state key literally named `__proto__` is silently dropped

`redactByPath` / `walk` in `src/deployment/secret-redaction.ts` build their
result with `out[k] = ...`. For `k === '__proto__'` that assignment writes the
object's prototype instead of an own property, so the key **vanishes from the
persisted state**.

Measured during the #1917 review: it fails SAFE — no plaintext leaks, the key is
simply lost. But losing a key from state silently is its own defect: the next
deploy sees the property as absent and may re-create or reset it.

`Object.hasOwn` is already used as the READ-side guard a few lines away
(`secret-redaction.ts:905`), so the module already knows about this class; only
the write side is unguarded. Fix with `Object.defineProperty` or a `null`-
prototype accumulator, and pin it with a test — a key named `__proto__` arrives
readily from a `JSON.parse`d state document.

## Verification

[...]

PR #3020, opened 2026-09-11
fix(deploy): narrow the no-change re-check's desired side by the --prefer-sdk-route allow set

### LINKED ISSUE #2809: fix(deploy-engine): narrow the no-change re-check's DESIRED side like its stored side, so a silent-drop key stops forcing an update
## What

`DeployEngine`'s post-resolution no-change re-check compares an UNNARROWED desired bag against a NARROWED record, so it cannot skip a provider call for an SDK-routed resource whose bag carries a removable silent-drop key.

Since issue #2750 / PR #2788 the stored side of that comparison is `currentPropsAsWritten` — the record minus the keys the SDK route cannot write:

```ts
const currentPropsAsWritten =
  currentResource.provisionedBy === 'cc-api'
    ? currentProps
    : withoutSilentDropProperties(resourceType, currentProps);
```

The desired side is the full `resolvedProps`, which still carries the key. When both are otherwise equal the two strings differ, the re-check does not fire, and the provider is called.

## Why it is not already visible

`DiffCalculator` narrows BOTH sides (`withoutAcceptedSilentDropProperties` for the desired side, `withoutSilentDropProperties` for the current one), so a difference CONFINED to the dropped key is never reported as a change and the UPDATE arm is not entered at all. The residual needs the diff to report a change for one of the reasons the re-check exists to absorb — an intrinsic that resolves to the value already stored — while the same bag also carries a removable silent-drop key.

## Impact

RE-SCOPED after review of go-to-k/cdkd#2753: this is worse than the wasted call first described here, and for one resource type it is a REPLACEMENT.

The attribute-only branch sits INSIDE the re-check's true arm, so for an SDK-routed resource whose bag carries an allow-listed removable drop the re-check can never fire, and a `DeletionPolicy` / `UpdateReplacePolicy`-only flip — which touches no AWS-side property at all — falls through to a real `provider.update()` with identical property bags on both sides.

`AWS::SNS::Subscription` is the live case: `Region` is a non-create-only silent drop, and `SNSSubscriptionProvider.update()` is implemented as unsubscribe-then-resubscribe. So an attribute-only flip on such a subscription destroys and recreates it, which changes the subscription ARN and drops whatever the old one had in flight.

For every other type the earlier reading still holds — a redundant `update()` and a misleading `↻` line, with the record itself correct, since `propertiesToRecord` narrows what it writes.

The trigger is unchanged and still narrow: the resource must be SDK-routed, its bag must carry a removable silent-drop key, and the deploy must reach the UPDATE arm.

## Suggested fix

Narrow the desired side of the re-check the same way the record side is narrowed, so the comparison is like-for-like — the diff already does exactly this on its own two sides.

## Verification

A unit case in the shape of `tests/unit/deployment/deploy-engine-silent-drop-record.test.ts`: an SDK-routed resource whose type has a removable silent drop present in the template bag, whose other properties resolve to exactly what the record holds, asserting `provider.update` is NOT called. It fails today. A second case should drive the attribute-only branch on a type whose `update()` recreates, so the replacement half is pinned rather than argued.

Dup-check: searched open issues for `silent drop`, `withoutSilentDropProperties` and `no-change re-check`; #2750 is the parent that introduced the record-side narrowing and is closed, and nothing open covers the re-check's own operand.
Session-fit: next (not this session) — found while resolving PR #2753's conflict with #2788; it is #2788's own interaction, in a function that PR does not otherwise touch, and folding it in would widen a diff already through three review rounds.
Severity: medium — for `AWS::SNS::Subscription` an attribute-only flip becomes a destroy-and-recreate of the subscription; for other types it stays a redundant call with a correct record.
Effort: small (S) — one operand plus the unit case above; it rides the existing suite.
Estimate: ~30-60 min — most of it writing the case that fails first.

nix-tkobayashi:
## Summary

Closes #2809.

Since go-to-k/cdkd#2750, the UPDATE arm's no-change re-check in `provisionResourceBody` compared two bags that could not match. The record side is narrowed to what the SDK route can write. The desired side still carried an allow-listed removable silent drop (a create-only drop stays on both sides). For an SDK-routed resource carrying one, the skip was unreachable, and so was the attribute-only branch nested inside it. A `DeletionPolicy` / `UpdateReplacePolicy`-only flip therefore took a full provider `update()` instead of the state-only refresh. For `AWS::SNS::Subscription`, where `Region` is such a drop, that `update()` is the provider's delete-then-create.

The desired operand is now narrowed by the **diff's own desired-side rule**, not the record side's:

```ts
const desiredForSkipCheckAsWritten =
  currentResource.provisionedBy !== 'cc-api' && allowedSilentDrops
    ? withoutAcceptedSilentDropProperties(resourceType, desiredForSkipCheck, allowedSilentDrops)
    : desiredForSkipCheck;
```

**This deliberately differs from the issue's suggested fix** (narrow "the same way the record side is narrowed", i.e. every removable drop). An un-allowed drop auto-routes the resource to Cloud Control, which does write the key. Removing that key from the desired side would hide a real difference and skip an update that must be sent. The existing case "does not skip the resource when the ONLY difference is the never-written key" fences this. The claim comment on the issue records the correction before any edit.

One consequence, deliberate: a record a pre-go-to-k/cdkd#2750 binary wrote still holds the allow-listed key, and a flag-ful policy-only deploy now takes the attribute-only branch, which keeps the recorded properties, so that leftover key stays. The redundant `update()` used to rewrite it away. The diff's stored side and `currentPropsAsWritten` narrow the record before comparing, so neither is misled by it. `cdkd drift` compares against the recorded properties only for a record with no observed baseline. A deploy without the flag is the healing path. The test file's doc comment and the changelog record this.

## Changes

- `src/deployment/deploy-engine.ts`: the re-check's desired operand, plus the `property-coverage` import. The comment block also records why the recreate flags, read below the skip (the go-to-k/cdkd#2651 class), are not bypassed here. For a resource whose type is unchanged, pre-flight refuses both combinations with `RECREATE_TARGETS_INVALID`: `ambiguousIntent` for `--recreate-via-cc-api`, and `blockedAlreadySdk` for `--recreate-via-sdk-provider`. A `Type` change does reach the skip, and the comment points at go-to-k/cdkd#3036 for that.
- `tests/unit/deployment/deploy-engine-silent-drop-record.test.ts`: a new describe block for this issue, with six cases:
  - the SDK-route skip;
  - the same skip for a pre-v7 record with no `provisionedBy`, which counts as SDK exactly as it does for `currentPropsAsWritten`;
  - a Cloud Control control, where the record does hold the key;
  - a registry double without `getAllowedUnsupportedProperties` (the `?.()` fallback the diff call shares) compares the full bag and calls the provider, rather than throwing;
  - an SNS premise guard: `Region` is a silentDrop, the other keys are handled, and `withoutSilentDropProperties` of the desired bag equals the written bag;
  - an SNS `DeletionPolicy` flip that must reach none of `update` / `delete` / `create`, refresh `deletionPolicy`, keep `physicalId`, and leave the narrowed `properties` exactly as written, own keys included (a probe recording the unnarrowed desired bag fails it, and so does one storing `Region: undefined`).
- `tests/unit/provisioning/sns-subscription-provider.test.ts`: `update() unsubscribes the old subscription before subscribing a new one` pins the premise the SNS case rests on (Unsubscribe first, then Subscribe, a different ARN, `wasReplaced: true`).
- `tests/integration/sdk-to-cc-autoroute/`:
  - **Phases 4b and 4c** (new): an `UpdateReplacePolicy`-only flip under `--prefer-sdk-route`, then the flip back. Both assert the verb `updated (metadata)` (the bare `updated` is the sentinel for "the provider was called"), the record's policy, and an unchanged `AlarmConfigurationUpdatedTimestamp` read from AWS. Phase 4b also asserts `provisionedBy: sdk`, that the record still lacks `EvaluationWindow`, and the threshold. A synth premise guard pins that the alarm resource in `allowmeta` differs from the one in `allowdrop` only by `UpdateReplacePolicy` (the resource object is compared whole, `DeletionPolicy` and `Metadata` included; the rest of the template is not compared).
  - **`EvaluationWindow` is now read through the SDK, not `aws cloudwatch describe-alarms`** (phases 2b, 4, 5). The CLI prints only fields its bundled model knows, and aws-cli 2.34.37's CloudWatch model has no `EvaluationWindow`. On that CLI, phase 2b and phase 5 fail on a correct deploy, and phase 4's absence control passes vacuously. Measured: Cloud Control's `get-resource` showed the property set while the CLI printed `null`. Phase 2b's positive read goes through the same reader, which is what makes phase 4's `null` evidence.
  - The stack doc's phase-4 wording said `--allow-unsupported-properties`; it now says `--prefer-sdk-route`.
- `.claude/rules/layout-provisioning.md`: the `withoutAcceptedSilentDropProperties` sentence names both desired-side consumers, `DiffCalculator` and the UPDATE arm's no-op skip, under one gate: the record is not on `'cc-api'`, and an allow set exists (only a test double lacks one; the real registry always returns one, which is why the changelog does not mention it). The adjacent phrase "whose SCOPE is the decision" now reads "whose SCOPE is deliberate", because "the decision" there could be read as the routing decision.
- `changelog.d/entries/2026-09-11-2809-recheck-desired-narrow.md`.

## Verification

- **Unit A/B**: against the pre-fix `deploy-engine.ts`, exactly the two fix-target cases fail (`skips the provider when the ONLY difference is the allow-listed drop`, `refreshes DeletionPolicy in state and never reaches the provider`); the Cloud Control control and the SNS premise pass. This A/B was taken when the file held 15 cases; with the fix, the final file passes 17/17.
- **Mutation probes** (one mutation each, tree restored byte-exact between probes):
  - the issue's prescription, `withoutSilentDropProperties(resourceType, desiredForSkipCheck)` on the desired side: 1 of 15 fails, the pre-existing fence `does not skip the resource when the ONLY difference is the never-written key`;
  - the route gate removed (`allowedSilentDrops` alone as the condition): 1 of 15 fails, `still skips on the Cloud Control route, where the record DOES hold the key`;
  - the allow-set guard removed (the route test alone as the condition): 1 of 16 fails, `compares the FULL desired bag when the registry double has no allow-set method`;
  - the route test keyed on `=== 'sdk'` instead of `!== 'cc-api'`: 1 of 17 fails, `also skips for a record with NO provisionedBy marker, which counts as SDK`;
  - `SNSSubscriptionProvider.update()` skipping its delete: 1 of 14 fails, `update() unsubscribes the old subscription before subscribing a new one`.

  Each probe ran against the file as it stood when that probe's target case was added, which is why the engine counts differ (15, 16, 17).
- **Live A/B** on `sdk-to-cc-autoroute`, same tree except `src/deployment/deploy-engine.ts`:
  - pre-fix build: phases 1-4 pass, then **phase 4b FAILS**. The deploy line reads `[1/1] ✓ AutorouteAlarm (AWS::CloudWatch::Alarm) updated` (no `(metadata)`), and the fixture prints `FAIL (phase 4b): the policy-only flip reached the provider's update() ...`. The timestamp assertion comes after the verb assertion, so the run did not reach it: whether a `PutMetricAlarm` with unchanged settings moves `AlarmConfigurationUpdatedTimestamp` is unmeasured. The verb is what this A/B shows; the timestamp check is a second observable that has not been shown to discriminate.
  - fixed build: PASS, all phases including 4b/4c, destroy `1 deleted, 0 errors`, 0 object versions left under the stack prefix.
- **Broad**: `lambda` PASS, destroy `9 deleted, 0 errors`, 0 versions left.
- **Final runs, on this branch's tree after it moved onto `1e2afd18`** (go-to-k/cdkd#3008 landed mid-lane and also touches `deploy-engine.ts`, so the earlier runs no longer described the tree under review): `sdk-to-cc-autoroute` PASS in 275 s, and `lambda` PASS in 99 s. Both left 0 errors and 0 versions. These are the ledger rows.
- Full suite on the final tree: `vp test run`, 991 files, 21839 passed, 1 skipped. `vp run check`, `typecheck:test`, `build`, `gen:all-matrices` and `audit:coverage:check` are clean.
- Codex, delta review, 2 rounds:
  - round 1 found a test comment calling a mocked intrinsic UPDATE "the only way into the UPDATE arm", and "could never match" overclaiming (a create-only drop stays on both sides, so it now says "removable");
  - round 2: NO ISSUES.
- Codex, maintainer-checklist proxy, 3 rounds:
  - round 1 found the undefined allow-set arm unpinned, the SNS re-create premise asserted rather than pinned, the synth guard comparing only `Properties`, and imprecise route wording;
  - round 2 found the absent-marker arm unpinned, and the changelog missing the pre-v7 clause;
  - round 3: NO ISSUES.

  Every finding above is fixed in this PR. The mutation probes cover four new cases and one pre-existing fence. The two fix-target cases (the SDK skip and the SNS policy refresh) are covered by the unit A/B instead, and the SNS premise case is a guard on generated coverage data rather than on code.

## Related

- go-to-k/cdkd#3019, filed while working on this: `docs/cli-deploy-safety.md` described the go-to-k/cdkd#2750 record defect as still open. Closed by go-to-k/cdkd#3030.
- go-to-k/cdkd#3036, filed during review: the UPDATE arm's no-op skip swallows a resource `Type` change whose property bags compare equal. It happens on `main` too for bags neither narrowing changes; this PR widens it slightly, and it is not fixed here.


go-to-k:
Reviewed at head `7f6d9f0`. Four reviewers — spec, code, test, and an additive security pass (the hunk moves a `redactSecretsForState` result through a new narrowing helper, so a sensitive-value bag flows through the changed lines). Tier was biased up from the heuristic's `1-reviewer`: the code this PR edits shipped the defect it is fixing, two merges ago in go-to-k/cdkd#2788.

**The central decision is right.** I checked the deviation from the issue's prescribed fix independently, and it is required rather than merely preferable: with `withoutSilentDropProperties` on the desired side, an un-allowed drop on a record still marked `sdk` is stripped from *both* operands, the skip fires, and the Cloud Control auto-route that delivers the key never runs — re-opening go-to-k/cdkd#2750 through the new gate. Recording the correction on the issue before the first commit was the right call.

Two things my reviewers raised that I then **refuted by probe**, so you do not need to chase them:

- The helper choice is NOT unfenced. Swapping in `withoutSilentDropProperties` leaves all six new cases green, but reds the pre-existing `does not skip the resource when the ONLY difference is the never-written key` — exactly as your mutation-probe section claims. Reproduced: 1 of 17 fails.
- Your A/B claim is accurate. Reverting the fix on the final tree reds 3, not 2, but the third is `also skips for a record with NO provisionedBy marker`, which you added after that probe was taken — and the PR body already says the counts differ because each probe ran against the file as it stood. No change needed.

## Required before merge

**M1 — `.claude/rules/layout-provisioning.md`, the rewritten `withoutAcceptedSilentDropProperties` sentence.** Two defects in one clause, and this is an agent-instruction file, so a wrong rule here propagates to every future session.

The sentence currently reads:

> `withoutAcceptedSilentDropProperties` narrows the diff's DESIRED side and, unless the record is on `'cc-api'` or the registry exposes no allow set (a test double), the UPDATE arm's no-op skip's (#2809); both must keep an un-allowed drop because Cloud Control does write it.

- The trailing `the UPDATE arm's no-op skip's (#2809)` is an ellipsis whose head noun ("desired side") is 14 words back with the issue ref wedged in. It reads as a typo.
- More importantly, the two conditions are attached to the re-check ALONE, but `src/analyzer/diff-calculator.ts:291-299` gates its desired-side narrowing on the same two (`sdkRouted && allowedUnsupportedProperties`). As written, the next reader takes the diff's narrowing to be unconditional — the opposite of this PR's own thesis.

Please spell the second consumer out and scope the conditions to both.

**M2 — `changelog.d/entries/2026-09-11-2809-recheck-desired-narrow.md`, the headline.** It says the fix covers "a resource whose **only** difference from state is a `--prefer-sdk-route` silent drop". That case cannot reach the UPDATE arm at all: the diff narrows both of its own sides, so a difference confined to the dropped key is `NO_CHANGE`. Your own test file says so at `tests/unit/deployment/deploy-engine-silent-drop-record.test.ts:496-497` — "What it never reports is a property change confined to the dropped key". The entry's body states the real trigger correctly; the headline is the part that ships.

## Take-or-leave

**m3** — `src/deployment/deploy-engine.ts:5438`: the skip now short-circuits ahead of the `recreateViaCcApi` / `recreateViaSdkProvider` read at 5497, so a `--recreate-via-*` target of this shape can be silently ignored (the go-to-k/cdkd#2651 class, slightly widened). Reaching it needs `--prefer-sdk-route` and `--recreate-via-*` on the same resource plus a policy-only flip, which is self-contradictory input — worth one clause in the 5484 comment so the next reader does not re-derive it.

**m4** — the changelog entry could carry one clause on the healing consequence the test file's doc comment already records: a pre-go-to-k/cdkd#2750 record keeps the stale key where the redundant `update()` used to narrow it away. Harmless for the engine's own readers, but `cdkd drift` compares the unnarrowed recorded bag, so this lengthens that pre-existing window.

**m5** — the attribute-only unit case asserts `deletionPolicy` and `physicalId` but not `saved.properties`, so nothing at the unit level pins that the narrowed record survives the refresh unmodified. Integ phase 4b's `RECORDED_META` does cover it.

## Notes, no action

The `AlarmConfigurationUpdatedTimestamp` assertions in phases 4b/4c are not shown to discriminate, as your PR body says — the verb assertion is carrying those phases alone. That is fine and honestly disclosed; I mention it only so it is not counted as a second fence later.

Security pass came back clean. Specifically checked and OK: the hoist is behaviour-neutral (the old call sat in an `if` condition, which always evaluates, so redaction ran exactly once either way); the `markSameGenerationBag` mark dropped by the helper's spread is already consumed by `redactSecretsForState` before the narrowing runs, and a future wrong reorder degrades to a spurious update rather than a leak; and the skip path re-persists an already-redacted record through the same choke point.

Please push M1 and M2 and I will merge — I will re-run the review on the new head, then the broad integ, since markgate markers are per-worktree and your ledger rows are runs on your own account.

nix-tkobayashi:
Thanks for the review. Everything is in `f954c17`, on top of `7f6d9f0`. I took both required items and all three take-or-leave items.

**M1**: the rules sentence in `.claude/rules/layout-provisioning.md` now names both desired-side consumers under the one gate they share: the record is not on `'cc-api'`, and an allow set is passed. The consumers are `DiffCalculator` (`sdkRouted && allowedUnsupportedProperties`, `src/analyzer/diff-calculator.ts:291-299` as you cited) and the UPDATE arm's no-op skip. The sentence also says who passes no allow set: `cdkd diff` (the same paragraph already explains why that is correct) and, in a deploy, only a test double. The trailing elliptical clause is gone.

**M2**: the changelog headline now names the real trigger: "a change the no-change re-check should absorb -- a `DeletionPolicy` / `UpdateReplacePolicy`-only flip, or an intrinsic resolving to the stored value". A difference confined to the dropped key is `NO_CHANGE` at the diff, as you said, and the headline no longer claims it.

**m3**: added a clause to the #2809 comment block in `deploy-engine.ts`. A `--recreate-via-*` target the skip absorbs is silently ignored, because the recreate flags are read below it (the go-to-k/cdkd#2651 class). Reaching it needs `--prefer-sdk-route` and a recreate flag on the same resource, which is contradictory input, plus any change the skip absorbs. That includes an intrinsic resolving to the stored value, not only a policy-only flip.

**m4**: added one sentence to the changelog entry, scoped more narrowly than your note after checking `src/cli/commands/drift.ts`. Drift's comparison baseline is `observedProperties` when the record has one, and the recorded `properties` only when it does not (`useObserved`). Drift also reads the recorded properties for other purposes, such as secret and unknown-path collection. So the sentence says drift "compares against" the stale key only for a record with no observed baseline. It also avoids "the engine narrows it on every read": the replacement guard and the replacement `delete()` deliberately keep the unnarrowed recorded bag. It says the diff and the re-check narrow it before comparing. To stay under the 2000-character cap, the live-test sentence in the entry was shortened.

**m5**: the SNS attribute-only case now asserts `saved.properties` equals the written bag. I probed it: making the attribute-only branch record the resolved, unnarrowed desired bag fails exactly that case (1 of 17), and the tree was restored byte-exact.

The only source change in this round is the comment clause. Since the ledger runs, the fixture and the executable code are unchanged, so I did not re-run the integs on my account. Your broad run on the new head is the one that counts for the merge gate anyway.

Checks: `vp run check`, `typecheck:test`, `build` and the full suite are green (991 files, 21839 passed, 1 skipped), and the changelog and rules-payload fences pass. Codex on this round's diff took 2 rounds: round 1 found four overclaims in the new text (the three corrected above, plus the recreate clause crediting only a policy-only flip), all fixed; round 2 returned NO ISSUES.

go-to-k:
Reviewed `f954c17` (delta `7f6d9f0..f954c17`). M1 and M2 are correctly resolved, and I verified m4 and m5 rather than taking them on trust.

- **M1 is clean.** The gate is byte-identical at both consumers (`diff-calculator.ts:290-296` and the re-check), both read the same source at `deploy-engine.ts:2922`, and there are exactly two call sites so "two consumers" is exhaustive. One thing worth naming because it is easy to lose later: moving "Both must keep an un-allowed drop" from the two HELPERS to the two CONSUMERS is a correction, not just a rewrap — the old sentence was false of `withoutSilentDropProperties`, which does not keep one.
- **m4** — your narrowing is better than my note. `drift.ts:2434-2435` selects `observedProperties` as the baseline when the record has one, so "only for a record with no observed baseline" is right and my "compares the unnarrowed recorded bag" was not.
- **m5** — probed it: recording the unnarrowed desired bag in the attribute-only branch reds exactly that case, 1 of 17. Restored byte-exact.

## Required before merge

**M3 — `src/deployment/deploy-engine.ts:5425-5429`, the clause I asked you to add. It is wrong, and the error is mine: I relayed a reachability characterisation without checking it.** Two things:

- The scenario is not "silently ignored" — it is refused at pre-flight. `--recreate-via-cc-api` on a resource whose template carries an allow-listed drop pushes `ambiguousIntent` (`src/deployment/recreate-targets.ts:388-391`), and `--recreate-via-sdk-provider` pushes `blockedAlreadySdk` for exactly the records this narrowing applies to, i.e. `provisionedBy !== 'cc-api'` (`:414-419`). Both reach `throw new CdkdError(..., 'RECREATE_TARGETS_INVALID')` at `src/cli/commands/deploy.ts:861`.
- "contradictory input" is wrong for one of the two flags. `recreate-targets.ts:716-720` explicitly instructs the user to pass `--prefer-sdk-route <Type>:<Prop>` *alongside* `--recreate-via-sdk-provider`, so that pairing is the documented remedy, not a contradiction.

Please replace the clause with the refusals, e.g. "Not reachable from the CLI: `--recreate-via-cc-api` + `--prefer-sdk-route` on the same resource is `ambiguousIntent` (`RECREATE_TARGETS_INVALID`) whenever the template carries the allow-listed drop, and `--recreate-via-sdk-provider` is `blockedAlreadySdk` for exactly the records narrowed here — see `src/deployment/recreate-targets.ts`."

**M4 — `changelog.d/entries/2026-09-11-2809-recheck-desired-narrow.md`, a regression from the cap trim.** "Live-tested by `sdk-to-cc-autoroute` phases 4b/4c." now follows the `AWS::SNS::Subscription` delete-then-create sentence, so it reads as if the SNS case was live-tested. Phases 4b/4c run on `AWS::CloudWatch::Alarm` / `EvaluationWindow` (`verify.sh:379-407`); the SNS re-create is pinned by a unit test only. Restoring the type fits: "Live-tested on `AWS::CloudWatch::Alarm` by `sdk-to-cc-autoroute` 4b/4c." takes the entry to 1992/2000.

## Take-or-leave

**m6** — `tests/unit/deployment/deploy-engine-silent-drop-record.test.ts:719`: `toStrictEqual` would also catch a `Region: undefined` regression that `toEqual` lets through. Consistent with the neighbouring assertions as written, so your call.

Push M3 and M4 and I will merge. `vp run check`, `typecheck:test` and `build` are green on `f954c17` here; after your push I will re-run the full pass plus the broad integ on the new head, since the `integ-broad` marker is per-worktree and your ledger rows are runs on your own account.

nix-tkobayashi:
Thanks. M3 and M4 are in `e67d273`, on top of `f954c17`. On m6 I kept the intent but not the matcher; the reason is below.

**M3**: I replaced the clause with the refusals, close to your wording, after checking each citation:
- `--recreate-via-cc-api` with `--prefer-sdk-route` on the same resource pushes `ambiguousIntent` for every allow-listed drop the template carries (`src/deployment/recreate-targets.ts:388-391`).
- `--recreate-via-sdk-provider` pushes `blockedAlreadySdk` for every record not on `'cc-api'` (`:414-419`). That is a superset of the records this narrowing touches, since the narrowing also needs an allow set and a removable drop. So the comment says "a superset" rather than your suggested "exactly".
- Both reach `RECREATE_TARGETS_INVALID` at `src/cli/commands/deploy.ts:861`.
- You are right that `--prefer-sdk-route` alongside `--recreate-via-sdk-provider` is the documented remedy (`recreate-targets.ts:716-720`), so "contradictory input" is gone.
- "Not reachable" needed one more qualifier: the resource's TYPE must be unchanged. The `ambiguousIntent` check reads drops against the RECORDED type, while the narrowing uses the desired one. A `Type` change does reach this arm, because `diff-calculator.ts` emits it as an UPDATE carrying `Type` with `requiresReplacement`, not as a separate DELETE + CREATE. The comment now scopes the claim to an unchanged type.

That led to a separate finding, which I have **not** fixed here. This skip compares properties only, before `propertyDrivenReplacement`, so it swallows a `Type` change whose property bags compare equal.

I measured it at unit level with a mocked diff: a record on `AWS::SQS::Queue`, a template on `AWS::CloudWatch::Alarm`, identical properties. The deploy completed, made zero provider calls, and the record kept the old type. By the code path, `main` behaves the same for bags that neither narrowing changes. This PR widens the case to bags that differ only by an allow-listed removable drop of the new type.

I filed it as go-to-k/cdkd#3036. It is related to go-to-k/cdkd#2668 but has a different root cause: the skip discards the replacement intent before any routing happens. The obvious one-line fix is to not take the skip when `currentResource.resourceType !== resourceType`. But that would send the change into the replacement path, whose delete still routes on the new type (#2668), so it should be ordered against that routing decision. The comment in `deploy-engine.ts` points at the new issue.

Thanks too for the note on M1: moving "must keep an un-allowed drop" from the helpers to the consumers was the correction that mattered, since `withoutSilentDropProperties` does not keep one.

**M4**: the sentence now reads "Live-tested on `AWS::CloudWatch::Alarm` by `sdk-to-cc-autoroute` 4b/4c.", so it no longer reads as covering the SNS case. The entry is 1992 of 2000 characters.

**m6**: `toStrictEqual` fails on the unmutated code. The persisted `properties` holds exactly the three written keys, but it is a null-prototype copy: the secret-redaction walks in `src/deployment/secret-redaction.ts` build `Object.create(null)` bags (the #1943 class). `toStrictEqual` rejects it on the prototype alone. So the case keeps `toEqual` and adds an own-key assertion, `Reflect.ownKeys(saved.properties)` against the written bag's keys, which catches the `Region: undefined` regression you described. I probed it: storing `Region: undefined` in the attribute-only branch fails exactly that assertion (1 of 17). The redaction walk keeps the `undefined`-valued own key, so the regression does reach the saved record, and `toEqual` alone lets it through. The tree was restored byte-exact.

One more correction, found while re-reviewing the PR body: the test file's doc comment, from the previous round, said the leftover key is harmless "because every reader narrows the record first". That is not true of `cdkd drift`, which compares against the recorded properties when a record has no observed baseline. It now names the two deploy comparisons that do narrow (the diff's stored side and `currentPropsAsWritten`), matching the changelog, and the PR body's summary says the same.

The only source change is the comment clause, so there is again no executable change since the ledger runs. Checks: `vp run check`, `typecheck:test`, `build` and the full suite are green (991 files, 21839 passed, 1 skipped). Codex on this round's diff took five rounds:
- round 1 narrowed "exactly" to "a superset";
- round 2 flagged the missing same-type qualifier;
- round 3 corrected my first answer to round 2, since the diff emits a `Type` change as an UPDATE, not as a DELETE + CREATE;
- rounds 4 and 5 returned NO ISSUES.

The body of go-to-k/cdkd#3036 took two Codex rounds, and the second returned NO ISSUES.

go-to-k:
Merged. Thanks — this was a well-run PR, and three things in particular are worth naming.

**You corrected me twice, with evidence, and both times you were right.** On m4 you narrowed my claim after reading `drift.ts:2434-2435`: drift's baseline is `observedProperties` when the record has one, so "compares the unnarrowed recorded bag" was wrong and "only for a record with no observed baseline" is right. On M3 you replaced my suggested "exactly" with "a superset", because `blockedAlreadySdk` fires for every record not on `'cc-api'` while this narrowing also needs an allow set and a removable drop. A contributor who takes a maintainer's wording verbatim would have shipped both errors.

**The find while fixing M3 is the best part.** Chasing whether the recreate-flag scenario was reachable turned up that the skip compares properties only and swallows a `Type` change whose bags compare equal — measured, not reasoned, and filed as go-to-k/cdkd#3036 rather than fixed inline. That was the right call: the one-line guard sends the change into the replacement path, whose delete still routes on the new type, so it has to be ordered against go-to-k/cdkd#2668. Correctly identifying that a fix belongs in a different PR is harder than writing the fix.

**The m6 answer was better than my suggestion.** I proposed `toStrictEqual`; you found it fails on unmutated code because the redaction walks build `Object.create(null)` bags, and used `Reflect.ownKeys` instead. I reproduced it: storing `Region: undefined` in the attribute-only branch passes `toEqual` and fails the own-key assertion, 1 of 17.

Verification on my side, for the record:

- Re-ran your mutation probes rather than relaying them. The issue's prescribed fix reds the pre-existing `does not skip the resource when the ONLY difference is the never-written key` (1 of 17) — so the helper choice was fenced all along, contrary to one of my reviewers, and your mutation-probe section already said so.
- Your A/B claim checked out. Reverting the fix on the final tree reds 3, not 2, but the third case was added after that probe and the body already says the counts differ per probe.
- `sdk-to-cc-autoroute` PASS in 168 s on my account, phases 4b and 4c both rendering `updated (metadata)`; destroy 1 deleted, 0 errors, 0 orphans.
- `lambda` (broad set) PASS in 101 s; destroy 9 deleted, 0 errors, 0 orphans.
- Full suite 991 files / 21839 passed on the merged head.

One note so the ledger is not misread: the `integ-last-run.tsv` rows in this PR are your runs. Mine superseded them locally, and I am recording those in a separate PR rather than pushing a maintainer commit onto your branch.

PR #3031, opened 2026-09-11
ci: correct the fork-decoy reach, guard an empty head ref, cover a hostile one

go-to-k:
Follow-up to go-to-k/cdkd#3025, which merged with the same defect it was written to fix: a
comment that overstates what the code does. Two reviewers, on the two sibling
repos carrying this file, found it independently. Same fixes are in
go-to-k/cdk-local#721 and go-to-k/cdk-real-drift#1902.

## The reach was one step too long

The bullet said a lone fork decoy "would read, and could comment on and disarm,
a fork's PR". The read is right. The disarm is not reachable:

```
head_ref  = the FORK's branch name
git fetch --no-tags --quiet origin "${head_ref}"   # resolves against THIS repo
                                                   # which does not carry it
  -> exit 1
```

So the honest reach is a **second DoS** wearing a misleading cause ("may have
been deleted"). Comment/disarm needs *both* an identically named branch still
present in this repo *and* a write-access user having armed auto-merge on the
fork PR — which its author cannot do.

## The debunked story survived in the test

go-to-k/cdkd#3025 rewrote the `[0]`-picks-the-decoy model in the workflow only. The test
carried it verbatim in two places, and that is the copy a reader hits first.
Both now describe what the `count != 1` refusal actually does.

The eviction case's "same suppression as the decoy" is wrong in both files too:
the decoy is the **loud** failure (`count != 1` → exit 1) and truncation is the
silent one. It also said the decoy block was "below" when it is above.

## An empty head ref was not fail-closed

`armed` already refused an unreadable answer; `head_ref` did not. A `gh` exiting
0 with empty stdout reached `git fetch origin ""` and was reported as a branch
that "may have been deleted" — pointing at the repo when the fault was the CLI.

## The hardening had protected the stub, not the workflow

`headRefName` answered a **constant**, so no case could drive an attacker-shaped
branch name into `git fetch` — the one place a head ref reaches a shell. The
stub now reads the matched PR's ref back out of the same canned list, and a case
drives `release-please--branches--main$(touch pwned)` through it.

Measured rather than asserted, because the obvious claim is false:

| mutation | result |
| --- | --- |
| drop the quotes (`origin ${head_ref}`) | **no change** — bash does not re-scan a variable's value |
| `eval "git fetch … $head_ref"` | reds the new case, and **only** that case (17 others green) |
| neutralize the empty-`head_ref` guard | reds the new empty-ref case, and only that one |

All three probes run in all three repos.

## Smaller

- The remaining interpolations into the stub's bash source (the canned `armed`
  answer, the log path) move to a file and an env var.
- The fetch error names transient failure as a cause, since it is not retried.
- The header's case count is replaced with the enumeration — the number was
  already off by two.

No changelog entry: CI-only, no shipped-binary behavior delta.

## Round 4 — the head ref is now validated against git's own grammar

Two findings, both "right conclusion, wrong reason" — the same class as the
three rounds before.

- The hostile fixture drove `release-please--branches--main$(touch pwned)`
  through the head-ref path. Measured: `git check-ref-format --branch`
  **rejects** that name (it contains a space), so the case probed a branch name
  that cannot exist. `$(>pwned)` is ref-format **valid** and still writes the
  file under a parser — that is the payload now.
- The safety argument said dropping the quotes changes nothing "because bash
  does not re-scan a variable's VALUE". True but not sufficient: unquoted
  expansion still word-splits and globs, and the old payload's space would have
  split it. The load-bearing reason is that no real ref can carry those
  characters — `check-ref-format` rejects space, `*`, `?`, `[`, `:`, `\`, `~`,
  `^`.

So the guard became what it already claimed to be. `[ -z "${head_ref}" ]` was
described as "the same fail-closed rule as `armed`", but `armed` accepts an
enumerated set while that tested **one** bad shape — whitespace-only and
option-shaped answers went straight through. It is now
`git check-ref-format --branch`, and the fetch names the ref in full
(`refs/heads/...`).

Also: the fork-reach bullet was short one condition (the branch must be STALE
too), the metacharacter case gained a message assertion it could otherwise pass
vacuously, and every count in the test docblock — "six invocations" (ten), "the
ten" (twelve), "those eight" (fourteen) — was deleted rather than corrected.
Three wrong revisions is enough evidence that nothing re-checks a number in
prose.

### DIFF
--- a/.github/workflows/release-pr-staleness.yml
+++ b/.github/workflows/release-pr-staleness.yml
@@ -80,9 +80,17 @@ jobs:
           # the two remaining reaches are:
           #   * DoS. One fork PR wedges every run into that refusal, turning a
           #     guard that should be quiet into a permanent red anyone can set.
-          #   * Acting on a fork PR. With no real release PR open, a lone decoy
-          #     is `count == 1`, and this job holds `pull-requests: write` -- it
-          #     would read, and could comment on and disarm, a fork's PR.
+          #   * Reading a fork PR. With no real release PR open, a lone decoy
+          #     is `count == 1`, so the run picks it up -- and then hard-fails,
+          #     because `git fetch origin "${head_ref}"` below resolves the
+          #     FORK's branch name against THIS repo, which does not carry it.
+          #     So the reach is a second DoS wearing a misleading cause ("may
+          #     have been deleted"), not a write. The comment/disarm this job's
+          #     `pull-requests: write` allows needs BOTH an identically named
+          #     branch still present here AND someone with write access having
+          #     armed auto-merge on that fork PR -- a fork author cannot arm it.
+          #     Stated at that length because two earlier drafts of this bullet
+          #     overstated the reach, each time describing a chain nobody ran.
           #
           # A deleted fork leaves `head.repo` null, which `// ""` turns into a
           # non-match rather than a jq error -- correct, since it was a fork.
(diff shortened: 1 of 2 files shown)


go-to-k:
Merged. This one existed because go-to-k/cdkd#3025 shipped the defect it was written to fix — a comment that overstates what the code does — and the same thing happened again inside this PR, twice. Four rounds total, each a correction of the previous correction.

Two things worth keeping:

- The fork-decoy comment now records that history in its own text, so a fifth round starts from it instead of re-deriving the chain.
- Every count in the test docblock was wrong at some point (`six invocations` was ten, `the ten` was twelve, `those eight` was fourteen). They were deleted rather than corrected — nothing re-checks a number in prose.

The substantive change is that `head_ref` is validated with `git check-ref-format` rather than a `-z` testing one bad shape, which also made the test's safety argument structural instead of incidental.

Issue #1910, opened 2026-08-14
fix(state): four sibling redaction writers still collapse coinciding secret expressions (journal / outputs / scrub / import)


go-to-k:
## Context

Follow-up to the PATH-based redaction that closed (#1904) / (#1900). Reported independently by the code and security reviewers of that PR.

## Gap

That change gave the deploy engine's state-save choke point a POSITION source (the unresolved template bag), so a leaf is redacted back to ITS OWN expression. Four sibling writers still call `redactSecretsForState` / `scrubResourceRecord` with the value-keyed map ALONE, even though the unresolved bag is available at each call site — so each still collapses two expressions that resolve to the same value onto whichever was recorded last:

- `src/deployment/deploy-engine.ts` — `redactOperationsForJournal` (rollback-journal ops). **The most consequential of the four**: the journal is REPLAYED to AWS by `resolveReplayProps`, so a staged (`:AWSCURRENT`) expression swapped onto an unstaged leaf re-resolves to the wrong secret VERSION and ships it to the live resource, once the two versions diverge.
- `src/deployment/deploy-engine.ts` — the stack Outputs redaction. Two outputs resolving one secret collapse onto one expression.
- `src/cli/commands/scrub.ts` — has the synthesized template in hand and passes no source, so `cdkd scrub` can bake the WRONG expression into state while claiming to clean it.
- `src/cli/commands/import.ts` — same shape on the import path.

## Direction

Each site already has the unresolved bag; pass it as the third argument, with the `PathSourceKind` the site's provenance implies (`'template-derived'` for all four — none is an AWS readback). No new plumbing is needed, which is why these were left out of the first change rather than being hard: the deploy choke point was where the reported defect was observed, and widening the blast radius of a security-path change in the same PR was not worth it.

Note the residue the first change deliberately kept, since it bounds what fixing these buys: two `SecureString` **ssm** references sharing one resolved value still collapse, because the value map keeps only the last expression and (unlike `secretsmanager`) ssm secret-ness is not derivable from the spelling. Closing that needs the resolver to record secret EXPRESSIONS in their own set rather than only as map values — worth doing in the same pass, since it is what makes the path gate exact for every reference kind.

## Verification

[...]

PR #3044, opened 2026-09-12
fix(deploy): keep a secret recorded after the drain cap released a failure, and redact the outputs against it

### LINKED ISSUE #2814: A part that is merely SLOW, not hung, can record after the drain cap released the rejection
## What

`allSettledKeepingFirstRejection` in `src/deployment/intrinsic-function-resolver.ts` waits for the remaining concurrent parts once a rejection is in hand, and that wait is capped at `DRAIN_AFTER_REJECTION_MS` (60 s) for the whole resolution. The cap is a hang guard: without it a part that never settles holds `resolveOutputs` — which runs after every resource exists in AWS, before `saveState`, with the S3 lock held and its heartbeat renewing it — so an unbounded wait costs a deploy its state, and on a first deploy every created resource becomes invisible to cdkd.

The cap bounds the WAIT, not the LATENESS. Nothing cancels the sibling: it keeps running and can record `plaintext -> expression` into `context.recordedSecretValues` after the rejection was released. So a sibling that has not finished recording when the rejection is released reopens exactly the window go-to-k/cdkd#2563 closed — the recording lands after `DeployEngine`'s `Export.Name` `finally` copied its private map into the pass map, and the plaintext is not a needle for the masking of later refusal messages or for `redactOutputs` at persist.

## Why 60 s is not obviously enough

The constant is sized against the largest FIXED wait in the file, the `Fn::GetAtt` `Ipv6CidrBlocks` poll's sleep budget (15 attempts x 2 s). That budget is a floor rather than a ceiling: when every attempt still reads `associating` the poll sleeps 30 s and awaits 15 AWS calls on top (it can also return earlier), and one part can drive several lookups in sequence through object properties or `Fn::Sub` variables. Two healthy shapes measured during review of go-to-k/cdkd#2797 already exceed the cap with no hang and no failure of their own:

- three sequential `Ipv6CidrBlocks` polls in one part: 3 x (15 x 2 s) = 90 s
- five throttled dynamic references: 5 x (1+2+4+8 s) = 75 s, at `MAX_DYNAMIC_REFERENCE_THROTTLE_RETRIES` = 4

Both need a concurrent sibling to fail before any of this matters. go-to-k/cdkd#2805 tracks a different consequence (which failure is SELECTED across nested drains), not this one.

## The floor is LOWER than "slower than 60 s" — corrected 2026-09-09

The original framing of this issue, and of the JSDoc, was that the exposure needs a healthy part slower than the whole cap. That is wrong, and the review of go-to-k/cdkd#2797 found it.

The budget is shared across a resolution, so a drain that ARMS once it is spent computes `Math.max(0, remaining - openWindow)` = 0 and releases its rejection on the next macrotask. That applies to any drain not ALREADY armed when a rejection reaches it, and the ordinary case is an inner drain that spends the full budget and throws: each of its not-yet-armed ancestors then arms against an exhausted budget and waits not at all. go-to-k/cdkd#2797's own depth-3 case shows it — levels 2 and 3 wait about 0.

Being precise about who is exposed, because the first version of this section overstated it: a sibling still pending at that moment has itself been running at least as long as the budget, so this is not "a fast sibling gets dropped". What gets no wait is that sibling's own RECORDING — a lookup begun late inside a long-running part is fast in itself and still lands after the rejection was released. The 90 s / 75 s shapes above are the cheapest way to reach the same exposure with NO nesting at all, but they are not the floor.

**Corrected again at go-to-k/cdkd#2797 round 6.** The shared budget is no longer an absolute DEADLINE but REMAINING WAIT, charged only while a drain is actually waiting and only once for waits that overlap. That matters to this issue's floor: under the deadline model, ordinary resolution time between drains spent the budget, so a later output could reach zero grace after nothing but clean work. It cannot now — reaching zero takes a full cap of drain WAIT inside one pass. Read the section below with that correction: the exposure is real, but getting there costs actual waiting rather than wall clock.

**Wider again since go-to-k/cdkd#2797 round 5.** The outputs pass wraps its whole loop in ONE budget -- as do `cdkd scrub`'s three loops under one hoisted budget, `cdkd import`'s per-resource loop and `cdkd export`'s parameter loop. Each holds a lock; the engine, scrub and import also write state downstream, while export's justification is the lock plus the CloudFormation import parameters it builds -- because otherwise `#outputs` multiplied the hold on `saveState` — `#outputs x` the cap, 200 minutes at CloudFormation's 200-output limit, with the S3 lock held. The cost lands here: a later iteration starts FRESH siblings against a budget an earlier failing output already spent, and those need not be long-running at all. The nesting case at least required the exposed sibling to have been running a while; the loop case does not.

## Options, none of them obviously right

1. **Raise the constant.** Cheapest, and buys nothing structural: any fixed number has a shape that exceeds it.
2. **Cancel the siblings instead of waiting.** Cancellation could end the wait and the lateness together, but the resolver has no cancellation hook today: it would need plumbing across the AWS calls, the sleeps, the retry wrappers and the recording itself, and the cancellation would have to be ACKNOWLEDGED before the rejection is released. Scope needs investigating before this is called cheap or complete.
3. **Make the sibling's recording survive its own lateness.** The record could go somewhere the pass reads later rather than a map copied at one instant — the shape go-to-k/cdkd#2531 took for `cdkd scrub` (a live VIEW of the pass map). A view preserves a late entry for consumers that have not run yet; it cannot help the masking or the persist that already ran, so the recording must still precede those. It narrows the window rather than closing it.
4. **Accept and document.** What ships today.

## Reproduction sketch

No live repro is attached, because provoking it needs a >60 s healthy lookup beside a failing one. The unit seam is there: `concurrentDrainCap.ms` shrinks the cap, and `tests/unit/deployment/intrinsic-resolver-concurrent-drain.test.ts` already drives held / hung / slow parts against it. A case at `ms = 20` with a sibling delayed 50 ms would pin whichever option is chosen.

Session-fit: next (not this session) -- go-to-k/cdkd#2797 documents the trade and ships the cap; changing it is a design decision with its own verification.
Severity: medium -- corrected from low. It still needs a concurrent failure, but NOT a whole part slower than the cap: any drain arming after the budget is spent -- every not-yet-armed ancestor of a drain that timed out, and every later output in the wrapped outputs pass -- abandons its pending siblings immediately, so a recording begun late inside one of them lands unwaited. The exposure is an unmasked secret in a failure message or a persisted output, the same class as issue #2563.
Effort: medium (M) -- option 1 is minutes, options 2 and 3 touch the resolver's call surface and the engine's copy shape.
Estimate: ~2-4 h -- deciding between the options and pinning the choice, not the edit.
Dup-check: searched open issues for "drain", "cap", "concurrent", "2563" -- go-to-k/cdkd#2805 is the sibling residual (error SELECTION across nested drains) and go-to-k/cdkd#2531 is the shared-map view for `cdkd scrub`; neither covers a late record surviving the cap. go-to-k/cdkd#2797's JSDoc points at THIS issue by number.

nix-tkobayashi:
## Summary

The go-to-k/cdkd#2563 drain waits for a failing resolution's other parts, but the wait is capped by one budget per resolution. A part it stops waiting for keeps running and records `plaintext -> expression` late. The cap costs the part its ORDER, not the write. So this PR fixes the readers that run after such a record arrives, and it reports when the cap stops waiting.

Two engine readers missed a late record:

- **`Export.Name` (the record was dropped).** The block resolved into a per-output local map and copied it into the outputs pass map in a `finally`. A record that landed after the copy died with the local, so a later export name carrying the plaintext was published as a state key. The block now resolves through `ForwardingSecrets`. It keeps its own entries, which is the exact "substituted into this name" set `exportNameSecretExposure` reads. It also writes each entry through to the pass map as it is recorded. Resolved pairs stay on the instance, following the existing entries-only rule (go-to-k/cdkd#2485).
- **The outputs readers (the record arrived too late).** The outputs bag was redacted when the pass ended and read unchanged afterwards. A record that lands during the observed-capture drain or during the final save's await missed the saved outputs, the exports index and the deploy summary.
  - `redactOutputs` now folds the outputs pass map in every time it runs (`absorbOutputsPassSecrets`, entries and pairs), instead of trusting a copy taken in the pass's `finally`.
  - The final save already redacted the bag again (`withParentInfo`). The exports index and the deploy summary now do the same at the moment they read, on both the changes and the no-change path.
  - That is not a promise they equal the saved copy: they run after the save's await, so a record arriving during it reaches them and not the copy the save already took. The residual test pins exactly that asymmetry.
  - Redacting an already-redacted bag is not idempotent in one pathological case — when one secret's plaintext IS another's complete reference token, the second pass rewrites the first pass's token, converging on the third. This PR does not introduce it (`withParentInfo` already redacted the saved bag a second time before this change) and it makes the index and the summary agree with what state holds rather than diverging from it.

When `allSettledKeepingFirstRejection` releases a rejection with inputs still pending, it reports how many, at most once per drain budget. The resolver turns that into a warning.

## What stays open, and why it is not a follow-up

A reader that took its copy before the record arrived cannot see it without waiting. That means a failure message already printed, or a state save already sent. The wait is exactly what the cap bounds. One new test pins this residual as its premise: the saved state still carries the plaintext while the exports index and the summary written after it do not.

Cancelling the part instead (option 2 on the issue) would not close this. A cancelled part never records, so its plaintext would be missing for every reader instead of only for the early ones. The new warning makes that case visible.

## Verification

- New cases in `tests/unit/deployment/intrinsic-resolver-concurrent-drain.test.ts`. The cases order events by what they observe (a warning fired, a record landed, a redaction ran, a save entered), not by racing timers:
  - the resolver warns once, with the count, and the late record still lands;
  - the count includes every pending part;
  - the list call site reports too;
  - nested drains released by one failure warn once;
  - an independent resolution warns again, and two resolutions under one shared budget warn once;
  - no warning when every part settled inside the cap;
  - the engine refuses a later export name that carries a plaintext recorded after the drain released;
  - a plaintext recorded during the observed-capture drain is redacted from the save, the exports index and the summary, and no warning carries it;
  - a plaintext recorded while the final save is in flight is still redacted from the exports index and the summary;
  - the no-change path prints its kept bag redacted, as its save writes it;
  - a reused engine does not carry the previous deploy's pass maps.
- Mutation probes, one mutation per probe, tree restored byte-exact between them:

| Probe | Mutation | Result (3 files, 84 cases) |
|---|---|---|
| R3 | pending count ignores `settled` | KILLED: 4 failed |
| R5 | fulfilled input not counted as settled | KILLED: 1 failed |
| R6 | rejected input not counted as settled | KILLED: 4 failed |
| R7 | list call site passes no report | KILLED: 2 failed |
| R8 | join call site passes no report | KILLED: 14 failed |
| R10 | singular / plural wording swapped | KILLED: 3 failed |
| E1 | `ForwardingSecrets` does not write through | KILLED: 4 failed |
| E2 | `ForwardingSecrets` keeps no own entry | KILLED: 4 failed |
| E3 | `nameSecrets` back to a plain `Map` | KILLED: 4 failed |
| E5 | `redactOutputs` does not absorb the pass map | KILLED: 22 failed |
| E7 | absorb drops resolved pairs | KILLED: 3 failed |
| E8 | absorb drops entries | KILLED: 22 failed |
| E9 | pass-map list not reset per deploy | KILLED: 1 failed |
| R1 | per-budget dedupe: `abandonReported.add` removed | KILLED: 2 failed |
| R2 | `pending > 0` -> `pending >= 0` | KILLED: 1 failed |
| R4 | `onAbandoned(pending)` call removed | KILLED: 10 failed |
| R9 | dedupe check bypassed | KILLED: 2 failed |
| R11 | dedupe keyed per module instead of per budget | KILLED: 10 failed |
| E4 | pass map not registered for later redaction | KILLED: 22 failed |
| E10 | exports index not re-redacted | KILLED: 2 failed |
| E11 | deploy summary not re-redacted | KILLED: 2 failed |
| E14 | no-change summary not re-redacted | KILLED: 1 failed |

All 22 killed; both source files restored byte-exact after every probe (sha1 verified).

- Real-AWS gating runs on this branch, both clean, both recorded in the ledger: `export` (262 s, deploy -> CFn adoption -> destroy, state cleared, no orphans) and `lambda` (121 s, broad set, 9 deleted / 0 errors). `export` is the fixture closest to this change — it drives `Export.Name`, the exports index and the outputs bag — and `lambda` covers the cross-cutting deploy/destroy paths this diff touches. Markers: `integ-destroy` and `integ-broad`.
- No live AWS run REPRODUCING THE ISSUE ITSELF. Reaching the cap for real needs a healthy lookup slower than `DRAIN_AFTER_REJECTION_MS` (60 s) beside a failing sibling, and the issue already records that no live repro was feasible. The cases above drive the real `IntrinsicFunctionResolver` and the real `DeployEngine`, with only the Secrets Manager client faked, and use the existing `concurrentDrainCap.ms` seam at 20 ms. This follows the issue's own reproduction sketch.
- Full suite on the final tree, rebased onto `origin/main` after go-to-k/cdkd#3020 merged: 992 files, 21,896 passed, 1 skipped, no type errors (root-asserted, one run). `vp run check`, `typecheck:test`, `build`, `gen:all-matrices` and `audit:coverage:check` are clean, with no generated artifact left dirty.
- `scripts/check-resolver-mask-coverage.ts` requires every value interpolated into a resolver message to be masked or annotated. So the new warning carries two `not-in-class` notes: `parts` is a count in fixed wording, and `capSeconds` is the cap. Its exactly-pinned population counts move from 131 / 155 / 98 to 132 / 155 / 100, measured with `scan()` on the final tree (0 findings, 0 unconsumed markers). The bands are unchanged.
- Codex review of the uncommitted change, 7 delta rounds plus 2 maintainer-checklist passes. Round 1: re-redacting an already-redacted bag is not idempotent when one secret's plaintext is another's complete reference token, and a record landing during the final save's await still missed the exports index and the summary. Rounds 3 and 4: the no-change path wording overclaimed, then understated what its summary prints. Checklist pass 1 (seven groups): the no-change summary, clauses no test could pin (a fallback arm, an unused reset, optional-callback and presence guards), per-budget warning lifetime, weak premise and summary assertions, and stale comments / rule-file text. Checklist pass 2: the warning-plaintext check read only the resolver's logger, and one comment promised the index equals the saved copy. Rounds 2, 5, 6 and 7 returned NO ISSUES.

- Reviewers (spec / code / test / security, in parallel against this diff). Every finding is applied:
  - the no-change path's exports-index publish now re-redacts like its sibling on the changes path. It cannot differ today — a released drain always leaves an output unresolved, so `resolutionFailed` is true and both flags guarding that block are false — and the comment there says so, because three reviewers had to re-derive that gate to clear the site;
  - `absorbOutputsPassSecrets` no longer claims unconditional idempotence: repeating is a no-op for entries, while a LATE pair can mark an expression conflicting and cost a leaf its positioning (still masked);
  - `redactOutputs` documents that it folds the pass map into `outputSecrets`, and `ForwardingSecrets` documents why it overrides only `set` where `cdkd scrub`'s view overrides `delete` / `clear` too;
  - two comments in `deploy-engine-output-failure-masking.test.ts` that described the deleted `finally` merge were corrected;
  - a new case renders the DEFAULT 60 s cap wording (every other case shrinks the cap through the test seam), and the save-in-flight case says why it asserts a plaintext in state.

Closes #2814


go-to-k:
Reviewed at head `8f1a5eac`. Spec / code / test / security ran in parallel against this diff; CI is green on every check. The engine change is correct as written — I traced the plaintext from the late record to every reader I could enumerate (state save, both `updateForStack` sites, both summary paths, `exportNameSecretExposure`, the failure message, deployment events, the rollback journal, `cdkd diff` / `scrub` / `export` / `import`) and found no reader left unredacted, and no fail-open: `absorbOutputsPassSecrets` has no catch, so a throw fails the save rather than serving a plaintext bag.

One item must change before this merges, one should, and four are yours to take or leave.

## M0 — required before merge: the refold is unfenced

`redactOutputs` folding the pass map in on **every** call is the central behavioral claim of this PR, and no case pins it. The probe table's E4 / E5 / E8 all delete the fold entirely, so they only pin the **first** call — the one reached while `outputSecrets` is still empty.

I ran the narrower mutant that keeps the fold but skips the **refold**, in a `git archive` export of `8f1a5eac` (`src/deployment/deploy-engine.ts:1612`):

```ts
if (this.outputSecrets.size === 0) this.absorbOutputsPassSecrets();
```

| Selection | Baseline | Mutant |
|---|---|---|
| the PR's 3 test files | 123 passed | **123 passed** |
| `tests/unit/deployment` + `tests/unit/scripts` | 126 failed / 6288 passed (256 files) | **126 failed / 6288 passed — byte-identical** |

Zero new failures across 6414 tests. (The 126 are environmental in an export with no `.git` / `.claude`; both columns are the same run shape, and the file was restored byte-exact — sha1 `95073c5e5ade1f7341e22142cf39c1abd316fda2` before and after.)

The reason it survives: all three late-record engine cases (`intrinsic-resolver-concurrent-drain.test.ts:2302`, `:2365`, `:2434`) build templates in which the outputs pass resolves **no** secret successfully, so `outputSecrets` is empty when the readers run and the first fold is also the only one. A real stack with two secret-bearing outputs — one resolving cleanly, one failing — is exactly where a future narrowing would silently drop the late needle from the save, the index and the summary.

Please add one case of that shape: the `:2302` template plus a second output resolving a secret cleanly during the pass, so `outputSecrets` is non-empty by the time the late record lands. It reds under the mutant above and passes on the real code.

## m1 — required before merge: a rule file went stale in this diff

`.claude/rules/layout-deployment-secrets.md:364` still says the outputs bag is "consumed by all **three** outputs-redaction sites through one `redactOutputs` helper", and that "**two of the three** sites hand it a bag that is NOT this generation's". Measured:

```
base c06f3e56: 3 call sites (2122, 2995, 3951)
tip  8f1a5eac: 7 call sites (2226, 3099, 3275, 3319, 3452, 3471, 4092)
```

The substantive claim (one helper, `TEMPLATE_SOURCED_RULES`, more than one foreign-generation bag) survives; both counts do not. This PR edits two other rule files but not this sentence, and agent-instruction files get no down-bias here — a wrong count propagates to every future session. Please recount both numbers.

## Out of scope for this PR

`intrinsic-resolver-concurrent-drain.test.ts:463` (`a LOOP of resolutions wrapped in one budget…`) flaked 2 of ~20 runs on an idle machine during this review. It is unmodified from `c06f3e56` and is already filed as #2907. Nothing for you to do here.

## Optional — take or leave, none of these block

- **m2** `deploy-engine.ts:8151` / `:8155`: `context.recordedSecretValues ?? EMPTY_SECRETS` pushes the process-wide singleton at `:157` onto the per-deploy `outputsPassSecretMaps`. Inert today (nothing writes it), but it puts a cross-deploy object in a per-deploy needle list. `if (context.recordedSecretValues) this.outputsPassSecretMaps.push(...)` closes it at the callee.
- **m3** `deploy-engine.ts:3262-3264`: the comment's first reason is wrong twice. "a second pass over one is idempotent" contradicts the non-idempotent case this PR documents at `absorbOutputsPassSecrets`, and on the `exportSetChanged`-only arm `refreshedState.outputs` is `persistedOutputs` (`:3194`) — the previous deploy's bag, not a `redactOutputs` product of this process. The block is still unreachable for the second reason you give (`resolutionFailed` ⇒ both guards false), which I verified holds: no swallowing `catch` in the resolver encloses an `allSettledKeepingFirstRejection`. Only the first reason needs dropping.
- **m4** `intrinsic-function-resolver.ts:1734-1738`: when `shared === undefined`, `reportKey = {}` is unreachable the instant the statement ends, so `abandonReported.add(reportKey)` can never be observed. `if (shared === undefined) onAbandoned(pending); else if (!abandonReported.has(shared)) {…}` says the same thing without the dead add.
- **m5** `deploy-engine.ts:4102`: `rememberRecoverableMaskedOutputs` has exactly one call site, at the end of the outputs pass. An output that a **later** `redactOutputs` newly masks therefore never enters the recoverable cache, so an in-process cross-stack consumer can be refused on `***`. Not an exposure, and strictly the fail-safe direction — worth a sentence wherever you land on it.

Push M0 and m1 and I will re-review; the four optional items need no reply unless you want one.

nix-tkobayashi:
Thank you — both required items are pushed, and I took all four optional ones.

## M0 — the refold is now fenced

You are right that nothing pinned it. One correction to the mechanism, found by a delta review of my own fix: the three late-record cases you name do leave the pass with no successfully resolved secret, but the export-name case above them DOES record — both of its parts settle before the pass ends. It misses the mutant for a different reason, namely that those recordings land before the first fold. So the gap is "no case has a needle in the bag BEFORE the late one arrives", and the new case's comment says it that way.

New case, `folds the pass map in AGAIN when the needle bag is already non-empty`. Its template is the observed-capture one plus a `Clean` output that resolves a secret cleanly during the pass, so `outputSecrets` is non-empty by the time the late record lands. Its premise assertion reads the bag directly (`outputSecrets.get(valueOf(CLEAN_ID))`) rather than inferring it, so the case cannot quietly become the empty-bag shape again.

Measured against your exact mutant (`if (this.outputSecrets.size === 0) this.absorbOutputsPassSecrets();`) on the same file:

| Run | `intrinsic-resolver-concurrent-drain.test.ts` |
|---|---|
| baseline | 54 passed |
| mutant | 2 failed / 52 passed |

The two are the new case and `a drain arming while another WAITS gets what is left of the open window` (`:700`). To be precise about the second one, since it is NOT the case you filed: that is `a LOOP of resolutions wrapped in one budget…` (`:463`). Both are timing-sensitive and neither is touched by this PR, and `:700` passed in the baseline run of the same file seconds earlier (54 passed), so I read it as the same class of flake under load rather than a second mutant kill — but I am not claiming it is go-to-k/cdkd#2907 itself. The new case is the one the mutant kills deterministically. Source restored byte-exact (sha1 `9dbd83145beb63652fcddbf519baa0b337c269ed` before and after).

## m1 — rule file recounted

`.claude/rules/layout-deployment-secrets.md` now says seven outputs-redaction sites, three of which can hand `redactOutputs` a bag that is not this generation's: the persist walk (`redactStateForPersist`, which on the no-change and failure paths walks the previous deploy's bag), the no-change path's exports index, and the no-change path's summary. The substantive claim — one helper, `TEMPLATE_SOURCED_RULES`, more than one foreign-generation bag — is unchanged.

## The four optional items, all taken

- **m2** — the pass map is pushed under `if (context.recordedSecretValues)`, so the `?? EMPTY_SECRETS` singleton never lands in the per-deploy list.
- **m3** — dropped the first reason. You are right twice over: on the `exportSetChanged`-only arm the bag is `persistedOutputs`, and this PR's own `absorbOutputsPassSecrets` paragraph says a repeat is not unconditionally idempotent. The comment now gives the `resolutionFailed` reason alone, plus the fail-safe direction.
- **m4** — the dedupe no longer adds an unreachable key: a drain with no store reports directly, and only a drain with a budget joins the set.
- **m5** — stated at the single call site: an output a later redaction newly masks never enters the recoverable store, so an in-process cross-stack consumer is refused on `***` rather than served the plaintext.

## Verification on this head

- Full suite: 992 files, 21,898 passed, 1 skipped, no type errors, root-asserted. (One earlier run of the same tree reported every test passing but exited 1 on a vitest worker that died; the re-run is the clean one.)
- Codex delta review: two rounds on this delta. The first found the false statement corrected above; the second returned NO ISSUES.
- CI: re-run on this push.

go-to-k/cdkd#2907 — understood, nothing changed here for it.

go-to-k:
Reviewed at head `fb60f297`. **M0 and m1 are both closed**, and all four optional items landed correctly. One required item remains, and it is the other half of m1.

## M0 — closed, verified independently

I re-ran my exact mutant against your head in a fresh export (source sha1 `9dbd83145beb63652fcddbf519baa0b337c269ed` before and after, matching yours):

| Run | `intrinsic-resolver-concurrent-drain.test.ts` | whole `tests/unit/deployment` |
|---|---|---|
| baseline | 54 passed | 2789 passed |
| mutant | **1 failed** / 53 passed | **1 failed** / 2788 passed |

`folds the pass map in AGAIN when the needle bag is already non-empty` is the only failure in 2789 tests, and it fails on the disclosure assertion rather than vacuously on its premise. The premise reading the bag directly is the right call — that is what stops it drifting back to the empty-bag shape.

Your mechanism correction is right and my round-1 rationale was wrong: the export-name case does record, and it misses the mutant because those recordings land before the first fold, not because nothing records. The reviewers reproduced that independently.

On `:700` — agreed it is a flake, and your caution about not claiming it as #2907 was warranted. It never constructs a `DeployEngine`, so `redactOutputs` is structurally unreachable from it and the mutant cannot change its verdict; it passed 10/10 on both arms here and in both 2789-test runs. It is a wall-clock sibling of the family #2907 describes rather than a new phenomenon. Nothing for you to do.

## m1 — the rule file is right, three siblings in `src/` are not

`.claude/rules/layout-deployment-secrets.md` now reads correctly: I counted seven `this.redactOutputs(` sites (2226, 3099, 3276, 3320, 3453, 3472, 4093), and exactly the three you name take a foreign-generation bag (2226, 3276, 3320). The other four take this generation's.

But this PR is what moved that count from three to seven, and three other places in the repo still state the old one. Two reviewers converged on the first independently:

- **`src/deployment/deploy-engine.ts:1617`** — "which two of this method's **three** callers cannot say". This is `redactOutputs`'s own doc comment, in the method this PR rewrote. Now seven callers, three of which cannot say it. Note the "two" was already wrong at `c06f3e56`: the base had three sites (2122, 2995, 3951) of which only the persist walk at 2122 was foreign, so it should have read "one of three" then. Both numbers need setting, not just the one this PR moved.
- **`src/deployment/secret-redaction.ts:1798`** — the rules table row for deploy `redactOutputs`, which says "(3 sites)". Now 7.
- **`src/cli/commands/scrub.ts:4433`** — "the **2 of 3** `redactOutputs` sites whose bag WAS freshly resolved from today's template". Now 4 of 7 (3099, 3453, 3472, 4093).

Please recount all three. Same reason m1 was required: a count that is wrong is worse than a count that is absent, because the next reader budgets their audit against it.

## Optional — take or leave

- **m6** The new case's comment at `:2374-2378` is true but understates the export-name case. Probing it: with the fold deleted **entirely**, `:2241` still passes — so it is insensitive to `absorbOutputsPassSecrets` at all, not merely to the refold. One clause would say that.
- **m7** Same comment, `:2371`: "both its parts settle before the pass ends" does not match the fixture. `Exporter`'s name is `Fn::Join` over `[ref(SLOW_ID), ref(FAIL_ID)]` and `FAIL_ID` is in `control.fails`, so only `SLOW_ID` records; the second recorder is the separate `Spacer` output via `SPACER_ID` (`:2258-2266`).

## Checked and not raising

The two unfenced arms a probe found — the `if (context.recordedSecretValues)` guard at `deploy-engine.ts:8166` and the `shared === undefined` arm at `intrinsic-function-resolver.ts:1740` — are both documented as unreachable in production (`buildResolverContext` always sets the field; no public entry point reaches a storeless drain), and fencing either would mean poking internals to build a state the code cannot enter. Correct as they are.

Security and code reviewers both came back with nothing at any severity on this delta.

Push the three recounts and I will merge; m6 and m7 need no reply.

nix-tkobayashi:
All three recounts are pushed, plus m6 and m7, and a sweep found two more statements of the same count that your list did not name.

## The three you named

Measured from `grep -n "this.redactOutputs(" src/deployment/deploy-engine.ts` at this head (`acf766b5`): seven sites — 2237 (persist walk), 3110 (no-change resolution), 3287 (no-change index), 3331 (no-change summary), 3464 (changes index), 3483 (changes summary), 4104 (outputs pass). Three can take a foreign-generation bag (2237, 3287, 3331); four take this pass's (3110, 3464, 3483, 4104). Your numbers were right at `fb60f297`; these comment edits moved them.

- `src/deployment/deploy-engine.ts` `redactOutputs` doc — now "three of this method's seven callers cannot say", naming them, and recording that the "two" was wrong before this PR too.
- `src/deployment/secret-redaction.ts` rules table — "(7 sites)".
- `src/cli/commands/scrub.ts` — "the 4 of 7 `redactOutputs` sites whose bag WAS freshly resolved".

## Two more the sweep found

A grep for other statements of the same count turned up two your list did not include. Both are fixed in the same commit:

- `tests/unit/deployment/deploy-engine-outputs-generation-rules.test.ts` — "one method behind three call sites, and two of them". Now seven and three. My first rewrite of this one was itself wrong: I wrote that the no-change index takes the previous bag "whenever a resolution failure keeps today's from landing", which cannot happen — a failure disables both flags guarding the index write. It now says the SUMMARY takes it on the failure path and the INDEX on the `exportSetChanged`-only arm.
- `src/deployment/deploy-engine.ts` `rememberRecoverableMaskedOutputs` doc — "The other two `redactOutputs` callers hand it a bag from a previous generation". Now: of the six others, three take a foreign bag, while the no-change path performs the FIRST redaction of its own resolution and the changes path's two are re-redactions of a bag this pass produced.

## One left alone, deliberately

`changelog.d/_archive.md` repeats "Two of its three call sites cannot say that" — historically false in the same way (at that base only the persist walk took a foreign bag). I did not touch it: it is an archived record of what shipped, entries here are forward-only, and editing a shipped claim is your call rather than mine. Say the word and I will correct it in place or add a follow-up entry, whichever you prefer.

## m6 / m7

Both taken, in the new case's comment. m7 is exactly as you describe — the failing name records only `SLOW_ID` and the second recorder is the `Spacer` output via `SPACER_ID`. For m6 I re-measured rather than quoting you: with `absorbOutputsPassSecrets` deleted outright, the export-name case still passes (1 passed / 53 skipped, filtered by name), so it is insensitive to the fold altogether — it asserts a REFUSED export name rather than a redacted bag. The comment says that now.

## Verification on this head

- Full suite: 992 files, 21,898 passed, 1 skipped, no type errors, root-asserted.
- Codex delta review: three more rounds on this round's edits. Rounds 14 and 15 each found a false statement in wording I had just written — the no-change index claim above, then "wrong on both halves", which over-corrected a count that had been right. Round 16 returned NO ISSUES.
- CI: re-run on this push.

go-to-k:
Reviewed at head `acf766b5`. Every correction in this round is TRUE — I verified each against the code rather than against your reply — and the two you found on your own sweep were real. One statement of the same count is still standing, and my own sweep missed it too.

## Verified this round

All three axes came back with the corrections holding:

- Seven `this.redactOutputs(` sites (2237, 3110, 3287, 3331, 3464, 3483, 4104); foreign-capable exactly {2237, 3287, 3331}; freshly-resolved exactly {3110, 3464, 3483, 4104}. Your "4 of 7" in `scrub.ts` and "(7 sites)" in `secret-redaction.ts` both follow.
- The historical parenthetical checks out at `c06f3e56`: three sites, of which only the persist walk took a foreign bag — so "two" was wrong then.
- `rememberRecoverableMaskedOutputs`'s 6-others / 3-foreign / 3-not-this-pass breakdown is right, and it does have exactly one caller.
- The generation-rules header's summary/index split is right on the second attempt, including "a failure disables both flags guarding the index write". That is not prose-only: mutating the guard to `outputsChanged || exportSetChanged || resolutionFailed` reds two cases (`deploy-engine-outputs-only-change.test.ts` and `deploy-engine-skipped-outputs.test.ts`).
- m6's insensitivity claim was re-measured here rather than taken from either of us: with `absorbOutputsPassSecrets`'s body deleted, `:2241` is among the 49 that still pass while 5 siblings go red.
- No regression: `tests/unit/deployment` is 2789 passed / 0 failed at this head, and the round-1 mutant still dies on the new case alone (1 failed / 2788).

## m8 — required: one more statement of the count, in the same file

**`src/deployment/deploy-engine.ts:1527`**

```
* A single entry point because THREE call sites redact this same bag — the
* state-persist choke point, the no-change re-check, and the post-deploy
* publish that feeds the exports index / deploy summary —
```

Present tense, 18 lines above the corrected "Of the six other `redactOutputs` callers" at `:1546` and 97 above "three of this method's seven callers" at `:1624`. It is the base's three-site enumeration, and #2814 is what made it false. A reader arriving at `:1527` first carries "three" into both corrections below it.

The third clause also needs splitting now: at base the index and the summary were one post-deploy publish; this PR made them two separate re-redacting sites.

**Why neither of us caught it**: my sweep grepped for lines containing `redactOutputs`, and this line does not contain the word. A needle that finds it is the phrase, not the symbol:

```bash
grep -rnE 'redact this same bag|outputs-redaction site' src tests docs .claude scripts
```

I re-ran the sweep that way across `src/`, `tests/`, `docs/`, `.claude/` and `scripts/`, and `:1527` is the only remaining one. The other count claims in that file are about `resolveOutputs`, which genuinely has two call sites (`:3113`, `:4086` — `:8458` and `:8480` are both correct), and `:1101` is about `perResourceSecrets`' population sites.

## The archive — decided, nothing for you to do

Leave it. `changelog.d/_header.md:61-68` makes "no lane edits the archive again" the standing convention, and the archive is a record of what shipped rather than a live reference. You were right to ask rather than edit. For completeness there are **two** instances there, not one: `:672` ("Two of its three call sites") and `:720` ("all three outputs-redaction sites route through one `redactOutputs` helper"). Both stay.

## Optional — take or leave

- **m9** `deploy-engine.ts:1629` is 91 columns inside a block otherwise wrapped at 73-80 — the sentence head moved and the tail was not rewrapped. `printWidth` is 100 so oxfmt will not flag it, and formatters do not reflow comment text, so nothing will fix it automatically. Same class a reviewer raised last round.
- **m10** `deploy-engine.ts:1546` says three callers "hand it a bag from a previous generation"; `:3287` and `:3331` only *can* — each is this generation's on its other arm. The two sibling sentences you wrote this round both use the conditional ("cannot say" / "can hand").
- **m11** `deploy-engine-outputs-generation-rules.test.ts:133-143` now names three foreign-bag sites, but that file builds its engine with no `exportIndexStore` and asserts only `saveState` calls, so it exercises just the persist walk. The other two are covered elsewhere (`intrinsic-resolver-concurrent-drain.test.ts:2513`, `deploy-engine-outputs-only-change.test.ts:535/566`). No coverage gap — the header just reads as though the file covers all three.
- **m12** `intrinsic-resolver-concurrent-drain.test.ts:2373`: `SPACER_ID` sits in `Spacer`'s **Export Name**, not its `Value`, so "its `Spacer` output resolves `SPACER_ID`" is loose; and that line uses `--` where the rest of the block uses `—`.

## Not raising

The three stacked `/** */` blocks at `:1516-1522`, `:1523-1533` and `:1534-1555` all precede one method, so the first two are attached to nothing. That predates this PR (identical at `c06f3e56`), so it is not yours to fix — but m8 lands you inside that exact block, and reattaching `:1523-1533` to `redactOutputs` is nearly free if you want it. `secret-redaction.ts:1835`'s "one row per write site" against a row folding seven sites is also pre-existing.

Push m8 and I will merge.

nix-tkobayashi:
m8 is pushed, and I took m9 through m12 as well.

## m8

`redactOutputs`' own doc opened with the base's enumeration — "THREE call sites redact this same bag — the state-persist choke point, the no-change re-check, and the post-deploy publish that feeds the exports index / deploy summary". It now reads SEVEN, with the third clause split the way this PR split it: each path's index and summary are separate sites that redact when they read. A parenthetical records that it was three when the sentence was written and what #2814 changed.

Your point about why both sweeps missed it is the useful part, and it is now the one I will reuse: the line never says `redactOutputs`, so a symbol grep cannot see it. I re-ran your phrase sweep across `src/`, `tests/`, `docs/`, `.claude/` and `scripts/` after fixing it. What it returns now is that corrected line, `.claude/rules/layout-deployment-secrets.md` (already seven), and `docs/changelog-cdkd.md` — the assembled, gitignored changelog, whose text comes from the archive entries you decided stay.

## m9 through m12

- **m9** — rewrapped, and I swept the class rather than the instance: every comment line this PR added above ~92 columns is now wrapped, including one at 102 and the two `not-in-class(...)` notes. I had assumed those two had to stay one-per-line; they do not — `scripts/check-resolver-mask-coverage.ts` parses markers per comment RANGE and takes several per range, and `scan()` still reports 132 / 155 / 100 with 0 findings and 0 unconsumed markers after the wrap. The one long line left is the rules-table row in `secret-redaction.ts`, whose columns are aligned.
- **m10** — now "three CAN hand it a bag from a previous generation", with the persist walk as the one that always may and the other two qualified to the arms where that path keeps the previous bag.
- **m11** — the header now says this file covers the persist walk only (no exports-index store, asserts `saveState` alone) and names the two files that cover the other two sites.
- **m12** — corrected: it is the `Spacer` output's EXPORT NAME that resolves `SPACER_ID`, not its `Value`. Em dash fixed too.

## The archive

Understood, and thank you for the second instance — I had only found `:672`. Both stay.

## The stacked doc blocks

Leaving them. You are right that m8 put me inside that block, but reattaching `:1523-1533` changes what a reader sees above a method this PR already rewrites, and it is pre-existing rather than something this change broke. If you would rather have it done while the file is open, say so and I will push it as its own commit.

## Verification on this head

- Full suite: 992 files, 21,898 passed, 1 skipped, no type errors, root-asserted.
- Codex delta review on this round: NO ISSUES.
- CI: green on this head (every check pass, 0 fail).

go-to-k:
Reviewed at head `08fb1d8d`. m8 through m12 all landed; m9 is proven rather than trusted — `scan()` at this head reports 132 / 155 / 100 with 0 findings and 0 unconsumed markers, and deleting the first line of the wrapped `not-in-class(parts)` note produces a finding at the site, so the wrapped marker is genuinely consumed, not passing for another reason. Full suite here: 992 files, 21,898 passed, 1 skipped, no type errors. The stacked doc blocks stay as you decided.

Two sentences written this round are false, and one of them is my wording. Both are the last items; exact replacement text is below so this does not take another round.

## m13 — required: m8's history clause is false for four of the seven

`src/deployment/deploy-engine.ts:1527-1535` now reads "SEVEN call sites redact this same bag — … — and before this they each spelled the value-only redaction separately. Two outputs resolving one secret collapsed onto whichever expression was recorded last at every one of them."

That was true of the THREE #1910 unified. The four sites #2814 added never spelled any redaction: at `c06f3e56` they read the bag as-is — `importableOutputs(refreshedState)` (3154), `buildDisplayOutputs(template, persistedOutputs)` (3193), `importableOutputs(newState)` (3313), `buildDisplayOutputs(template, newState.outputs ?? {})` (3331). Moving the subject from three to seven made the clause claim a history four of them do not have, and the parenthetical does not rescue it because the main sentence is what a reader parses.

Also "split the post-deploy publish into a separate index and summary on each path" is not what happened: the outputs-pass site (base 3951, now 4110) still exists, and the no-change path's index and summary sit downstream of the re-check (base 2995), not of the publish.

Replacement for the paragraph:

```
* A single entry point because SEVEN call sites redact this same bag — the
* state-persist choke point, the no-change re-check, that path's exports
* index and deploy summary, the changes path's index and summary, and the
* outputs pass itself. The first three are the ones issue #1910 unified:
* before it they each spelled the value-only redaction separately, and two
* outputs resolving one secret collapsed onto whichever expression was
* recorded last at all three. The other four are issue #2814's: each read
* the bag unredacted until then, and now redacts at the moment it reads.
```

## m14 — required: m11's coverage claim is mine, and it overclaims

`tests/unit/deployment/deploy-engine-outputs-generation-rules.test.ts:145-146`: "The other two are covered by `intrinsic-resolver-concurrent-drain.test.ts` and `deploy-engine-outputs-only-change.test.ts`."

The summary half is true — `intrinsic-resolver-concurrent-drain.test.ts:2513` asserts the previous-generation bag is re-redacted. The index half is not: `deploy-engine-outputs-only-change.test.ts:535` / `:566` pin that the `exportSetChanged`-only arm REACHES the index with the carried bag, and the file asserts nothing about redaction (one incidental match for the whole secret/redact/mask vocabulary). That is consistent with the site's own comment: its re-redaction is the fail-safe one no case can pin today, because a late needle cannot reach it. I wrote "covered elsewhere" in my last comment without checking what the second file asserts; that is on me.

Replacement for the two sentences:

```
* THIS file covers the first of the three: its engine is built with no
* exports-index store and it asserts `saveState` alone, so the persist walk
* is the site it exercises. The no-change summary's re-redaction of the
* previous bag is asserted by `intrinsic-resolver-concurrent-drain.test.ts`;
* the no-change index's arm is reached by
* `deploy-engine-outputs-only-change.test.ts`, but its re-redaction is the
* fail-safe one no case pins — see the comment at that call site.
```

Push those two and I will run the gating integs from this head and merge.

nix-tkobayashi:
Both pushed. I used your replacement text for m14 verbatim and had to change one clause of m13 — that clause is the only thing in this reply that needs your eye.

## m13 — applied, with one correction to the replacement text

Your paragraph enumerates the seven sites ending with the outputs pass, then says "The first three are the ones issue #1910 unified". By that order the first three are the persist choke point, the re-check and the no-change EXPORTS INDEX — and the index is one of the four this PR added. The trio #1910 unified is the persist walk, the re-check and the outputs pass, which your sentence lists last (`c06f3e56`: 2122, 2995, 3951 — I re-read all three in that tree rather than trusting the numbers).

So I kept your structure and replaced the positional reference with a named one:

```
* outputs pass itself. THREE of those are the ones issue #1910 unified —
* the persist choke point, the re-check, and the outputs pass, which the
* list above ends on rather than opens with: before it they each spelled
* the value-only redaction separately, and two outputs resolving one secret
* collapsed onto whichever expression was recorded last at all three. The
* other FOUR are issue #2814's, both paths' index and summary: each read
* the bag unredacted until then, and now redacts at the moment it reads.
```

Everything else in your text stands: the four added sites read the bag unredacted at the base (I checked each of the four lines you cited), and the "split the post-deploy publish" framing is gone.

## m14 — applied verbatim

Your replacement is in as written. I verified the half that changed the claim rather than taking it: `deploy-engine-outputs-only-change.test.ts` has zero matches for redact / secret / mask / `{{resolve`, so "covered" was wrong and "reached, but its re-redaction is the fail-safe one no case pins" is right.

## Verification on this head

- Full suite: 992 files, 21,898 passed, 1 skipped, no type errors, root-asserted.
- Two intermediate runs each failed one or three script-critic cases at 6-7 s (`gen-handled-property-wiring`, `gen-sdk-attr-coverage`, `withretry-interrupt`). Each passed in isolation — 163 and 34 at the same heads. Two of the three scan `src/provisioning/**`, which this PR does not touch; the third, `gen-sdk-attr-coverage`, does read `src/deployment/intrinsic-function-resolver.ts`, so it is not an untouched subject — but its failing case is the floor over SDK-backed types carrying an Arn/Url attribute, which this PR's comment-only edits to that file cannot move, and it passed in isolation at that head. I read all three as load timeouts on this machine and re-ran rather than recording a green over them. Flagging it in case you see the same shape on your side.
- Codex delta review: the round that applied your text caught the "first three" slip above; the round after the fix returned NO ISSUES.
- CI: green on this head (every check pass, 0 fail).

go-to-k:
Merged at `f2e370d2` — thank you. Five rounds, and every one of them made the change more honest: the refold fence you added on round 2 is the case this file was missing, and catching my own "first three" slip on round 5 is exactly the kind of reading this PR needed. Both gating integs ran clean from this head on the maintainer side (`export` 243 s, `lambda` 80 s, 0 orphans); the ledger rows land in a follow-up.

PR #3052, opened 2026-09-12
fix(secret-redaction): position a sub-floor secret embedded through an Fn::Join / Fn::Sub source, and mark cdkd import's own resolution bag

nix-tkobayashi:
## Summary

A 1-3 character secret sits below the value scan's `MIN_NEEDLE_LENGTH` floor, so wherever the position arms refuse, the plaintext persists (go-to-k/cdkd#2745). This PR closes two of its three sites; the third (the nested-stack parameter carry in `inheritedSecretsCarriedBy`, `src/deployment/intrinsic-function-resolver.ts`) is queued behind go-to-k/cdkd#3044, which holds that file, and follows as its own PR once that merges. So this is **part of go-to-k/cdkd#2745**, not its close; the issue closes with the second PR. A trial merge against go-to-k/cdkd#3044's head is clean (the two share only `.claude/rules/layout-deployment-secrets.md` and the integ ledger, on different lines).

**Site 1 — an `Fn::Join` / `Fn::Sub` source.** The dominant CDK shape, `'port:' + secret.secretValueFromJson('pin')`, synthesizes (probed on aws-cdk-lib 2.268.0) as an `Fn::Join` whose first part FUSES the prefix into the token's opening and whose ARN `Ref` sits INSIDE the token. The skeleton arm positions only a whole-token leaf and the literal span arm needs a source string, so the leaf fell to the value scan — a sibling's spelling for a 4+ character middle, plaintext below the floor. `positionByIntrinsicFrame` (`src/deployment/secret-redaction.ts`) renders the intrinsic's literal parts through the skeleton's own parser (`intrinsicSkeletonSegments`, one parser for both readers; a literal carrying the NUL placeholder refuses), finds the one token with `singleSpanFrame`, requires a wholly literal frame (a `Ref` outside the token — the shape the issue defers — refuses), builds the token pattern through `anchoredSkeletonPattern` (the skeleton's wildcard cap and candidate length bound kept), requires exactly one candidate over the skeleton's two stores PLUS the pass-local pair table, skips a candidate recorded under another plaintext, requires the pair to equal the middle, and then takes the literal arm's bound and mark through `writeFramedTokenWithinScanBound`, extracted from `positionByEmbeddedSpan` so the two arms cannot drift. A whole-token intrinsic the skeleton REFUSED reaches it too and is written back as itself on pair evidence (an unpinned `ssm` token that lost the map slot); where the skeleton refused for two matches, so does this arm.

Why the pair table is a candidate source: the map's values hold one survivor per plaintext and the process-wide set holds only PINNED expressions, so the leaf's own collapsed token can be absent from both — and absent from the uniqueness count. The first draft had that hole and, in the two-candidate unit case, wrote the sibling's expression (the go-to-k/cdkd#1910 class); the pair table is exactly "what this pass resolved".

**Site 3 — `cdkd import`'s own resolution.** `resolveImportedProperties` (`src/cli/commands/import.ts`) now marks the bag its own resolver produced with `markSameGenerationBag` before redaction. What the mark claims is pair provenance — this bag was produced from this source, in this pass, by the resolver that recorded these pairs — which import's resolve satisfies exactly as the create arm does; the record holds the redacted copy, unmarked, and the observed capture never sees the marked object.

## Verification

- Unit: `tests/unit/deployment/secret-redaction-intrinsic-frame.test.ts` (38 cases) and `tests/unit/cli/import-sub-floor-secret.test.ts` (3 cases, the real resolver; the two cases that carry a reference also assert that the resolve ran — no refusal, no resolver warning, the secret fetched — since the literal shape's expected leaf equals the template's; the third case pins that a bag with no reference is kept as is). Mutation probes, run in this session against the branch (one per new clause, source restored byte-exact after each; the probe outputs are session evidence, not part of the diff), all killed: literal-placeholder refusal, prefix arm, suffix arm, cap guard, pair-table candidates (6 cases), process-wide candidates, poisoning skip, two-match refusal, pair equality, literal escaping, mark requirement (5), conflicting pairs counted, frame arm unwired (12), import mark (2).
- Review: Codex delta loop to NO ISSUES (9 rounds) plus three maintainer-proxy passes with the clause-to-test mapping; the passes found and this PR fixed the pair-table candidate gap, two unpinned guards, two tests crediting clauses they did not pin, an unsound render in the fixture's env-key ordering premise (every resource `Ref` rendered empty — now only an `AWS::SecretsManager::Secret` `Ref`, whose ARN cannot spell `:SecretString:`; the pre-existing outside-env occurrence check and its lenient render are unchanged), and failure diagnostics that could print a value.
- Integ, on the worktree at e851c09a (the fix commit; the commits after it touch the ledger, docstrings and unit tests only): `secrets-dynamic-ref` PASS (550 s) — `DB_PORT_SUB` / `DB_PORT_JOIN` / `PortJoin` hold their own expressions in `properties`, `observedProperties` and `outputs`, whole-document grep for the framed plaintext clean, 0 surviving state versions; `import-secret-observed` PASS (202 s, 20 assertions) — `SubFloorParam` persisted as the ARN-form expression, observed baseline equal, 0 surviving versions. Ledger rows in `docs/_generated/integ-last-run.tsv`. Neither fixture is in the broad set and `deploy-engine.ts` / the resolver are untouched, so no broad run is required.
- Review round 1 on this PR (spec / code / test / security reviewers): four docstrings the extraction left stale reworded, the residual above stated and pinned, and a refusal case added for a source the parser cannot segment on a marked bag (38 + 3 unit cases now).
- Review round 2 (code / test / security reviewers on the fix delta; test: clean, the others prose nits): the residual bounded to a token carrying a `Ref` and compared against the value scan precisely, the frame helper's third sharer counted, the changelog naming the sibling as same-service. Prose only, no behavior change.
- Maintainer round 1 (reviewed at da67ad78): M0 (the replay consumer named on the docstring, here and on the issue) and M1 (the rules-file parenthesis) taken; the optional items too — m2 (`anchoredSkeletonPattern` takes the segment form and escapes inside, behavior-identical: the regex-metachar and `${!Literal}` cases red when the escape is removed), m3 (a refusal case for a competitor visible only through the map's survivor slot — reds when the map's values are dropped from the candidates; `isSameGenerationBag` exported and the import cases assert the persisted properties are unmarked — both red when the marked bag is assigned; `DB_PORT_SUB` / `DB_PORT_JOIN` re-asserted in the fixture's post-rollback and re-capture phases — `secrets-dynamic-ref` re-run on the final tree, PASS in 553 s, 4 deleted / 0 errors, 0 surviving versions, ledger row updated), m4 (`plaintextIndexOf`'s third sharer counted, `UNKNOWN_PART_PLACEHOLDER` moved beside `UNKNOWN_PART`). Out-of-scope non-string `Fn::Join` part: filed as go-to-k/cdkd#3055.
- Maintainer round 2 (reviewed at 585b6fbc): M0 addendum — the residual names all THREE live consumers, the cross-stack read (`reresolveCrossStackValue`) added, on the docstring, the changelog, the issue and here; m5 (the "this consumer" slips and the stale case count), m6 (the import mark fence's comment names the mutation that isolates it) and m7 (`--accept` is no way out) taken. Prose only.
- Measured live in this session's `import-secret-observed` run log and stated rather than closed: the fixture's NOTE line fired — the `--verbose` import log carries `port:q7`, the go-to-k/cdkd#2453 masking-floor residual; the fixture masks it on the terminal and does not assert on it, so the ledger row cannot show it.

## Residuals

- A nonliteral intrinsic frame (a region `Ref` outside the token) still falls to the value scan — deferred by the issue.
- `maskSecretsInText` keeps the four-character floor (go-to-k/cdkd#2453).
- A leaf whose own `ssm` reference resolved PUBLIC, with a `Ref` inside its token, beside a same-service SECRET sibling whose value coincides with the middle takes the sibling's expression — a wrong reference in the STORED artifact, never a plaintext there (the whole-value scan's own class for a whole-leaf coincidence; at 4+ characters the arm can change which wrong reference the scan would have taken, below the floor on a marked bag it adds one where the scan wrote nothing). At the LIVE resource it does not stay a reference: `resolveReplayProps` (`src/deployment/rollback-executor.ts`) re-resolves every `{{resolve:` token in a replayed bag regardless of which reference the leaf named, so a later failed deploy plus `cdkd rollback` resolves the sibling's expression and writes its CURRENT plaintext into the live property — the transformed-value-meets-inverse-transform shape of GHSA-p5qg-v9gv-hc7w's consumer, reached through a wrong reference. Three live consumers re-resolve a stored expression this way, each with a weaker precondition: `cdkd drift --revert` needs no failed deploy (and `--accept` refuses to persist the masked AWS-side value, so it is no way out); a consumer stack's cross-stack read (`reresolveCrossStackValue`: `Fn::ImportValue` / `Fn::GetStackOutput` / a parent's `Fn::GetAtt Nested.Outputs.X`) needs nothing at all, since the outputs bag is marked and an ordinary deploy of the consumer after the sibling rotates hands the sibling's plaintext to its provider. Bounds: the class pre-exists at 4+ characters through the value scan, the coincidence means the value already sat in the public parameter, and candidates come only from secret-treated resolutions. Named on the arm's docstring and on go-to-k/cdkd#2745 so the PR closing site 2 weighs these consumers. Found by the security reviewer; stated on the frame arm's docstring and pinned by a unit case rather than closed, since nothing records a public resolution.


go-to-k:
Reviewed at head `da67ad78`. Tier `3-axis` + the security add-on: `loc = 1688` / 11 files is already 3-axis by size, and two up-bias triggers fired on top of it — the PR is a secret-disclosure fix, and both touched `src/` files carry a recent-defect history (`secret-redaction.ts` 3 of its last 3 commits are `fix:`, `import.ts` 2 of 3). Spec, code, test and security reviewers all ran. **No blocker.** CI green.

What the reviewers confirmed rather than assumed, worth recording because each was a way this could have been wrong:

- The extraction is behaviour-identical for `positionByEmbeddedSpan` — same guards, same order, `scanned` still computed exactly once per path from the same `(bag, secrets)`.
- The frame arm runs LAST, after the cross-stack and skeleton arms, so it only ever sees their refusals. No 4+ character regression.
- The candidate widening does not reopen the `isKnownSecretExpression` question `positionByIntrinsicSkeleton`'s docstring settles: `recordResolvedPair` is gated on `isSecret && resolved` at both resolver seams (`intrinsic-function-resolver.ts:8587`, `:8731`), so a definitively PUBLIC parameter never enters the pair table.
- Fixture realism was checked by live synth on aws-cdk-lib 2.268.0, not by reading: `'port:' + secret.secretValueFromJson('pin')` does emit the fused-prefix `Fn::Join` the fixtures encode, byte for byte.
- 19 mutation probes against the tip, in a throwaway copy; the new clauses are pinned, including the pair-table union member (6 failures), `bagIsSameGeneration` (5) and the import mark (2).

## Required before merge

**M0 — residual 3's "a wrong REFERENCE, not a disclosure" understates it, and the sentence is load-bearing for the follow-up PR.**

`resolveReplayProps` (`src/deployment/rollback-executor.ts:1177`) re-resolves every dynamic reference in a replayed bag and hands the plaintext to `provider.update` / `create` / `delete`. The walk is gated on `v.includes('{{resolve:')` — not on the service — so it does not discriminate the reference the frame arm wrote from the one the leaf actually named. The consequence: where check 2's residual puts a same-service SECRET sibling's expression on a leaf whose own reference resolved public, a later failed deploy plus `cdkd rollback` resolves that expression and writes the sibling's CURRENT plaintext into the live property. Below the floor the value scan wrote nothing, so this is the arm adding a reference where there was none.

The code reviewer reproduced the write at tip: source `{"Fn::Join":["",["port:{{resolve:ssm:/app/",{"Ref":"Env"},"/port}}"]]}`, bag `port:q7`, marked, with sibling `{{resolve:ssm:/app/secure/port}}` recorded → `"port:{{resolve:ssm:/app/secure/port}}"`.

The claim is accurate for the STORED artifact and not for the live resource, and the difference is exactly the transformed-value-meets-inverse-transform shape GHSA-p5qg-v9gv-hc7w's consumer had. I am not asking for a behavior change in this PR — the bounds you already state (the class pre-exists at 4+ characters through the value scan, the coincidence precondition means the value already sat in the public parameter, candidates come only from secret-treated resolutions) are what keep it out of blocker territory. What I am asking for is that the frame arm's docstring, the PR body's Residuals bullet and issue #2745 name `resolveReplayProps` as the consumer that converts the wrong reference into a live write, so the PR that closes site 2 weighs it instead of re-deriving it.

**M1 — `.claude/rules/layout-deployment-secrets.md` has one unbalanced `)`.**

```
    change on every deploy; a failed op's journaled `attemptedProperties`);
    and `cdkd import`'s own resolution bag (#2745, per the docstring)),
```

The `(the update arm's no-change re-check ...` parenthetical now closes at `` `attemptedProperties`) ``, the new clause opens and closes its own `(#2745, ...)`, and the trailing `)` has no opener (verified by a paren-depth scan over the paragraph: one unmatched close, final depth 0). The six-site count itself is right; only the punctuation broke, and the clause reads as sitting outside the parenthetical it belongs to.

## Optional (take-or-leave, no re-review needed)

- **m2** — `anchoredSkeletonPattern` (`secret-redaction.ts:2454`) takes `readonly string[]` and silently requires pre-escaped input. Both callers escape (`:2443`, `:3238`), so nothing is wrong today; a third caller passing raw literal text would build a pattern with live metachars and get no type error, in the function whose job is bounding a regex over template-controlled text. Taking `Array<string | typeof UNKNOWN_PART>` and escaping inside would make that structural.
- **m3** — three unfenced-but-correct spots the mutation run surfaced: dropping the `...secretExpressions` union member at `:3255` left 37/38 green (no case pins that a competitor visible only through the map's values breaks uniqueness); the "the record holds the redacted copy, unmarked" claim on `import.ts:1795` has no test, so a refactor assigning the marked `resolved` straight to `properties` stays green; and `secrets-dynamic-ref/verify.sh`'s plain-deploy readback block re-asserts `DB_PORT_LITERAL` but not the new `DB_PORT_SUB` / `DB_PORT_JOIN`.
- **m4** — doc placement: `plaintextIndexOf`'s docstring still names two sharers and the frame arm is the third; `UNKNOWN_PART_PLACEHOLDER` (`:3121`) sits ~650 lines from `UNKNOWN_PART` (`:2465`) and the parser that renders it. Also worth knowing before the next edit: the changelog entry is 1977 of the 2000-byte cap.

## Out of scope for this PR

- A non-string scalar `Fn::Join` part becomes an unknowable segment, so `["port", 8080, ":{{resolve:...}}"]` refuses where `String(part)` is knowable. Pre-existing — `origin/main:src/deployment/secret-redaction.ts:2431` has the same `typeof part === 'string'` test — so it is not this PR's to fix; file it separately if it is worth closing.
- `tests/unit/deployment/intrinsic-resolver-concurrent-drain.test.ts` failed once under a full-directory parallel run and passed alone. Pre-existing timing sensitivity, unrelated to this diff.

Push M0 and M1 and I will re-run the round on the new head.

nix-tkobayashi:
Thanks — pushed as `5745dc11` (reviewed at `da67ad78`). Every item taken, the optional ones included.

**M0** — the frame arm's docstring (check 2's residual paragraph), the PR body's Residuals bullet and go-to-k/cdkd#2745 (a new comment) now name `resolveReplayProps` as the consumer that converts the wrong reference into a live write on `cdkd rollback`, with the GHSA shape and the three bounds stated as you put them, so the site-2 PR weighs it rather than re-deriving it.

**M1** — the unmatched `)` after `(#2745, per the docstring)` is gone; the paragraph's depth scan is balanced.

**m2** — `anchoredSkeletonPattern` takes `ReadonlyArray<string | typeof UNKNOWN_PART>` and escapes inside; both callers pass raw segments. Behavior-identical (the frame arm's split-on-placeholder segments produce the same pattern); removing the escape inside reds the regex-metachar case and the `${!Literal}` case (2 failed / 70 passed over the two unit files).

**m3** — three fences added: a refusal case for a competitor visible only through the map's survivor slot (no pair, not pinned) beside a valid pass-local candidate — dropping `...secretExpressions` from the candidates now reds it (2 failed / 36 passed, the other being the length-bound case); `isSameGenerationBag` exported and both reference-carrying import cases assert the persisted `properties` are unmarked — assigning the marked bag reds both (2 failed / 1 passed); `secrets-dynamic-ref/verify.sh` re-asserts `DB_PORT_SUB` / `DB_PORT_JOIN` in the post-rollback (Phase 1e) and re-capture (Phase 1g) blocks, re-run on the final tree: PASS in 553 s, all four new lines OK, 4 deleted / 0 errors on destroy, 0 surviving state versions (ledger row updated).

**m4** — `plaintextIndexOf`'s docstring counts the frame arm as its third sharer; `UNKNOWN_PART_PLACEHOLDER` now sits directly under `UNKNOWN_PART`, beside the parser that renders it. The changelog entry now names the live-write consequence too and sits at 1995 of the 2000-character cap (trimmed elsewhere to fit).

**Out of scope** — the non-string `Fn::Join` part is filed as go-to-k/cdkd#3055 (the resolver's `String(resolved)` coercion is the rule to copy). The `intrinsic-resolver-concurrent-drain` timing sensitivity is noted; I did not touch it.

Verification this round: full suite green on the final tree (994 files, 21,925 tests); Codex delta loop to NO ISSUES in 2 rounds (round 1 caught a wrong fixture comment: a scan-only replay leaves the two-character value in plaintext, it does not collapse it) plus two maintainer-proxy passes (the first added the live-write consequence to the residual test comment and the changelog, a case with exactly three unknowable parts inside the token — which pins the first-segment guard, since a leading wildcard pushes it over the cap — and a sub-floor premise that iterates lengths 1-3, which reds when `MIN_NEEDLE_LENGTH` is lowered to 3; the second was NO ISSUES); CI on the new head.

nix-tkobayashi:
Follow-up push, `585b6fbc` (prose only): the round-2 delta reviewers on `5745dc11` found the residual named one of two live consumers — `cdkd drift --revert` (`resolveStateSecretExpressions`) re-resolves the same stored record with no failed deploy required — so the docstring, the residual's unit-case comment, the changelog entry, the PR body and the go-to-k/cdkd#2745 record now name both `resolveReplayProps` and `drift --revert`; two docstring sentences that still said the readers spell a segment now name `anchoredSkeletonPattern` as the one speller; `isSameGenerationBag` carries the module's test-only-export note. No behavior change; full suite green (994 files, 21,925 tests).

go-to-k:
Round 2, reviewed at head `585b6fbc` (delta `da67ad78..585b6fbc`). Same tier — `3-axis` plus the security add-on; a fix round adds a `fix:` commit and the stats moved (1809 LOC), so the tier was recomputed rather than carried. CI green. Every round-1 item landed and was verified rather than taken on the reply:

- M1 paren fix: depth scan over the paragraph is 0 unmatched / final depth 0, and the six-site enumeration matches six `markSameGenerationBag(` call sites at tip.
- m2 refactor: 22 shapes (frame arm 14, skeleton arm 8) compared old-vs-new `anchoredSkeletonPattern` output in a throwaway copy — 0 mismatches, including a placeholder at the token's start / end (empty literal dropped by `joinSkeletonSegments` exactly as before) and adjacent placeholders (the collapse still fires because the empty literal is dropped first).
- m3 fences: all six mutation claims reproduced at the counts you gave (`...secretExpressions` drop 2/36; escape removal 2/70 across the frame and source files; three-part case 1/37; `MIN_NEEDLE_LENGTH = 3` 1/37). One wording note under m6 below.
- M0: the two named consumers are mechanically accurate — `resolveReplayProps`'s walk is gated on `v.includes('{{resolve:')` with no service discrimination, and `drift --revert` re-resolves `observedProperties ?? properties` into `desiredProperties` (`drift.ts:4999-5020`) which reaches `provider.update` (`:5289`) with no failed deploy required.

## Required before merge

**M0 (round-2 addendum) — the residual names two of THREE live consumers, and the missing one has the weakest precondition.**

`reresolveCrossStackValue` (`src/deployment/intrinsic-function-resolver.ts:6463`, callers at `:4085`, `:6881`, `:7010`, `:7674`) re-resolves a PRODUCER's `state.outputs` expression when a consumer stack reads it through `Fn::ImportValue` / `Fn::GetStackOutput` / a parent's `Fn::GetAtt Nested.Outputs.X`, and hands the plaintext to the consumer's `provider.create` / `update`. Its own docstring (`:6451-6453`) calls the walk "A THIRD COPY" of `resolveReplayProps` and `resolveStateSecretExpressions`. The outputs bag IS marked (`deploy-engine.ts:8374`), so the frame arm's sub-floor write reaches `state.outputs` — your own `PortJoin` integ assertion shows it there. So where the residual puts a sibling's expression on a public output, an ORDINARY deploy of the consumer stack after the sibling rotates writes the sibling's current plaintext into the consumer's live property — no failed deploy, no drift, no `--revert`.

Same request as round 1, one more entry: the frame arm's docstring, the changelog entry and the #2745 comment should say three consumers and name this one. Prose only; no behavior change, no integ.

## Optional (take-or-leave)

- **m5** — two prose slips from the delta: the docstring (`secret-redaction.ts:3209-3210`) and the PR body Residuals bullet end "weighs this consumer" after naming two (the issue addendum already says "both"); and the PR body's Verification bullet 1 still says the frame file has "36 cases" where the tip has 38 (+3 import). Fold into the M0 edit if convenient.
- **m6** — the mark fence on `import-sub-floor-secret.test.ts:152` is real, but the mutation your reply credits it with (assign the marked bag straight to `properties`, dropping the redaction) is caught first by the pre-existing `toEqual` at `:159` / `:166`, not by the new assertion. The mutation that isolates the new line is `markSameGenerationBag(redactSecretsForState(...))` — redaction kept, output marked — which reds `:152` alone (2 failed / 1 passed). Worth one sentence in the test comment so the next reader does not credit the wrong line.
- **m7** — on the drift consumer, `--accept` does not offer a way out: the AWS-side value is masked because the path is secret-classified, so `runAccept` refuses, and the user is left with `--revert` (the harmful one) or a redeploy. A clause in the residual saying so would steer a reader away from the wrong button.

Push the M0 addendum and I will close the round on the new head.

nix-tkobayashi:
Pushed as `9463a63b` (reviewed at `585b6fbc`). Prose only, as asked.

**M0 addendum** — the frame arm's docstring, the changelog entry, the PR body's Residuals bullet and a second go-to-k/cdkd#2745 comment now name three live consumers, ordered by precondition: `resolveReplayProps` (a failed deploy plus `cdkd rollback`), `drift --revert` (a rotation), and `reresolveCrossStackValue` (`Fn::ImportValue` / `Fn::GetStackOutput` / a parent's `Fn::GetAtt Nested.Outputs.X` — an ordinary deploy of the consumer, since the outputs bag is marked and the sub-floor write reaches `state.outputs`, as the `PortJoin` integ assertion shows).

**m5** — "weighs this consumer" is "weighs these consumers" on the docstring and in the PR body; the Verification bullet now says 38 frame cases (+3 import).

**m6** — the import fence's comment now names the mutation that isolates it (`markSameGenerationBag(redactSecretsForState(...))`, redaction kept, output marked — reds that line alone, 2 failed / 1 passed) and says that dropping the redaction is caught first by each case's `toEqual`. Re-measured on this tree: 2 failed / 1 passed, both on `expected true to be false` at that line.

**m7** — the docstring's drift clause says `--accept` is no way out: the AWS-side value of a secret-classified path is masked, `runAccept` refuses to persist the mask, and the user is left with `--revert` or a redeploy. The PR body and the issue addendum carry the same sentence.

Verification: full suite green on the final tree (994 files, 21,925 tests; two earlier runs of it lost 5-second-timeout cases in load-sensitive script tests (cause not established; a 15-minute load average of 11.6 was recorded right after the second) and those files passed alone); Codex delta loop to NO ISSUES in one round; CI on the new head.

go-to-k:
Merged at `9463a63b`. Thank you — three rounds, and every request came back verified rather than asserted: the mutation counts reproduced exactly, and the live-consumer restatement is now the record the site-2 PR can weigh. Gating runs on the maintainer side, this tree: `secrets-dynamic-ref` PASS (4 deleted / 0 errors / 0 surviving state versions) and `import-secret-observed` PASS (20 assertions / 0 surviving); ledger rows follow in a separate chore PR.

Issue #3056, opened 2026-09-12
SecretsManager: a malformed GenerateSecretString MEMBER still throws on the update path (ExcludeCharacters), and the other members silently default


go-to-k:
Residual of go-to-k/cdkd#3048, found by its round-2 code review (measured with a mutation probe, not predicted).

## What

#3048 guards the `GenerateSecretString` CONTAINER on the update path: a malformed container is SKIPPED rather than thrown. One level down, the container's MEMBERS are still read by `generateSecretString` (`src/provisioning/providers/secretsmanager-secret-provider.ts`) exactly as before:

- `ExcludeCharacters` goes through `readConfigString(config, 'ExcludeCharacters', '', path)` with no `onUnusable`, so `GenerateSecretString: { ...valid, ExcludeCharacters: null | 123 | true | {Ref: ...} }` THROWS `... ExcludeCharacters must be a non-empty string` from `update()`. That is the #1544 un-rollbackable hazard #3048 closed for the container, reached through a member: the rollback executor's revert arms hand `update()` a cdkd STATE record the user cannot edit from the template.
- `PasswordLength` (`(x as number) || 32`), `ExcludeUppercase` / `ExcludeLowercase` / `ExcludeNumbers` / `ExcludePunctuation` (truthiness casts), `SecretStringTemplate` / `GenerateStringKey` (below) are cast, never validated -- a malformed member silently DEFAULTS, which is the #1471 / #1493 substitution class (`ExcludePunctuation: 'false'` is truthy, so the CFn string `'false'` EXCLUDES punctuation).

## Why not folded into #3048

Each member needs its own answer (skip the whole value / coerce the CFn boolean via `coerceCfnBoolean` / refuse), and the answer for `ExcludeCharacters` on the update path is the same SKIP #3048 chose for the container -- but the `SecretStringTemplate` / `GenerateStringKey` pair decides whether the value is a JSON document at all, so a wrong default there mints a RAW password as `AWSCURRENT`. That is a review cycle of its own, on the same #2212 fence.

## Verification

`vp test run tests/unit/provisioning/secretsmanager-secret-provider-update-value-source.test.ts` -- add one case per member on the update path (STATE-record shape) and one on the create path; the #2212 fence (`tests/unit/scripts/integ-secret-fixture-sweep.test.ts`) must stay green.

Session-fit: now (this session) -- re-classified: the session that shipped go-to-k/cdkd#3058 still holds the provider, the guards and the #2212 fence in context
[...]

Issue #3032, opened 2026-09-11
requireConfigObject accepts an unresolved intrinsic as a container, so the block silently empties


go-to-k:
`requireConfigObject` (`src/provisioning/config-shape.ts`) refuses a container that is not a plain object. An UNRESOLVED CloudFormation intrinsic — `{ "Ref": "X" }`, `{ "Fn::If": [...] }` — **is** a plain object, so it passes the guard, every member probe indexes to `undefined`, and the caller proceeds with an empty block. That is exactly the silent-DROP shape the guard family exists to refuse, reached through a value the guard cannot see.

Measured on `AWS::S3::Bucket`'s `ObjectLockConfiguration.Rule.DefaultRetention.DefaultEventHold` while landing go-to-k/cdkd#3002: a template declaring `DefaultEventHold: { Ref: 'X' }` sends `DefaultEventHold: {}` to `PutObjectLockConfiguration` — a declared WORM setting silently not applied, from a Put that succeeds. Pinned there by `tests/unit/provisioning/s3-bucket-provider-roundtrip.test.ts` ("an UNRESOLVED intrinsic still sends an empty hold (known residual)") so the limitation is a recorded fact rather than something a reader assumes is closed.

It is NOT specific to that member. Every `requireConfigObject` call site shares it — the S3 lifecycle `Filter`, the analytics `StorageClassAnalysis` / `.DataExport`, and the replication `Filter`, whose empty-block fall-through is the widest of them (`Filter: {}` is valid CFn for "replicate EVERY object").

Fixing it at one site alone would diverge from all the others, which is why go-to-k/cdkd#3002 recorded the residual instead of special-casing its own member.

## What would settle it

The guard would need to recognise an intrinsic-shaped object — a single key matching `Ref` or `Fn::*` — and treat it as UNUSABLE rather than as a container. Two things to establish first, neither assumed here:

1. **Whether an unresolved intrinsic can reach a provider at all in practice**, or only through a resolver defect. If the resolver always resolves or fails first, this is a defence-in-depth guard rather than a live bug, which changes the severity but not the correctness argument.
2. **Whether the refusal belongs in `requireConfigObject` or in a sibling predicate** — the string guards (`readConfigString` / `configStringRefusal`) already refuse a non-string, so an intrinsic OBJECT is refused there today. Only the container guard has the hole, which suggests a narrow addition rather than a new family.

[...]

Issue #2212, opened 2026-08-25
test(scripts): the secret-fixture sweep lint has no generateSecretString pattern, so four fixtures seed generated credentials into versioned state unswept


go-to-k:
## Summary

`tests/unit/scripts/integ-secret-fixture-sweep.test.ts` lints that a fixture whose CDK app puts secret material into state must sweep S3 object versions on teardown. Its pattern list has no entry for **`generateSecretString`** -- a Secrets Manager secret whose value AWS generates. cdkd resolves and persists that value into `state.json` exactly like a hand-written one, and the state bucket is VERSIONED, so `aws s3 rm` writes a delete marker and every version the fixture ever wrote stays readable through `GetObjectVersion`.

The lint's own doc block already names "service-generated credentials with no marker in the source" as blind spot 3, listing `generateSecret: true`, `iam.AccessKey`, `appsync.CfnApiKey` and `addApiKey`. `generateSecretString` belongs on that list and is not there -- and unlike those four it IS a syntactic marker in the source, so it is mechanically detectable rather than a genuine blind spot.

## Fixtures currently in the gap

Found while adding the arm for go-to-k/cdkd#2189, which made `local-run-task-from-state` a seeding fixture through a different pattern (`unsafePlainText`) and closed its own exposure incidentally. These four remain:

| fixture | shape | sweeps? |
| --- | --- | --- |
| `composite-stack` | `generateSecretString` | no |
| `event-driven` | `generateSecretString` | no |
| `full-stack-demo` | `generateSecretString` | no `verify.sh` at all |
| `secrets-rotation-schedule` | `generateSecretString` | has a `verify.sh`, does not source the helper |

`full-stack-demo` is the awkward one: with no `verify.sh` there is nothing to source the helper INTO, so it needs either a `verify.sh` or an explicit carve-out with a recorded reason.

## Before fixing, MEASURE

Do not take the table above as the finding. The go-to-k/cdkd#2096 audit that produced this lint was itself wrong twice, both times from a bad sampling shape, and the numbers are in the lint's doc block: of `AppSyncStack`'s 557 state versions the 12 NEWEST carried no key while 17 of versions 12..45 did. So:

- Sample ACROSS the version range, not the newest N. The newest versions come from the most recent run, which is the most likely to be already-fixed or to have failed early.
- Derive the needle PER FIXTURE (each spells its own literal) or, better, assert on a needle-independent observable -- surviving version count == 0.
[...]

Issue #1654, opened 2026-08-12
fix(lambda): the Lambda URL AuthType warn-and-substitute records the malformed desired value (#1612 item 3)


go-to-k:
Found by the issue #1612 audit (item 3: "check the same question for the non-S3 warn-and-skip sites").

## The gap

`LambdaUrlProvider.update`'s `AuthType` guard warns and then SUBSTITUTES rather
than skipping (issue #1551): on an unusable desired value it sends the PREVIOUS
`AuthType`, or OMITS the field entirely when the previous side is unusable too
and lets `UpdateFunctionUrlConfig`'s merge semantics retain the live auth type.
Both arms are correct — defaulting to `'NONE'` would flip a live IAM-guarded
function URL to PUBLIC.

But the engine still records the DESIRED bag, so state holds the malformed
`AuthType` while AWS holds the previous / live one. `readCurrentState` reads
`AuthType` back (`resp.AuthType`), so the mismatch is a permanent phantom drift:
re-reported by every `cdkd drift` and re-triggered by `drift --revert`, which
calls `update()` again.

`.claude/rules/providers.md` already names this exact shape — "A SUBSTITUTED
value is the same class as a dropped key" (#1633) — and `update()` returns no
`effectiveProperties` today (`grep -c effectiveProperties
src/provisioning/providers/lambda-url-provider.ts` is 0).

## The fix

Return `effectiveProperties` carrying the `AuthType` actually SENT:

- previous-value arm -> the previous `AuthType`;
- OMITTED arm -> this one needs a decision, because nothing was sent and the
  live value is not read at that point. Either drop the key (state stops
  claiming an auth type cdkd cannot vouch for) or read the live value back
  before answering. Settle it in the PR rather than guessing here.

The `.claude/rules/providers.md` carve-out the #1612 PR adds applies: this is
the SUBSTITUTION shape, so unlike a pure narrowing it must NOT gain a
`canonicalizeDesiredProperties` twin — canonicalizing the desired side would
compare a previous side holding a valid `AuthType` against a desired side with
the key removed and derive a REMOVAL.

## Scope

- `src/provisioning/providers/lambda-url-provider.ts` update path.
- Unit pin plus a live test (deploy a URL with `AuthType: AWS_IAM`, hand-patch
  the state record's `AuthType` to a malformed value, redeploy, and assert two
  consecutive `cdkd drift` runs converge AND the URL is still IAM-guarded).

Split out of #1612 rather than bundled because it is a different provider file
[...]

PR #3058, opened 2026-09-12
fix(secretsmanager): skip a malformed GenerateSecretString on update and retain the previous block in state

### LINKED ISSUE #3048: A malformed GenerateSecretString is un-rollbackable: skip the value instead of throwing
`SecretsManagerSecretProvider.changedSecretValue` refuses a malformed `GenerateSecretString` container on the UPDATE path (it reads every member through `generateSecretString`, and go-to-k/cdkd#3032 widened that refusal to the unresolved-intrinsic shape). The refusal itself is right and should stay: the fallback would index every member to `undefined`, minting a bare default-charset password AND — because `generateStringKey` / `secretStringTemplate` both go `undefined` — returning the raw password instead of the JSON document the template declared. That value becomes the new `AWSCURRENT`, breaking every consumer reading `{"username":...,"password":...}`, recoverable only via `AWSPREVIOUS`.

What is NOT settled is the go-to-k/cdkd#1544 side of it. `update()` is reached by the rollback executor's revert arms and by `cdkd drift --revert` with a cdkd STATE record as the desired bag, which the user cannot edit from the template — so a record carrying a malformed `GenerateSecretString` is un-rollbackable at this site.

## The remedy is NOT `onUnusable`

That is the point worth recording, because it is the obvious move and it is wrong here. An `onUnusable` downgrade warns and proceeds, and proceeding is what mints the bad secret. The two real answers are:

1. **Return `undefined` from `changedSecretValue`** — skip the value entirely. `UpdateSecret` merge semantics leave the live secret untouched, so the rollback completes and the secret keeps the value AWS already holds. This is the `retainPrevious` shape the S3 appliers use, expressed through this provider's own "no value to send" return.
2. Leave the throw and accept that such a record needs a hand-edit.

Option 1 is the better answer and is what this issue is for. It is a new arm touching the helper `create()` shares, so it wants its own change and its own review rather than riding the guard PR that surfaced it.

## What to verify

- The skip must not be confused with an UNCHANGED config: `changedSecretValue` already returns `undefined` for "nothing to send", so the new arm needs a test distinguishing "unchanged, send nothing" from "malformed, send nothing" — otherwise a regression that always skips looks correct.
- `create()` has no live secret to fall back to, so it must keep refusing; confirm the shared helper does not carry the skip into it.
- The existing case `refuses an UNRESOLVED INTRINSIC GenerateSecretString rather than generating from it` (`tests/unit/provisioning/secretsmanager-secret-provider-update-value-source.test.ts`) pins today's behaviour and will need updating as part of the change, not before it.

## Classification

Session-fit: next (not this session) — surfaced while landing go-to-k/cdkd#3032, whose scope is the shape guard itself; this is a provider-behaviour change on a secrets path with its own blast radius.
Severity: low — reachable only on a rollback or `drift --revert` over a record whose `GenerateSecretString` is malformed, and the current behaviour is a loud refusal rather than a wrong value.
Effort: medium (M) — one arm plus its distinguishing tests, and a re-read of the create/update split in the shared helper.
Estimate: ~1-2 h — most of it establishing that the skip cannot leak into `create()`.
Dup-check: searched open issues for "GenerateSecretString" and "changedSecretValue" — no match.

go-to-k:
Closes #3048

## What

After go-to-k/cdkd#3032 widened `readConfigString` to refuse an unresolved-intrinsic container, a malformed `GenerateSecretString` block THREW from `SecretsManagerSecretProvider.update()`. That path is reached by the rollback executor's revert arms with a cdkd STATE record as the desired bag, which the user cannot fix from the template -- the go-to-k/cdkd#1544 un-rollbackable hazard.

- **Update path SKIPS the value** instead of throwing (`UpdateSecret` goes out with no `SecretString`; the service's merge semantics keep the live value) -- and NOT via an `onUnusable` downgrade, because proceeding is the harm: a malformed container indexes every member to `undefined`, so `generateSecretString` mints a bare default-charset password and returns it RAW as the new `AWSCURRENT`.
- **The recorded bag retains the previously-applied block** via `effectiveProperties` (both reviewers' round-1 blocker): recording the malformed desired bag would make the next deploy compare equal, take the `unchanged` early return, and leave the secret un-regenerated with no warning. The previous side runs the same `requireConfigObject` predicate as the wire; when it is absent or unusable the key is DROPPED and the drop is WARNED (the go-to-k/cdkd#1654 rule -- a record with no block later reaches the reverse-replacement replay-create).
- **`create()` keeps refusing** (no live value to fall back to; the deviation from the replay-downgrade rule is recorded at the refusal and in `.claude/rules/provider-replay-and-refusals.md`), now through the same `requireConfigObject` guard behind a `!= null` gate, so a falsy malformed block (`''` / `0`) no longer falls through to the literal read and creates a valueless secret; it also warns when the bag carries no value source at all.
- `cdkd drift --revert` does NOT reach the skip arm (`getDriftUnknownPaths` keeps the key out of the comparison) -- comments and changelog say so.
- `EffectivePropertiesResult.effectiveProperties` admits an explicit `undefined` (exactOptionalPropertyTypes) so the always-present-key spelling the go-to-k/cdkd#2212 fence reads typechecks; every consumer gates on truthiness or `=== undefined`.

## The go-to-k/cdkd#2212 fence

`tests/unit/scripts/integ-secret-fixture-sweep.test.ts` asserts the generated secret is never persisted by reading the provider's return literals. `effectiveProperties` had to be admitted, and the fence was TIGHTENED rather than relaxed: the exact `skippedGenerate ? this.retainPreviousGenerateBlock(logicalId, properties, previousProperties) : undefined` expression is pinned (else arm included), the helper body may not mention `generateSecretString` / `SecretString` / `secretString`, may not reach `this` beyond the logger, exactly two `generateSecretString` call sites exist, no instance field is assigned outside the constructor, and the module-level escape tokens (`let`/`var`, `static`, `globalThis`, `Reflect.`, `Object.assign(this`) are refused by absence. Eleven single-mutation probes across the three rounds each reddened a case, tree restored byte-exact between them; three of those mutants were GREEN under the round-1 fence.

## Verification

- Unit: 994 files / 21952 tests green; the corpus-floor ratchet in `rule-file-payload.test.ts` fired on CI (main grew under the PR) and is re-derived against the merge measurement; new cases in `secretsmanager-secret-provider-update-value-source.test.ts` (skip vs unchanged distinguished by warning, retain, copy-not-alias, drop + warn, absent-previous arm, falsy gate on both paths, create refusal, no-value warning + control).
- Review: 3-axis tier -- `pr-spec-reviewer`, `pr-test-reviewer`, `pr-code-reviewer` (4 rounds) and `pr-security-reviewer` on fable (8 rounds, closing verdict: mergeable); no blocker left, every should-fix applied, the residual bounds recorded in the fence header.
- Live: `/run-integ secretsmanager-update-value-source` PASS twice (111 s, then 76 s on the final src tree), destroy 0 errors / 0 orphans / 0 surviving state versions; it exercises the ordinary create / update paths (no provider warning fired), not the skip / retain arms, which a template cannot reach.

## Residual

- go-to-k/cdkd#3056 -- the container's MEMBERS (`ExcludeCharacters` throws on update; the others silently default) are the same class one level down.


go-to-k:
Landed on main as dc8e3dc6 (the squash-merge call returned a 502 after the commit was pushed, and GitHub left this PR object open with a stale base; the branch tree is byte-identical to main). Closing by hand to match. Thanks to the reviewers -- eight rounds on the #2212 fence, and the runtime half that came out of round 6 is the part that will keep paying.

Issue #3018, opened 2026-09-11
fix(state): consumers outside the three state views still abort on a malformed resources bag, and an overlay preview slices by code unit


nix-tkobayashi:
Residuals of go-to-k/cdkd#2947, found while fixing it. That issue's fix makes `cdkd state show`, `cdkd state resources` and `cdkd state list` render a state record whose shape violates its declared types rather than abort on it, and deliberately leaves INNER record-shape validation out of `parseStateBody`: that is the read path for deploy, destroy, rollback and the recovery commands, so beyond invalid JSON and the existing schema-version check it refuses only a root that is not a JSON object. The consequence is that each consumer tolerates a malformed shape where it dereferences one, and the consumers outside those three views still do not.

**The population is a CLASS, not the three sites first listed here** (found by the code review of go-to-k/cdkd#3067, which re-derived it rather than trusting the count). `Object.entries(state.resources)` / `Object.keys(state.resources)` with neither a `?? {}` nor a null-entry guard appears at 21 sites across 9 files: `src/cli/commands/destroy-runner.ts` (5), `src/deployment/deploy-engine.ts` (4), `src/cli/commands/import.ts` (4), `src/cli/commands/scrub.ts` (2), `src/cli/commands/diff-recursive.ts` (2), and one each in `src/cli/commands/diff.ts`, `src/cli/commands/orphan.ts`, `src/cli/commands/prefix-migration-check.ts` and `src/analyzer/orphan-rewriter.ts` (measured on `17b8e9d1`; re-run the grep rather than trusting the number). Two of those matter more than the rest and are called out below, because they are the RECOVERY commands go-to-k/cdkd#2947's placement decision leans on: `cdkd state destroy` (`destroy-runner.ts`) and `cdkd state orphan` (`orphan.ts`) both abort on a `null` or absent `resources` bag before they can clean up the very record that is broken. The read boundary deliberately lets such a record through so that each of them is one guard away; today none of them has taken it.

The three sites originally listed here are kept below because each carries a decision the others do not.

## 1. `cdkd state refresh-observed` dereferences a `null` resources entry

[...]

Issue #3003, opened 2026-09-11
fix(state): state show / state resources error paths still forge rows the rendered output no longer can


go-to-k:
## What is wrong

go-to-k/cdkd#2922 made every value `cdkd state show` and `cdkd state resources` RENDER from a
state record control-safe, because those views are joined by newlines and a
record is an unchecked cast that anyone with `s3:PutObject` on the state bucket
can write. The rule stops at the rendered rows. The ERROR paths both commands
throw through still interpolate record-derived and key-derived text raw, and
`handleError` prints `error.message` unmodified (`src/utils/error-handler.ts`),
so a refusal can forge exactly the row the rendered output no longer can.

Five sites, none of them in go-to-k/cdkd#2922's diff:

1. `src/state/s3-state-backend.ts` — `parseStateBody`'s invalid-JSON arm embeds
   `JSON.parse`'s own message, which quotes raw body bytes (V8 includes roughly
   26 characters of the offending input, ESC included). The repo already
   sanitizes this exact class about 100 lines away in `probeLegacyState`; this
   arm is the unsanitized sibling, and it sits directly on both commands'
   `getState` path.
2. `src/state/lock-manager.ts` — `getLockRecord`'s catch wraps the
   `JSON.parse` SyntaxError from a malformed `lock.json` the same way, and
   `cdkd state show` surfaces it.
3. `src/cli/commands/state.ts` — `resolveSingleRegion` interpolates region
   strings raw. They are raw S3 KEY SEGMENTS from `listStacks`, and S3 keys
   admit any UTF-8 including newline and ESC. Planting
   `cdkd/<victimStack>/<hostile>/state.json` makes the ambiguity error forge
   rows. The same values already go through `displaySafe(..., {asciiOnly: true})`
   in `formatStackRefSafe` and in the lock-contention messages, so the fix has a
   settled shape in this file.
4. `src/cli/commands/export.ts` — `walkCdkdStateStackTree`, reached by
   `state show --show-nested`: the missing-child error embeds a child stack name
   built from a record-body resources KEY, and the region-mismatch error embeds
   `childResult.state.region`, both raw.
5. `src/state/s3-state-backend.ts` — the unsupported-version refusal embeds
   `String(v)` raw. This one is ALREADY recorded on go-to-k/cdkd#2947, listed
   here only so the set is complete; it is the same class and wants the same fix.

Related but distinct, and already filed: go-to-k/cdkd#2947 covers shapes that
[...]

Issue #2772, opened 2026-09-07
feat(state): render skippedOutputs in cdkd state show's text view, the field that explains a missing diff row


nix-tkobayashi:
## What is wrong

`cdkd state show <stack>` renders `outputs`, the parent-stack fields and the resources, but not `skippedOutputs` — the field issue go-to-k/cdkd#2740 added, which is the one that explains "why does `cdkd diff` render no row for this Output" (the last deploy could not resolve it and recorded a digest of its template inputs). Today the only way to see it is `aws s3 cp s3://<bucket>/cdkd/<stack>/<region>/state.json -` or `cdkd state show --json`, which prints the raw record.

`src/cli/commands/state.ts` (the `show` text renderer, around the block that prints `Outputs:`) enumerates the fields it prints by hand; the new optional field is not in the list.

## Why it matters

The record is informational and self-healing, but its effect on `cdkd diff` is deliberately silent (no row, no warning), so the text view is where a user would look for the reason. Without it, a "missing" Output in the diff has no visible explanation short of reading raw JSON. Low impact: the deploy's own warn names the broken output at deploy time when the resolver THREW; a resolver that returns `undefined` outright is recorded with no OUTPUT-SPECIFIC warning (the no-change path still emits its generic "Outputs changed but one or more could not be resolved" line when the maps differ — `tests/unit/deployment/deploy-engine-skipped-outputs.test.ts` pins only the absence of the per-output line), and `--json` shows the field either way.

## Proposed fix

Print a `Skipped outputs:` block after `Outputs:` when the record is present: one line per key, the key and the first 12 hex characters of its digest (the full digest is noise in a text view; `--json` has it), with a one-line legend saying the last deploy could not resolve these and `cdkd diff` previews each as absent while it is still absent from `outputs` and its template inputs are unchanged (a no-change deploy can keep an earlier value beside the record, and the diff then ignores the record for that key). Absent field → no block (pre-go-to-k/cdkd#2740 records, and stacks that skipped nothing). Unit case in the `state show` suite: present renders the block with the key; absent renders nothing; the `--json` shape is unchanged.

## Classification

Dup-check: searched open and closed issues for `state show skippedOutputs`, `state show outputs render`, `2740` — no hit; go-to-k/cdkd#2740 added the field and its docs, not the text renderer.
Session-fit: next (not this session) — a `cdkd state show` renderer change is outside go-to-k/cdkd#2740's diff/deploy scope; surfaced as a review nit on that PR.
[...]

PR #3067, opened 2026-09-13
fix(state): render a state or lock record whose shape violates its types instead of aborting

### LINKED ISSUE #2947: state show / state resources abort on a record shape no display guard can reach
## What happens

cdkd reads a state or lock record as an unchecked cast (`JSON.parse(...) as StackState`), and
the boundary between "bytes from S3" and "typed data" carries no COMPREHENSIVE shape check --
`parseStateBody` validates the schema version and `getLockRecord` treats a null, primitive or
array root as no lock, and that is all -- so in practice every consumer is the validator. `cdkd state show` and `cdkd state resources` therefore ABORT --
no rendered output, a failure from the error handler, non-zero exit -- on a record that
violates its own declared types.

PR go-to-k/cdkd#2922 routed the field VALUES those human paths render through a guard that
strips control characters and tolerates a wrong type, which closed that class. (`version` is
the deliberate exemption: it is interpolated directly, because `parseStateBody` already
refuses any value but a readable schema number or `undefined`.) It does not
reach a throw that happens EARLIER: during a dereference, a coercion, or an enumeration
upstream of the guard. Those sites are what this issue is about.

Measured on the tree at the time of filing, and deliberately **not an exhaustive list** --
that nothing enumerates them is the point:

- **lock `expiresAt` that cannot be coerced to a number.** `formatLockSummary` computes
  `lockInfo.expiresAt - Date.now()`; `{"toString": null}` and `[{"toString": null}]` both
  throw a `TypeError` during the subtraction. A numeric string, `null` or a boolean coerces
  and renders.
- **lock `owner` or `operation` that cannot be coerced.** `LockManager.getLockRecord` passes
  them through `displaySafe`, whose `String(value)` is unguarded, so the throw happens
  before `state show` reaches a renderer at all -- `--json` included.
- **`resources` holding `null` for a key.** Reading `resource.resourceType` throws, in
  `state show` and in all three `state resources` modes (`--json` too: the per-resource
  detail objects are built before the JSON branch). A number, string or boolean there does
  NOT throw -- its missing fields render as `undefined` or their defaults.
- **a top-level JSON body of `null`.** `parseStateBody` throws at `parsed.version`.
- **`resources` absent or `null` under `--show-nested`.** `walkCdkdStateStackTree` in
  `src/cli/commands/export.ts` calls `Object.entries(state.resources)` with no fallback,
  where `renderStateBlock` uses `?? {}`. A nested child's uncoercible `region` also throws
  while building the region-mismatch message.
- **`parseStateBody`'s own rejection message.** For `{"version": {"toString": null}}` the
  version check is correctly reached and `String(v)` then throws, so the user gets a
  primitive-conversion `TypeError` instead of the sentence about the schema version the code
  is trying to print.

Out of scope: a throwing enumerable getter in an `attributes` / `outputs` / `resources` bag
throws during `Object.entries` the same way, but neither command can receive one -- both
read through `S3StateBackend.getState`, and JSON has no getters.

## Why it is separate from go-to-k/cdkd#2922

That PR's scope is the `skippedOutputs` block plus the rows its review reached, and its
answer -- a total value-rendering guard -- is the right answer for the class it closed and
the wrong shape for this one. Wrapping each throw site in a `try` is the third-best answer.
What needs deciding first is WHERE the validation belongs: a shape check at the read
boundary (`parseStateBody` / `getLockRecord`), so every consumer inherits it and the error
names the offending field; or per-consumer tolerance, which keeps the renderers printing
something for a record no binary wrote. The read boundary also has to stay cheap, because
`cdkd state list --tree` and `cdkd gc` read every record in the bucket.

## Verification plan

One case per site, driven through the real read path rather than a mocked `getState` --
mocking the read is exactly what makes `version` look unguarded when it is not, so a case
that mocks it proves nothing about this class. Each asserts the command either renders the
record or refuses it with a message NAMING the offending field, never an empty payload; the
`resources: {"R": null}` case and the lock `owner` / `operation` cases need a `--json` arm too,
since those modes throw as well -- an uncoercible `expiresAt` does NOT, because `--json`
serializes the lock without calling `formatLockSummary`. A mutation probe per guard or check added.

Dup-check: searched open and closed issues for `state show` / `state resources` render
crashes, for `parseStateBody` / `displaySafe` validation, and for `unchecked cast`.
go-to-k/cdkd#2772 is the feature this was found under and covers the `skippedOutputs` block
only; nothing covers record-SHAPE validation at the read boundary.
Session-fit: next (not this session) -- found during review round 4 of go-to-k/cdkd#2922,
and the fix needs the placement decision above rather than the guard swap that PR is making.
Severity: low -- every shape needs a hand-edited record; one cdkd wrote cannot reach any of
them, and the commands fail loudly rather than printing something wrong.
Effort: medium (M) -- the read-boundary option touches every consumer of a parsed record.
Estimate: ~2-4 h -- most of it the placement decision and its blast radius, not the guards.

nix-tkobayashi:
Closes #2947

## Summary

A state record and a lock are unchecked casts: anyone with `s3:PutObject` on the state bucket, or a hand edit, can put any JSON there. `cdkd state show`, `cdkd state resources` and `cdkd state list` rendered malformed VALUES since go-to-k/cdkd#2772, but a record malformed in its SHAPE still aborted them with a raw `TypeError` before anything rendered.

**Placement — decided here, as the issue asked.** Inner record-shape validation stays OUT of the read boundary. `parseStateBody` is the read path for deploy, destroy, rollback and the recovery commands (`state orphan`, `state destroy`), so a refusal there would make a broken record unreadable by every one of them at once, including the two whose job is to clean it up — an argument about what stays FIXABLE, not a claim about what already works, since those two still abort at their own dereference today (go-to-k/cdkd#3018). Normalising one would be worse still, though not for the reason first given here: the deploy path's `previousState` lookup already reads a `null` entry as it reads an absent one (a truthiness test), so dropping the entry changes nothing at that lookup. What it changes is a consumer that enumerates KEYS — the destroy count, `state orphan`'s `id in resources` — for which a `null` entry keeps the logical id visible while a dropped one erases it from the record silently. So beyond invalid JSON and the existing schema-version check, `parseStateBody` refuses only a root that is not a JSON object, by name, and each consumer tolerates an inner shape where it dereferences one. Plain `cdkd state show --json` stays the verbatim mode — the comparand the issue named. The one cost of that root refusal is stated rather than hidden: `cdkd state list --long` aborts on a stack whose body is `0` / `false` / `""`, where it used to print a 0-resource row. That is the same abort the unsupported-version refusal has always produced from the same `Promise.all`, so it is consistent with the view's existing behaviour rather than a new hazard — and making that listing survive a peer's failure is go-to-k/cdkd#3069, filed from this PR's review.

| shape | before | after |
| --- | --- | --- |
| `resources: {"R": null}` | `TypeError` in `state show`, in all three `state resources` modes, and in `state show --show-nested` (plain and `--json`) — each dereferences the entry. `state list --long` (counts keys) and plain `state show --json` (serialises the record) already tolerated it | renders in every mode: the human views print `Type` / `PhysicalID` as `undefined` and the rest as their defaults, `state show --json` keeps the stored `null`, `state resources --json` lists the entry |
| `resources` absent or `null` | the COMPLEMENT of the row above: `TypeError` in `state list --long` and in `state show --show-nested` (plain and `--json`), the two sites that had no `?? {}`; `state show` and `state resources` already had one | renders as an empty bag |
| lock `expiresAt` whose number coercion throws | `TypeError` in the lock read, and a takeover deleted the lock and THEN threw | read as `NaN`: the expiry check treats it as expired, the destroy-time warning as unknown |
| lock `owner` / `operation` whose string coercion throws | `TypeError` | renders through `Object.prototype.toString`'s tag — `[object Object]` for the object, `[object Array]` for an array holding one |
| `version` whose string coercion throws | `TypeError` while BUILDING the refusal | refused, naming the value |
| root is `null` / an array / a primitive | `null` threw a `TypeError` at the `.version` read; an array or a primitive passed it and was returned as a record to every consumer, deploy included | refused with a `StateError` naming what the body parsed to |
| `skippedOutputs` digest with a surrogate pair at the cut | preview ended on a lone high surrogate | cut by code point |

The single mechanism behind three of those rows is `displaySafe`: its `String(value)` threw on `{"toString": null}`, so every caller inherited the throw. It now falls back to `Object.prototype.toString`, which cannot throw for any value `JSON.parse` produces (a hand-built object with a throwing `Symbol.toStringTag` getter still can, and no state record can carry one). That is also the half of go-to-k/cdkd#3016's version refusal it leaves to this issue by name. This PR was built on top of go-to-k/cdkd#3016 and keeps its sanitizer class (`displaySafe(..., { asciiOnly: true }) || UNRENDERABLE`) at every site it shares, rather than carrying a second spelling. One line needed a real decision: go-to-k/cdkd#3016 coerces the version with `String(v)` BEFORE sanitizing, so a `null` version prints as the word `null` rather than the stand-in — and that coercion is exactly the throw this issue removes. The refusal now maps `null` to `'null'` and hands every other value to `displaySafe` directly, so both go-to-k/cdkd#3016's `version null` case and this PR's uncoercible-version case hold.

`truncateCodePoints` is new in `src/utils/display-safe.ts`, so the code-point rule has one owner, per the maintainer's note on the issue — the note offered `Intl.Segmenter` or a code-point-aware slice, and this is the second. The defect named there is a LONE SURROGATE, i.e. a string that is not well-formed; grapheme segmentation solves a different problem (a split ZWJ sequence or a combining mark), costs an ICU-backed allocation on a path that renders one 12-character window per skipped output, and would make "12" mean a different number of code points per record. If a site ever renders user text where a split cluster matters, that is a second helper beside this one, not a widening of it.

## Residuals

go-to-k/cdkd#3018 — the consumers outside these three views still do not tolerate a malformed shape, and the review of this PR re-derived that as a CLASS rather than the three sites the issue first named: an unguarded `Object.entries`/`keys(state.resources)` appears at 21 sites across 9 files (measured on `17b8e9d1`). Two matter most and are now called out in that issue by name — `cdkd state destroy` and `cdkd state orphan` both abort on a `null` or absent bag, which are the recovery commands the placement argument above leans on. The comment at the read boundary was reworded to say that plainly: it argues what stays FIXABLE, not what already works. The issue's `Effort` and title were widened with it.

go-to-k/cdkd#3069 — `cdkd state list --long` still loses the whole detail listing to one stack it cannot read. A per-row tolerance was written during this PR's review round and REVERTED: it grew two JSON fields, a render arm, a helper and a bound inside a PR about record-shape rendering, and the maintainer-checklist pass then showed its reason string could carry state-body bytes into stdout, which sanitizing does not redact. That is a design question about what a row may disclose, not a wording fix, so it is its own issue — carrying what the attempt measured (both reads must degrade and report separately; the reason must be class-level, not the caught message; `renderTreeMode` is the precedent) so the next attempt does not rediscover it. Plain `state list`, `state list --json` and `--tree` are unaffected.

## Test plan

- [x] Full unit suite, root-asserted to this worktree — 1000 files, 22,146 passed, 1 skipped, no type errors, on this branch rebased onto `main` after go-to-k/cdkd#3016 merged.
- [x] `tests/unit/cli/state-record-shape.test.ts` (new) drives the REAL `S3StateBackend` and `LockManager` over a mocked S3 client that answers with raw bytes — what the issue's verification plan asks for, since `state-show.test.ts` mocks `getState` / `getLockInfo` wholesale and so skips exactly the code this class lives in. `process.exit` is mocked to throw, so the oracle is two-sided: an earlier cut without that mock passed three render cases against the UNFIXED source.
- [x] Premise: reverting ONLY the `displaySafe` fallback (probe 1 below) reds 8 cases across 4 files, enumerated: the two `displaySafe` unit cases (`renders an object with a non-callable toString instead of throwing`, `renders an array holding one through its own tag`), THREE uncoercible-owner lock cases (`reads an uncoercible owner and operation without throwing` in `LockManager` plus the `state show` and `state show --json` renders), the region-uncoercible legacy-record skip, the nested-child region-mismatch refusal, and `names an uncoercible schema version in its refusal instead of throwing`. That fallback is what closes the half go-to-k/cdkd#3016 defers.
- [x] Mutation probes — 21 probes, one mutation each, the first 18 re-run on the rebased tree because go-to-k/cdkd#3016 moved `displaySafe` under this lane. **All 21 KILLED**; the source files were restored byte-exact between probes (sha1 compared before and after every run). Baseline on the final tree for the five files the table names (`state-record-shape`, `state-show`, `display-safe`, `lock-manager`, `s3-state-backend`): 238 passed. Six further probes covered a `state list` tolerance that was later reverted; they are dropped here with the code they pinned.

  | # | mutation | reds |
  | --- | --- | --- |
  | 1 | `displaySafe`: drop the throw fallback, back to a bare `String(value)` | 8 cases / 4 files (the premise above) |
  | 2 | `displaySafe`: fallback renders `''` instead of `Object.prototype.toString` | 7 — probe 1's eight minus the legacy-record region skip, which the empty string still satisfies |
  | 3 | `truncateCodePoints`: `<=` → `<`, so an exactly-window-long value is "cut" | 3 — `reports no truncation for a value exactly the window long`, `marks a digest ONLY when it was actually cut`, `counts an astral character as ONE …` |
  | 4 | `truncateCodePoints`: `Array.from` → `split('')`, i.e. count UTF-16 units | 3 — `never splits a surrogate pair at the cut`, the same case at the digest preview, `counts an astral character as ONE …` |
  | 5 | `truncateCodePoints`: report `truncated: false` when it did cut | 8 — both surrogate-pair cases plus six Skipped-outputs render cases (the digest-and-key render, both control-character cases, the non-string digest, `marks a digest ONLY when it was actually cut`, and the legend indent) |
  | 6 | `renderStateBlock`: digest preview back to `slice()` by code unit | 1 — `never splits a surrogate pair in a skippedOutputs digest preview` |
  | 7 | `stateListCommand`: drop the `?? {}` on `resources` | 2 — `state list --long` counts a `null` bag, and an absent one, as zero |
  | 8 | `stateResourcesCommand`: drop the `null`-entry tolerance | 3 — `state resources` renders a null entry in plain, `--long` and `--json` |
  | 9 | `renderStateBlock`: drop the `null`-entry tolerance | 2 — `state show renders a null resource entry instead of aborting`, and the `--show-nested` walk |
  | 10 | `parseStateBody`: drop the `root === null` arm | 2 — `refuses a body that parses to null, naming it (null)`, and its end-to-end twin |
  | 11 | `parseStateBody`: drop the `Array.isArray(root)` arm | 2 — `refuses a body that parses to [], naming it (an array)`, and the stack-name sanitisation case |
  | 12 | `parseStateBody`: drop the `typeof root !== 'object'` arm | 4 — `true`, `false`, `5` and `"text"` each refused by name |
  | 13 | `parseStateBody`: refusal stops naming what the body parsed to | 8 — the six root cases (`null`, `true`, `false`, `[]`, `5`, `"text"`), the stack-name sanitisation case, and the one end-to-end null-root case |
  | 14 | version refusal: drop the `null` → `'null'` mapping | 1 — `reports a null VERSION as the word rather than the unrenderable stand-in` (go-to-k/cdkd#3016's own case) |
  | 15 | `getLockRecord`: drop the `expiresAt` normalisation | 3 — `reads an uncoercible expiresAt as NaN …`, `completes the expired-lock TAKEOVER … instead of crashing after the delete`, `state show renders a lock whose expiresAt cannot be coerced` |
  | 16 | `nanIfUncoercible`: convert (`return Number(value)`) instead of probing | 1 — `passes every COERCIBLE expiresAt through untouched, numeric strings included` |
  | 17 | `walkCdkdStateStackTree`: drop the `?? {}` on `resources` | 2 — `--show-nested` over a record whose bag is `null`, and one where it is absent |
  | 18 | `walkCdkdStateStackTree`: drop the `entry?.` optional chain | 2 — `--show-nested walks past a null resource ENTRY`, and the `--json` twin |
  | 19 | `displaySafe`: catch arm RETURNS its tag instead of sanitizing it | 1 — `sanitizes the FALLBACK tag too, rather than returning it straight` |
  | 20 | `displayName`: ASCII allowlist weakened to the plain denylist | 4 — the new not-an-object case plus three go-to-k/cdkd#3003 cases that already pinned it |
  | 21 | `isLockExpired`: coerce before the finite test | 1 — `takes over a lock whose expiresAt is a FUTURE numeric string, since the raw value is not finite` |
  Probes 19–21 and the second case of 16 come from the maintainer-checklist
  pass: three obligations the first 18 left **asserted rather than pinned**.
  Each got a case, and each case was then probed. Only probe 20's clause turned
  out to be pinned already — by go-to-k/cdkd#3003's cases, not by this PR's, so
  the new refusal's own case now carries a non-ASCII character too rather than
  leaning on them.

- [x] **Live-tested against the real state bucket** through `node dist/cli.js`, with hand-edited records planted under a throwaway stack: a `null` resource entry in `state show`, `state show --json`, `state resources` (plain, `--long`, `--json`) and `state show --show-nested` (plain and `--json`), all exit 0 and `--json` keeps `{"Broken": null}`; an absent `resources` bag reads `Resources: 0` in `state list --long`; a lock with uncoercible `owner` / `operation` / `expiresAt` renders `locked by [object Object] (operation: [object Object]), expired NaNmNaNs ago`, and its `--json` carries `expiresAt: null`; an uncoercible version, a `null` version, an array root and a `true` root are each refused (exit 1) naming the value; a `skippedOutputs` digest with an emoji at the cut prints `abcdefghijk😀…`; a `2\nStack: FORGED` version is refused on one line. Every planted object version was swept afterwards, and the sweep asserted 0 remain. The Placement trade was measured the same way rather than reasoned: with a body of `0` planted for one stack, `state list --long` exits 1 with `State file for stack '<name>' is not a JSON object (it parses to a number)`, while plain `state list`, `state list --json` and `state list --tree` each still enumerate that stack and exit 0 — which is also the scope line in go-to-k/cdkd#3069. That object's versions were swept to 0 as well.
- [x] Codex: delta review 3 rounds (round 1 found three overclaims in comments and the changelog, fixed; rounds 2 and 3 NO ISSUES, round 3 after re-applying onto go-to-k/cdkd#3016's second commit), then the maintainer-checklist pass 3 rounds (round 1: six items, round 2: two, all fixed with a probe each; round 3 NO ISSUES), one more delta round over the rebase onto merged `main` — NO ISSUES — and a body review plus a second maintainer-checklist pass before opening. The body review found six wrong claims (two rows of the shape table naming views that never threw, three probe enumerations, one over-general "renders as `[object Object]`"), all corrected against the pre-image and re-measured probes. The checklist pass found the three unpinned obligations above plus three text problems: `displaySafe`'s header claimed EVERY caller keys a decision on emptiness (`ConsoleLogger.formatMessage` and `sameLockIdentity` do not), and `docs/cli-state.md` both conflated "cannot be coerced" with "coercion THROWS" — only the throwing arm is replaced, `{}` and `"soon"` still pass through as stored — and never mentioned the absent/`null` resources bag. Round 4 of the delta review then caught a false sentence in one of those new test comments; round 5 NO ISSUES. A further maintainer-checklist round (round 5) raised seven items against the `state list` tolerance, one of them a plaintext-disclosure blocker its sanitization could not answer; that code was reverted in full rather than patched, and the remaining delta was re-reviewed after the revert. The issue body for go-to-k/cdkd#3069 took two rounds (round 1: the impact was overstated to every `state list` mode when only `--long` does the reads, plus two wrong mechanism claims; round 2 NO ISSUES).
- [x] Review: 3-axis + the additive security reviewer (12 files / 979 LOC is 3-axis by size; every touched `src/` file also carries 2-3 `fix:` commits in its last 3, and the subject is the terminal sanitizer over attacker-writable input). Security: clean, with the lock takeover, the refusal messages and the `null`-version mapping each traced. Spec: clean against the issue's shape list and verification plan. Code: three minor items — two taken (the read-boundary comment now argues fixability rather than asserting a benefit two commands do not yet deliver, and go-to-k/cdkd#3018 is re-framed as the 21-site class); the third, `state list` surviving one unreadable record, was implemented, measured to be a larger and partly undecided change than this PR should carry, and reverted in favour of go-to-k/cdkd#3069 — with the trade stated in the Placement section above. Test: three gaps, all closed — the `--json` `expiresAt` arms this PR's docs promise, the rendered SHAPE of a null entry (`Type` / `PhysicalID` read `undefined`; `state resources --json` emits the entry with its absent fields absent), and two legacy cases relabelled as CHARACTERIZATION of go-to-k/cdkd#3016's sanitization rather than of this diff.
- [x] No integ: none of the five changed `src/` files is in the `integ-destroy` include, `integ-broad`'s cross-cutting regex, `integ-local` (`src/local/**`) or `integ-schema-migration` (`src/types/state.ts`).


go-to-k:
Reviewed head `209d7ee3` (merge-base `08e563c2`) with the 3-axis set plus the security reviewer, since `displaySafe` is the terminal sanitizer and state/lock records are attacker-writable. The mechanism is sound: the `displaySafe` fallback is sanitized on both arms and can only emit a fixed ASCII tag for JSON-derived input; `isLockExpired` still decides on the raw value, so the takeover predicate is unchanged and a live numeric lock cannot be taken; none of the four relaxed dereferences reach a write path; every sampled mutation probe discriminates as the table says. **No blocker.** Two small text fixes before merge, two optional ones.

**Required before merge**

- **M1 — changelog overclaims one delta.** `changelog.d/entries/2026-09-11-2947-state-record-shape.md` says "The legacy-fallback debug line sanitizes both regions and the stack name it prints." That sanitization is already on the merge base (`git show 08e563c2:src/state/s3-state-backend.ts`, line 1344, shipped by #3016); this PR does not touch `tryGetLegacy`. What this PR adds there is only that an uncoercible region no longer throws, via the `displaySafe` fallback — say that instead. While you are in the entry, the residual sentence names the three consumers #3018 started with; #3018 has since been widened to the whole class and records that `cdkd state destroy` / `cdkd state orphan` still abort on a `null` or absent `resources` BAG — the recovery commands the placement rationale leans on. Point the residual at #3018's current scope rather than the three-site list.

- **M2 — `parseStateBody` rationale claims a mechanism the code does not have.** `src/state/s3-state-backend.ts:1413-1415`: "Normalising one would be worse — a dropped `null` resource entry reads as "no resource", and a deploy then creates a second one." Deploy already reads a `null` entry that way today without any normalising — `src/deployment/deploy-engine.ts:3528` tests `currentState.resources[logicalId]` for truthiness, and `null` and absent are the same there. So "would be worse" is not supported by that path; the decision (no inner validation at the boundary; tolerate at the dereference sites; keep `state show --json` verbatim) stands on the recovery-command and verbatim-comparand arguments alone. Drop the "second one" clause rather than replace it with another causal claim.

**Optional (take-or-leave)**

- **m3 —** `src/cli/commands/export.ts:3399-3403` says the tolerance matches `renderStateBlock` and names `--show-nested`, but `cdkd export` also reaches this walker (`export.ts:2662`). It is still safe there — the import plan is built from the synth template, and `buildImportPlan` (`export.ts:3871`) still throws on a `null` bag — but one clause naming the export caller would stop the next reader re-deriving it.
- **m4 —** two behaviours are live-tested only: `expiresAt: NaN` serialising as `null` in `state show --json` (`tests/unit/cli/state-record-shape.test.ts:307` uses a numeric `expiresAt` for the `--json` case), and the non-object-root refusal through the legacy-key path (`tryGetLegacy`'s `Failed to get legacy state …` wrap). A unit case for each would pin them.

**Out of scope, noted so it is not re-litigated**

- `expired NaNmNaNs ago` (`formatLockSummary` / `formatDuration`) is the row `{}` already produced before this PR, while `formatRemaining` in `lock-contention-message.ts` prints "at an unknown time" for the same non-finite input. Cosmetic and pre-existing; not this PR's.
- The `?? {}`-guarded-but-entry-unguarded family (`drift.ts:2232`, `:2985`, `:3044`; `destroy-runner.ts:398` `countProtectedResources`; `orphan-adoption.ts:527`, where the throw is caught and the sibling's physical ids are silently not claimed) is outside #3018's grep. I am adding those to #3018 myself.

Push the M1/M2 fix and I will re-review the new head.

go-to-k:
Re-reviewed head `428216ff`. M1 resolved (the base-shipped sentence is gone and the residual now points at the class). M2 resolved: the "second one" clause is gone. The `--json` shape case and the `NaN`-vs-`{}` case cover the half of m4 that mattered. CI green. One word in the replacement text is a claim the code contradicts, so one more small round:

**Required before merge**

- **M3 — "actionable" overclaims.** `src/state/s3-state-backend.ts`, the new rationale: "a `null` entry keeps the logical id visible and actionable while a dropped one erases it from the record silently." Visible, yes — `id in resources` is true and `Object.keys` counts it. Actionable, no: `cdkd state orphan Broken` on `{"Broken": null}` passes the `in` check in `orphan.ts:228` and then throws at `src/analyzer/orphan-rewriter.ts:463-466` (`if (!r) throw new Error(... not found in state.resources)`), so the one command that could act on the id refuses it today. That is #3018's business, but the comment must not say the id is actionable now. Drop "and actionable" (or say "visible — and nameable to a command, even though `state orphan` still refuses it today (#3018)"). The rest of the sentence, and the keys-enumeration argument it makes, hold.

Everything else on this head is clean; push that and I will merge.

go-to-k:
Applied M3 myself as `d329bb18` (comment-only: "visible and actionable" → "visible (nameable to a command, even though `state orphan` still refuses the entry at its own dereference today — go-to-k/cdkd#3018 again)"), since the rest of the head was clean and it was one clause. Merging once CI is green on that head; nothing further needed from you.

go-to-k:
Merged — thank you. The shape table, the 21-probe matrix and the live run made this one of the easier large PRs to verify: every reviewer claim I re-derived matched the table. Of the optional items, m3 (naming the export caller in the walker comment) and the legacy-path half of m4 are consciously left as-is; the review thread records why each is safe. The `NaNmNaNs` row is now #3083.

Issue #3062, opened 2026-09-13
secret-redaction: a nested-stack parameter embedding a 1-3 character secret through an Fn::Join / Fn::Sub source still reaches the child in plaintext


nix-tkobayashi:
## Summary

A nested-stack `Parameters` value that embeds a 1-3 character secret through an `Fn::Join` / `Fn::Sub` SOURCE (`Pin: { 'Fn::Join': ['', ['port:{{resolve:secretsmanager:', { Ref: 'Secret' }, ':SecretString:pin::}}']] }` -- what `'port:' + secret.secretValueFromJson('pin')` synthesizes, the prefix fused into the token's opening part and the ARN `Ref` inside it) still reaches the child's `state.json` as `port:q7`. The literal spelling of the same shape (`Pin: 'port:{{resolve:...:pin::}}'`) is closed by go-to-k/cdkd#2745's nested-stack site; this is the object spelling, which that issue scoped out ("spelled as a literal embedding the token").

## Where it stops today

`recordNestedStackParameterExpressions` (`src/deployment/secret-redaction.ts`) writes the sub-floor carry -- `'port:q7' -> 'port:{{resolve:...}}'` as a whole-value entry of the parent's bag -- under three conditions, and the first is `positioned[name] === sourceLeaf` with `typeof sourceLeaf === 'string'`: on a literal source `positionByEmbeddedSpan` returns the source verbatim, so equality proves the pass certified the leaf. An object source has no such identity to test. The parent's OWN record of the row is already correct for this shape (`positionByIntrinsicFrame`, go-to-k/cdkd#2745's first site, runs over the marked copy and writes `port:{{resolve:...}}`), so what is missing is only the carry: the parent persists the frame, the child persists `port:q7`, and `cdkd diff --recursive` reports the child leaf as a change on every run.

## Proposed fix

[...]

Issue #2347, opened 2026-08-27
fix(deployment): coerceParameterTypedValue recognises only two list types, so a List<AWS::...> child parameter resolves to a string


go-to-k:
Dup-check: searched open issues for `coerceParameterTypedValue`, `parameter type coercion`, `List<Number>`, `CommaDelimitedList` and `nested stack parameter type`, plus a body-window scan for `coerceParameterTypedValue` / `List<AWS::` / `isListType`. The only neighbour is go-to-k/cdkd#2327, whose root cause is the secret-redaction COLLAPSE for two list-typed parameters sharing one plaintext -- a different defect that happens to be reachable through the same coercion. Nothing covers which types the coercion recognises.

Session-fit: next (not this session) -- widening the switch is a behaviour change in `src/deployment/intrinsic-function-resolver.ts`, which is in the `integ-broad` cross-cutting scope, and its downstream blast radius is unmeasured. The run that found this was closing go-to-k/cdkd#2327 and deliberately did not widen into it.
Severity: medium -- a child parameter of a COMMON declared type receives a string where the CDK's own model says a list, and nothing warns. It is not a disclosure and there is a workaround (declare `CommaDelimitedList`), but the divergence is silent.
Effort: medium (M) -- the switch itself is three lines; what it drags is a re-review round on a cross-cutting file plus measuring what actually consumes the coerced value.
Estimate: ~2-4 h -- most of it is establishing the downstream consequence, not the edit.

## cdkd holds TWO contradictory views of which parameter types are LISTS

**View 1 -- `coerceParameterTypedValue`** (`src/deployment/intrinsic-function-resolver.ts`, the function whose own JSDoc says "ONE definition of parameter-type coercion"):

```ts
switch (type) {
  case 'Number':             return Number(value);
  case 'List<Number>':       return value.split(',').map((v) => Number(v.trim()));
  case 'CommaDelimitedList': return value.split(',').map((v) => v.trim());
  case 'String':
  default:                   return value;
}
```

So exactly TWO types coerce to an array. `List<String>`, `List<AWS::EC2::Subnet::Id>`, `List<AWS::EC2::SecurityGroup::Id>` and every other `List<...>` fall to `default` and come back as the STRING.

[...]

Issue #2087, opened 2026-08-20
fix(deployment): the #1903 child-engine seed hands every child resource the parent's secret bag, so an unrelated literal containing the plaintext is spliced into an expression


go-to-k:
## Context

Found by the security review of branch `fix/1903-2055-nested-stack-secret-flow` (pushed, no PR), which fixes go-to-k/cdkd#1903 by seeding the parent's resolved secret pairs into the child `DeployEngine`.

## The gap

`deploy-engine.ts:972` seeds the inherited pairs into EVERY child resource's `perResourceSecrets`, not only into the resources that actually consumed the parameter. `redactSecretsForState` then substring-matches at or above `MIN_NEEDLE_LENGTH` (4 characters).

Trigger: the secret plaintext is `production`. An unrelated child resource declares `BucketName: my-production-bucket`. The child's state persists `my-{{resolve:secretsmanager:...}}-bucket`.

That is the same class as go-to-k/cdkd#1904 and the "value-based redaction over-redacts a coinciding literal" lesson, reached from a new direction: not two expressions colliding on one value, but ONE expression's value colliding with an unrelated literal in a resource that never referenced it.

## Why it does not self-heal

`redactParametersForDiff` rewrites only the PARAMETERS, so the desired side of the comparison keeps the literal `my-production-bucket` while the persisted side holds the spliced expression. Every subsequent deploy therefore sees a change:

- on an ordinary property, a perpetual UPDATE
- on a create-only property, a perpetual REPLACEMENT of a resource nothing asked to replace

## The in-code rationale is wrong and should not be kept as written

The seam carries a note saying this is "the same over-approximation the parent already accepts". It is not. The parent scopes its bag to the ONE resource whose resolution produced the secret (`perResourceSecrets` is keyed by logical id). The child seed hands every resource the same bag, so the blast radius is the whole child stack rather than one resource.

## Direction

Record the pair at RESOLUTION time instead of pre-seeding every context: when the child's resolver substitutes a `{Ref: <Param>}` whose value is a key of `inheritedSecrets`, record it into that resource's own map. That restores the parent's per-resource scoping and keeps the go-to-k/cdkd#1903 fix, because the resources that genuinely carry the secret are exactly the ones that resolve that `Ref`.

## Verification this needs

[...]

PR #3070, opened 2026-09-13
fix(secret-redaction): carry a sub-floor literal frame across the nested-stack handoff as a whole-value entry

### LINKED ISSUE #2745: fix(secret-redaction): a 1-3 character embedded secret still persists in plaintext for an Fn::Join / Fn::Sub source, a nested-stack parameter value that embeds it, and cdkd import's own resolution

nix-tkobayashi:
## Summary

Closes #2745 -- its last open site, the nested-stack parameter carry. The first site (an `Fn::Join` / `Fn::Sub` source) and the third (`cdkd import`'s own resolution) landed in go-to-k/cdkd#3052.

A parent `Parameters` value spelled as a LITERAL frame around one token (`Pin: 'port:{{resolve:secretsmanager:...:pin::}}'`) resolves to `port:q7`. The child's carry (`inheritedSecretsCarriedBy`) records an inherited pair for a whole-value match at any length or a substring match at/above `MIN_NEEDLE_LENGTH`, and the framed value was neither: not a key of the parent's map, middle below the floor. So the child's `{Ref: Pin}` leaf persisted `port:q7` in `state.json`, and `cdkd diff --recursive` reported that leaf as a change on every run (the recursive diff binds the child's `Pin` to the framed string; measured in the new unit case, not reasoned).

## What changes

`recordNestedStackParameterExpressions` (`src/deployment/secret-redaction.ts`) is the only point holding the literal frame beside the pair. Two additions, both inside it:

1. The position pass runs over a MARKED shallow copy of the row's resolved `Parameters` (`markSameGenerationBag({ ...resolvedParameters })`, the seventh mark site), so `positionByEmbeddedSpan` writes the sub-floor middle as its token there on pair evidence. The caller's own object is never marked here; whether it is marked later is the caller's decision, as before.
2. A second walk records `resolvedValue -> sourceLeaf` (`'port:q7' -> 'port:{{resolve:...}}'`) as an ordinary WHOLE-VALUE entry of the parent's own per-resource bag, under five conditions stated on the docstring: (i) the pass certified the leaf; (ii) the value scan is silent on it; (iii) the frame's token is the map's survivor for its middle; (iv) every spelling of the value in the row is this one frame carrying this pass's pair; (v) no other leaf of the row would be rewritten by the entry (equality anywhere outside `Parameters` or inside a list-valued sibling; containment only for a value at or above the floor). Written after the walk. No pair and no pin are recorded for the entry.

Every consumer then reads the entry through floorless whole-value arms that already exist, with no change in `deploy-engine.ts`, `intrinsic-function-resolver.ts`, `diff-recursive.ts` or `rollback-executor.ts`: the child's carry and persist, the child's outputs pass, `redactParametersForDiff`, `cdkd diff --recursive`, and the rollback replay's three recorder sites (`STATE_DERIVED_RULES`).

Docs: `markSameGenerationBag`'s site list (seventh site), `recordedExpressionsOf` (the second non-token class in the candidate unions), the refusal-1 note (now states what `extractParameters` does since go-to-k/cdkd#2347), one-line pointers in `.claude/rules/layout-deployment-secrets.md` / `layout-deployment.md` (byte-capped), a changelog entry.

## Residuals, stated on the recorder's docstring

- (a) Over one middle with two tokens, (iii) records the survivor's frame only. The CHILD's loser leaf takes the survivor's frame when the two frames match, and stays plaintext when they differ; the parent's record stays per leaf by (iii) and (iv). On a match this is the wrong reference the three live consumers named on `positionByIntrinsicFrame` re-resolve (`cdkd rollback`, `drift --revert`, a consumer's cross-stack read).
- (b) The framed value is a 7-character substring needle in every child resource that consumed the parameter (the go-to-k/cdkd#2087 direction, scoped the same way); on the parent's own row (v) refuses the entry instead, over the leaves visible at record time -- a readback leaf AWS later rewrote to contain the value is spliced like any other 4+ character needle's.
- (c) A frame longer than `MAX_SKELETON_CANDIDATE_LENGTH` makes the intrinsic arms refuse that child resource's intrinsic-sourced leaves.
- (d) A child OUTPUT carrying the framed value re-resolves in the parent to `port:q7`; the cross-stack seam refuses a non-token, and the consumer's leaf persists the plaintext.
- (e) An `Fn::Join` / `Fn::Sub`-spelled parameter source is out of this arm's reach -- go-to-k/cdkd#2745 scopes this site to the literal spelling; filed as go-to-k/cdkd#3062 with the certification argument it needs.
- (f) A child record persisted before this carry keeps `port:q7` until the child is next redeployed (the parent's row already held the frame, so a parent deploy with an unchanged child row never re-runs the child); `cdkd scrub` cannot repair it.

## Tests

- `tests/unit/deployment/secret-redaction-nested-parameter-source.test.ts`: the per-condition discriminators -- the positive carry read back through a mirror of the child's two carry arms; the fixed 1-3 range plus the four-character control; a frame shorter than the floor in total (whole-value arm alone); the survivor case with matching frames and the plaintext half with different frames; (iv)'s boundary-collision, shared-prefix, three-token, object, plain-literal and public-sibling shapes; (v)'s containing sibling, plain literal, `TemplateURL` (containing and equal), list sibling, and the tiny-frame split; the "map no resolver populated" case for (i); `STATE_DERIVED_RULES`; the copy boundary (`isSameGenerationBag`); and a refusal table whose cases pin shapes, not single conditions (said so in place).
- One consumer each: `intrinsic-resolver-inherited-parameter-secrets` (the real carry, whole-value arm alone), `deploy-engine-nested-stack-inherited-secrets` (child persist, the pre-fix discriminator, the diff binding), `deploy-engine-nested-stack-shared-parameter-plaintext` (parent engine, CREATE and UPDATE sites, the parent's own record framed), `rollback-executor-nested-stack-secret-scope` (the replay's bag), `tests/unit/cli/diff-recursive-nested-stack-secret.test.ts` (NO_CHANGE on the framed record; UPDATE on the pre-fix record, measured), `secret-redaction-intrinsic-frame` (the framed entry beside the frame arm; the length-cap residual).
- Fixture `tests/integration/nested-stack-secret/`: `SubFloorPin` (parent, literal `port:` + a two-character token), `PinParam` (child `{Ref}`) and `ChildPinOutput`; premise guards on the synthesized spelling (a string, not an `Fn::Join`) and on the pin's length; state, `observedProperties` and output framed, live value `port:q7`, the named defect checked before the equality assert; phase 2c re-resolves `PinParam` through the UPDATE site; `mask()` masks values shorter than the floor whole.

## Verification

- Unit suite on the rebased tree: 999 files, 22168 passed, 1 skipped (root-asserted run, green). One flake seen while measuring, in `tests/unit/deployment/intrinsic-resolver-concurrent-drain.test.ts` (go-to-k/cdkd#3044's timing cases, untouched by this PR): on this branch it failed in an earlier full-suite run and on the first solo run, then passed the next three solo runs; on a pristine `origin/main` worktree it failed a DIFFERENT case on the first solo run and passed the next two. Recorded as measured; the mechanism was not established.
- Mutation probes, each red on at least one case with the tree restored byte-exact between them: the mark on the copy; the copy boundary (marking the caller's object instead); (i); (ii); (iii); (iv)'s frame gate, its length split, its suffix, its pair evidence; (v)'s containment, its outside-`Parameters` equality, its list-leaf walk, its floor split (`>=` to `>`), and (v) skipped below the floor; pinning the framed value as an expression; dropping the `set`; `MIN_NEEDLE_LENGTH` raised to 5; each of the three replay recorder sites dropped in turn. Stated as equivalent rather than pinned, on the docstring: `set` inside the loop; a pair for the framed entry.
- Codex: the follow-up issue body, 3 rounds to NO ISSUES; the delta, 15 rounds in all -- 8 to the first NO ISSUES (rounds 1-6 became conditions (iv) and (v), residual (a)'s two halves, the refusal-1 note's `extractParameters` correction and the `UNFRAMED_SPELLING` pair test) and 7 more on the fix rounds that followed the proxy and reviewer passes, each ending NO ISSUES; a maintainer-proxy pass on the delta, 3 rounds (frame-key collision cases, the copy-boundary case, the fixed 1-3 range with the four-character control, the tiny-frame and at-floor cases, `mask()` whole-masking below the floor, docstring parity); the fixture sentinel and ledger delta, 2 rounds (round 1 asked for `gone_probe` too; declined as above); this body, 3 rounds plus a maintainer-proxy pass whose findings became the here-string conversion of the negative scans, residual (b)'s record-time qualification, and the corrected statement that a framed carry entry CAN match the skeleton pattern of a join spelling the same frame (naming that leaf's own frame; `isKnownSecretExpression` reads a map value as secret regardless).
- Reviewer agents, twice: on the working-tree diff -- `pr-code-reviewer` clean plus one nit (the sentinel's declaration order, fixed); `pr-security-reviewer` clean with the residuals traced from the write to every reader, plus two nits (residual (f) added; the pre-entry `--verbose` line is the masking-floor residual `positionByEmbeddedSpan`'s docstring already states); `pr-test-reviewer` two gaps (the (v) floor boundary and the two other replay sites), both fixed and probed -- and again on the PR head (14 files, 3-axis by file count, the spec axis replaced by the security add-on since no design document exists): `pr-code-reviewer` no blocker, two items fixed (a duplicate `lambda` ledger row a rebase had merged in; the `redactSecretsForState` comment that still named a nested-stack `Parameters` block as an unmarked sub-bag) and the changelog cap noted; `pr-test-reviewer` two low gaps fixed and probed (a list-valued sibling at sub-floor length, a nested `Tags[].Value` leaf outside `Parameters`); `pr-security-reviewer` clean, one nit declined (`gone_probe`, as above); and a third `pr-code-reviewer` pass scoped to that fix delta: clean, one assertion nit applied.
- Real AWS (`us-east-1`), on the rebased tree: `nested-stack-secret` PASS in 280 s on the final tree (all nine stack-owned SSM parameters gone, 0 surviving state versions on both the parent and the child prefix), after a PASS in 267 s on the tree with only the three positive sentinels fixed; broad `lambda` PASS in 107 s (clean destroy, prefix swept to 0 versions). The two runs before those failed on the fixture's own positive-match `printf | grep -q` sentinels (go-to-k/cdkd#2582's race; `printf: write error: Broken pipe` in the log); this PR turns every `printf | grep -q` scan in the fixture into a here-string -- the negative plaintext scans too, whose early match would have read as "no match", the fail-open half -- leaving only the canonical `gone_probe` block, which `integ-verify-probe-not-found.test.ts` pins and go-to-k/cdkd#2582 closes centrally. The go-to-k/cdkd#2745 assertions passed in every run that reached them. Ledger rows added.
- Gates: `check`, `docs`, `integ-destroy` (empty in-scope delta: this PR touches no file the gate covers) and `integ-broad` (the lambda run) recorded; `/verify-pr` bound to the PR head.


go-to-k:
Merged -- thank you. Reviewed at head 4e3992a with code, test and security passes; all three came back without a blocker, and the fixture passed on the maintainer account too (`nested-stack-secret`, 206 s, clean destroy, 0 surviving state versions). Condition (iv)'s row-wide frame gathering and condition (v)'s leaf walk were the parts that made the carry safe to record without touching any consumer -- nicely scoped. Residual (a) now has its own number, #3079, with the review's docstring / test-gap notes attached there.

Issue #3068, opened 2026-09-13
SecretsManager: GenerateSecretString.IncludeSpace and RequireEachIncludedType are declared members the local generator never reads


go-to-k:
Found while working go-to-k/cdkd#3056 (the member-level shape guards), which deliberately leaves this out: it is a FIDELITY gap, not a shape question.

## What

`SecretsManagerSecretProvider.generateSecretString` (`src/provisioning/providers/secretsmanager-secret-provider.ts`) mints the password LOCALLY (Secrets Manager's `GetRandomPassword` is never called), so the recipe is whatever the provider implements. Two members the registry schema declares on `GenerateSecretString` are never read:

- `RequireEachIncludedType` (service default **true**): the service guarantees at least one character from every included class; the local generator draws uniformly, so a short password can miss a class -- a `PasswordLength: 8` with punctuation included has no punctuation ~(63/89)^8 ≈ 6% of the time.
- `IncludeSpace` (default false): the local charset never contains a space, so a template that opts in gets a password without one.

Neither is malformed-value handling -- a well-formed template simply gets a different password DISTRIBUTION than CloudFormation would produce. No silent-drop critic sees it: `GenerateSecretString` is a handled top-level property and the members are consumed by a local algorithm, not forwarded to an SDK call, so `gen-nested-key-coverage` has no SDK member to compare against.

## Remedy

Read both through the shared `config-shape.ts` boolean predicate (as #3056 does for the other members), add `' '` to the charset under `IncludeSpace`, and under `RequireEachIncludedType` (default true) re-draw until every included class is present (or place one character of each class at random positions). Alternatively call `GetRandomPassword` and let the service own the recipe -- which also retires the local implementation's divergence class wholesale, at the cost of one more API call per generate.

## Verification

`vp test run tests/unit/provisioning/secretsmanager-secret-provider-update-value-source.test.ts tests/unit/scripts/integ-secret-fixture-sweep.test.ts` -- the #2212 fence pins exactly two `generateSecretString` call sites and the helper's allow-list, so the change must stay inside the generator.

Session-fit: now (this session) -- re-classified: the generator, its tests and the #2212 fence are in context from go-to-k/cdkd#3074; the local recipe is kept
Severity: low -- a well-formed template gets a valid password of the declared length; only its class guarantee and the space option diverge from CloudFormation
Effort: small (S) -- one function, plus unit cases on the charset and the per-class guarantee
Estimate: 1h -- deciding local-vs-service is the only real question
[...]

Issue #1653, opened 2026-08-12
fix(dynamodb): the GlobalTable StreamSpecification warn-and-skip records the malformed desired value (#1612 item 3)


go-to-k:
Found by the issue #1612 audit (item 3: "check the same question for the non-S3 warn-and-skip sites").

## The gap

`DynamoDBGlobalTableProvider.update` guards `StreamSpecification` with a
`readConfigString` whose `onUnusable` downgrade SKIPS the block (issue #1551 —
defaulting would silently re-point a live stream's view type, so the skip is the
right call):

```ts
let streamSpecUnusable = false;
const streamViewType = readConfigString(properties['StreamSpecification'], 'StreamViewType', ...);
if (!streamSpecUnusable) {
  flatUpdate.StreamSpecification = { StreamEnabled: true, StreamViewType: streamViewType };
  flatChanged = true;
}
```

The deploy then SUCCEEDS, so the engine records the DESIRED bag — malformed
`StreamSpecification` and all — for an update that never reached AWS.
`readCurrentState` can never match it, every later `cdkd drift` re-reports the
same difference, and `drift --revert` re-issues the same skipped call. That is
the permanent phantom drift `.claude/rules/providers.md` records for the EC2
Route warn arm (#1591), reached through a SKIP instead of a narrowing.

`update()` returns no `effectiveProperties` today — `grep -c effectiveProperties
src/provisioning/providers/dynamodb-globaltable-provider.ts` is 0.

## The fix

Per the carve-out the #1612 PR adds to `.claude/rules/providers.md`: on the
UPDATE path retain the PREVIOUS value (AWS still holds the previously applied
stream configuration) and return it as `effectiveProperties`. Do NOT implement a
`canonicalizeDesiredProperties` twin — for a SKIP that arm derives a REMOVAL and
would disable the live stream.

## Scope

- `src/provisioning/providers/dynamodb-globaltable-provider.ts` update path.
- Check the create path's sibling guard (around line 400) for the replay-CREATE
  answer: DROP the key.
- Unit pin plus a live test (a GlobalTable deploy, then a redeploy whose state
  record carries a malformed `StreamSpecification`, asserting two consecutive
  `cdkd drift` runs converge instead of re-reporting the same difference).

Split out of #1612 rather than bundled because it is a different provider file
and needs its own real-AWS verification.

go-to-k:
[...]

Issue #1544, opened 2026-08-11
Three create-path config-shape refusals do not downgrade under CreateContext.replayingState


go-to-k:
## Summary

Surfaced by the 3-axis review of PR #1542. `CreateContext.replayingState` exists so a provider's create-path PRE-FLIGHT REFUSAL downgrades to a warning when the properties come from a cdkd STATE record — the rollback executor's reverse-replacement arm revives the old resource from `previousState.properties`, which the user cannot edit from their CDK code. A refusal there leaves the old resource unrestorable with only a hand-edit of `state.json` as a remedy.

Three create-path refusals do not downgrade, because their providers never declared the optional `context` parameter:

- `src/provisioning/providers/wafv2-provider.ts:526` — `requireConfigString(properties['Scope'], 'REGIONAL', …)`, no options bag.
- `src/provisioning/providers/lambda-url-provider.ts:69` — `requireConfigString(properties['AuthType'], …)`, no options bag.
- `src/provisioning/providers/s3-directory-bucket-provider.ts:157` — `requireConfigString(properties['DataRedundancy'], …)`, no options bag.

A second, narrower instance lives inside `dynamodb-globaltable-provider.ts` itself, which DOES now receive a context: its sibling guards still hard-throw because the helpers they use take no options —

- the `StreamSpecification` guard (`readConfigString`, which has no `ConfigStringOptions` parameter at all), and
- `toSdkGlobalSecondaryIndexes`'s `requireConfigArray`.

So a state record carrying `StreamSpecification: ''` still makes a reverse-replacement create un-rollbackable.

## What this needs

1. Thread `context?: CreateContext` into the three providers' `create()` signatures and spread `replayWarn(this.logger, context)`.
2. Decide whether `readConfigString` / `requireConfigArray` should grow a `ConfigStringOptions` parameter so the nested-container guards can downgrade too, or whether the container class is different enough to stay strict. `config-shape.ts`'s header records the DESIRED-side-only rule that governs the decision.
3. Each change needs a unit test asserting both polarities, as in `tests/unit/provisioning/dynamodb-globaltable-provider-billing-mode-shape.test.ts`.

Note this is a downgrade-on-replay gap, not a silent-drop: the refusals are correct on a template-path create. The blast radius is limited to rollback of a replacement whose OLD state record carries a value the guard refuses.

Refs #1463, #1513

go-to-k:
[...]

PR #3074, opened 2026-09-13
fix(secretsmanager): refuse or skip a malformed GenerateSecretString member instead of minting a wrong secret

### LINKED ISSUE #3056: SecretsManager: a malformed GenerateSecretString MEMBER still throws on the update path (ExcludeCharacters), and the other members silently default

go-to-k:
Closes #3056

## What

The member half of go-to-k/cdkd#3048. That PR guarded the `GenerateSecretString` CONTAINER; its members were still read by cast, and the defaults they took were not inert:

- `PasswordLength: 'abc'` (truthy) reached `new Uint8Array('abc')` and minted an EMPTY password;
- `ExcludePunctuation: 'false'` (truthy) EXCLUDED punctuation;
- a `SecretStringTemplate` that did not parse fell into a `catch` and the bare password went out RAW where a JSON document was declared -- each as the new `AWSCURRENT`, silently;
- `ExcludeCharacters: null` THREW from `update()`, the go-to-k/cdkd#1544 hazard one level below the container.

Now every member is read through the shared `config-shape.ts` predicate for its type -- a new `configIntegerRefusal` / `coerceCfnInteger` pair for `PasswordLength` (floor 1, cap 4096), the boolean pair for the `Exclude*` flags, `configStringRefusal` for the three strings -- plus a pair rule for `GenerateStringKey` / `SecretStringTemplate` (together or not at all, the template a JSON object, the key not `__proto__`). The rules live in one module-level `generateMemberRefusal`, shared by the generator AND `retainPreviousGenerateBlock`, so a previous block with a malformed member is dropped rather than re-recorded (the go-to-k/cdkd#1653 same-predicate rule). On the create path a refusal throws; on the update path it is the same SKIP the container takes, announced once, with the previous block retained via `effectiveProperties`.

## Measured on CloudFormation (us-east-1, 2026-09-13), not read off the docs

- `SecretStringTemplate` without `GenerateStringKey`: refused (`SecretStringTemplate and GenerateStringKey must both be set or removed`) -- the docs pin only the other direction.
- `PasswordLength`: `"+12"` and `"007"` pass validation; `" 12 "` fails (`found: String`), `"1e1"` fails (`found: Float`). `coerceCfnInteger`'s string grammar is exactly that (`/^[+-]?\d+$/`, no whitespace tolerance), where `Number()` would have minted from a template the service rejects.

## The go-to-k/cdkd#2212 fence

The helper's identifier allow-list and the module's top-level allow-list each gain the one shared predicate's name (a free function so the recording helper reaches it without `this`). A new arm refuses a property write on every value binding the module holds (`generateMemberRefusal.cache = v`, `CreateSecretCommand.stash = v` -- measured GREEN before); the cast spelling is a recorded bound the runtime half closes on the read.

## Verification

- Unit: 999 files / 22199 tests green. 21 malformed-member rows on both paths, 6 CFn-spelling coercion rows, the template write, the retention drop, the integer guard's own describe; ~20 single-mutation probes across three review rounds, each red one case.
- Review: code + test + security (fable) round 1, code + security delta round 2, security closing round 3 -- no blocker left, every should-fix applied, bounds recorded in the fence.
- Live: `secretsmanager-update-value-source` PASS (79 s) on the final src tree, destroy 0 errors / 0 orphans / 0 surviving state versions.

## Residual

- go-to-k/cdkd#3068 -- `IncludeSpace` / `RequireEachIncludedType` are declared members the local recipe never reads (fidelity, not shape).


go-to-k:
Merged as 1300b9a5. Thanks to the reviewers -- the round-1 catch that the retention ran only the container predicate, and the push to MEASURE the integer grammar and the pair rule on CloudFormation rather than read them off the docs, are what made this the member half of #3048 rather than a second pass over it.

Issue #3081, opened 2026-09-13
hooks: close_paren / subst_open do not read `)#` as a comment, so a `cd` inside $( ) is emitted top-level


go-to-k:
## Summary

`close_paren` and `subst_open` in `.claude/hooks/lib/command-match.sh` treat `#` as a comment only after `[ \t;&|(]`. bash 5 and zsh also start a comment after `)`, so in

```
x=$( (echo a)# note )
  cd /tmp
)
echo POISON > docs/_generated/ledger.tsv
```

the `)` inside the comment is counted as the substitution's closer. `subst_open` reports the span CLOSED at line 1, and `gate_segments_marked` emits `cd /tmp` at mark 0 (top level) -- measured on `origin/main` and on go-to-k/cdkd#3040's branch alike:

```
1	echo a)# note
0	cd /tmp
0	)
0	echo POISON > docs/_generated/ledger.tsv
```

bash 5 runs the `cd` INSIDE the substitution (the parent shell stays in the repo), so `main-tree-edit-gate`'s ordered walk follows a `cd` the real shell never took and resolves the write against `/tmp`: the round-4 code review of go-to-k/cdkd#3040 drove this through the real `main-tree-edit-oracle` harness and got `written=yes rc=0` -- a live fail-open of that gate, inherited (`origin/main` rc=0 too). Under bash 3.2 the shape is a syntax error and nothing runs.

## Why it is not in go-to-k/cdkd#3040

That PR fixed the same class in `last_heredoc_opener` (round 3, `)#`; round 4, `` `# ``), which is new code with its own fences. `close_paren` / `subst_open` are the segmenter every gate shares; the fix is two characters in two classes, but it changes segmenting for every consumer and has to be priced by `command-match-differential.test.sh` and driven through `main-tree-edit-oracle.test.sh` on its own, not folded into a fourth review round of a heredoc change. The file's own comments say the three state machines must agree; today they do not.

## Verification

```
bash -c 'source .claude/hooks/lib/command-match.sh; gate_segments_marked "$(printf "%s\n" "x=\$( (echo a)# note )" "  cd /tmp" ")" "echo POISON > docs/_generated/ledger.tsv")"'
```

expects `cd /tmp` at mark 1 after the fix. Add the shape to `main-tree-edit-oracle.test.sh`'s corpus (bash 5 ground truth: file NOT written outside the repo... the gate must still refuse the top-level `echo POISON >`), and the backtick twin `` x=`#note` ``.

[...]

PR #3082, opened 2026-09-14
chore(hooks): gate a cdk-local bump on integ-local, and stop a heredoc inside $( ) reading as commands

### LINKED ISSUE #3040: chore(hooks): integ-local's scope is source paths only, so a cdk-local version bump — the change most certain to move local-execution behaviour — passes ungated
## What happened

`.claude/hooks/integ-local-gate.sh` decides scope with

```
LOCAL_SCOPE_REGEX='^src/local/|^src/cli/commands/local-[A-Za-z0-9_-]*\.ts$|^tests/integration/local-'
```

A PR that bumps `cdk-local` in `package.json` matches none of those, so the merge is not gated — even though `cdk-local` IS the local-execution engine and `src/local/**` is largely a set of shims over it. The change guaranteed to move `cdkd local` behaviour is the one change the gate cannot see.

## The live instance

The `cdk-local` 0.147.7 -> 0.148.4 bump (superseding dependabot go-to-k/cdkd#2725) carries two user-visible deltas, neither of them in any file the regex matches:

- Three partitions (`us-isof-` / `eu-isoe-` / `eusc-`) stop resolving COMMERCIAL, which changes the ECR host every `cdkd local` image-URI path synthesizes. That is go-to-k/cdkd#1821, closed by the bump.
- A cross-stack resolver error's `message` is no longer interpolated into the per-key unresolved reason.

Both were caught only because cdkd carries unit cases that pinned the OLD upstream behaviour on purpose. Nothing structural required a `local-*` integ run; the lane ran one by judgment rather than by gate.

## Why the inverse is not the answer

Adding `^package\.json$` wholesale would fire the gate on every dependabot PR in the repo, which is the shape that gets a gate disabled rather than obeyed. The scope wants the cdk-local DEPENDENCY specifically — a changed `cdk-local` version string in `package.json` / `pnpm-lock.yaml`, not any manifest edit.

That is a precise check rather than a path match, so it needs the treatment `integ-schema-migration-gate.sh` gives the `StackState.version` literal: grep the PR diff for a change to that one dependency, so a `package.json` edit touching anything else passes through.

## Related but distinct

- go-to-k/cdkd#2260 is about a dep bump not being able to trigger a RELEASE. This is about a dep bump not being able to trigger an INTEG.
- `integ-broad-gate.sh` and `integ-destroy-gate.sh` derive scope the same source-path way, so the same question should be asked of them once the shape is settled — a cdk-local bump does not reach their subjects, but this issue is the reason to look.

## A second, separate defect found while filing this

`gh issue create` for THIS issue was blocked by `integ-local-gate.sh`, because the hook's command matcher read a merge-verb out of the heredoc carrying the issue BODY. The body is data, not a command; a gate that matches its own prose is a false positive that will train sessions to route around it. Same class as the `commit-msg-heredoc-gate` carve-out CLAUDE.md already documents. Worth fixing in the same pass, or splitting out if the two remedies diverge.

## Acceptance

Either the gate recognizes a `cdk-local` dependency change and blocks on a fresh `integ-local` marker, or the decision that it should not is recorded with its reason. The check's own probe must break REAL code, per `.claude/rules/testing.md` — a hook that cannot be shown to fire on an actual cdk-local bump diff is not evidence of anything.

Dup-check: searched go-to-k/cdkd issues (open and closed) for `integ-local gate scope`, `gate scope dependency bump` and `cdk-local bump integ`. Nearest hits are go-to-k/cdkd#2260 (dep bump vs RELEASE, not integ), go-to-k/cdkd#2445 (review-vs-integ ORDER, not scope), go-to-k/cdkd#2687 (commit-prefix-scope-gate resolving the wrong tree — a different hook and a different defect) and go-to-k/cdkd#2600 / go-to-k/cdkd#1821 (cdk-local behaviour, not gating). No issue covers this.

Session-fit: next (not this session) -- the lane that found it is the cdk-local bump itself, and gating that bump on a hook the same PR introduces would make the hook's first evidence its own author's run. It also wants the `integ-broad` / `integ-destroy` question answered in the same pass, which is wider than a dependency lane.
Severity: medium -- a local-execution behaviour change can reach users with no real-AWS / real-Docker run demanded at any point, and the two deltas in the live instance were caught by unit cases that happened to pin upstream behaviour rather than by anything structural.
Effort: small (S) -- one hook, one predicate, plus the real-diff probe; it rides an existing gate's shape rather than inventing one.
Estimate: ~1-2 h -- most of it building a real cdk-local-bump diff to prove the hook fires on it and does not fire on an unrelated package.json edit.

### LINKED ISSUE #3066: fix(hooks): a heredoc opener flushed from a substitution body leaks pending_tag into the top-level latch, which then swallows lines up to an unrelated later terminator
## What happens

`lib/command-match.sh`: a `pending_tag` set while `drain_extra()` flushes a substitution BODY leaks into `run()`'s top-level heredoc latch, which then reads the NEXT bare delimiter line anywhere later in the input as that heredoc's terminator and drops every line up to it -- including a real gated verb. Pre-existing on `origin/main`; found by code review of go-to-k/cdkd#3040 while checking a comment that claimed the latch "never fires" for a substitution-borne opener, and reproduced there:

```
x=$(cat <<EOF
p
EOF
)
git commit -m x
cat <<EOF
q
EOF
```

`gate_matches "$cmd" "$GATE_RE_GIT_COMMIT"`: **NOMATCH** on `origin/main` and on the go-to-k/cdkd#3040 branch alike. Bash runs that commit. The control -- the same five lines with the trailing top-level heredoc removed -- is MATCH, so the trailing heredoc's bare `EOF` is what the leaked latch consumes.

## Mechanism

1. `run()` joins the still-open `$(` lines into one logical line and, on the `)` line, calls `flush_line(line)`; the body is queued into `extra`.
2. `__bodies = __bodies drain_extra()` flushes that body through `flush_line` again, and the body's `cat <<EOF` sets the GLOBAL `pending_tag = "EOF"`.
3. Back in `run()`, `if (pending_tag != "" && terminated(pending_tag, i + 1) > 0) tag = pending_tag` runs for the enclosing `)` line. `terminated()` searches to the END of the input, finds the trailing top-level heredoc's bare `EOF`, and latches.
4. Every line from `git commit -m x` down to that `EOF` is dropped as heredoc body.

The heredoc the opener belongs to was already fully consumed inside the substitution; the latch is answering for a terminator that belongs to a DIFFERENT heredoc.

## Fix shape

`drain_extra()` should save and restore `pending_tag` around its flushes the way it already saves and restores `q` (its own comment explains why `q` needs that: "the bodies are a separate scan, not a continuation of the quoting on the enclosing line" -- the same is true of the opener state). Then add the shape above and its control to `command-match.test.sh`, and price it in `command-match-differential.test.sh`, which must be re-run because this is a segmenter change (the vendored baseline predates the drain, so the fence sees a segment-count cell, not the verb cell -- the same blind spot the SUBST_HEREDOC legend records).

Not folded into go-to-k/cdkd#3040 because it is pre-existing and orthogonal: that lane's `ptag` latch is a SEPARATE state that does not touch `pending_tag`, and its scope was already three review rounds wide. The comment there that overstated the latch as never firing is corrected in that lane.

Dup-check: searched go-to-k/cdkd issues (open and closed) for `pending_tag`, `drain_extra`, `heredoc latch` and `command-match heredoc`. Hits are go-to-k/cdkd#3040 (the substitution-body false positive, which is where this was found and which does not cover it), go-to-k/cdkd#2027 / go-to-k/cdkd#2129 / go-to-k/cdkd#2650 (the segmenter rounds that introduced the drain and the per-line body flush, none of which names the leak). No issue covers this.

Session-fit: next (not this session) -- found in review of a lane already three rounds deep in the same file; the fix touches `drain_extra`, a function that lane deliberately left alone, and needs its own differential run.
Severity: medium -- a silent fail-open in the shared matcher every blocking gate sources: a real `git commit` between a substitution-borne heredoc and any later top-level heredoc reaches no gate. The shape needs two heredocs with the same delimiter in one command, which is uncommon in agent-written commands but not contrived (`EOF` is the universal delimiter).
Effort: small (S) -- a save/restore pair in one function plus two cases; the cost is the differential re-run and the both-bash hook runner.
Estimate: ~1 h -- most of it the differential fence and confirming the cell it produces is the intended one.

### LINKED ISSUE #3261: hooks: subst_open keeps the outer double-quote state inside $( ), so a ) in an inner single-quoted span closes the frame early
## Summary

`subst_open` never resets its quote context when it enters a `$(` from inside a double-quoted span, so within `"$( … )"` a `'` is literal to it while bash treats it as opening a span. A `)` inside that inner single-quoted span then closes the frame, the line is never joined with the next, and the verb after the closing quote is swallowed as quoted top-level data.

Measured by the round-25 security review of go-to-k/cdkd#3082; 68 records, all three shells RUN the verb, and **identical on `origin/main`** — pre-existing. Decode and run directly:

```
base64 -d <<'B64' > t.sh
ZWNobyAiJChlY2hvICdhXCdiJykKJyA7IGdoIHByIHhtZXJnZSAxCikiCg==
B64
```

which is

```
echo "$(echo 'a\'b')
' ; gh pr xmerge 1
)"
```

bash 5.3.9, bash 3.2.57 and zsh 5.9 all run `gh pr xmerge 1` (a stub logging to a file — a stdout marker inside `$( … )` is captured and invisible). `gate_matches` says no match on `origin/main` and on the go-to-k/cdkd#3082 head. Two more records of the same class: `ZWNobyAiJChlY2hvICJhXCIpCiIgOyBnaCBwciB4bWVyZ2UgMQopIgo=` and `ZWNobyAiJChlY2hvICcpCicgOyBnaCBwciB4bWVyZ2UgMQopIgo=`.

**The class is bounded**: the realistic spellings, where no `)` sits inside the inner quoted span, are caught today.

## Fix shape

`.claude/hooks/lib/command-match.sh:491-500` falls through from the double-quoted state into the `$(` arm at `:548`, which pushes a frame without saving `q`. Save and reset the quote state per frame there, the way `last_heredoc_opener` does with `lho_OQ[]` — the precedent is in the same file, added for exactly this class of desync.

This is a SIBLING of go-to-k/cdkd#3258 but not the same fix: that one is the comment / frame-kind reading in `flush_line`'s quoted branch, and it does not touch `subst_open`'s quote stack.

Cases: the three records above (want MATCH), plus controls that a `)` inside a quoted span at top level and the already-caught no-`)` spellings keep their verdicts. Mutants beside `so-frame-glue`.

## Verification

```
bash -c 'source .claude/hooks/lib/command-match.sh; cmd=$(base64 -d <<<"ZWNobyAiJChlY2hvICdhXCdiJykKJyA7IGdoIHByIG1lcmdlIDEKKSIK"); gate_matches "$cmd" "$GATE_RE_GH_PR_MERGE" && echo MATCH || echo no'
```

prints `no` today and must print `MATCH`.

## The four TODO fields

Session-fit: now (this session) -- `subst_open`'s quote walk and the `lho_OQ[]` precedent it should follow are both in `.claude/hooks/lib/command-match.sh`, open in this session for twenty-five review rounds, and the corpora that measured this class are in its scratchpad. Worked as a follow-up PR after go-to-k/cdkd#3082 merges, with go-to-k/cdkd#3258.
Severity: medium -- it needs a `)` inside a single-quoted span inside a double-quoted substitution, which nothing here writes by accident; when it occurs every blocking gate is silent on a line all three shells run.
Effort: small (S) -- a per-frame save/restore of one variable; the differential cells are the cost.
Estimate: 2 hours -- the fence cells and the both-bash runner.
Dup-check: searched `is:issue subst_open quote`, `double-quoted substitution`, `quote reset frame` -- go-to-k/cdkd#3258 is the nearest and is the `)#` comment reading in a different function; nothing else open or closed.

go-to-k:
Closes #3040
Closes #3066

## What

Two gaps in `integ-local-gate`, found while merging go-to-k/cdkd#3053 (a cdk-local bump that reached `main` ungated), and the shared-matcher defect the same session hit:

1. **A cdk-local version bump now engages `integ-local`.** cdk-local IS the local-execution engine, and a bump moves what `cdkd local` does with zero lines under `src/local/**`. When no path matches, the gate reads the PR diff (`gh pr diff <N> --color never`; on the `git merge` branch `git diff --no-relative --no-color --no-ext-diff --src-prefix=a/ --dst-prefix=b/ HEAD...<ref>`) for a `-`/`+` pair on the `"cdk-local":` line inside a `package.json` block. Deliberately narrow: a lockfile-only re-resolve, an unrelated dep bump beside the line, and docs quoting the pair all stay out. `.markgate.yml`'s `integ-local.include` gains `package.json` so the bump also STALES the marker; `cross-cutting-list-sync.test.ts` gains an integ-local section fencing the two lists by mechanism. Every flag on those readers is pinned against a measured user-config fail-open: `diff.noprefix`, `color.ui=always` (colours a PIPED diff), `diff.relative=true` from a subdirectory (hides the root manifest), and gh's `--color auto` under `GH_FORCE_TTY`.

2. **A heredoc inside `$( )` no longer refuses prose.** `gh issue create --body "$(cat <<'EOF' … EOF)"` whose prose quoted `gh pr merge` was refused: the segmenter joins the lines of a still-open substitution BEFORE any heredoc is recognised, so the body arrived as `;`-separated commands. The join now latches onto a QUOTED delimiter and drops the body through its terminator.

Closes go-to-k/cdkd#3261.

## The latch took thirty-two review rounds, and the design is what came out of them

Each round measured, through real bash with a stub `git` on PATH, shapes bash EXECUTES a guarded verb for that the previous cut answered NO MATCH on — new fail-opens in the matcher every blocking gate sources. Read the commit messages for the per-round measurements; the design that survived:

- **Quoted-delimiter only, inside `$( )` only.** An unquoted `<<EOF` body is EXPANDED by bash, and two rounds each found a body-line shape a fall-through still dropped, so an unquoted opener anywhere in the substitution is a STICKY bail, and so is a frame that closes on the opener's own line, an unterminated `$((` / `${`, or a quote, backtick or backslash inside a `${…}` or `$((…))` span (round 15: a non-sticky return left the rest of that line unscanned, and a backtick opened after the `)` went unseen; round 16: `${a:-"}` hid an open quote). A BACKTICK frame is skipped textually — bash ends a backtick substitution at the next backtick on any line, quotes inside it protecting nothing (rounds 14 and 16) — so an opener inside one is never seen and its body is read as commands, exactly as `origin/main` read it: a false refusal at worst, never a miss. Round 15 traded the opener-time bail for the end-of-line check alone and a backtick closing after the opener on the same line latched (three reviewers, one finding); and the `)` closing a `$( )` ends a WORD, so `$(true)#"` glues the `#` where the bare `(true)#` is a comment (round 16, security) — as does a backslash-escaped character (`\)#"`, round 17), and `$(( ))` ends at its own closer, found by a paren walk rather than the first `))` (round 17). The TOP-LEVEL latch skips a `<<` sitting in a `#` comment, which reading the delimiter word properly had made reachable (round 18, security).
- **The scan's lexer state carries across the physical lines of one substitution** (quote, backtick, per-depth frame stack, the bail) and resets when it closes. A per-line scan lost what bash carries (an unquoted opener on line 1, a quote left open, a nested `$(` closed on line 2); the incremental scanner is linear and measures below `origin/main` at every size tried.
- **Bail on doubt**: an unbalanced quote at end of line, a new `$(` opened after the delimiter, or the opener's own frame closing on the same line (`y=$(cat <<'EOF') ; z=$(` — which shell reads the next line as body is version-dependent, so it is not modelled). `#` is a comment after an unescaped space or a `)` closing a bare `( )`, never after a character consumed as word glue (`$(x)#`, `<(x)#`, `\)#`), and an `a=( )` frame's `)#` is a BAIL because bash and zsh disagree about it. A body line that begins with the delimiter and carries a `)` ends the latch: bash 5 and 3.2 close BOTH the heredoc and the `$( )` there and run what follows — the join receives only the text after the delimiter, sliced from a left-trimmed copy so an escaped trailing space is not a continuation (rounds 9–11; and a `\`-continuation is recognised only on an ODD run of trailing backslashes, as bash does — round 12; bash 3.2 alone also closes on ANY `)` in a body — recorded in hooks.md as not modelled).
- **The delimiter is the whole word as bash reads it after quote removal** (`heredoc_word`: `<<'EOF'x` is `EOFx`, `<<E"O"F` is `EOF`, `<<'a"b'` keeps its inner quote, `<<"E\xF"` keeps its backslash, `<<\EOF` counts as quoted), in BOTH the `$( )` arm and the top-level one — the top-level arm had the same defects and the restore below made them reachable (round 7); it latches an unquoted word only when a whole identifier — `origin/main` latched the identifier PREFIX of any word that had one, a decoy that dropped an expanded body on a later bare `EOF` line (round 8). And `drain_extra` restores `pending_tag` around its body flushes — the latch had WIDENED go-to-k/cdkd#3066 (a body carrying a backtick used to keep the verb after the `)` matched on `origin/main`), and the restore closes the issue's own unquoted shape as well.
- Every arm of the scan reds its own case under deletion — the matrix lives in `command-match-mutants.sh` now (102 mutants, in their own four-shard CI job since the shared 60-minute one was cancelled with the harness still running; the script exits 0 only when every one fails a case — by name, timing cases excluded — that the unmutated run did not), not in prose. The differential fence prices the change as `SUBST_HEREDOC` (14 declared cells); ids 245–249 carry the round-3 and round-6 shapes.

## Verification

- `bash .claude/hooks/run-tests.sh`: every suite green under bash 5.x and 3.2, run alone (a run with `vp run lint` / `build` in parallel failed two LATENCY budgets under 3.2; the same case measures identically with `origin/main`'s library).
- `command-match.test.sh` 1188/1 — the one failure is `a CASE-variant spelling of THIS repo -> NOT foreign`, which go-to-k/cdkd#3387 owns: `gate_repo_slug` (main's code, untouched here) trims `.git` before it case-folds, so the case fails on any clone whose `origin` carries the suffix and passes in CI, where `actions/checkout` writes none. `command-match-mutants.test.sh` 21/0; `integ-local-gate.test.sh` 67/0; differential 29/0 over 265 inputs (SUBST_HEREDOC 14 declared cells, COMMENT_FRAME 10); `vp test run` green apart from the five over-cap paths on `main` that go-to-k/cdkd#3295 owns, against which this branch's delta is 0 B.
- **Where the two bashes read one line differently, this branch reads it as `origin/main` does and FILES the class.** An escaped character before a `#` inside an open `$( )` is word glue for bash 5.x and zsh and a comment for bash 3.2's `$( )` pre-scan, and the shells that RUN the two resulting shapes are disjoint. Round 30 tried to serve both readings and round 31 withdrew it: the marking it used is invisible to `main-tree-edit-gate`'s raw-`cd` union compensation (measured: a write into the MAIN tree from a feature worktree went rc 2 → 0), and a quote opened inside the comment region flips data and commands in a way no `cd`-context device can express (measured: `branch-gate` 2 → 0 on a shape bash 3.2 really pushes). go-to-k/cdkd#3303 carries all four shapes with their bytes, which shells execute each, both failed designs, and the fix shape that remains. What ships is the narrow half: the glue record is for an escaped `)` alone — the member round 23 added to `subst_open` and the one character all three shells agree glues (SO7 / CP-BS2 / SO9, mutants `so-bs-glue` / `so-bs-glue-wide`).
- A paren OUTSIDE every substitution carries a kind too, so the `)` of an `a=( )` glues its `#` while a bare `( )`'s still opens a comment — both bashes open that substitution and run its `cd` in the child (SO8 / SO-ctl3, mutant `so-depth0-frame`), and the differential fence gains the four shapes, two of them declared `COMMENT_FRAME` cells.
- Thirty-two review rounds (three 3-axis + security, one security-only, one security + code, then twenty-seven `/review-pr` 3-axis + security passes); every finding fixed here or filed (the seventeen issues listed below).
- Live, through the REAL hooks with PreToolUse payloads: the original false positive passes; a real `gh pr merge 3053` refuses through a real `gh pr diff`; `branch-gate` on a `main` cwd refuses every round-3/4 shape and passes quoted prose.
- Not run: `.claude/hooks/lib/false-refusal-survey.sh` (a 90-minute measurement, not in `run-tests.sh` or CI).
- Two rule-file splits fund this branch's own prose rather than trimming anyone's text: the heredoc-latch detail to `hooks-command-match-heredoc.md` (matcher + its three suites) and the ten one-shot foot-gun gates to `hooks-foot-gun-gates.md` (those gates + their suites), each behind a pointer, with `CORPUS_BYTES_MIN` re-derived 966,000 → 1,010,000 and a representative budget row for the new roster. The only payload failure left is the five over-cap paths on `main` that go-to-k/cdkd#3295 owns; this branch's delta against each is 0 B.

## Out of scope, recorded

- go-to-k/cdkd#3081 — `close_paren` alone still does not read a `)#` as a comment (`subst_open` gained it here, with the frame kind); a `cd` inside such a `$( )` is emitted top-level. Pre-existing, and the issue now carries the narrowed fix shape: carry the frame KIND, do not just add the class member.
- go-to-k/cdkd#3099 — `strip_noncommand_spans` (the other awk program) still reads a heredoc delimiter the old way; reach is two bypass-token detectors and a non-blocking detector.
- go-to-k/cdkd#3132 — `run()`'s continuation arm still joins a `\` that ends a `#` comment or precedes a CR, and refuses an even run inside backticks; pre-existing on `origin/main`, the same arm round 12 touched.
- go-to-k/cdkd#3204 — a leading redirection (`>/dev/null git commit`) defeats every blocking gate: `gate_strip_prefix` strips no redirection. Pre-existing, measured identical on `origin/main`.
- go-to-k/cdkd#3205 — a newline inside a `bash -c` string is joined away, so the second line's verb is never matched. Pre-existing, same measurement.
- go-to-k/cdkd#3219 — a MALFORMED opener (`x=$(a=(cat <<'E'`) is a syntax error whose recovery runs the following lines, while the latch drops them as body. The latch's safety argument holds for a well-formed opener only.
- go-to-k/cdkd#3228 — a `#` comment inside a multi-line `$( )` swallows the next command line: the join replaces the newline that ends it. Pre-existing, all three shells run the verb.
- go-to-k/cdkd#3258 — a `)#` inside a DOUBLE-QUOTED multi-line substitution comments out the closer, so bash 5.x and zsh run the next line unmatched. Pre-existing.
- go-to-k/cdkd#3259 — `run()`'s `$( )` join is O(n²); ~800 body lines cross the 10 s PreToolUse timeout, and a killed hook disarms every gate. Pre-existing (9.92 s on `origin/main` at the same point).
- go-to-k/cdkd#3260 — a `case` arm written `(y)` is not a command position; it is the only spelling bash 3.2 parses inside `$( )`. Pre-existing.
- go-to-k/cdkd#3262 — a bare `(( ))` is not skipped, so `1<<X` reads as a heredoc opener. Pre-existing.
- go-to-k/cdkd#3263 — `/bin/sh -c`, `bash -c --`, a function body and `eval` all put a guarded verb past every gate. Pre-existing.
- go-to-k/cdkd#3267 — the differential fence has no detector for a DEAD allow-row, which pre-authorises a lost match. Pre-existing.
- go-to-k/cdkd#3299 — a `"` inside a single-quoted span inside `"$( )"` carries the quote across the line end; what is LEFT of the class go-to-k/cdkd#3261 was filed for, which this PR closes. Pre-existing.
- go-to-k/cdkd#3300 — `close_paren` and `flush_line` still model bash alone for the PID-then-quote shape, and miss an unbalanced quote in a backtick frame. Pre-existing.
- go-to-k/cdkd#3387 — `gate_repo_slug` trims `.git` before it case-folds, so a case-variant remote spelling makes a checkout of THIS repo answer FOREIGN and `integ-schema-migration-gate` exit 0. Main's code, untouched here; it is why the matcher suite reports one failure on a `.git`-suffixed clone and none in CI. Pre-existing.
- go-to-k/cdkd#3228 gained the continuation-line route into the `#`-inside-`$( )` class (`x=$(echo \` / `#b"` / a verb), measured rc=0 on this head AND on `origin/main`. Pre-existing.
- go-to-k/cdkd#3303 — an escaped character before a `#` inside an open `$( )` reads two ways, and each reading leaves fail-opens the other closes; four shapes with their bytes, the two designs measured to open new holes, and the forking-segmenter fix shape that remains. Pre-existing (this branch reads the class exactly as `origin/main` does).
- Won't-do: in `cross-cutting-list-sync.test.ts`, removing `package.json` from the include list reds the include-count floor before the IFF case's message; both red, only the first message differs.

https://claude.ai/code/session_01RRGFrvqtpHvaqnmij7jkdQ


go-to-k:
Merged. Thirty-two review rounds, and the reviewers earned every one of them — so a note on what actually worked, since the shape is reusable.

**The measurements outranked the reasoning, repeatedly.** Round 30 built a device to serve both shell readings of an escaped character before a `#`; round 31's security pass measured two fail-opens the device itself opened, and it was withdrawn rather than patched. The class is filed as go-to-k/cdkd#3303 with all four shapes' bytes, both failed designs, and what remains — which is worth more to the next attempt than a half-closed hole would have been.

**Two of the last round's blockers were mine, not the code's.** `git rebase -X ours` reads `ours` as the UPSTREAM side, so the rebase that flattened this branch silently discarded every edit to the differential fence — including the pricing go-to-k/cdkd#3066's own fix shape requires. Neither the suite nor CI could see it; the test and spec reviews found it independently by measuring the fence instead of reading the claim about it. That is exactly the failure mode this PR spent thirty rounds closing in the matcher, arriving through the tooling.

**The instrument needed an instrument.** The mutation harness gained CI sharding, and the shard selector promptly shipped a silent full run (`0/08` aborts a bash `if` compound as an octal literal and exits 0). It has its own suite now, asserting the property no single shard can observe — exact cover.

Net effect, measured on a 46,298-record corpus against `origin/main`: 870 fail-opens closed, 0 opened, with the false-refusal cost priced too. Seventeen residuals are filed rather than waved past.

PR #3086, opened 2026-09-14
fix(secretsmanager): honour IncludeSpace and RequireEachIncludedType and match the service's password recipe

### LINKED ISSUE #3068: SecretsManager: GenerateSecretString.IncludeSpace and RequireEachIncludedType are declared members the local generator never reads

go-to-k:
Closes #3068

## What

The `GenerateSecretString` password is minted LOCALLY (never read back -- the premise the go-to-k/cdkd#2212 fence protects), so the recipe is whatever cdkd implements, and it diverged from Secrets Manager's `GetRandomPassword` four ways:

- `IncludeSpace` and `RequireEachIncludedType` were declared members the generator never read -- and the latter's service default is TRUE, so a short password could miss a class the service guarantees;
- the punctuation class was a 26-character subset of the service's 32, so an `ExcludeCharacters` naming `"` `'` `/` `\` backtick `~` was inert here while it meant something to the service;
- a fully excluded charset fell back to lowercase the template had excluded.

Now one module-level `generateCharset` serves the member predicate and the generator: the four classes minus `ExcludeCharacters`, the 32-character service punctuation set, the space under `IncludeSpace` (taken back by `ExcludeCharacters`), and the REQUIRED classes under `RequireEachIncludedType`. The guarantee is satisfied by PLACING one character of each required class at a random distinct position, so it holds per draw at any length; every draw is uniform (rejection sampling replaces `byte % n`).

## Measured on CloudFormation (us-east-1, 2026-09-14), not read off the docs

- `PasswordLength: 2` with the four default classes required: refused, `Password length is too short based on the required types`.
- A REQUIRED class emptied by `ExcludeCharacters`: refused, `All characters of the desired type have been excluded`; the same template with `RequireEachIncludedType: false` reaches CREATE_COMPLETE, so the rule is scoped to required classes.
- Every class switched off: refused, `All characters have been excluded from selection`.

cdkd refuses the same three shapes (throw on create, the usual SKIP with the previous block retained on update); a template the service accepts is never refused.

## Also found while probing

Removing the too-short refusal in a mutation probe made `randomIndex(0)` spin forever (`limit` is `NaN`) and hung the test process; `randomIndex` now refuses an empty range so that mutant fails in 5 ms instead.

## Verification

- Unit: 999 files / 22219 tests green. Refusal rows for the five new shapes; the 32-character set (every member present over 2000 draws); `ExcludeCharacters` stripping fenced per class (the six formerly inert characters, half the digits, a required class reduced to one letter); `IncludeSpace` admitted and taken back; the per-draw class guarantee at length 4 over 200 draws and its absence under `false`; a non-required emptied class (measured CREATE_COMPLETE); deterministic fences for the rejection sampling and the placement step by stubbing `crypto.getRandomValues`. Eleven single-mutation probes red one case each.
- Review: code + test + security (fable) round 1, code closing round 2 -- no blocker, every should-fix applied.
- Live: `secretsmanager-update-value-source` PASS (74 s) on the lane, destroy 0 errors / 0 orphans / 0 surviving state versions -- the new recipe's output accepted by the live service.

## Fence note

The go-to-k/cdkd#2212 fence's two allow-lists gain the two new free functions (`generateCharset`, `randomIndex`). The punctuation backtick is built from its char code because that fence's string stripper takes template literals first and prettier rewrites a `\u` escape back into the raw character.


go-to-k:
Merged as a4780526. Thanks to the reviewers -- the round-1 catch that the six formerly inert characters were proven PRESENT but never proven EXCLUDABLE is the one that made this fix the issue's headline, and the deterministic placement fence closed a 1.6%-per-run survivor the probabilistic rows would have let through.

Issue #3097, opened 2026-09-14
fix(ec2): a SecurityGroup declared without VpcId records the attribute as '' and the resolver arm answers undefined instead of reading the group's VPC


go-to-k:
`EC2Provider`'s two `AWS::EC2::SecurityGroup` result maps (`src/provisioning/providers/ec2-provider.ts`, the create and update `attributes` literals) record `VpcId: (properties['VpcId'] as string) ?? ''` — the TEMPLATE's own property, not a read-back — so a group declared without `VpcId` (a hand-written L1 landing in the default VPC) records `VpcId: ''`, and `Fn::GetAtt [SG, VpcId]` resolves to the empty string where CloudFormation returns the default VPC's id. They are the two sites go-to-k/cdkd#3077 deliberately left on the old shape, and the reason is the second half of this issue: the resolver's `AWS::EC2::SecurityGroup` arm in `src/deployment/intrinsic-function-resolver.ts` answers `VpcId` with `return undefined; // Would need API call`, and `resolveJoin` runs `String(part)` over every part, so merely omitting the key moves the value from `''` to the literal `'undefined'` inside any `Fn::Join` / `Fn::Sub` — worse, not better. The fix is the API call that comment defers: read the group's `VpcId` live (`DescribeSecurityGroups` on the physical id, cached per deploy like the EC2 Instance arm), record it from the create response rather than from the template (`CreateSecurityGroup` returns the group id; a follow-up `DescribeSecurityGroups` returns `VpcId` for both the explicit and the default-VPC case), and then retire the two allow-list entries in `tests/unit/provisioning/attribute-map.test.ts`, whose third case fails until they are removed.

Dup-check: searched open issues for `SecurityGroup VpcId GetAtt` and the body window for `SecurityGroup.*VpcId` -- none covers this root cause
Session-fit: now (do it in this session) -- re-classified at claim: both files the fix touches were read in the go-to-k/cdkd#3103 session, and the default-VPC control the fixture arm needs exists in the account; verify with `/run-integ ec2-instance` (its `VpcId` Output) after adding a default-VPC security group to that fixture
Severity: low -- reachable only by a template that omits `VpcId` on a security group AND reads the attribute back; CDK's L2 always emits `VpcId`, so the recorded value is the correct one for every CDK-authored stack
Effort: small (S) -- one describe call at two provider sites plus a resolver arm, riding the fixture the resolver follow-up already needs
[...]

Issue #3096, opened 2026-09-14
fix(resolver): the EC2 Instance live-lookup arm returns the instance id for a still-unassigned IP / DNS / AZ attribute instead of refusing


go-to-k:
`IntrinsicFunctionResolver.constructAttribute`'s `AWS::EC2::Instance` arm (`src/deployment/intrinsic-function-resolver.ts`, the `PrivateIp` / `PublicIp` / `PrivateDnsName` / `PublicDnsName` / `AvailabilityZone` cases) runs a live `DescribeInstances` because, as its own comment says, the physical id is the WRONG value for those attributes (an ELBv2 IP target group rejects `i-...`). Yet when the live read finds the field still unassigned — an instance still `pending` under `--no-wait` — the arm warns `returned no <attr>; returning physical ID` and returns the instance id anyway, and the same fallback covers a `DescribeInstances` failure. So a same-deploy `Fn::GetAtt` over such an attribute (a downstream property, a stack Output, an exported alias) is fed `i-xxxx` in the position of an IP address or a DNS name, loud only where AWS rejects it. Issue go-to-k/cdkd#3077 stopped the providers RECORDING `''` for those fields, which is what routes this case to the live arm at all; it deliberately left the arm alone, so the residual is that a reference to a not-yet-assigned address degrades to the instance id instead of refusing. The fix is a refusal (an `IntrinsicResolutionRefusalError`, `markNonRetryable` per the class's rule in `.claude/rules/architecture.md`) or a deferred re-read, for exactly the five names the arm already enumerates — never for the `InstanceId` case, where the physical id IS the answer. Evidence: `tests/unit/deployment/intrinsic-functions.test.ts`, the case `returns the physical id, and caches NOTHING, while the live read still finds the attribute unassigned`, which pins today's shape so it is visible.

Dup-check: searched open issues for `returning physical ID`, `physical id fallback GetAtt`, `cachedEc2InstanceAttributes` (title search and the body window) -- none covers this root cause; go-to-k/cdkd#1852 is the neighbouring stale-record case and does not name this arm
Session-fit: now (do it in this session) -- re-classified at claim: the `--no-wait` phase that catches the instance while `pending` now exists in `tests/integration/ec2-instance/verify.sh` (go-to-k/cdkd#3103), and the resolver arm was read in that PR's review; verify with `/run-integ ec2-instance`
[...]

Issue #1852, opened 2026-08-13
fix(deployment): an issue-1824-era stale attribute map is never healed, and the resolver "not enriched for this resource type" message is now false for the two fixed types


go-to-k:
Issue https://github.com/go-to-k/cdkd/issues/1824 made
`AWS::RDS::DBSubnetGroup.DBSubnetGroupArn` and `AWS::SSM::Parameter.Arn` resolve
for NEW deploys, by caching each ARN in the provider's create / update / import
attribute map. It does NOT heal a resource that a PRE-FIX binary already
deployed, so the exact scenario issue 1824 reports still hard-fails after an upgrade —
and the resolver's error message now tells the user something false.

## Reproduction (the scenario issue 1824 opens with)

1. Deploy a stack containing an `AWS::SSM::Parameter` (or an
   `AWS::RDS::DBSubnetGroup`) with any cdkd release BEFORE the issue 1824 fix. The
   state record's `attributes` is `{Type, Value}` — no `Arn`.
2. Upgrade cdkd to a release carrying the issue 1824 fix.
3. Add the output the issue names, changing nothing else:

   ```ts
   new cdk.CfnOutput(this, 'ParamArn', { value: param.attrArn });
   ```

4. `cdkd deploy`.

The deploy FAILS (or, for an Output, exits non-zero under `--strict-getatt` /
warns and keeps the previously persisted outputs) on the resolver's `*Arn`
shape guard, because `constructAttribute` reads the CACHED
`resource.attributes['Arn']` and the record still has none.

## Why the upgrade does not heal it

`src/deployment/deploy-engine.ts:2829` skips a resource whose resolved
properties equal the state record with NO provider call at all:

```ts
if (JSON.stringify(resolvedProps) === JSON.stringify(currentProps)) {
  ...
  this.logger.debug(`Skipping ${logicalId}: no actual changes after intrinsic function resolution`);
  if (counts) counts.skipped++;
  break;
}
```

Adding an Output changes no resource property, so the parameter / subnet group
is skipped, `update()` never runs, and the attribute map is never re-recorded.
The record heals only when the user happens to change one of that resource's
properties — which is not something the failing template asks them to do, and
not something the error message tells them to do.

## ...and the error message is now FALSE for these two types

`src/deployment/intrinsic-function-resolver.ts:2585`
(`guardedPhysicalIdFallback`) says:

> `attributes are not enriched for this resource type, and the physical ID
[...]

PR #3103, opened 2026-09-14
fix(providers): omit an attribute the read-back left unassigned instead of recording ''

### LINKED ISSUE #3077: fix(providers): --no-wait records an unassigned EC2 attribute as '' rather than omitting it, so the empty value persists, exports, and never self-heals
## What happens

Under `--no-wait`, `EC2InstanceProvider.create` skips the `running` wait and then describes the instance while it can still be `pending`. Every IP / DNS field falls back to the empty string:

```ts
const attributes: Record<string, unknown> = {
  InstanceId: instanceId,
  PrivateIp: runningInstance?.PrivateIpAddress ?? '',
  PublicIp: runningInstance?.PublicIpAddress ?? '',
  PrivateDnsName: runningInstance?.PrivateDnsName ?? '',
  PublicDnsName: runningInstance?.PublicDnsName ?? '',
  AvailabilityZone: runningInstance?.Placement?.AvailabilityZone ?? '',
};
```

(`src/provisioning/providers/ec2-provider.ts`, in the block whose own comment calls this the accepted `--no-wait` trade-off.)

The fallback converts UNKNOWN into KNOWN-EMPTY, and three separate consumers key off exactly that distinction, so all three take the wrong branch:

1. **State.** The empty string is persisted as the attribute value, and an Output over it is persisted as `''`.
2. **Exports.** The export pass skips only an output whose value is `undefined` (`if (value === undefined) continue;` in `deploy-engine.ts`), so `''` is published and a downstream stack's `Fn::ImportValue` resolves to an empty string. A consumer that interpolates it into an endpoint or a security-group rule gets a silently wrong value, not a failure.
3. **No self-healing.** The intrinsic resolver prefers the cached state attribute and gates that on `flatValue !== undefined`, so `''` wins on every later resolution. No ordinary later deploy re-reads the field from AWS: the empty value stands until the instance is updated or replaced.

## Why omitting the key is the fix

Leaving an unassigned attribute OUT of the map (rather than writing `''`) makes all three consumers behave correctly with no new mechanism and no added wait:

- the export is skipped instead of publishing an empty alias;
- the resolver's cache lookup misses, so the value is resolved again on a later deploy and heals itself;
- a template that genuinely needs the value fails loudly at resolution time instead of deploying an empty string.

This is deliberately NOT a proposal to wait when a skipped resource is referenced. That rule is too coarse to be useful: references to attributes that ARE final in the create response dominate real stacks (a route table referencing `NatGatewayId`, a listener referencing a `ServiceArn`), so a resource-granularity "wait if referenced" would reintroduce the NAT / RDS / CloudFront waits `--no-wait` exists to skip and collapse the flag into the default mode. An attribute-granularity refinement (the analyzer already extracts the `(logicalId, attributeName)` pairs to build implicit edges, so it could refuse or wait only for an attribute that is not final before stabilization) is a separate, larger question and should not block this.

## Scope to check while implementing

- Whether any reader assumes these keys are always present on an EC2 `ResourceState.attributes` (drift comparison, `cdkd export`, `cdkd state show`, the events payload).
- Whether sibling `--no-wait`-gated providers have the same `?? ''` shape (NAT Gateway, ELBv2, RDS) — the fix should be consistent across them rather than EC2-only.
- A unit test pinning that an unassigned field is ABSENT from the recorded attributes, and one pinning that the export pass then publishes no alias for an Output over it.

## Current documented behaviour

`docs/cli-deploy.md` now describes the empty-string outcome accurately under "What a second deploy started too early runs into" (PR #3076), so this issue is about changing the behaviour, not about a docs gap.

Dup-check: searched open and closed issues for `--no-wait`, EC2 attribute and empty-string wording. #1277 is the related but distinct fix that gated the `running` wait on `CDKD_NO_WAIT` in the first place; it introduced the fallback and did not revisit what gets recorded. No issue covers the recording behaviour.

Session-fit: next (not this session)
Severity: high — a downstream stack's `Fn::ImportValue` resolves an IP / DNS value to an empty string and deploys with it, and state keeps a value that is not true of AWS; the wrong result never heals on its own, so this is a wrong result rather than a missing capability even though it is scoped to `--no-wait` runs.
Effort: small (S) — the change itself is dropping a fallback, but it touches the state / export / resolver boundary, so the sibling providers and the present-key assumptions have to be swept before it is safe.
Estimate: 1-2 hours — most of it the consumer sweep and the two pinning tests, not the provider edit.

## Review

Both halves get a review pass on Fable: the DESIGN decision (omit the key vs keep a sentinel vs the attribute-granularity wait) before any code is written, and the IMPLEMENTATION once the sweep above is done. Dispatch the reviewers with the Fable model rather than the session default, since the value at stake here is a cross-boundary invariant (`undefined` means "not known", `''` means "known to be empty") that three separate consumers depend on, and a plausible-looking wrong call propagates to state, exports and every later resolution.

go-to-k:
Closes #3077

## What changes

An attribute a provider could not read back is now **absent** from the recorded `attributes` map instead of being recorded as `''`. One new leaf helper spells the rule, `src/provisioning/attribute-map.ts` (`definedAttributes`, dropping `undefined` / `null` — an empty string AWS itself reported is a KNOWN value and is kept; `stringifyIfAssigned`, so a numeric `Endpoint.Port` is never recorded as the literal `'undefined'`), and every attribute-position `?? ''` / `|| ''` reachable from an `attributes` value in `src/provisioning/providers/**` is routed through it:

- `ec2-provider.ts` — `AWS::EC2::Instance` create and update, through one state-gated builder `describedInstanceAttributes`: while the instance is `pending` (a `--no-wait` create) an empty or absent `PublicIp` / `PublicDnsName` is UNKNOWN and omitted (the API reports `PublicDnsName: ''` in that window); on a settled instance the public pair is recorded as what AWS reports, `''` included — a running private-subnet instance gets `''` for both, CloudFormation's own answer, so a reference does not degrade to the instance id on every resolution (round-1 review, major). The private address, private DNS name and AZ are assigned at launch and are recorded whenever present.
- `rds-provider.ts`, `docdb-provider.ts`, `neptune-provider.ts` — cluster and instance, create and update (`Endpoint.Address` / `Endpoint.Port` / `ReadEndpoint.Address` / `Arn` / `DBClusterResourceId` / `ClusterResourceId`)
- `ec2-provider.ts` VPC create: `DefaultSecurityGroup` was a `let … || ''` and `DefaultNetworkAcl: ''` a permanent placeholder nothing reads back (round-2 review) — the first now omits, the second is not recorded
- `ecr-provider.ts` (create + update), `rds-dbproxy-provider.ts`, `rds-dbproxy-endpoint-provider.ts`, the four `agentcore-*-provider.ts` files (create / update / import), and the sites the first fence could not see (round-1 review): `elasticache-provider.ts` (builder element writes, create + update), `codecommit-repository-provider.ts` (`toAttributes` helper), `servicediscovery-provider.ts` (`|| ''`), `appsync-provider.ts` / `cloudfront-distribution-provider.ts` / `eventbridge-bus-provider.ts` / `s3-vectors-provider.ts` (a hoisted `const x = … ?? ''` one hop above the literal)

The population was derived with a TypeScript-compiler walk from every `attributes` value, following the literal, a same-function `const` / `let` an identifier resolves to and its reassignments, builder writes onto it, a same-file helper's `return`s, the arguments of a call at those positions, and a spread — **96 sites in 18 files on `origin/main`** (the first, literal-only cut saw 71 / 11; round 1 found four more shapes, round 2 three more — a hoisted `const` passed INTO `definedAttributes({ K: x })`, a `let x = ''` reassigned with `|| ''`, and a bare `DefaultNetworkAcl: ''` — and the maintainer round the bare-`''` positions beyond a property value: a `const` / `let` initializer, an `=` right-hand side, a `?:` branch, plus the empty-template-literal fallback and a method call's receiver). The same walk is `tests/unit/provisioning/attribute-map.test.ts`, a source-shape fence with a floor per followed shape AND a fixture-driven probe (a fake provider under a temp dir, one fixture per shape) that reds every shape the fence claims to see and pins the declared misses. Its documented reach: a call nested inside a property value's expression is not resolved to its callee's body (a fallback spelled inline there IS seen), a per-member arrow helper is not followed (`publicMember` is that shape), and a helper is resolved by bare name (first same-named method in the file). DLM's `arn = await this.fetchPolicyArn()` already omits by truthiness and is not a member. The rule itself was already prose in `docs/provider-development.md` ("Never store an empty-string placeholder") and `docs/state-management.md`; this PR makes the providers obey it.

## Why omitting, and what a reader sees instead

Three consumers key off `undefined`-vs-present, and `''` put all three on the wrong branch: state persisted it; the Outputs pass exported it to `Fn::ImportValue` consumers; the resolver served it forever, because a stored attribute is a hit whenever it is not `undefined`, so the `AWS::EC2::Instance` live `DescribeInstances` arm (and the RDS-family rebuilt-`Arn` arm) never ran.

Sibling sweep (issue scope item b): NAT Gateway (create records only `NatGatewayId`) and ELBv2 (`?.replace(...)` yields `undefined`, never `''`) were swept and had ZERO attribute-position sites, as did ACM and Lambda MicrovmImage.

Reader sweep (issue scope item a): no reader outside the resolver assumes these keys are present — `drift.ts` passes `res.attributes ?? {}`, `state.ts` prints `Object.entries(attributes ?? {})`, `export.ts` reads only `Id` / `*Arn` by `typeof`, `orphan-rewriter` / `local-*` guard by `hasOwn`; the Cloud Control enrichment already guards every string by truthiness.

Design review (Fable, before any code) settled: omit the key (A) over a sentinel (B) or an attribute-granularity wait (C); sweep the RDS family in the same change; keep the `update()` map REPLACING the record (a public IP the describe no longer reports is gone from AWS — a stop, an EIP change — and a merge would resurrect a dead address); and leave `intrinsic-function-resolver.ts` / `deploy-engine.ts` untouched.

One claim in the issue body does **not** hold and is corrected here rather than restated: "a template that genuinely needs the value fails loudly" and "the export pass then publishes no alias". An `Fn::GetAtt` over an absent key never resolves to `undefined`. For EC2 the live arm re-describes the instance: the real address once it is `running` (the Outputs pass runs after the DAG, so this is a same-deploy improvement — with `''` the Output was empty even when AWS had the address by then); while it is still `pending`, the instance id with a warning and NOTHING cached, so the next resolution re-reads. For the RDS family there is no live arm: `Endpoint.*` resolves to the instance identifier with a warning until the next `update()` of that resource re-records it — a no-change deploy does not (that is the existing heal-on-no-change root cause, folded into go-to-k/cdkd#1852). Both residuals are documented in `docs/cli-deploy.md` and pinned by tests so their shape is visible.

## Follow-ups filed

- go-to-k/cdkd#3096 — the EC2 live arm should refuse (or defer) rather than return the instance id for a still-unassigned IP / DNS / AZ attribute; contested resolver file, needs its own live arm. Round 3 folded a second row into it: the VPC `DefaultSecurityGroup` arm answers the VPC id with NO warning when the key is absent (a failed post-create `DescribeSecurityGroups`), the same silent physical-id class. **Re-classified `Session-fit: now`** after this PR's live test: the `--no-wait` phase it said had to be written now exists in `tests/integration/ec2-instance/verify.sh`, and the resolver arm was read in this PR's review — taken in the same session, as its own PR, right after this merge.
- go-to-k/cdkd#3097 — the two `AWS::EC2::SecurityGroup` `VpcId` sites are deliberately NOT converted (allow-listed in the fence): the value is the template's own property, not a read-back, and the resolver arm answers `undefined`, which `Fn::Join` stringifies to `'undefined'` — omitting there is worse than `''` until the arm reads the VPC live. **Re-classified `Session-fit: now`** for the same reason (both files it touches were read in this session; the default VPC its control needs exists in the account) — taken after go-to-k/cdkd#3096.
- go-to-k/cdkd#1852 (existing) — one checklist row: an RDS-family `Endpoint.*` omitted under `--no-wait` heals only on `update()`.

## Tests

- `tests/unit/provisioning/ec2-instance-no-wait-gate.test.ts` — a direct state table over `describedInstanceAttributes` (every settled state → `''` for the public pair; `pending` / no `State` → omitted; a reported public address kept whatever the state), plus realistic `DescribeInstances` shapes (private members present while `pending`, `PublicDnsName: ''`): pending create omits the public pair and keeps the launch-time members; running control records all six; running private-subnet create records `PublicIp: ''` / `PublicDnsName: ''`; `update()` on running / stopped (known-empty `''`), pending (omitted), and a State-less read-back (`InstanceId` only).
- `tests/unit/provisioning/rds-provider.test.ts` — DBInstance create under `--no-wait` while `creating` → `{ Arn }` only; DBInstance update with no `Endpoint`; the assigned-Port control (`'5432'`); DBCluster create AND update when the describe returns no cluster → `{}`. DocDB and Neptune: one DBInstance create case each, plus cluster create AND update with a Port-less describe (pins `Endpoint.Port` never becoming `'undefined'` at all four cluster sites). ECR `update()` (no repository → `{}`, plus the control), AgentCore runtime `import()` (no `agentRuntimeName` → key absent), ElastiCache `update()` (Redis endpoint without a port) and create + update (memcached `ConfigurationEndpoint` without a port), CodeCommit create (metadata without a KMS key).
- `tests/unit/deployment/intrinsic-functions.test.ts` — the consumer half: a stored `''` is served WITHOUT re-describing (the shadowing the providers must not manufacture); an absent key re-reads AWS even when the record holds other attributes; a still-unassigned live read returns the physical id and caches nothing (the go-to-k/cdkd#3096 residual, pinned).
- `tests/unit/provisioning/attribute-map.test.ts` — helper semantics (`''` kept, `undefined` / `null` dropped) plus the source-shape fence (floors ~12% under the measured 376 attribute values / 220 hoisted resolutions / 40 helper calls / 86 builder writes).

Mutation probes, each restored byte-exact (`cp` snapshot back, `git diff HEAD` empty):

| mutant | red |
| --- | --- |
| `?? ''` re-added INSIDE `definedAttributes(...)` at the EC2 create block | fence only — the helper drops the `''`, which is the property the module buys |
| helper BYPASSED at the EC2 create block (pre-fix shape) | `omits every unassigned address…` + fence |
| helper bypassed at the EC2 update block | both `update()` cases |
| helper bypassed at the RDS DBInstance create block (`String(x ?? '')`) | `DBInstance create under --no-wait…` + fence |
| identity `definedAttributes` (drops nothing) | every provider case + the helper case — after switching the whole-map assertions to `toStrictEqual`; under `toEqual` the RDS / DocDB / Neptune cases and one EC2 case stayed GREEN, because `toEqual` ignores a present-but-`undefined` key |
| round 2: `settled` forced true (pending treated as settled) | the pending create case + the pending update case |
| round 2: `settled` forced false (running treated as pending) | the running-private create case + the running / stopped update cases |
| round 2: one hoisted `?? ''` restored (eventbridge-bus), one builder write (elasticache), the helper body (codecommit), the `\|\| ''` site (servicediscovery) | the fence, each time |
| round 3: `?? ''` restored in a followed call's ARGUMENT (agentcore-runtime import), the VPC `let … \|\| ''` reassignment, a bare `DefaultNetworkAcl: ''`, and a `?? ''` reachable only through `{ ...this.toAttributes(m) }` | the fence, each time (the spread needed the depth cap raised from 2 to 4 — measured green at 2) |
| round 4: VPC `DefaultSecurityGroup: defaultSgId ?? 'sg-unknown'` and `String(defaultSgId)` (both fence-INVISIBLE shapes), bare `DefaultNetworkAcl: ''` restored | the new VPC behaviour cases in `ec2-provider-partial-create-cleanup.test.ts` |
| round 4: `DEPTH_DROPPED_CAP` 25 → 17 (measured drops: 18) | the parser-floor case — the depth cap is no longer a silent drop |
| maintainer round: `stringifyIfAssigned` → `String(x)` at each DocDB / Neptune CLUSTER site (4 probes, one site each) | the matching cluster create / update case, each time |
| maintainer round: the `!== undefined` guard dropped at each ElastiCache `ConfigurationEndpoint.Port` site (2 probes) | the matching create / update case |
| maintainer round, the fence's own recognisers reverted one at a time: `isManufacturedEmpty` back to property-only / template-literal fallback unrecognised / receiver not scanned / allow-list back to `some()` | 5 / 1 / 1 / 1 fixture cases red |
| maintainer round, real code: the pre-fix VPC `let defaultSgId = ''` restored (no `\|\| ''`) | the fence (`ec2-provider.ts:925: ''`) |
| `stringifyIfAssigned` stringifying unconditionally (`'undefined'`) | all five RDS-family cases + the helper case |

## Live test (real AWS, us-east-1, 2026-09-14)

`tests/integration/ec2-instance/verify.sh` gained the arm the fixed path needs (no fixture entered the `--no-wait` pending window before): **Phase 0** deploys with `--no-wait` and asserts, from `state.json`, that (a) no `AWS::EC2::Instance` record carries an empty-string attribute and at least one instance took the pending path (`PublicIp` / `PublicDnsName` absent — both did), (b) the launch-time members `PrivateIp` / `AvailabilityZone` / `InstanceId` are still recorded (the negative control against an over-broad fix), and (c) the `PublicIp` Output resolved to something other than `''` (the live arm answered the real address); it then waits for `running`, and the existing default deploy becomes **Phase 1**, the no-change second deploy whose Outputs pass re-resolves the Output — asserted equal to the live `PublicIpAddress`. `docs/cli-flag-coverage.md` gains the `ec2-instance` row under `--no-wait`.

| run | result |
| --- | --- |
| `ec2-instance` (fix tree) | PASS, 169 s — Phase 0 omission observed on both instances, Output `44.213.110.24` from the live arm, Phase 1 healed to the same address; destroy 0 errors, 0 orphans |
| `ec2-instance` (MUTANT: pending path restored to `value ?? ''`) | **red at Phase 0 (a)**: `recorded an empty-string attribute (PublicInstance.PublicIp PublicInstance.PublicDnsName Instance.PublicIp Instance.PublicDnsName)`; the trap destroyed the stack (10 deleted, 0 errors); provider restored byte-exact, dist rebuilt |
| `agentcore-tools` | PASS, 113 s, 5 deleted 0 errors 0 orphans |
| `rds-full-stack` | PASS, 742 s — the `GetAtt`-computed endpoint walks the RDS `Endpoint.*` sites through `stringifyIfAssigned` live; 16 deleted 0 errors 0 orphans |
| `docdb-neptune` | PASS, 1545 s — cluster + instance `Endpoint.*` sites live; 14 deleted 0 errors 0 orphans |
| `rds-aurora` | PASS, 2299 s — DBCluster attributes live; 23 deleted 0 errors 0 orphans |

Account-wide sweep after the runs: 0 `state.json` records, 0 RDS / DocDB / Neptune clusters or instances, no fixture VPCs. `integ-destroy` set from these runs; neither `integ-broad` nor `integ-local` is in scope (no cross-cutting or local file in the diff).

## Docs

`docs/cli-deploy.md` (wait-semantics row + "What a second deploy started too early runs into" — now states which members can be unassigned and that a running private-subnet instance records `''`), `docs/provider-development.md` (names the helper, the `''` rule and the fence's reach), `.claude/rules/layout-provisioning.md` (module entry), `.claude/rules/provider-replay-and-refusals.md` (the replace-vs-heal decision), `changelog.d/entries/2026-09-14-3077-omit-unassigned-attribute.md`.

## Review rounds

Maintainer delta round (code + test on Fable, @50b853b8): no blockers. Code — `isManufacturedEmpty` verified not to over-match legitimate `''` (16 probe shapes; two contrived accumulator / `JSON.stringify` shapes red and now named in the reach paragraph), three spellings neither followed nor declared (`const { arn = '' } = resp`, `??=` / `||=`, `Object.assign`) plus the present-but-`undefined` shape are now DECLARED non-reaches with `hit: null` fixture rows (a widening goes red), two stale measurement comments re-measured (`hoistedResolved` 220, `depthDropped` 19). Test — the round-1 gaps confirmed closed with 12 mutation receipts (each cluster / ElastiCache site red only on its own case; allow-list `some()` red on the third-copy case), the declared-miss fixtures pin the miss in the widening direction. Final touch: the helper-semantics `toEqual({})` became `toStrictEqual({})` (an identity helper passed the former).

Maintainer round (independent 4-reviewer pass @9b9054fb, zero blockers; all findings applied in one commit): the fence now reads a bare `''` at every position an attribute value is built and the template-literal fallback and a call's receiver, with a fixture-driven probe per shape; the four DocDB / Neptune cluster sites and both ElastiCache `ConfigurationEndpoint` sites carry `toStrictEqual` cases; the allow-list is an equality (a third copy of an allow-listed text fails, fixture-probed); the `describedInstanceAttributes` docstring names the `--no-wait` + private-subnet residual (go-to-k/cdkd#3096); the depth-cap assertion carries its own message; the changelog states the `*Url` / `*Arn` refusal direction; CloudFront `DomainName` and DBProxy(-Endpoint) `VpcId` were added to go-to-k/cdkd#3096. **Declined**: a deploy-engine outputs/export-boundary test for an absent key — the resolver-level pin in `intrinsic-functions.test.ts` covers the consumer half and `deploy-engine.ts` is untouched; the real-resolver engine harnesses in `tests/unit/deployment/` each mock a different provider surface and none reaches an `AWS::EC2::Instance` `Fn::GetAtt` without a new EC2 client mock, so it is not a 20-line addition.

Round 4 (delta 18d57b54 → b707d748, one code reviewer on Fable): clean; two optional nits taken as the final delta (a `toHaveBeenCalledTimes(3)` pin on the two VPC cases, and a reworded depth-cap comment) without another author-side round — the maintainer-side `/review-pr` round is where that last delta gets independent eyes.

Round 3 (delta 1eb60351 → 18d57b54, code + test on Fable): no blockers. Code — the VPC comment claimed a warning the `DefaultSecurityGroup` arm does not emit (comment corrected, row folded into go-to-k/cdkd#3096) and the depth cap silently dropped 18 nodes with no floor (now counted and capped at 25); both reviewers asked for a VPC create behaviour test (two cases added, probed against the fence-invisible `?? 'sg-unknown'` / `String(x)` shapes). Test — the walk's documented non-reach now names `publicMember` as the deliberate example.

Round 2 (delta c22a5f42 → 1eb60351, four reviewers on Fable): security clean; spec — a stale count in the module header and four wording nits (all taken); code — one minor (`createVpc`'s manufactured `''` pair, fixed) and three nits (the export now has a direct table test; `findCallee`'s bare-name ambiguity is documented; CodeCommit's live `getAttribute()` answering `undefined` for a missing member is named in the commit message); test — two medium fence gaps (a call's arguments, a followed helper's arguments) and the spread shape, all now followed and probed.

Round 1 (spec / code / test / security, all on Fable, head c22a5f42): spec clean (two nits, both taken); security clean; code — one major (running private-subnet instance degraded to the instance id: fixed by the state gate above) and one minor (fence blind spots: fixed, 22 more sites converted, four shapes followed); test — the same blind spots plus three unfenced bypass mutants (ECR update / AgentCore import / RDS cluster update: cases added) and a fixture-realism nit (fixed). Won't-do: `asg-provider.ts`'s `getAttribute()` returning `''` for a missing ARN — that is the live `getAttribute` contract read by `orphan-rewriter`, not the persisted `attributes` map this PR is about; noted for the fence's follow-up.


go-to-k:
Merged — thanks. The state-gated `describedInstanceAttributes` (settled → known-empty `''`, pending → omitted) and the fixture-driven fence over the 18 providers are what made this safe to take across the whole family, and the new `--no-wait` phase in `ec2-instance` gives the pending path a live arm at last.

Issue #3064, opened 2026-09-13
fix(rollback): rollback-journal content is rendered raw, and its plan preview can be forged


go-to-k:
`rollback-journal.json` is an unchecked cast in the same bucket as `state.json`, written by anyone with `s3:PutObject`, and two of its readers render it raw.

Found by the security sign-off on go-to-k/cdkd#3016, which hardened the equivalent `state.json` / `lock.json` paths. Pre-existing on `main`; that PR neither introduced nor touched this.

## Two readers

**1. The parse failure prints the journal's own bytes.** `parseRollbackJournal` (`src/types/rollback-journal.ts`) interpolates the raw V8 `SyntaxError`, which QUOTES the offending input. `loadRollbackJournal` (`src/state/s3-state-backend.ts`) does not wrap it in a `CdkdError`, so `formatError`'s non-`CdkdError` branch prints it with none of the sanitising go-to-k/cdkd#3016 added one function away. cdkd's output is line-oriented, so an injected newline invents a line that reads like a real one — the same mechanism as go-to-k/cdkd#3003.

Reachable two ways: `cdkd rollback` reading the journal, and `appendRollbackJournalSegment` on a failed deploy.

**2. Worse: `cdkd rollback`'s CONFIRMED plan preview renders journal fields raw.** The preview interpolates each entry's `logicalId` and `resourceType` (`src/cli/commands/rollback.ts`). That is a step beyond a forged diagnostic line — a planted journal can forge the plan ROWS the user is being asked to confirm, so the confirmation attests to something other than what will run.

## Why this is not go-to-k/cdkd#3027

That issue is `severity:low` and scoped to raw `stackName` interpolations in `s3-state-backend.ts` and `lock-manager.ts`. Neither its title nor its categories cover journal CONTENT, and the plan-preview forgery is a different severity class from a mangled log line. A comment was left there for cross-reference; this is the issue that owns it.

## Suggested shape

Reader 1 is the same fix go-to-k/cdkd#3016 applied: wrap the parse failure in a `CdkdError` (so `formatError`'s sanitised branch handles it) or sanitise at the interpolation. Reader 2 needs the preview's per-entry fields routed through `displaySafe(..., { asciiOnly: true })` — `logicalId` and `resourceType` both have a known charset, so the allowlist is a no-op on every legitimate value.

[...]

Issue #3027, opened 2026-09-11
fix(state): 38 more raw stackName log and throw sites outside the two read commands


go-to-k:
## What is wrong

go-to-k/cdkd#3003 guarded the values `cdkd state show` and `cdkd state resources`
interpolate into their REFUSALS, so a state record cannot forge a row in the
diagnostic the way go-to-k/cdkd#2772 stopped it forging one in the rendered
output. That scope was the two READ commands.

The same two files carry many more sites on OTHER paths that interpolate a raw
`stackName` (and often a raw `region`) into a thrown message or a log line.
Measured on `4f76097e`, excluding the ones go-to-k/cdkd#3003 fixed and
excluding key-construction and dedupe-key uses:

| file | raw-`stackName` sites |
| --- | --- |
| `src/state/s3-state-backend.ts` | 11 |
| `src/state/lock-manager.ts` | 27 |

They are on `saveState`, `deleteState`, the rollback-journal helpers, and the
lock acquire / release / renew paths — so they surface through
`cdkd deploy`, `cdkd destroy`, `cdkd rollback` and the lock-contention
messages rather than through the two read commands.

## Why it matters

It is the same mechanism and the same threat model: a `stackName` reaches
these from an S3 key segment or from a CLI argument, the rows of cdkd's own
output are joined by newlines, and a thrown message reaches the terminal
through `handleError`. What differs is only WHICH command prints it.

It is filed separately rather than folded into go-to-k/cdkd#3003 because the
reachability differs per site and has to be judged per path — several of these
take a name the user just typed, where the guard is defence in depth rather
than a hole being closed, and a few are on paths where an operator may WANT
the exact bytes. A sweep that applies the guard uniformly without that
judgement would be the kind of change that looks complete and is not.

## Proposed fix

Per site: decide whether the value is attacker-influenceable on that path, and
where it is, route it through the file's existing predicate
(`S3StateBackend.displayName`, `lock-manager`'s `safeSegment`) — both already
exist, so the change is a call per site plus a case per path.

The wider question this raises, and the reason it is worth doing as one pass
rather than incrementally: cdkd now has THREE names for one expression
(`displayName`, two `safeSegment`s, and `safe` in `src/cli/commands/state.ts`)
plus roughly eighteen surviving inline copies across `s3-state-backend.ts`,
`state.ts`, `lock-contention-message.ts` and `custom-resource-provider.ts`.
`src/utils/display-safe.ts`'s own header argues that a rule widened by hand
[...]

PR #3104, opened 2026-09-14
fix(rollback): sanitize every journal field the executor and the plan preview render, with a visible boundary and a cap

### LINKED ISSUE #3092: fix(rollback): rollback-executor replay logs render journal fields raw, and the plan preview has no length cap
`src/deployment/rollback-executor.ts` renders rollback-journal fields raw in its replay logs -- 39 interpolations of `op.logicalId` / `op.resourceType` / `op.changeType` into `logger.*` lines and thrown messages, including the `--orphan ${op.logicalId}` remedy a user is told to paste.

Found by the security sign-off on go-to-k/cdkd#3072, which closed the same class in `cdkd rollback`'s PLAN PREVIEW and in `parseRollbackJournal` (go-to-k/cdkd#3064). The executor is the layer BELOW those two: it runs the plan the preview showed, and it is reached by the standalone `cdkd rollback` AND by the deploy engine's automatic rollback, so its lines print on every failed deploy, not only on the rollback command.

## Threat, briefly

`rollback-journal.json` is a sibling of `state.json` in the same bucket, writable by anyone with `s3:PutObject`, validated no more than an unchecked cast. cdkd's output is line-oriented, so an injected newline invents a line that reads like a real one. The `--orphan ${op.logicalId}` remedy is the sharpest of the 39: it is a command the user is invited to run, so a planted id forges the ARGUMENT of that command.

## What go-to-k/cdkd#3072 established, so this does not re-derive it

- One predicate per module, ASCII allowlist for known-charset values (a logical id, a CFn type, a change type), `|| UNRENDERABLE` so an all-invisible value is not shown as absent. `UNRENDERABLE` now lives in `src/utils/display-safe.ts`, importable from `src/deployment/` without a layering inversion.
- Sanitize DISPLAY only. The executor keys its own state mutations and the AWS calls on the raw id; a sanitized lookup or a sanitized argument to a provider is the go-to-k/cdkd#2170 class in the wrong direction. Enumerate which of the 39 are renders and which feed a call before touching any.
- Fence both dimensions: the guard removed (hole) AND `asciiOnly` swapped for the denylist (class). The class needs a zero-width space in the fixture -- a control byte is in both classes and discriminates neither.
- A SOURCE-SHAPE fence closes the per-arm wiring at once: no bare `${op.logicalId}` / `${op.resourceType}` / `${op.changeType}` in the file, with a floor on the wrapped form so the fence proves it read the file. go-to-k/cdkd#3072's `SOURCE SHAPE` case in `tests/unit/cli/commands/rollback.test.ts` is the template.

## Also in scope, from the same review: same-line spoofing in the preview

An all-ASCII `logicalId` such as `X (AWS::RDS::DBInstance) -- already reverted` survives the allowlist and plants cdkd's own annotation wording INSIDE a real row; there is no length cap, so a long value can push the genuine trailing `(${type})` off a narrow terminal. It cannot add or hide an OPERATION -- every rendered row still maps 1:1 to an op the replay will attempt -- which is why go-to-k/cdkd#3072 left it out. A fix is a length cap through `truncateCodePoints` (go-to-k/cdkd#3067's helper) plus quoting the id so its boundary is visible. Listed here rather than as its own issue because the executor's lines have the same exposure and the same remedy.

Dup-check: searched open and closed issues for `rollback-executor replay log logicalId raw`, `rollback plan preview lookalike annotation length cap` and `same-line spoof plan row`; no hits. go-to-k/cdkd#3027 covers raw `stackName` interpolations in `s3-state-backend.ts` / `lock-manager.ts` / `state.ts` and does not name this file or these fields.
Session-fit: next (not this session) -- the session that found it merged go-to-k/cdkd#3072 on the command and parser layers; this is the executor layer, a different file with 39 sites needing a render-vs-argument judgement each, and the evidence is a grep that survives in this body. Verification a fresh session runs: `vp test run tests/unit/deployment/rollback-executor.test.ts tests/unit/cli/commands/rollback.test.ts`, then the mutation table over `tests/unit/`.
Severity: medium -- a forged log line on a failed deploy, and a forged `--orphan <id>` remedy the user may paste; the plan the user CONFIRMS is already protected by go-to-k/cdkd#3072, so the confirmation itself is no longer forgeable.
Effort: medium (M) -- 39 sites with a per-site render-vs-argument call, one predicate, one source-shape fence, and a re-review round; rides the existing rollback fixtures, no new integ.
Estimate: ~2-3 h -- most of it the per-site judgement and the two-dimension fences; the executor is reached by the deploy engine, so its test file is the larger of the two.

go-to-k:
Closes #3092

## Summary

`rollback-journal.json` is a sibling of `state.json` in the same bucket — writable by anyone with `s3:PutObject`, validated no more than an unchecked cast. go-to-k/cdkd#3064 closed the terminal-injection class one layer UP, in `cdkd rollback`'s plan preview and the journal parser. This is the layer that RUNS the plan: `src/deployment/rollback-executor.ts`, reached by the standalone command AND by the deploy engine's automatic rollback, so its lines print on every failed deploy. Round 2 (review) also closes the issue's second half — the same-line spoof and the missing length cap — in BOTH layers.

### What was exposed

53 interpolations of a journal operation's `logicalId` / `resourceType` / `changeType` into `logger.*` lines and thrown messages were raw. cdkd's output is line-oriented, so an injected newline invented a line that read exactly like one of the executor's own. Measured on the pre-fix binary against a journal planted in a real bucket:

```
  Rollback: Deleting created resource RealDB (AWS::RDS::DBInstance) — resource no longer i...
```

That is a forged executor line, standing at column zero, from a journal whose only real op was a skipped UPDATE. The sharpest of the 53 is the `--orphan ${op.logicalId}` remedy a user is invited to paste — a planted id forges the command's ARGUMENT.

Two things an allowlist cannot see, from the issue's second section: an all-ASCII id such as `X (AWS::RDS::DBInstance) -- already reverted` plants the line's own annotation wording INSIDE a real row, and there was no length cap.

### What changed

**Population first, field names second.** The enumeration that missed `roleArn` on go-to-k/cdkd#3072 had grepped for the fields it expected; here every `${op.<field>}` interpolation was listed regardless of field (58) and classified by its enclosing statement. Round 2 widened the population by DATA FLOW, which is how the review found the three the receiver-name enumeration missed: `retainedSurvivorMessages(logicalId, resourceType, ...)` renders its PARAMETERS, `prev = op.previousState` carries a journal `physicalId`, and `withRetry(..., logicalId)` is a cross-module reader.

- **One helper, `displayIdent`** (`src/utils/display-safe.ts`), and both layers' `safe()` call it: the ASCII allowlist with `<unrenderable>` for a value that sanitizes to nothing; a cap of `IDENT_MAX_CODE_POINTS` (255, CloudFormation's logical-id maximum) that names how much it withheld; and a JSON-quoted rendering for a value that is not a plain identifier, so its BOUNDARY is visible — `"X (AWS::RDS::DBInstance) -- already reverted" (AWS::S3::Bucket)`. **The quoting is conditional on purpose**: every legitimate id / type / stack name / region / run id / ARN is a plain identifier and renders exactly as before, so none of the 97 `grep` hits on executor wording across 20 unit files and none of the 4 integ greps (`rollback-replacement-retain`, `rollback-deletion-policy-snapshot`) move. A direct suite pins identity on every legitimate shape.
- **53 renders** in the executor and every plan-preview render in `rollback.ts` take `safe()`.
- **5 free-form SDK messages** take `displaySafe()` directly — the denylist, as `formatError` does for a `cause`. The name-collision `msg` is classified RAW by `isNameCollisionError` and rendered sanitized; the two USES are split, not the value.
- **The pasted `--orphan` remedy** prints its id only when it IS a CloudFormation logical id (`typeof op.logicalId === 'string' && PASTEABLE_LOGICAL_ID.test(op.logicalId)`, `/^[A-Za-z0-9]{1,255}$/`, rounds 4-5 — the `typeof` because `RegExp.test` coerces, and the journal parser validates no per-op field, so a numeric `logicalId` would otherwise print `--orphan 123`). Stricter than "`safe()` is the identity on it" on purpose: identity already refuses the TRIM (an id differing from a legitimate one only by a leading invisible rendered identically), the quote, the cap and the placeholder — but `~user` / `=x` are plain identifiers the user's shell expands before cdkd sees them. Otherwise the line says why the id is withheld and where to read it (`cdkd events`).
- **`retry.ts`** sanitizes its label once at `withRetry`'s entry (allowlist, NOT `displayIdent`: a label there is not always an identifier — `Fn::GetAtt [X, Arn]`, `<id> (<dimension>)` — so boundary quoting would change the resolver's and the providers' ordinary lines). Both executor retry helpers (`updateWithRollbackRetry`, `createWithRollbackRetry`) render the label themselves before handing it over, so the journal path gets the cap and the boundary too; callers pass the raw id.
- **The collision message** — the one line that carries the pasted remedy — renders `prev.physicalId` / `current.physicalId` through `safe()` too (round 3): the outer catch's denylist stops a newline, not a planted physical id that READS as the remedy ahead of the guarded one. Round 4 also wraps `current.physicalId` — from `state.json`, same bucket, same trust class — on its three raw `logger.*` lines (`displaySafe`), and fence (3) now covers both aliases.
- **6 bare `${op.*}` stay on purpose**: 5 in the `reason` / `survivorReason` strings handed to `ctx.recordEvent` — PERSISTED into `deployments/*.jsonl`, rendered by `cdkd events` through `safeText` + `<unrenderable>` (`events.ts:603`), so sanitizing here would double it at render — and the charset-gated remedy. Two cases pin the reasons raw; the source-shape fence pins the count at exactly 6.

Provider calls and `stateResources` keys receive the raw id — `${}` never touches them, and a case asserts the provider saw the hostile id verbatim.

**A hostile `changeType` cannot reach a classified arm.** The classifier routes on the literal; an unknown one lands on the UPDATE arm, which does not render the field. It reaches exactly one line — the catch after a provider throw — and that is where the case drives it.

### Not changed, and why

Follow-up filed: go-to-k/cdkd#3140 — `parseRollbackJournal` accepts a non-string per-op `logicalId` (the round-5 security review's out-of-delta nit; the render side is closed here, the USE side — a `{"toString":null}` id throws at the first bracket lookup — is the parser's to refuse).

`src/types/rollback-journal.ts` keeps its round-1 `safeJournalText` (allowlist + `<unrenderable>`): its renders already single-quote the stack name, its parse `detail` is free text, and the value class is go-to-k/cdkd#3027's. `PLAIN_IDENT` deliberately excludes the IAM-path characters `!#$%&'()*` (the boundary-forging set), so a role ARN carrying one renders quoted — cosmetic, recorded in the regex's doc. `.claude/rules/layout-utils.md` gained a 382 B pointer only — the `src/utils/aws-client-defaults.ts` rule-path budget had 500 B of headroom on `origin/main`; the decisions live in `displayIdent`'s doc comment.

## Test plan

- [x] **Unit tests pass** — 1003 files / 22,376 passed / 1 skipped, root asserted to this worktree, exit 0.
- [x] **Suites**: `tests/unit/deployment/rollback-executor-log-injection.test.ts` (new, 22 cases: delete / three skip / revert / `--orphan` / three failed-op arms, the catch line, the placeholder, the readopt-Retain WARN through the helper parameter, BOTH `--orphan`-remedy branches through the collision path, the retry label chain on the update arm and on BOTH re-create arms, the same-line spoof + cap, two raw-event pins, four source-shape fences), `tests/unit/utils/display-safe.test.ts` (+5: identity on every legitimate shape incl. the nested-stack `Parent~Child` name, quoting + JSON escape, cap, placeholder, never-identity-on-a-changed-value), `tests/unit/cli/commands/rollback.test.ts` (+2: spoof + cap in the preview; 8 pins re-anchored on the quoted boundary, 7 more in the executor suite), `tests/unit/deployment/retry-transient-server-error.test.ts` (+1: a hostile label through the REAL `withRetry`).
- [x] **Mutation table** — measured LAST, on the final tree, one mutation per probe over the whole `tests/unit/`, hash-verified restore, failing cases named (below the table).

  | mutation | cases red |
  | --- | --- |
  | executor `safe()`: guard dropped (`String(value)`) | 14 |
  | executor `safe()`: CLASS (`asciiOnly` -> denylist) | 14 |
  | executor `safe()`: round-1 form (allowlist only, no cap / quote) | 12 |
  | `displayIdent`: boundary quoting dropped | 18 |
  | `displayIdent`: length cap dropped | 4 |
  | `displayIdent`: `<unrenderable>` fallback dropped | 3 |
  | `--orphan` remedy: gate dropped (always prints the id) | 4 |
  | `retainedSurvivorMessages` WARN half un-wrapped | 2 |
  | `${prev.physicalId}` re-adopt info line un-wrapped | 2 |
  | `updateWithRollbackRetry` -> `withRetry` label passed raw | 2 |
  | `retry.ts` `shownId` raw | 1 |
  | one label arm un-wrapped (source-shape fence) | 2 |
  | one label arm un-wrapped BEHIND a `// reason:` comment (fence comment hole) | 2 |
  | event `reason` SANITIZED (the regression) | 2 |
  | catch line: free-form message RAW | 3 |
  | catch line: free-form message `asciiOnly` | 2 |
  | `PLAIN_IDENT` without `~` (nested-stack child name quoted) | 2 |
  | `createWithRollbackRetry`: rendering dropped (`shownId = logicalId`) | 3 |
  | `createWithRollbackRetry`: INNER loop label raw | 3 |
  | `createWithRollbackRetry`: OUTER loop label raw | 1 |
  | collision message: `prev.physicalId` raw (forged remedy ahead of the guarded one) | 2 |
  | collision message: `current.physicalId` raw | 1 |
  | re-adopt info line un-wrapped BEHIND a comment inside `logger.info(` (fence-3 comment hole) | 2 |
  | `--orphan` paste gate: CFn charset -> `safe()` identity (round-3 form; `~x` pasteable) | 2 |
  | `${current.physicalId}` un-wrapped on a logger line (delete-new-first info) | 2 |
  | `--orphan` paste gate: `typeof` guard dropped (`RegExp.test` coerces a non-string) | 2 |
  | preview `safe()`: round-1 form (allowlist only) | 8 |

  - **executor `safe()`: guard dropped (`String(value)`)** (14): `a value past the identifier cap is cut, and the cut is named`; `an all-ASCII id cannot plant the line's own annotation wording -- its boundary is quoted`; `an id that sanitizes to NOTHING renders the placeholder, not an empty slot`; `the --orphan path`; `the UPDATE revert path`; `the catch line: a hostile changeType is flattened and the provider error takes the DENYLIST`; `the completed-CREATE delete path`; `the delete-first re-create arm hands withRetry the executor's rendering too`; `the failed-op paths under --revert-failed (force-revert, delete-failed-create, skip-failed-noop)`; `the pasted `--orphan` remedy is WITHHELD when sanitizing would change the id`; `the re-create path hands withRetry the executor's rendering too`; `the readopt-Retain WARN: a field passed through a helper PARAMETER is still sanitized`; `the retry label handed to withRetry is sanitized where retry.ts renders it`; `the three SKIP paths (already reverted / physical id changed / absent)`
  - **executor `safe()`: CLASS (`asciiOnly` -> denylist)** (14): `a value past the identifier cap is cut, and the cut is named`; `an all-ASCII id cannot plant the line's own annotation wording -- its boundary is quoted`; `an id that sanitizes to NOTHING renders the placeholder, not an empty slot`; `the --orphan path`; `the UPDATE revert path`; `the catch line: a hostile changeType is flattened and the provider error takes the DENYLIST`; `the completed-CREATE delete path`; `the delete-first re-create arm hands withRetry the executor's rendering too`; `the failed-op paths under --revert-failed (force-revert, delete-failed-create, skip-failed-noop)`; `the pasted `--orphan` remedy is WITHHELD when sanitizing would change the id`; `the re-create path hands withRetry the executor's rendering too`; `the readopt-Retain WARN: a field passed through a helper PARAMETER is still sanitized`; `the retry label handed to withRetry is sanitized where retry.ts renders it`; `the three SKIP paths (already reverted / physical id changed / absent)`
  - **executor `safe()`: round-1 form (allowlist only, no cap / quote)** (12): `a value past the identifier cap is cut, and the cut is named`; `an all-ASCII id cannot plant the line's own annotation wording -- its boundary is quoted`; `an id that sanitizes to NOTHING renders the placeholder, not an empty slot`; `the --orphan path`; `the UPDATE revert path`; `the catch line: a hostile changeType is flattened and the provider error takes the DENYLIST`; `the completed-CREATE delete path`; `the delete-first re-create arm hands withRetry the executor's rendering too`; `the failed-op paths under --revert-failed (force-revert, delete-failed-create, skip-failed-noop)`; `the pasted `--orphan` remedy is WITHHELD when sanitizing would change the id`; `the re-create path hands withRetry the executor's rendering too`; `the retry label handed to withRetry is sanitized where retry.ts renders it`
  - **`displayIdent`: boundary quoting dropped** (18): `an all-ASCII id cannot plant the line's own annotation wording -- its boundary is quoted`; `an all-ASCII id cannot plant the row's own annotation wording -- its boundary is quoted (#3092)`; `is never the identity on a value it changed`; `quotes a value that could plant the surrounding line's own annotation`; `the --orphan path`; `the CONFIRMATION PROMPT itself is sanitized`; `the REGION reaches the plan header and the prompt sanitized`; `the UPDATE revert path`; `the catch line: a hostile changeType is flattened and the provider error takes the DENYLIST`; `the completed-CREATE delete path`; `the delete-first re-create arm hands withRetry the executor's rendering too`; `the failed-op paths under --revert-failed (force-revert, delete-failed-create, skip-failed-noop)`; `the multi-journal CANDIDATE LIST cannot inject a row`; `the pasted `--orphan` remedy is WITHHELD when sanitizing would change the id`; `the previewState LOOKUPS stay keyed on the RAW logicalId`; `the re-create path hands withRetry the executor's rendering too`; `the retry label handed to withRetry is sanitized where retry.ts renders it`; `the stack name and region reach every message sanitized on the no-arg path`
  - **`displayIdent`: length cap dropped** (4): `a value past the identifier cap is cut and the cut is named (#3092)`; `a value past the identifier cap is cut, and the cut is named`; `cuts a value past the identifier cap and says how much it withheld`; `is never the identity on a value it changed`
  - **`displayIdent`: `<unrenderable>` fallback dropped** (3): `a field that sanitizes to NOTHING renders the placeholder, not an empty slot`; `an id that sanitizes to NOTHING renders the placeholder, not an empty slot`; `renders nothing-left and absent values as the placeholder, unquoted`
  - **`--orphan` remedy: gate dropped (always prints the id)** (4): `SOURCE SHAPE: a bare journal-field interpolation exists only in an event-bound statement`; `the pasted `--orphan` remedy is WITHHELD for a NON-STRING id (the gate must not coerce)`; `the pasted `--orphan` remedy is WITHHELD for a plain id the shell would expand`; `the pasted `--orphan` remedy is WITHHELD when sanitizing would change the id`
  - **`retainedSurvivorMessages` WARN half un-wrapped** (2): `SOURCE SHAPE: the aliases the first round missed are wrapped`; `the readopt-Retain WARN: a field passed through a helper PARAMETER is still sanitized`
  - **`${prev.physicalId}` re-adopt info line un-wrapped** (2): `SOURCE SHAPE: the aliases the first round missed are wrapped`; `the readopt-Retain WARN: a field passed through a helper PARAMETER is still sanitized`
  - **`updateWithRollbackRetry` -> `withRetry` label passed raw** (2): `SOURCE SHAPE: both retry helpers render the label themselves, and their callers pass the raw id`; `the retry label handed to withRetry is sanitized where retry.ts renders it`
  - **`retry.ts` `shownId` raw** (1): `renders the label through the ASCII allowlist, so a journal-sourced id cannot forge a line (issue #3092)`
  - **one label arm un-wrapped (source-shape fence)** (2): `SOURCE SHAPE: a bare journal-field interpolation exists only in an event-bound statement`; `the UPDATE revert path`
  - **one label arm un-wrapped BEHIND a `// reason:` comment (fence comment hole)** (2): `SOURCE SHAPE: a bare journal-field interpolation exists only in an event-bound statement`; `the UPDATE revert path`
  - **event `reason` SANITIZED (the regression)** (2): `SOURCE SHAPE: a bare journal-field interpolation exists only in an event-bound statement`; `the event RECORD keeps the raw value: it is persisted, and `cdkd events` sanitizes at render`
  - **catch line: free-form message RAW** (3): `SOURCE SHAPE: every free-form error render takes displaySafe()`; `the catch line: a hostile changeType is flattened and the provider error takes the DENYLIST`; `the delete-first re-create arm hands withRetry the executor's rendering too`
  - **catch line: free-form message `asciiOnly`** (2): `SOURCE SHAPE: every free-form error render takes displaySafe()`; `the catch line: a hostile changeType is flattened and the provider error takes the DENYLIST`
  - **`PLAIN_IDENT` without `~` (nested-stack child name quoted)** (2): `is the identity on every legitimate identifier shape`; `the pasted `--orphan` remedy is WITHHELD for a plain id the shell would expand`
  - **`createWithRollbackRetry`: rendering dropped (`shownId = logicalId`)** (3): `SOURCE SHAPE: both retry helpers render the label themselves, and their callers pass the raw id`; `the delete-first re-create arm hands withRetry the executor's rendering too`; `the re-create path hands withRetry the executor's rendering too`
  - **`createWithRollbackRetry`: INNER loop label raw** (3): `SOURCE SHAPE: both retry helpers render the label themselves, and their callers pass the raw id`; `the delete-first re-create arm hands withRetry the executor's rendering too`; `the re-create path hands withRetry the executor's rendering too`
  - **`createWithRollbackRetry`: OUTER loop label raw** (1): `SOURCE SHAPE: both retry helpers render the label themselves, and their callers pass the raw id`
  - **collision message: `prev.physicalId` raw (forged remedy ahead of the guarded one)** (2): `SOURCE SHAPE: the aliases the first round missed are wrapped`; `the pasted `--orphan` remedy is WITHHELD when sanitizing would change the id`
  - **collision message: `current.physicalId` raw** (1): `SOURCE SHAPE: the aliases the first round missed are wrapped`
  - **re-adopt info line un-wrapped BEHIND a comment inside `logger.info(` (fence-3 comment hole)** (2): `SOURCE SHAPE: the aliases the first round missed are wrapped`; `the readopt-Retain WARN: a field passed through a helper PARAMETER is still sanitized`
  - **`--orphan` paste gate: CFn charset -> `safe()` identity (round-3 form; `~x` pasteable)** (2): `SOURCE SHAPE: a bare journal-field interpolation exists only in an event-bound statement`; `the pasted `--orphan` remedy is WITHHELD for a plain id the shell would expand`
  - **`${current.physicalId}` un-wrapped on a logger line (delete-new-first info)** (2): `SOURCE SHAPE: the aliases the first round missed are wrapped`; `the delete-first re-create arm hands withRetry the executor's rendering too`
  - **`--orphan` paste gate: `typeof` guard dropped (`RegExp.test` coerces a non-string)** (2): `SOURCE SHAPE: a bare journal-field interpolation exists only in an event-bound statement`; `the pasted `--orphan` remedy is WITHHELD for a NON-STRING id (the gate must not coerce)`
  - **preview `safe()`: round-1 form (allowlist only)** (8): `a field that sanitizes to NOTHING renders the placeholder, not an empty slot`; `a value past the identifier cap is cut and the cut is named (#3092)`; `an all-ASCII id cannot plant the row's own annotation wording -- its boundary is quoted (#3092)`; `the CONFIRMATION PROMPT itself is sanitized`; `the REGION reaches the plan header and the prompt sanitized`; `the multi-journal CANDIDATE LIST cannot inject a row`; `the previewState LOOKUPS stay keyed on the RAW logicalId`; `the stack name and region reach every message sanitized on the no-arg path`

  </details>

  Every row's count equals its named-case count (checked mechanically). History, kept because naming cases is the point: `updateWithRollbackRetry -> withRetry label raw` first measured 0 in round 2 (the case asserted flattening, which `retry.ts`'s own allowlist also supplies; it now pins the QUOTE); round 3's second `createWithRollbackRetry` call site measured 0 until the delete-first arm had a case; the round-4 table's first row carried an unrelated resolver-depth flake, and the round-5 harness once mis-parsed a summary line as 0 while naming two red cases -- both caught by the names, and the harness now refuses a run whose summary and named cases disagree. One row is fence-only (1): the create helper's OUTER loop (its own lines print only after the inner give-up under a collision / cooldown classification, which no unit fixture reaches).

- [x] **Source-shape fences**: all three share `statementAround`, which treats comment lines as TRANSPARENT (skipped on the walk back, dropped from the text) — round 1 walked raw lines, so a `// reason:` comment above an un-wrapped render satisfied it; round 2 BLANKED comment lines, so one inside a multi-line `logger.info(` read as the blank that ends a statement and cut the walk (both measured by review; both comment-hole rows above are the probes). (1) a bare `${op.logicalId|resourceType|changeType}` exists only inside an event-`reason` statement (`^\s*reason:` / `survivorReason =`) or the charset-gated remedy, exact count 6, wrapped count > 40; (2) every `<x>.message : String(<x>)` interpolation is `displaySafe(`-wrapped except the persisted `survivorReason`, wrapped count exactly 4 + the `msg` pair; (3) `${prev.physicalId}` is wrapped on every `logger.*` line, and the collision message's two physical ids take `displayIdent`; (4) both retry helpers render their label and all four callers pass the raw id. Known limitation, recorded: fence (1) matches the exact `${op.<field>}` form — `${op.logicalId!}` or `${op.logicalId ?? ''}` would be invisible to it; the `> 40` floor and the per-arm runtime cases are the backstop.
- [x] **Live-tested against the BUILT bundle and a real state bucket** — twice. Round 1: hostile journal planted at `cdkd/Cdkd3092Probe/us-east-1/` with an empty `resources` map so every op skips and no AWS resource is touched; pre-fix build 1 forged executor line / 1 invisible, post-fix 0 / 0. Re-run on the FINAL build (`7900a125`) with four planted ops — the forged newline+ZWSP id, the all-ASCII same-line spoof, a 300-char id, and `~RealDB`: 0 forged column-zero lines, 0 invisibles; the spoof renders `"X (AWS::RDS::DBInstance) -- already reverted" (AWS::SQS::Queue) — no longer in state` in the plan AND in the executor's skip line, the long id renders `L…L [cut: 45 more characters withheld]`, `~RealDB` renders plain (withheld only inside the pasted command, which the skip path never prints). Probe removed with `s3_purge_prefix_versions ... all`; `s3_assert_versions_swept` reports 0 surviving versions.
- [x] **Documentation** — `changelog.d/` entry (1984 chars), `.claude/rules/layout-utils.md` pointer.
- [x] **Integration** (`rollback-executor.ts` / `retry.ts` are in `integ-destroy`'s scope and `integ-broad`'s `CROSS_CUTTING_REGEX`), both on the round-5 build via `/run-integ`: **`rollback-command`** PASS, 158 s, destroy 3 deleted / 0 errors, 0 surviving object versions under both stack prefixes; **`lambda`** (broad set — `bench-cdk-sample` has no `verify.sh` and cannot run from an agent session) PASS, 91 s, destroy 9 deleted / 0 errors, IAM / Lambda / SQS / DynamoDB / layer sweep empty. Ledger rows committed.

https://claude.ai/code/session_01HbJf9kM7x3JWZT8F1yG4oV


go-to-k:
Merged as 0a6cb848 — thanks. Five review rounds, each of which found a real site the previous enumeration had missed (a helper parameter, a cross-module label, a second call site, a coercing gate); the named-case mutation table earned its keep twice by exposing a flake and a mis-parsed summary before they were published.

PR #3105, opened 2026-09-14
chore(deps)!: raise the published Node.js floor from 20 to 22.12 across engines, the pack target, the CI matrix and the docs

### LINKED ISSUE #3037: chore(deps): decide whether to raise the published Node floor past 20 (EOL 2026-04-30), which is what currently pins commander at 14
## What this is

A decision, not a defect: cdkd declares `engines.node: ">=20.0.0"` and `vp pack` targets Node 20, while Node 20 reached end of life on 2026-04-30. Nothing is broken today. What it costs is that cdkd cannot take a dependency whose own floor has moved past 20 without silently raising the floor for its users as a side effect of a dependency bump.

## The concrete instance

dependabot go-to-k/cdkd#2533 proposed `commander` 12.1.0 -> 15.0.0. Two things were measured while handling it:

- `commander@15.0.0` declares `engines.node: ">=22.12.0"`. `commander@14.0.3` declares `">=20"`, and `commander@13.0.0` `">=18"`.
- Taking 15 would have moved cdkd's effective runtime floor from 20 to 22.12 with no line in the diff saying so.

So the lane took **14** instead: it clears the two-copies problem that made #2533 fail (cdkd and `cdk-local` must resolve the SAME commander major, since cdk-local's command factories are spliced into cdkd's commander tree) while leaving the declared floor untouched. `commander@14.0.3` already carries `helpGroup` / `commandsGroup` / `saveStateBeforeParse`, the members whose absence produced #2533's type errors, so nothing is given up at the type level either.

## What the decision actually involves

Raising the floor is deliberately NOT a one-line `engines` edit. The surfaces that have to move together:

- `package.json` `engines.node` (and `cdk-local`'s, since the two ship as a pair).
- The `vp pack` build target (`node20` today).
- `.github/workflows/**` — the `runtime-compat` matrix runs 22 and 24 today; whatever it asserts about the floor has to be re-derived from the new floor rather than left as-is.
- README and `docs/**` wherever the Node requirement is stated.
- `CONTRIBUTING.md` if it states a supported range distinct from the dev pin (`.node-version` is 24.15.0 and is a separate question).

Under `bump-minor-pre-major: true` the bump lands as a MINOR, which is the intended mapping for a breaking change on a 0.x line.

## Arguments on each side

For raising it: Node 20 is past EOL, so a floor of 20 advertises support for a runtime that receives no security fixes. cdkd is 0.x and documented as a dev/test tool, which is the cheapest possible moment to move a floor. And every deferral of this decision re-appears as a dependabot PR that has to be answered the same way.

Against raising it right now: it is unrelated to any user-visible cdkd behaviour, and bundling it into a dependency bump is what this issue exists to avoid. Doing it on its own PR is what makes the change legible.

## Acceptance

Either (a) the floor is raised to a stated version across every surface above and `commander` moves to 15, or (b) the decision is recorded as "not yet" with the condition that would change it, and `commander` stays on 14. Either outcome closes this.

Dup-check: searched go-to-k/cdkd issues (open and closed) for `commander`, `engines`, `Node.js version floor` and `drop Node`. Hits on `commander` are all CLI-surface issues (#2523, #2556, #2573, #2577, #2588) with no dependency-floor content; hits on `engines` are the bash-hook-runtime pair (#2715, #2716), which is about the hook layer's language, not the published Node floor. No issue covers this.

Session-fit: next (not this session) -- the session that found it was handling three dependabot PRs, and the whole point of the finding is that the floor decision does not belong inside a dependency bump. It also spans two repositories and a CI matrix, none of which the commander lane touches.
Severity: low -- nothing is broken for any user today; the cost is that cdkd advertises support for an EOL runtime and has to keep answering the same dependabot proposal.
Effort: medium (M) -- small diffs, but across two repositories plus a CI matrix, and the matrix change needs a green run on the new floor before it means anything.
Estimate: ~2-3 h -- most of it verifying the runtime-compat matrix and the packed artifact on the new target, not editing the manifests.

go-to-k:
## Summary

Node 20 reached end of life on 2026-04-30. This raises the published Node.js floor to **22.12.0** on every surface at once, so no single surface keeps advertising the old one:

- `package.json` `engines.node` → `>=22.12.0`
- `vite.config.ts` `vp pack` target → `node22`
- `.github/workflows/ci.yml` `runtime-compat` matrix → `['22.12', '24']` — the first row is the **exact** floor, so the smoke executes the bundle on the promised minimum (`vp env exec --node 22.12` resolves to v22.12.0; a bare `22` resolved to v22.23.2). The `ci-ok` comment's expanded-name example follows it (`runtime-compat (22.12)`).
- README, `docs/getting-started.md`, `docs/testing.md`, `CONTRIBUTING.md` + `docs/contributing.md`, `tests/benchmark/README.md`, the cdkd-skills plugin page, CLAUDE.md's Node.js section
- the two `src/cli/commands/local-*.ts` comments that pinned the `fs.cpSync` contract as "Node 20+" (comment-only). Review found the old comment's `verbatimSymlinks` claim was FALSE — it defaults to `false` in every Node, and a relative layer symlink is rewritten to an absolute host path (dangling in the container). The comments now state that fact and point at the behaviour fix, go-to-k/cdkd#3106 (cdk-local twin: go-to-k/cdk-local#727).
- `changelog.d/entries/2026-09-14-3037-node-floor-22.md`

Why 22.12 and not 22.0: it is the floor `commander@15` declares, so the next commander major cannot move the floor again as a side effect of a dependency bump — the thing the issue (#3037) exists to prevent. Under `bump-minor-pre-major` this lands as a MINOR release (`chore(deps)!:` / `BREAKING CHANGE:` footer).

## Fence

`tests/unit/scripts/node-floor-sync.test.ts` pins the surfaces to one literal `FLOOR` and derives everything else from it: engines (JSON-parsed, exact `>=22.12.0`), the pack target (exactly one `target: 'node<N>'`, fail-closed on 0 or 2), the matrix (YAML-parsed; quoted-string rows, first row equals the floor's major.minor, minimum equals the floor major, the dev pin's major from `.node-version` is present), the `ci-ok` example name (the floor row and no other, at any precision), and eight prose files each pinned to its exact floor sentence in a Node.js context (line-break tolerant). Retired spellings are GENERATED for every major back to 18 as bounded regexes restricted to floor claims (`or/and later|higher|newer|up`, `+`, `runtime`, `>=`) at any precision, so an EOL note and a `>= 2026` date stay legal while a file carrying both floors goes red on the old one. A self-probe runs the generator with today's floor as the retired one against today's eight sentences, so the next bump's fence provably catches each of them.

`CONTRIBUTING.md` and `.node-version` join the `check` gate's include (`.markgate.yml`) because the test reads them; CLAUDE.md's rough guide follows.

Mutation probes, each restored byte-exact and each red on the first try (three rounds): engines back to `>=20.0.0`; target back to `node20`; matrix `[20, 22, 24]`; first row as a bare `'22'`; dropping the `'24'` row; unquoted floats `[22.12, 24]`; `ci-ok` example as `(20)`, `(22)` and `(22.12.0)`; a duplicated `target:` line; README `22.12` moved out of a Node context; README carrying both floors, at `20` and at `20.19`; CONTRIBUTING carrying `Node.js 20 and later`; CLAUDE.md's two "Node 20 runtime" spellings; the generator with its minor-precision arm removed and with its `and` arm removed (the self-probe catches both); `docs/getting-started.md` losing its `.0` patch. Controls that stay green: an EOL note ("Node 20 or earlier is past end of life"), a `>= 2026-09-14` date, a hard-wrap moved inside a floor sentence, the matrix respelled as a block sequence.

## Live test

- `vp env exec --node 22 node dist/cli.js --version` → `0.289.8` on v22.23.2; same on v24.21.0; `vp env exec --node 22.12 node -v` → `v22.12.0`.
- A tarball carrying this `engines` block: `npm install --engine-strict` under Node v20.20.2 → `EBADENGINE ... Required: {"node":">=22.12.0"}` (rc=1); under Node 22 → rc=0. Without `--engine-strict` npm warns and installs, which is what the changelog entry says.

## Review

3-axis (spec / code / test) at 8cec86f8, then re-reviews of each fix delta: code + test at 3a37de94, test at ad62ec58, test at 188b3cd1. Everything raised is fixed in this PR except: the cpSync symlink BEHAVIOUR → go-to-k/cdkd#3106 (comment corrected here); and two recorded won't-dos — a matrix ENUMERATION ("on Node 20 / 22 / 24") is a known bound of the fence's generator because a measurement in `tests/setup.ts` has the same shape (noted in the test's doc comment; a doc that enumerates the matrix is pinned positively instead), and the two `src/` comments that name the floor stay outside the docs fence (the bump's own grep finds them; a docs fence should not read code text). The final commit (3d833b2d) only converges the three repos' fences on one generator shape and adds the synthetic claims the round-4 reviewer asked for; every generator arm now has a claim that reds when the arm is deleted, so that delta was not re-dispatched.

## Not in this PR

- `commander` 12 → 15: cdk-local's command factories are spliced into cdkd's commander tree, so both must resolve the same major and cdk-local moves first. A separate dependabot round after both floors are released.
- `.node-version` (the dev / CI pin, 24.15.0), `@types/node`, the `node:20-slim` base images of the local-runtime integ fixtures (container images, not the CLI's runtime), and the `nodejs20.x` Lambda runtime mentions.

Sibling lanes, same floor, same session: go-to-k/cdk-local#722 and go-to-k/cdk-real-drift#1905.

Closes #3037


go-to-k:
Merged — thanks. Four review rounds, and the one that mattered was the cpSync probe: the floor bump is the reason the comment got re-read, and go-to-k/cdkd#3106 is the bug that had been hiding under it.

Issue #3101, opened 2026-09-14
fix(diff): a no-change stack whose previously resolved output now fails suppresses an Outputs delta the partial persist writes, so cdkd diff --fail exits 0


nix-tkobayashi:
## What is wrong

Since go-to-k/cdkd#2771 the no-change deploy path persists the outputs that resolved while a sibling fails, and keeps a failed output's stored value. `cdkd diff` did not move with it: a bag it cannot fully resolve still sets `resolutionFailed`, and `computeStackDiff` then reports no outputs delta at all. That now disagrees with the deploy in one shape, and `cdkd diff --fail` exits 0 for a change the next deploy makes.

The shape:

- the stack has no resource change;
- an output that resolved on an earlier deploy now fails, and the diff reproduces the failure — for example an `Fn::GetAtt` whose attribute cannot be constructed, which is unresolved for the diff too;
- because that output HAS a stored value, the go-to-k/cdkd#2740 record does not bind for it (the reader checks absence from state first, `resolveTemplateOutputs` in `src/analyzer/outputs-diff.ts`), so the key fails in the diff instead of being skipped;
- a sibling output is added or changed.

`cdkd diff --fail` suppresses the Outputs section with the "may have changed, but one or more could not be resolved" warning and exits 0. `cdkd deploy` then writes the sibling.

Before go-to-k/cdkd#2771 the suppression matched the deploy, which kept the whole previous bag in exactly this case. The warning still prints, so the change is not silent, but `--fail` is the signal CI reads.

Found by the spec review of the go-to-k/cdkd#2771 PR; not fixed there because mirroring the merge on the diff side needs its own design (below).

## Proposed fix

When `resolutionFailed` is set AND the resource diff reports no change (the deploy will take the no-change path), preview the outputs through the same merge the deploy runs, `mergeNoChangeOutputs` in `src/deployment/no-change-outputs-merge.ts`, instead of suppressing the section. Keep the suppression when resources change, where a failure usually means a pending CREATE.

Two things need deciding before that is safe:

1. The diff's `failedKeys` also records a failed output's literal alias, and a failure on an intrinsic `Export.Name` records no key at all. The merge input must be built from output keys only, and a failure the diff cannot name must keep the suppression.
[...]

Issue #875, opened 2026-06-15
fix(deploy): Outputs-only change on an unchanged stack is not persisted, breaking cross-stack Fn::ImportValue


sakurai-ryo:
## Summary

When stack B newly references a property of an already-deployed stack A, CDK synth adds a new `Output` (with `Export.Name`, consumed via `Fn::ImportValue`) to A's template **without changing any of A's resources**. In this case, deploying A is treated as a no-op and the new export is **never persisted**, so B's subsequent `Fn::ImportValue` resolution fails.

This is a behavior divergence from the CDK CLI / CloudFormation, where an Outputs-only change is detected as a changeset diff and stack A is updated.

## Conditions to reproduce

The bug only fires when **A is already deployed** and a downstream reference is added **afterwards** (A's resources stay identical):

1. Deploy stack A (nothing references it yet, so A's `state.outputs` has no export).
2. Add stack B that references a property of A (e.g. `Fn::ImportValue` of A's resource attribute). CDK synth now adds a new Output/Export to A; A's `Resources` section is unchanged.
3. `cdkd deploy A B` (or deploy A, then B).
   - A's deploy is a no-op (no resource diff) -> the new export is not written to A's state / exports index.
   - B's deploy throws: `Fn::ImportValue: export '...' not found in any stack.`

Deploying A and B together **for the first time** works fine (A's resources are all `CREATE`, so the normal path runs and outputs/export are persisted).

## Root cause

1. **Diff ignores Outputs.** `DiffCalculator.calculateDiff()` compares only `currentState.resources` vs `template.Resources` (`src/analyzer/diff-calculator.ts:48-49`); `hasChanges()` inspects only the resource-change map. An Outputs-only change yields `hasChanges = false`.

2. **No-change path never persists outputs.** In `src/deployment/deploy-engine.ts:792-851`, the `!hasChanges` branch early-returns:
   - `resolveOutputs()` is not called (it only runs inside `executeDeployment`, `deploy-engine.ts:1340`).
   - The returned outputs use the stale `currentState.outputs` (`deploy-engine.ts:849`).
   - Even the observedProperties-refresh save writes `outputs: currentState.outputs` (`deploy-engine.ts:810`).
   - `exportIndexStore.updateForStack()` (`deploy-engine.ts:916-922`) is never reached.

[...]

PR #3107, opened 2026-09-14
fix(deploy): persist the outputs that resolved on a no-change deploy instead of keeping the whole bag when one fails

### LINKED ISSUE #2771: fix(deploy): a resolvable Output added beside a persistently failing one never lands on the no-change path, so cdkd diff --fail shows a real ADD forever
## What is wrong

While an Output keeps failing to resolve on every deploy — the shape issue go-to-k/cdkd#2740 records in `StackState.skippedOutputs` — a NEW sibling Output added to the template is previewed by `cdkd diff` as `ADD` on every run and never lands in state, because the no-change deploy path refuses to persist the outputs bag while any output in it is unresolved.

`DeployEngine.doDeploy`'s no-change branch (`src/deployment/deploy-engine.ts`, the `if (!hasChanges)` block) computes `resolutionFailed = Object.values(resolvedOutputs).some((v) => v === undefined)` and gates the Outputs-only persist on `outputsChanged = !resolutionFailed && !outputMapsEqual(...)`, keeping the PREVIOUS bag whole when any output failed (the issue go-to-k/cdkd#875 guard against overwriting good outputs with a partial map). So with `NeverResolves` still broken, adding `Plain2: { Value: 'x' }` next to it produces:

- `cdkd deploy`: "No changes detected", the bag is kept, `Plain2` is not persisted (the go-to-k/cdkd#2740 record for `NeverResolves` is written, but the record only describes the skipped key).
- `cdkd diff --fail`: `[+] Plain2`, rc 1 — on every run, until either the broken output is repaired or a resource change routes the deploy through the changed-resources path, which persists the whole bag regardless of failures.

Not the go-to-k/cdkd#2740 phantom: the `ADD` is real (the deploy really will not publish `Plain2` on the no-change path), which is why go-to-k/cdkd#2740 deliberately did not touch it. It surfaced while reviewing that PR.

## Why it matters

The workaround is a resource change or the repair of the broken output; until then a legitimate Outputs-only addition is unreachable through `cdkd deploy` and `cdkd diff --fail` stays red for a change the user cannot make land. Pre-existing since go-to-k/cdkd#875; go-to-k/cdkd#2740 narrowed the surrounding phantom but leaves this row as-is.

## Proposed fix

Persist the bag PARTIALLY on the no-change path instead of the all-or-nothing choice, with three rules spelled separately:

1. A key that RESOLVED this pass writes its value (replacing the stored one).
2. A key that FAILED this pass keeps its stored value if it has one and stays absent otherwise — the go-to-k/cdkd#875 guard's intent (never overwrite a good value with `undefined`) survives, because a failed key never writes.
3. A key the template no longer DECLARES (deleted, or condition-suppressed this pass) is REMOVED, exactly as the changed-resources path drops it — the merge must not turn a deletion into a permanent `REMOVE` row.

`exportNames` follows the merged bag: this pass's aliases for the resolved keys, the carried aliases for the failed-but-kept keys, none for the removed ones (or re-run the alias pass over the merged bag).

`computeOutputsDiff` DOES need a look before this ships: its record-level exoneration (`recordProvesPostGhsa` in `src/analyzer/outputs-diff.ts`) reads ONE stored secret-reference expression as proof the whole bag was rewritten post-GHSA, on the premise that `resolveOutputs` rewrites every key on every deploy. A merged bag breaks that premise — a newly redacted key can sit beside a carried legacy plaintext — so the exoneration must become provenance-aware (per-key, or keyed on which keys this pass wrote) or the merge must refuse to carry a key the record cannot prove redacted. Regression test for exactly that mixed-generation bag, in both directions.

Tests: a no-change deploy with one failing and one new resolving output persists the new key and keeps the failed key absent; a STILL-DECLARED failing output retains its stored value (a new case — today's failure case in `tests/unit/deployment/deploy-engine-outputs-only-change.test.ts` retains `Existing` although the template no longer declares it, so rule 3 REVISES that expectation rather than keeping it green: it must show `Existing` removed and the index republished); a deleted output leaves the bag; the export set and the index follow the merged bag; the go-to-k/cdkd#2740 record is still written for the failed key; the go-to-k/cdkd#875 guard (a partial map never overwrites a good value) re-pinned on the still-declared case; the mixed-generation exoneration case above.

## Classification

Dup-check: searched open and closed issues for `no-change path partial outputs persist`, `resolutionFailed outputsChanged sibling`, `Outputs-only change unresolved output` — no hit; go-to-k/cdkd#875 (the Outputs-only persist) and go-to-k/cdkd#2740 (the phantom ADD) are the neighbours and neither covers a resolvable sibling blocked by an unresolvable one.
Session-fit: next (not this session) — a change to the no-change persist semantics with its own go-to-k/cdkd#875-era regression net and an integ arm; found during go-to-k/cdkd#2740's review, out of that PR's scope.
Severity: medium — an Outputs-only addition cannot land while any sibling output is broken, and `cdkd diff --fail` stays red for it; condition-specific (a persistently failing output) with a workaround (repair it, or make any resource change).
Effort: medium (M) — the three-rule merge on the no-change path, the export-set consequence, the provenance-aware exoneration in `computeOutputsDiff`, unit cases on the new rules, the go-to-k/cdkd#875 guard and the mixed-generation bag, one integ arm.
Estimate: ~3-4 h — the export-set semantics and the exoneration rework for a merged bag eat most of it.

nix-tkobayashi:
## Summary

A resolvable Output added beside a persistently failing one never landed on the no-change deploy path, so `cdkd diff --fail` showed a real `[+]` row for it on every run. The no-change branch used to keep the previous outputs bag whole whenever any output failed.

It now persists a merge, decided in the new `src/deployment/no-change-outputs-merge.ts`:

1. A key that resolved this pass writes its value.
2. A key that failed keeps its stored value, or stays absent if it had none. Its literal export alias is carried only when the previous record published it.
3. A stored key this pass did not produce (deleted, or condition-suppressed) is removed, as the changed-resources path already does.

The export set and the exports index follow the merged bag.

Two shapes still keep the whole previous bag, and the warning names why:

- **A failed output with a stored value whose `Export.Name` is intrinsic.** The alias it published cannot be named, so rule 3 would remove a live export. An output that never had a stored value never published an alias, so it does not refuse.
- **A merge that would carry a value into a bag it gives its first secret-bearing expression.** `computeOutputsDiff` exonerates an unaccountable stored key when any stored value is such an expression. The check runs twice: on the merge, and again on the bag as the save will redact it, after the observed-state drain, so a needle recorded late cannot slip between them. The predicate now lives in the merge module and `outputs-diff.ts` imports it.

The save no longer positions a carried key, or a bag kept whole, by today's template. `redactByPath` returns a whole-secret-expression source leaf verbatim, so a carried value could be persisted as a reference it never came from.

`cdkd diff` stays the conservative side: a bag it cannot fully resolve still suppresses the Outputs section. One shape where that now disagrees with the deploy is filed as go-to-k/cdkd#3101.

## Verification

- Unit: a pure-rule suite with a case per rule, refusal and gate; engine cases for the issue's shape, the still-declared go-to-k/cdkd#875 guard, alias carry, both refusals and their warnings, an export-set change on the merged path, both position-source arms, and a needle recorded during the observed-capture drain; a deploy-then-diff round trip for the mixed-generation bag in both directions, with a control showing the unguarded bag exposes the plaintext in the returned rows.
- Mutation probes (author-reported; the scripts and logs are not committed): one mutation each, byte-exact restore, run on the pre-rebase tree in three rounds (the initial change, the review fixes, the proxy-pass fixes). Every guard and gate the change adds or keeps reddens at least one case; three guards no input could reach were removed as dead rather than tested (the carry's already-written check, the self-named alias early return, and export-set de-duplication), each with a comment saying why.
- Revised expectations that encoded the old whole-bag keep: the `Existing` failure case in `deploy-engine-outputs-only-change.test.ts` (now removed and the index republished, as the issue's rule 3 says), `(f)` in `deploy-engine-skipped-outputs.test.ts`, the default-mode case in `deploy-engine-strict-getatt.test.ts`, the spelling fence in `outputs-diff.test.ts`, and the writer allow-list in `deploy-engine-cross-stack-read-writers.test.ts`. The two `deploy-engine-outputs-generation-rules.test.ts` cases that need a previous-generation bag now reach it through the intrinsic-name refusal with a stored value.
- Review (author-reported): spec, code, test and security reviewers (one round), then Codex on the fix delta (NO ISSUES) and three maintainer-checklist proxy rounds. Round 1 of the reviewers found one security blocker (a late needle during the observed-capture drain could give the saved bag its first expression beside a carried value), one code finding (the save positioned carried keys by today's template), and test gaps; all are fixed in this PR. The spec review's `cdkd diff --fail` false negative is filed as go-to-k/cdkd#3101.
- Integ: `output-never-resolved-diff` gains phase 4b (an output added beside two failing ones lands, record unmoved, `diff --fail` exits 0) and phase 4c (a partial repair publishes its key while the sibling still fails); its offline premise compares each new variant's synthesized template to its base whole.
- Real AWS (us-east-1, on the rebased head), all `verify.sh` rc 0 with a clean destroy:

  | Fixture | Result | Duration |
  | --- | --- | --- |
  | `output-never-resolved-diff` (phases 4b / 4c included) | PASS | 295 s |
  | `lambda` (broad set) | PASS | 107 s |
  | `export` (broad set) | PASS | 269 s |
  | `secrets-dynamic-ref` | PASS | 529 s |

  After the run only the shared `_index/` remained in the state bucket; the `lambda` and `export` event-log prefixes were swept.

Closes #2771


go-to-k:
Merged, thank you. The three-rule merge module with its two refusals stated up front, the save-time re-check after the observed-capture drain, and the per-arm test matrix made this straightforward to verify: four reviewers (spec / code / test / security) and three real-AWS runs from this side (output-never-resolved-diff, lambda, export) all came back clean. The handful of polish nits they raised are small enough that I will carry them in a follow-up rather than ask for another round.

Issue #3083, opened 2026-09-14
cdkd state show renders a non-finite lock expiresAt as `expired NaNmNaNs ago` while the contention message says "at an unknown time"


go-to-k:
## Summary

Two renderers of a lock's `expiresAt` disagree on a non-finite value. `formatRemaining` in `src/state/lock-contention-message.ts` prints "at an unknown time" when the deadline is not a finite number, while `formatLockSummary` in `src/cli/commands/state.ts` (and the `LockManager` sites that go through `formatDuration`) print `expired NaNmNaNs ago` for the same input. `cdkd state show` on a lock whose `expiresAt` is `{}`, `"soon"`, or an object whose number coercion throws (`NaN` since go-to-k/cdkd#3067) therefore shows a literal `NaNmNaNs`, which reads as a rendering bug rather than as "the deadline is unreadable".

Pre-existing: `{}` already produced this row before go-to-k/cdkd#3067; that PR only widened the population that reaches it (the throwing coercion used to abort the read instead). Surfaced by the security reviewer on that PR as a cosmetic, non-blocking finding.

## Expected

One spelling for a non-finite deadline across the lock renderers — "expired at an unknown time" or similar — with `formatDuration` never receiving `NaN`. `docs/cli-state.md` currently documents the `NaNmNaNs` row and `tests/unit/cli/state-record-shape.test.ts` pins it, so both move with the fix.

## Where

- `src/cli/commands/state.ts` `formatLockSummary` / `formatDuration`
- `src/state/lock-manager.ts` — the expired-lock takeover warning and the two refusal lines that format `now - expiresAt`
- `src/state/lock-contention-message.ts` `formatRemaining` (the spelling to converge on)

## Classification

Session-fit: now (this session) — re-classified minutes after filing: the maintainer pointed out the session that reviewed go-to-k/cdkd#3067 already holds every site in context, so the lane is cheaper now than in a fresh session
Severity: low — a non-finite lock deadline renders as `NaNmNaNs` instead of "unknown"; the expiry decision itself is unaffected
Effort: small (S) — replace the arithmetic with a finite check at four sites, update one doc row and one pinned test; unit suite only, no integ scope
Estimate: 1 hour — most of it is locating every `formatDuration(now - expiresAt)` spelling and keeping the docs row and the pinned test in step
[...]

PR #3108, opened 2026-09-14
chore(session-report): make `now` the default Session-fit; `next` needs one of three reasons

go-to-k:
## Summary

At the wrap of nearly every recent session the maintainer has had to ask whether the leftover would not be cheaper to finish in the SAME session, with the context already loaded — and every time the answer was yes: the item was re-classified `now` and done there (go-to-k/cdkd#3083 is the latest, ~25 min because every file it touched was already read). This PR makes the rule pre-answer that question so it never needs asking again.

**`.claude/rules/session-report.md` → Session-fit** (the single source of truth):

- **`now` is the DEFAULT.** The CONTEXT TEST is written BEFORE the decision: list the files the fix touches or must read (tests and docs included); if this session read, edited or reviewed ANY of them — a reviewer's read set counts like an author's — the item is `now`. A fresh session pays the launch probe, install, build, module read and evidence re-derivation before its first edit; this session pays the edit alone.
- **`next` needs ONE of three reasons**, with stated precedence: (a) a NEW live-AWS verifier (an integ fixture, `Effort: large`) must be written — a unit case never qualifies; (b) external input (quota, upstream fix, credentials this host lacks, a maintainer decision already asked through `AskUserQuestion` and unanswered); (c) a COLD subsystem — nothing the fix touches or must read was read this session AND no `now` criterion fires. (a) and (b) ask whether the work CAN finish here and are decided first; (c) is what the test decides.
- The former third `next` arm — "an independent subsystem with no file overlap" — was the escape hatch every wrong `next` took, and is gone. "Right after a merge, `next` is the default for RESIDUALS" is deleted and reversed: residuals of a just-merged lane are the hottest context there is.
- **Nothing about the SESSION is a reason**: its length, the context left, "it has done enough", a wrap report already drafted, the PR already merged. The wrap reflex (file → classify → close) fires exactly when the context is richest, which is why it produced `next`. Before the final report, the context test is re-run on every `next`.
- Session-state clauses are reconciled with the new list: of the three cited incidents only "held by another open PR's diff" survives (reason (b), ending at that merge); "no integ run budgeted" and "no file overlap" are no longer reasons. "An integ that FAILS" is an `Estimate` line, not a reason. Both issue-body templates now show `now | next` with the context test first.

**Mirrored where the decision is actually made**: CLAUDE.md's wrap bullet; `/work-issues` triage §3-b (pick-time deferral), retro §10-0 (the promotion check — the edited-file query is a LOWER bound on the run's read set, so the context test runs beside it) and §10-c (session budget is not a `next` reason for a mirror landing); filing.md (a sweep residue is `next` only as reason (a)); `/verify-pr`'s residual-nit sweep (bucket (a) Fixed is the default; the "file a bundled follow-up issue NOW" fallback is gone).

**Budgets**: CLAUDE.md + session-report.md sit under the 58,200 B injected-context ceiling (58,196), funded by compressing incident narratives in the rule file to their citations and trimming restated field spellings from the CLAUDE.md bullets. `skill-file-payload.test.ts`'s MEASURED record and `MIN_REFERENCE_CORPUS_BYTES` (156_500 → 157_700) are re-derived per that file's own convention for non-leader growth.

Same change lands in cdk-local and cdk-real-drift in this session (their 10-c forbids raising the floor, so there the additions were paid by compression instead).

## Review

One read-only reviewer per repo (agent-instruction files get no down-bias). All three reported no blockers; every minor and nit was fixed in the same commit: precedence between the context test and reasons (a)/(b) stated; "unit case" removed from reason (a); "a maintainer decision" narrowed to one already asked via `AskUserQuestion`; "touches" widened to "touches or must read"; the issue-body templates default to `now | next`; "the one case where `next` is genuinely right" → "reason (a)"; "an integ that FAILS … defer on those" reworded; session-state clauses reconciled; the verify-pr nit-sweep fallback and the retro mirror-issue exception reworded.

## Test plan

- [x] `vp run check`, `vp run typecheck:test`, `vp run build`
- [x] `vp test run` — full unit suite green, rooted in this worktree
- [x] `claude-md-size.test.ts`, `rule-file-payload.test.ts`, `skill-file-payload.test.ts` green (the three size fences this change moves)
- [x] No `docs/**` page describes Session-fit; `hooks.md`'s pointer to the criteria still holds
- No runtime behavior change, so no changelog entry and no integ run

https://claude.ai/code/session_01P7zsF3eqvDGYDVQe6cUJpG


go-to-k:
Merged — thanks. Two review rounds converged on the precedence wording, and both siblings landed the same rule in the same session.

PR #3112, opened 2026-09-14
chore(session-report): a new fixture is never a `next` reason; loose ends and `Severity: high` are `now`

go-to-k:
## Summary

Follow-up to go-to-k/cdkd#3108, on the maintainer's correction. That PR kept "a NEW integ fixture must be WRITTEN" as `next` reason (a) — inherited from the calibration paragraph's "what is genuinely expensive". It is the wrong direction: a fixture is written cheapest while the subsystem is loaded, and deferred it is exactly the piece that never lands.

**`.claude/rules/session-report.md` → Session-fit**:

- `next` has **two** reasons: **(a) external input**; **(b) cold AND heavy** — nothing the fix touches or must read was read this session, no `now` criterion fires, AND doing it here is clearly worse than fresh (it needs a large body of context this session would load from zero anyway, or the context it holds would degrade the work). Cold alone is `now`. (b) is legitimate and never forced through — but must stay RARE: its reason names the context needed and why THIS session is the wrong one to load it; fired twice in one run, or on a loaded file, it is the reflex, not the reason.
- Two `now` criteria added: **leaving it loose compounds** (an unwritten fixture, a half-landed pattern, a guard with a known hole — the cost of undone grows with every session); and **`Severity: high`** (a wrong result, data loss, a security surface), which only (a) overrides.
- Writing a new fixture is `Effort: large` — a cost to record, never a Session-fit input. "Do not collapse Severity into Session-fit" now reads that `Severity` says what a USER suffers and `Session-fit` what THIS session does.

Every lettered reference follows: CLAUDE.md's wrap bullet, triage.md §3-b, filing.md (a sweep residue is `next` only on external input; a fixture it needs is written now), wrap-up.md §11. Funded under the 58,200 B ceiling by compressing the PR-shaped and session-state incident paragraphs to their citations; MEASURED / `MIN_REFERENCE_CORPUS_BYTES` re-derived (157_950 → 158_100). Same change lands in cdk-local and cdk-real-drift in this session.

## Test plan

- [x] `vp run check`, `vp run typecheck:test`, `vp run build`
- [x] `vp test run` — 1003 files / 22389 tests green, rooted in this worktree
- [x] The three size fences green
- No runtime behavior change

https://claude.ai/code/session_01P7zsF3eqvDGYDVQe6cUJpG


go-to-k:
Merged — thanks. The maintainer's correction (a new fixture is never a reason to defer) is what this one carries; the review round then closed the last two escape hatches in (b).

Issue #3113, opened 2026-09-14
test(integ): remove the Fn::Join-line exception from nested-stack-secret's verbose scan once the log masks it


nix-tkobayashi:
## Summary

`tests/integration/nested-stack-secret/verify.sh` on go-to-k/cdkd#3062's branch (`fix/3062-nested-parameter-object-carry`) scans the phase-1 `--verbose` output with a deliberate exception for go-to-k/cdkd#3100. `scan_verbose_output` drops the parent resolver's `Resolved Fn::Join:` lines before scanning for `PIN_JOIN_FRAMED_VALUE`. It then requires that a dropped line carry that plaintext, and fails otherwise, naming go-to-k/cdkd#3100 as the likely cause.

go-to-k/cdkd#3100's fix masks that line. Once both changes are on one tree, the requirement fails by design: the fixture's own signal that the exception must go. This issue is that follow-up, and the integ half of go-to-k/cdkd#3100's verification plan, which its PR could not take because the fixture change exists only on the go-to-k/cdkd#3062 branch.

## Impact

Whichever of the two PRs merges second leaves `/run-integ nested-stack-secret` failing in phase 1 until the exception is removed. The masking itself is proven meanwhile by the go-to-k/cdkd#3100 PR's own arm in `tests/integration/secrets-dynamic-ref/verify.sh` Phase 1, whose `DB_PORT_JOIN` is the same CDK `secretValueFromJson` `Fn::Join` shape over a two-character value. The PR that merges second can carry this change instead, in which case this issue closes with it.

## Proposed change

In `scan_verbose_output` on a tree carrying both changes:

1. Put `PIN_JOIN_FRAMED_VALUE` back into the full-text plaintext scan beside the other values.
2. Remove the go-to-k/cdkd#3100 exception block: the `Resolved Fn::Join:` line filter, the requirement that a dropped line carries the plaintext, and the filtered-text scan, together with the comments that describe them.
3. Keep the existing `Resolved Fn::Join:` presence check, and add an assertion that such a line in the phase-1 output reads `port:***` for the `SubFloorPinJoin` parameter's framed value, so the negative scan cannot pass because the line stopped firing or stopped masking.

## Verification plan

- `/run-integ nested-stack-secret` on a tree with both changes passes, with the full-text scan including `PIN_JOIN_FRAMED_VALUE`.
- With go-to-k/cdkd#3100's resolver change reverted locally, phase 1 fails on the full-text scan, which shows the restored needle is live.

[...]

Issue #3100, opened 2026-09-14
fix(resolver): the Fn::Join / Fn::Sub debug lines print a 1-3 character secret embedded in the resolved value at --verbose


nix-tkobayashi:
## Summary

At `--verbose`, the resolver's `Resolved Fn::Join:` / `Resolved Fn::Sub:` debug lines print a 1-3 character secret in the clear when an intrinsic embeds it in a longer string. `port:` + `secret.secretValueFromJson('pin')` synthesizes an `Fn::Join` that resolves to `port:k3`, and the line reads `Resolved Fn::Join: port:k3`.

## Where it happens

Both lines (`src/deployment/intrinsic-function-resolver.ts`, `resolveJoin` and `resolveSub`) pass the assembled result through `maskSecretsForLog`, which calls `maskSecretsInText` over the inherited bag and then `recordedSecretValues`. `maskSecretsInText` masks a recorded plaintext as a SUBSTRING only at or above `MIN_NEEDLE_LENGTH` (4); below it, only a message that IS the whole secret is masked. `k3` is recorded, but `port:k3` is neither the whole value nor a 4+ character needle, so the line prints it.

The floor is deliberate: a 1-3 character needle masked as a substring would rewrite unrelated text in every line. So the fix cannot be "drop the floor". What the resolver holds and the floor lacks is POSITION, but not always at the part level: in the measured fixture no single part is a dynamic reference. The `{{resolve:...}}` token is complete only after the parts are joined, and `resolveDynamicReferences` substitutes it into the assembled string. A position-aware mask therefore has to carry the spans that final substitution writes, not only the spans of parts or `${}` placeholders that resolved on their own.

## Measured

On go-to-k/cdkd#3062's branch, 2026-09-14, the `nested-stack-secret` integ gained a parent parameter spelled as `cdk.Fn.join('', ['port:{{resolve:secretsmanager:cdkd-nested-secret-', cdk.Aws.ACCOUNT_ID, ':SecretString:pinjoin::}}'])` over a 2-character value. Its `--verbose` scan (`scan_verbose_output`) failed in phase 1 on this line and no other, both with origin/main's recorder and with the fix. A temporary diagnostic printed the matching line with the value replaced by its name: `[IntrinsicFunctionResolver] Resolved Fn::Join: <PIN_JOIN_FRAMED>`.

[...]

Issue #3079, opened 2026-09-13
secret-redaction: two distinct secrets sharing one literal frame and one sub-floor value make the loser's nested-stack child leaf persist the survivor's expression


go-to-k:
## Where

`recordNestedStackParameterExpressions` (`src/deployment/secret-redaction.ts`), the sub-floor carry go-to-k/cdkd#3070 adds for go-to-k/cdkd#2745's nested-stack site. Residual (a) on its docstring, filed as an issue on the security reviewer's request so it is tracked rather than only documented.

## Shape

A parent nested-stack row spells two `Parameters` with the SAME literal frame around two DIFFERENT secrets that resolve to the SAME 1-3 character value:

```yaml
Parameters:
  PinA: 'port:{{resolve:secretsmanager:A:SecretString:pin::}}'   # -> port:q7
  PinB: 'port:{{resolve:secretsmanager:B:SecretString:pin::}}'   # -> port:q7
```

The bag is keyed by plaintext, so `q7` has one survivor token. Condition (iii) records only the survivor's frame; the loser's frame keys identically under (iv) (`resolvedPairsOf` holds the loser's pair, so `resolvedPlaintextOf(secrets, B) === 'q7'`), so the row records `'port:q7' -> 'port:{{resolve:...A...}}'` alone. In the child, BOTH `{Ref: PinA}` and `{Ref: PinB}` leaves match that whole-value entry and persist A's expression.

## Consequence

Before #3070 both child leaves persisted `port:q7` in plaintext. After it, `PinB`'s child leaf persists A's expression — a wrong reference. After secret B rotates, `cdkd rollback` / `drift --revert` of the child re-resolves A's value into B's live property. Same class as go-to-k/cdkd#2485 (top-level, closed), now at the nested-stack carry.

## Direction

The recorder has both tokens beside their frames; a per-leaf association (the `CrossStackAssociations` table the first walk already writes for whole-token sources) keyed by child parameter NAME would let `inheritedSecretsCarriedBy` / `recordInheritedParameterSecrets` bind each child `Ref` to its own expression instead of the map's survivor. Needs the (iv) reasoning re-done, since a per-name entry no longer needs the row-wide one-frame rule.

## Repro

`tests/unit/deployment/secret-redaction-nested-parameter-source.test.ts` has the survivor case with matching frames (asserts the current behaviour); the fixture `tests/integration/nested-stack-secret/` would need a second secret with the same 2-character value.

[...]

PR #3115, opened 2026-09-14
fix(secret-redaction): carry a sub-floor Fn::Join / Fn::Sub nested-stack parameter frame to the child

### LINKED ISSUE #3062: secret-redaction: a nested-stack parameter embedding a 1-3 character secret through an Fn::Join / Fn::Sub source still reaches the child in plaintext

nix-tkobayashi:
Closes #3062

## Summary

A 1-3 character secret a parent embeds in a nested-stack `Parameters` value through an `Fn::Join` / `Fn::Sub` source (what `'port:' + secret.secretValueFromJson('pin')` synthesizes) reached the child's `state.json` as `port:q7`. The sub-floor carry go-to-k/cdkd#2745 added required the parameter's SOURCE to be a string: condition (i) certified a leaf by comparing the positioned copy against the source text, which an object source does not have. The parent's own row was already framed; only the carry was missing.

## Change

`recordNestedStackParameterExpressions` (`src/deployment/secret-redaction.ts`):

- For an object source, condition (i) takes the spelling the position pass WROTE for the leaf, admitted only when the pass rewrote it and the source's own token spells a secret service (below).
- That is sound beside condition (ii). On a value the scan is silent on, the cross-stack and skeleton arms both need the whole leaf to be a key of the map, which would make the scan fire, so the frame arm is the only writer that can have rewritten the leaf.
- The same spelling feeds (iv)'s frame gathering, so a literal and an object spelling of one frame around one token count as one spelling.
- A frame that is not wholly literal (a `Ref` outside the token) is still refused by the frame arm, so (i) refuses it and the child leaf keeps the plaintext. Residual (e) on the docstring now says so.
- A child record persisted before this change keeps the plaintext until the child is next redeployed. A parent deploy whose child row is unchanged never re-runs the child, and `cdkd scrub` cannot repair it, since it walks the child's stored bag with no inherited bag (residual (f)).

- **The object arm certifies only a source whose OWN token spells a secret service.** The source is rendered the way the frame arm renders it, and its single token must start with `{{resolve:secretsmanager:` or `{{resolve:ssm-secure:` in literal text. The check reads the source, never what the pass wrote. The frame arm matches among RECORDED references, and a public `ssm` parameter records none. So a public leaf can match a recorded secret sibling, and the arm writes the sibling's reference into the parent's record; that is the frame arm's own documented residual. Carrying it would hand the wrong reference to the child, where `cdkd rollback` / `drift --revert` re-resolve it. Two source shapes allow it: a token spelling `ssm:`, and a token whose service is itself an intrinsic part (`port:{{resolve:` + `{Ref: Svc}` + `}}`). A SecureString spelled either way is indistinguishable from the recorder's side, so both stay uncarried. That residual is stated on the docstring and in the changelog. `ssm-secure` is accepted because the resolver never resolves a public parameter through it. The refusal sits in `frameSpellingOf`, so a refused leaf also counts as unframed when (iv) gathers frames.

## Review

3-axis (498 LOC / 7 files; the recent-defect bias fired on `secret-redaction.ts`) plus the security reviewer.

- Spec: clean.
- Code: one blocker, the public-`ssm` wrong reference above, found by a probe. The security reviewer reproduced the same shape and rated it minor. Fixed by the `secretsmanager`-only check.
- Code, minor and nits, all fixed: two docstring sentences the code did not support (the pair gate is implied on the object arm; a state-derived source CAN be an intrinsic object after `cdkd import`), the rules clause losing "literal", a fixture comment naming the wrong failing check, a local in the wrong branch.
- All three other reviewers: the verbose-scan exclusion hid every other printer of the value. `scan_verbose_output` now drops only `Resolved Fn::Join:` lines, fails if none exists, and scans the rest for `PIN_JOIN_FRAMED_VALUE`.
- Test, minor, all fixed: conditions (iii), (iv) and (v), plus `STATE_DERIVED_RULES`, had no case with an object leaf as the one being carried. One case was added for each.

Round 2, a re-review of that fix round by code, security and test:

- Code and security, one blocker found independently by both: the first fix checked the WRITTEN token, which is always a recorded secret's. A source whose service is itself an intrinsic part passes that check for a public leaf. Both reviewers ran it on `Fn::Join` + `{Ref: Svc}` and on `Fn::Sub` `${Svc}`. Fixed by reading the source's own rendered token, with one unit case per shape.
- Code, minor: `ssm-secure` was refused with no reason; it is now accepted, with a unit case. The verbose scan had stopped covering the other five values on `Resolved Fn::Join:` lines; it now scans the full text for those and the filtered text for the join value only.
- Test, medium: no case pinned that the refusal sits where (iv) sees it; a case now does. Low: a comment named a count the code does not keep, and the filtered text had no non-empty guard; both fixed.

Round 3, a re-review of the source-service check by code, security and test:

- Security: clean. It traced every path by which the leaf's own token could go unrecorded (fresh resolution, cache hit, a cross-region ARN, `skipDynamicReferences`, a failed or empty resolution, `ssm-secure` on a String parameter). Every successful, non-empty secret resolution records the pair in the pass's bag. `skipDynamicReferences` leaves the token unresolved and records nothing, which leaves no frame to carry, and an empty resolved value leaves an empty middle, which `singleSpanFrame` refuses.
- Code: clean on logic, after probing eight shapes (a placeholder inside the secret-id, a two-argument `Fn::Sub` service variable, two spans, and others). One nit, fixed: a comment line was over the print width.
- Test, medium: rendering unknown parts as an empty string passed every case. A fourth refusal row now covers a `Ref` inside the service prefix, and that mutant fails it. Low-medium, fixed: the verbose filter now also requires that a dropped `Resolved Fn::Join:` line carries the join value, so the exception fails loudly once go-to-k/cdkd#3100 is fixed. Minor, fixed: the `ssm-secure` case gained its premise line.

Round 4, a read-only re-review of that delta by test: the new row's premise and the `ssm-secure` premise hold. Low, fixed: the dropped-line check piped into `grep -q`, which under `pipefail` can take SIGPIPE and fail a correct tree; it now captures the lines first and greps a here-string. The same review pass corrected the per-arm test comment, which credited condition (ii) with refusing the cross-stack case.

Maintainer-checklist proxy pass (Codex, after the PR body reached NO ISSUES):

- Fixed: the value-scan per-arm case compared an object with a string, so its premise could not fail; it now asserts the scan's exact answer. The twin, (iv), (v) and unframed cases now assert their parent-record premise BEFORE the carry runs.
- Fixed: the fixture premise counted the account `Ref` anywhere in the join. It now pins ONE `secretsmanager` token in a wholly literal frame, with the `Ref` inside that token. The jq expression was run against the real synthesized template (true), against a `Ref` moved outside the token (false) and against an extra intrinsic part (false).
- Fixed: a test header and a describe title still credited (ii) for every per-arm case; docstring (iv) and (f) were incomplete; the changelog omitted the go-to-k/cdkd#3100 verbose-log residual; a comment said "three" spellings where there are four.
- Won't-do: the pass asked `.claude/rules/layout-deployment.md` to name the object arm beside its "sub-floor LITERAL frame arrives as a whole value" clause. The path's rule-file budget has 11 B left on `origin/main` after go-to-k/cdkd#3111, and the shortest wording that stays true (it must say sub-floor, and must not claim every join frame arrives) needs 15 B. That clause is left as on `origin/main`, where it is still true, and the object arm's rule and residuals live in the recorder's docstring.
- Won't-do, narrowed in round 2: the pass asked for cases pinning four guards in `frameSpellingOf`. `isPlainObject(sourceLeaf)` and `typeof written !== 'string'` narrow the types the calls below take. `written === resolvedValue` refuses an unrewritten leaf, which `singleSpanFrame` refuses as well, so no case can fail on its deletion; the comment above the function says so.
- Fixed in round 2 of the pass: round 1's won't-do also claimed the SPAN COUNT refuses nothing the frame arm has not refused, and that was wrong. The value scan also rewrites an object-sourced leaf whose whole value is a recorded plaintext, and frame gathering asks for the spelling before (ii) refuses that value. A source with no `{{resolve:` text then renders zero spans, and without the guard the recorder throws. A regression case now covers that shape and the comment states it.
- Round 3 of the pass, the last: two findings, both addressed. The span count's more-than-one arm survives a `=== 0` mutant. It is inert rather than unreachable: the value-scan rewrite that reaches it is refused by (ii) for every leaf holding the value, so no entry or association can differ. The comment states this, and no case claims to pin it. The fixture premise excluded `}}` but admitted a lone `}` in the opening part, which leaves the production scanner zero spans. It now allows no brace outside the token's `{{` and closing `}}`; the jq was run against the real synthesized template (true) and five negatives (lone `}`, stray `{`, a second token, a `Ref` outside the token, an extra part), all false.

## Verification

- Unit, `tests/unit/deployment/secret-redaction-nested-parameter-source.test.ts` (69 cases in the file):
  - one positive case per object spelling, and the 1-4 character range;
  - a literal and an object twin over one middle, in both survivor orders;
  - the no-pair refusal, and one case per other arm (skeleton, cross-stack, value scan);
  - the public-`ssm` refusal, and four refusals where the service is an intrinsic part or sits beside one;
  - an `ssm-secure` carry, and the refused leaf counting as unframed;
  - object-leaf cases for (iii)'s second arm, (iv), (v) and `STATE_DERIVED_RULES`.
- Pre-fix control: the ten positive cases run against origin/main's recorder, all ten fail.
- Mutation probes, one mutation each, restored byte-exact:
  - Object arm certifies nothing: 12 cases fail.
  - Frames loop treats an object source as unframed: 5 cases fail, both twin cases among them.
  - Condition (ii) skipped for an object source: 4 cases fail, the skeleton and value-scan per-arm cases, the 4-character control, and the embedded-secret refusal. The cross-stack per-arm case does not fail: its `Fn::ImportValue` source renders no token, so the source-service check refuses it first, as the test's comment states.
  - No source-service check: 6 cases fail, the public-`ssm` refusal, the four intrinsic-service refusals and the unframed case.
  - The check reads the WRITTEN token instead of the source's: 4 cases fail, the four intrinsic-service refusals.
  - Only `secretsmanager` accepted: 1 case fails, the `ssm-secure` carry.
  - The frames loop spelled without the check while certification keeps it: 1 case fails, the unframed case.
  - Unknown parts rendered as an empty string: 1 case fails, the `Ref` inside the service prefix.
  - Dropping the `written !== resolvedValue` test: no case fails. `singleSpanFrame` refuses a spelling equal to its bag as well, which the comment beside the test states.
- Integ `nested-stack-secret`, pre-fix control: the fixture unchanged, `src/deployment/secret-redaction.ts` swapped for origin/main's and rebuilt. It failed in phase 1 at the `--verbose` scan (`--verbose printed a resolved secret plaintext`). The scan's output is not captured in the log, so which needle matched is inferred rather than observed: the 05:31 run on the go-to-k/cdkd#3079 tree passed the same scan, and `port:k3` is the only needle this change adds. The run never reached the `PinJoinParam` state assertions. Its surviving state versions on both prefixes were swept to 0 by hand before the fixed run.
- That first control, and the first fixed run, both failed at the same scan. A masked diagnostic then showed the match: the PARENT resolver's `Resolved Fn::Join:` debug line prints `port:k3`, because the log mask has the same 4-character substring floor as the value scan. That is a resolver log residual independent of the state carry, filed as go-to-k/cdkd#3100. Its fix masks that line, so whichever of go-to-k/cdkd#3062 and go-to-k/cdkd#3100 merges second removes the exception block and returns `PIN_JOIN_FRAMED_VALUE` to the full-text scan (tracked as go-to-k/cdkd#3113). `scan_verbose_output` now scans the full text for the other five values, and the text without `Resolved Fn::Join:` lines for `PIN_JOIN_FRAMED_VALUE`; the state, diff and parent-record assertions cover the frame as before.
- Integ `nested-stack-secret`, pre-fix control after the scan scoping: FAILED in phase 1 at the child-state scan, `the child's state.json carries the Fn::Join-framed sub-floor plaintext` -- the named defect, measured. Its surviving state versions on both prefixes were swept to 0 by hand.
- Integ `nested-stack-secret`, final tree (after the third proxy round, on `b6ebcc07`): PASS in 281 s, with the brace-tight premise guard. Destroy deleted 9 + 4 with 0 errors, all eleven SSM parameters are gone, and 0 versions survive on either prefix.
- Integ `nested-stack-secret`, after the first proxy round: PASS in 287 s, with the first tightened premise guard. Destroy deleted 9 + 4 with 0 errors, all eleven SSM parameters are gone, and 0 versions survive on either prefix.
- Integ `nested-stack-secret`, round 4 plus the SIGPIPE-safe capture: PASS in 294 s. Phase 1 and phase 2c both ran the check that a captured `Resolved Fn::Join:` line still carries the join value. Destroy deleted 9 + 4 with 0 errors, all eleven SSM parameters are gone, and 0 versions survive on either prefix.
- Integ `nested-stack-secret`, round 4 tree before the capture fix: PASS in 298 s.
- Integ `nested-stack-secret`, round 3 tree (commit `255a59a6`, the source-service check and the split verbose scan): PASS in 292 s. The verbose scan checks the full text for five values, and checks the `Resolved Fn::Join:`-filtered text for `PIN_JOIN_FRAMED_VALUE`. Destroy deleted 9 + 4 with 0 errors, all eleven SSM parameters are gone, and 0 versions survive on either prefix.
- Integ `nested-stack-secret`, round 2 tree (rebased onto go-to-k/cdkd#3093): PASS in 287 s.
- Integ `nested-stack-secret`, fixed (HEAD `496a4d66`'s tree): PASS in 288 s. The go-to-k/cdkd#3062 block passed both premise guards (the synthesized `SubFloorPinJoin` is an object, an `Fn::Join` carrying the account `Ref`), the parent row, the live value, the child's `properties` and `observedProperties`, the phase-2 `cdkd diff --recursive --fail` exit 0, and the phase-2c UPDATE arm. Destroy deleted 9 + 4 resources with 0 errors, all eleven stack-owned SSM parameters are gone, and 0 object versions survive on either state prefix.


go-to-k:
Merged as a386124b — thank you. The object-arm certification reading the SOURCE's own token (not the written one) was the right cut, and the fixture's brace-tight premise guard is what makes the integ trustworthy on re-runs. I pushed one commit on top (a maintainer nested-stack-secret re-run in the ledger, plus three comment-level tightenings from the merge review); no code change.

For the record, two refactor-only review nits were declined rather than folded in: extracting the segments-to-text rendering (with the frame arm's NUL refusal) into one helper shared by `frameSpellingOf` and `positionByIntrinsicFrame`, and memoizing `frameSpellingOf` so an object leaf is rendered once instead of twice. Both are inert (the NUL case is refused one step later either way, measured), and a post-integ code edit for tidiness was not worth re-paying the real-AWS run.

Issue #3114, opened 2026-09-14
fix(resolver): two routes still print a sub-floor secret on the Resolved Fn::Join / Fn::Sub lines after the log twin


nix-tkobayashi:
## Summary

go-to-k/cdkd#3100's log twin masks a 1-3 character secret on the resolver's `Resolved Fn::Join:` / `Resolved Fn::Sub:` lines wherever a write in the same resolver pass put it there, and carries the mask through the pass by value (`LOG_TWINS_BY_PASS`). Two routes still reach those lines with the secret embedded and no twin, so the line falls to `maskSecretsForLog`'s needle floor (`MIN_NEEDLE_LENGTH`, 4) and prints it. Both were found by review of the go-to-k/cdkd#3100 change and left out of it deliberately. Each needs its own design, and neither is a regression against origin/main, which printed both lines in the clear.

## Route 1: a nested-stack child `Ref` to a parameter framed by an INTRINSIC in the parent

A parent parameter expression that is an `Fn::Join` or `Fn::Sub` framing a short secret (for example `port:` + a two-character `secretValueFromJson` key) resolves in the parent to `port:q7`. The child binds that string and treats `{ Ref: <Param> }` as a resolution product (`productLogTwin` / `logTwinOfProduct` in `src/deployment/intrinsic-function-resolver.ts`). A product is masked when a bag holds it WHOLE, or when a masked twin is registered for that exact string under the pass bag.

- A LITERAL parent frame does not leak: `recordNestedStackParameterExpressions` adds the whole framed value to the inherited bag, so the child matches `port:q7` whole and logs `***`.
- An intrinsic parent frame takes no such entry. `recordInheritedParameterSecrets` then copies the pair whole-value, or as a substring only at 4 characters or longer, so `port:q7` matches neither.
- The child context does carry the parent's pairs, as `inheritedSecrets`, but `logTwinBag` keys the registry lookup by the child's own `recordedSecretValues` whenever one exists, so a twin registered under the inherited bag is never searched.

The same missing whole-value entry is why the child's persistence redaction cannot position that embedded value either, which is go-to-k/cdkd#3062's residual. A fix for the log line must not alter persistence, and it does not by itself close the state side.

## Route 2: a string resolved in two stages loses its first stage's twin

[...]

PR #3116, opened 2026-09-14
fix(resolver): mask a sub-floor secret by position on the Resolved Fn::Join / Fn::Sub lines

### LINKED ISSUE #3100: fix(resolver): the Fn::Join / Fn::Sub debug lines print a 1-3 character secret embedded in the resolved value at --verbose

### LINKED ISSUE #3113: test(integ): remove the Fn::Join-line exception from nested-stack-secret's verbose scan once the log masks it

nix-tkobayashi:
## Summary

At `--verbose`, the resolver's `Resolved Fn::Join:` and `Resolved Fn::Sub:` lines printed a 1-3 character secret that the intrinsic embedded in a longer string. `port:` + `secretValueFromJson('pin')` printed as `port:q7`.

Both lines masked the resolved string with `maskSecretsForLog`, a needle mask. A needle shorter than `MIN_NEEDLE_LENGTH` (4) matches only when it is the whole text, and that floor stays. What the needle lacks is position, and only the code that writes the secret into the string holds it.

The resolver now builds a log twin beside the value at each such write, in `src/deployment/intrinsic-function-resolver.ts`:

1. The dynamic-reference substitution writes `***` into the twin for a secret verdict, on all three arms: a fresh lookup, a cache hit (the entry's own verdict), and a region-pinned sibling (the twin the sibling returns).
2. A string `Fn::Join` part or `Fn::Sub` variable resolves through that same substitution and keeps its twin. A template literal is never masked, even when it equals a recorded value.
3. An intrinsic part, variable or `Ref` / `GetAtt` placeholder is masked whole when its value is a recorded or inherited secret. A Join part's check waits until every part has settled, so a sibling part that records the secret later is still seen.
4. Every reference-bearing string and every Join or Sub result registers its masked twin for the pass, keyed by the pass's secret bag. An outer Join or Sub whose part is such a string keeps the inner mask. Only masked twins are registered, and two that disagree for one string register `***`, so a later write cannot unmask an earlier one.
5. `Fn::Split` splits the source string's registered twin by the same delimiter and registers each piece's twin, pairing pieces by position only when both splits give the same count and the delimiter does not occur in the mask, and masking every piece whole otherwise. The `Resolved Fn::Split:` line, which printed the same pieces, logs those twins. A source that is itself a recorded or inherited secret masks every piece.
6. Every line that prints a resolved value through the leaf mask (`maskValueLeaves`: the `Fn::Select`, `Fn::Split`, `Fn::Cidr` lines and the other callers) and the `Fn::Base64` line take the registered twin too, through a shared helper with the same whole-line guard, for object keys as well as values. The `Fn::Base64` encoding is masked whole when its input carries a mask, since it decodes back to that input. The thrown messages that interpolate a value through the leaf mask take the same position mask.

Both lines log the twin through the needle mask. When the needle mask also changes the value itself, the whole line is masked, because a 4+ character secret overlapping a masked span is no longer a contiguous needle in the twin.

The resolved value and state are unchanged, and no other log line changes. `resolveDynamicReferences` keeps its public signature.

## Verification

- Unit, in `tests/unit/deployment/intrinsic-resolver-sub-floor-log-mask.test.ts` (new): the issue's three spellings (a Join whose token completes only after assembly, with the secret id through a `Ref`; a Sub of the same; a Sub whose variable is a reference), a whole-reference Join part, and the literal negative for a Join part, a Sub variable and a Sub body. Further cases cover a public ssm value on the fresh and cache-hit arms, intrinsic parts and variables, a `Fn::Split` element, a Join over a Sub, a Sub over a Join, a `Fn::Select` over a reference string, the registry's pass scope, a later unmasked write, two disagreeing masks, a Join product whose sibling records later, the `GetAtt` and `Ref` placeholders, the inherited-only bag, the overlapping 4+ character secret, a list element carrying its own mask, and the cache-hit arm. Each log line is asserted as a whole string, so a mask-everything regression fails as well as a leak.
- Unit, for the maintainer review round: the `Fn.select(n, Fn.split(...))` idiom with its Join, Split and Select lines asserted whole; a split-secret piece picked by `Fn::Select`; a Select leaf whose registered twin splits a 4+ character secret; a `Fn::Split` over a whole inherited secret, and one case per bag (inherited only, recorded only); the `Fn::Base64` line with an unmasked control; a `Fn::Cidr` refusal message; and the Sub pseudo-parameter, escape and empty-placeholder arms, including that the empty arm logs no warning.
- Unit, in `intrinsic-resolver-assembled-secret-region.test.ts`: a two-character secret resolved by the region-pinned sibling is masked on the primary's Join line, a public ssm `String` resolved by the sibling prints, and a twin the sibling registers is visible to the consumer's Join.
- Mutation probes (author-reported; the script and logs are not committed): one mutation each against `intrinsic-function-resolver.ts`, byte-exact restore, run against the two test files above. Round 1 covered the initial change and killed 21 of 21. Round 2 covered the tree after the review fixes: 29 of 31 were killed. The two survivors were the post-drain product check, which gained a case whose lookup settles last, and the own-twin clause, reworked and pinned by two list-parameter cases; both were then killed. Each later Codex round's change was probed at the clauses it added: the `Fn::Split` pairing (6 of 6), its overlap guard (1 of 1), the registration moved into the dynamic-reference method (3 of 3, after one survivor got a pinning case), and the list stringification (6 of 6). The maintainer review round's clauses were probed the same way: 14 of 14 killed, one of them (the leaf helper's whole-line guard) only after its pinning case was added.
- Integ: `secrets-dynamic-ref` Phase 1 now deploys with `--verbose`, captured. It asserts that `Resolved Fn::Join: port:***` and `Resolved Fn::Sub: port:***` are present, that the framed pin is absent from the whole log, and that no `Resolved Fn::` line carries the bare pin. A bounded-pin helper is self-checked before the deploy. `DB_PORT_JOIN` is the CDK `secretValueFromJson` Join shape the issue measured.
- The issue's own integ plan, in `nested-stack-secret`: go-to-k/cdkd#3115 merged first with a verbose-scan exception that required the parent's `Resolved Fn::Join:` line to carry the framed pin. This PR, rebased onto it, removes that exception (go-to-k/cdkd#3113): `PIN_JOIN_FRAMED_VALUE` is back in the full-text plaintext scan, and the scan asserts that a `Resolved Fn::Join: port:***` line is present.
- Residuals, not regressions (origin/main printed both lines in the clear), filed as go-to-k/cdkd#3114: a nested-stack child `Ref` to a parameter the parent framed with an intrinsic, and a string resolved in two stages whose first stage's twin is not seeded into the second.
- Maintainer review, round 2 (go-to-k, on `df5740ac`): M5-M7 are in `b0f886a6` (the changelog no longer lists go-to-k/cdkd#3113 as a residual, object keys take the position mask, and a delimiter inside the mask never pairs split pieces), each new clause with a case; o4 and o5 are left, with the reasons in the reply.
- Maintainer review (go-to-k, on `ba8ba0b6`): M1-M4 are in this PR, o2 is taken, and o1 is left (it only over-masks); the reply on this PR has the detail. After those fixes (author-reported), Codex on the delta returned NO ISSUES on its second round, and three maintainer-checklist proxy rounds each found items, all fixed or recorded in that reply.
- Review (author-reported): spec, code, test and security reviewers, one round, no blockers. Their should-fix items are all in this PR: a later write could replace an earlier mask in the registry (code, security); the non-secret direction had no case, and the first spelling's case had no `Ref` (test); the region-pinned arm now takes the sibling's own twin (code); a string reaching a Join through another intrinsic, and a 4+ character secret split by a masked span (security); and the `nested-stack-secret` follow-up, filed as go-to-k/cdkd#3113 (spec).
- Codex, on the uncommitted delta (author-reported): 8 rounds, stopped at round 8 without reaching NO ISSUES. Rounds 1-3 and 5-7 each found a real gap, and each is fixed with a pinning case: the registry overwrite, a product checked before a sibling part recorded, the inherited-only bag, `Fn::Split` pieces, a Split piece splitting a 4+ character secret, a route that bypassed `resolveValue`, a region-pinned sibling registering on its own instance, and a list stringified into a part or placeholder. Round 4's shape, a public value that coincides with a secret the same Join records later, is documented in `logTwinOfProduct` as a decision. Round 8's two-stage re-resolution, together with a nested-stack child route found while enumerating the remaining inputs, is filed as go-to-k/cdkd#3114.
- Reviewer nits left as they are, each decided:
  - The twin's first-occurrence replace can put a value in the wrong place when template text literally equals an already-masked token (code). That affects only the order of values on a log line and discloses nothing.
  - Phase 1's framed-pin check scans the whole `--verbose` log, not only the `Resolved Fn::` lines (spec, code). The issue's plan asks for exactly that, so a leak at another log site fails the fixture too, and it passed live.
  - The docstring's divergence case (`{{resolve:${Pw}}}` over a secret `Pw`) has no unit case (test). The fixture's Phase 1b output exercises that shape live.
  - Same-pass over-masking by an equal string has no case (test). It only over-masks, and `logTwinOfProduct`'s doc names it.
- Real AWS (us-east-1), on the head rebased onto go-to-k/cdkd#3115, all `verify.sh` rc 0 with a clean destroy (0 errors):

  | Fixture | Result | Duration |
  | --- | --- | --- |
  | `nested-stack-secret` (the masked `Resolved Fn::Join: port:***` line asserted in phases 1 and 2c; 0 surviving state versions for parent and child) | PASS | 303 s |
  | `secrets-dynamic-ref` (Phase 1 `--verbose` arm: the pin masked on all 54 `Resolved Fn::` lines) | PASS | 555 s |
  | `lambda` (broad set) | PASS | 109 s |

  After the run (author-reported; the ledger does not record it) only the shared `_index/` remained in the state bucket, and the `lambda` event-log prefix was swept. The final PASS message in `nested-stack-secret/verify.sh` was reworded after the run, to drop its reference to the removed exception; no assertion reads it.

Closes #3100
Closes #3113


go-to-k:
Reviewed head `ba8ba0b6` with spec, code, test and security reviewers. The mechanism holds (no change to the resolved value, no unmask path, registry scoped by pass) and the Join / Sub lines are fixed. Not merging yet: four items below are required before merge, all in this PR. go-to-k/cdkd#3115 is clean and merges first (today), so please rebase onto it and fold go-to-k/cdkd#3113 in (M4).

**Required before merge**

- **M1 — `Resolved Fn::Select:` (and the other leaf-masked lines) print the plaintext the registry already masks.** `maskValueLeaves`'s string arm goes through `maskSecretsForLog` only, so `Fn::Select [0, {Fn::Split [',', {Fn::Join ['', ['port:', PIN_REF, ',x']]}]}]` (the `Fn.select(n, Fn.split(...))` CDK idiom) logs, measured on this head:
  ```
  Resolved Fn::Join: port:***,x
  Resolved Fn::Split: split by "," -> ["port:***","x"]
  Resolved Fn::Select: index 0 -> "port:q7"
  ```
  Two reviewers reached it independently (also via `Fn::Select [1, {Fn::Split [':', <dsn ref>]}]` -> `index 1 -> "q7"`). Fix: route `maskValueLeaves`'s string arm through `logTwinOfProduct({result: node, twin: node}, context).twin` (then the needle mask), which closes `Fn::Select`, `Fn::Equals`, `Fn::FindInMap` and `Fn::GetAZs` in one place. One unit case per the Select shape above, line asserted whole.
- **M2 — `splitLogTwins` does not see a source string that is WHOLE a recorded secret.** `sourceTwin` is a registry lookup only (`isRecordedSecretForLog` is never asked), so in a nested child `Fn::Split [':', {Ref: SecretParam}]` over a parent-passed `ab:cd` logs `Resolved Fn::Split: split by ":" -> ["ab","cd"]` and the consuming Join `ab/cd`, while the same `Ref` as a Join part or Sub variable is masked. Fix: `const sourceTwin = this.isRecordedSecretForLog(value, context) ? SECRET_MASK : ...registry...`, with a unit case.
- **M3 — the `Resolved Fn::Base64:` line prints the decoded plaintext.** `Fn::Base64` over `Fn::Join ['', ['port:', PIN_REF]]` logs `Resolved Fn::Base64: port:q7 -> cG9ydDpxNw==` on this head (the Join line above it is masked). Log the input through `logTwinOfProduct` / the registry like M1. The STATE side of the same site (the base64 form is persisted because `recordMaskOnlyValue` is gated on the needle mask) is a different reader and is filed separately as go-to-k/cdkd#3119 -- out of scope here unless you want to take it in the same push.
- **M4 — two test gaps, plus go-to-k/cdkd#3113.**
  - The region-pinned sibling arm has no negative case: replacing `foreign.twin` with `SECRET_MASK` at the `twin = twin.replace(fullMatch, () => foreign.twin)` line keeps all 49 cases green. One case in `intrinsic-resolver-assembled-secret-region.test.ts` with a public `String` parameter from the producer region embedded in a Join, asserting the value prints, closes it.
  - The Sub pseudo-parameter / `${!X}` escape / empty `${}` arms: setting their twin to `SECRET_MASK` keeps every case green. One case, `{ 'Fn::Sub': 'r:${AWS::Region}:${!Lit}:${}' }`, line asserted whole.
  - After rebasing onto go-to-k/cdkd#3115: remove the `Resolved Fn::Join:` exception from `tests/integration/nested-stack-secret/verify.sh` per go-to-k/cdkd#3113 (put `PIN_JOIN_FRAMED_VALUE` back into the full-text scan; assert the line reads `port:***`), and add `/run-integ nested-stack-secret` to this PR's runs beside `secrets-dynamic-ref` and a broad-set fixture. That closes go-to-k/cdkd#3113 with this PR.

**Optional (take or leave)**

- o1 — `splitLogTwins`: when the counts align but the delimiter occurs inside a mask (`*` delimiter, `***`), a secret-derived piece gets twin `''`; mapping `twinPieces[index] === '' && piece !== ''` to `SECRET_MASK` reads better. Over-masking only, no leak.
- o2 — `tests/integration/secrets-dynamic-ref/verify.sh` Phase 1: the `port:***` premise fails on an unfixed resolver with "line not logged" rather than "printed in the clear"; checking for the `Resolved Fn::Join: port:` prefix first would make the diagnosis match.
- o3 — go-to-k/cdkd#3114 route 1 also reaches `Resolved Ref to parameter: <Param> -> port:q7` (the `stringifyParameterForLog` site), not only the Join / Sub lines; I will add that to the issue body.

Everything else came back clean: the five decisions in the body, the value / state / journal / events / throw-site readers, the registry's pass scope, and the whole-line assertions (a mask-everything mutant fails 27 cases).

go-to-k:
go-to-k/cdkd#3115 is merged (a386124b), so M4's rebase + go-to-k/cdkd#3113 fold-in is unblocked.

nix-tkobayashi:
Thanks for the review. `df5740ac`, rebased onto go-to-k/cdkd#3115, addresses M1-M4. For the M4 fixture half, `nested-stack-secret/verify.sh` no longer drops the `Resolved Fn::Join:` lines: `PIN_JOIN_FRAMED_VALUE` is back in the full-text scan, and a `Resolved Fn::Join: port:***` line is asserted present. The PR now closes go-to-k/cdkd#3113 too. Real-AWS runs (`verify.sh` rc 0, clean destroy with 0 errors): `nested-stack-secret` 303 s, `secrets-dynamic-ref` 555 s, `lambda` 109 s (broad set). They ran on the working tree that `df5740ac` commits: the rebased fix commit `58b7ca43` plus the `nested-stack-secret/verify.sh` scan edit above, which was uncommitted during the runs. The run log (author-reported, not in the diff) shows the new `the framed pin's Join line is masked` OK line in phases 1 and 2c, and the runs were in us-east-1. After the runs, `df5740ac` added only the ledger rows and a rewording of the final PASS message, which no assertion reads.

- **M1:** `maskValueLeaves`'s string arm now goes through `logTextOfLeaf`: `logTwinOfProduct`, then `logTwinText`'s whole-line guard, then the needle mask. That covers every caller of `maskValueLeaves` in one place: the `Fn::Select`, `Fn::Split`, `Fn::Equals`, `Fn::FindInMap`, `Fn::GetAZs` and `Fn::Cidr` lines, and the VPC `Ipv6CidrBlocks` line. `maskValueLeaves` also feeds thrown messages (the `Fn::Cidr` argument refusals), so those take the same position mask; that is named in the helper's doc and the changelog, and pinned by a throw case. Cases: your `Fn.select(n, Fn.split(...))` shape with all three lines asserted whole, the split-dsn piece (`index 1 -> "***"`), and a leaf whose registered twin splits a 4+ character secret it still holds.
- **M2:** `splitLogTwins` asks `isRecordedSecretForLog` before the registry. Case: `Fn::Split` over a `Ref` to a parameter holding an inherited `ab:cd`, with the Split and Join lines asserted.
- **M3:** the `Fn::Base64` line logs its input through the same helper. When the input carries a mask, the encoding is masked whole, since it decodes back to that plaintext. Cases: your Join shape (`port:*** -> ***`) and an unmasked control. The state side stays with go-to-k/cdkd#3119.
- **M4:** a region-pinned public ssm `String` case in `intrinsic-resolver-assembled-secret-region.test.ts` asserts the value prints, and your `r:${AWS::Region}:${!Lit}:${}` case asserts the Sub line whole.
- **Found while closing these, also in this push:** the M2 case let either bag answer, because the `Ref` copies the inherited pair into the recorded one. Two cases now pin each bag alone, with nothing registered. The earlier Split overlap-guard case had become redundant under M2 (its source was a whole secret), so its source is now `hostA:<pin>,x`, which only the per-piece guard masks. The helper's doc and the changelog now describe the leaf-mask callers instead of listing an incomplete set. The `.claude/rules/layout-deployment-secrets.md` sentence on bounded masks is unchanged: it describes `maskSecretsInText`, which keeps its floor, and the rules budget on that path has 11 B of headroom.
- **Also closed:** the empty-`${}` case now asserts that no warning is logged. Skipping that arm falls through to a failed `Ref` that keeps the same text but warns, so the guard itself is pinned, not only its twin. The Phase 1 `resolved_line_carries_pin` self-check now probes the stripper's grammar, not examples: an empty parameter list, `;`-separated parameters, a `:` parameter, a private `?` parameter, and non-SGR final bytes (`K`, `h`, `A`). The `Fn::Cidr` result comment and the `Fn::Base64` recording comment no longer claim the needle mask is the only mask on those lines.
- **o1:** left as is. An aligned split giving `''` twin pieces needs the delimiter inside the mask with piece counts that still match; it only over-masks, as you say.
- **o2:** taken. Phase 1 now checks the unmasked `port:` prefix and the framed pin before the masked form, so an unfixed resolver fails as a leak.
- **o3:** noted, thanks.

Mutation probes on the new clauses (author-reported): one mutation each, byte-exact restore, 14 of 14 killed. One of them, the whole-line guard inside `logTextOfLeaf`, survived until the 4+ character case above was added. Your two M4 mutants (`foreign.twin` -> `SECRET_MASK`, and the three Sub arms -> `SECRET_MASK`), each whole-source bag alone, and skipping the empty-`${}` guard are among them and now redden a case.

go-to-k:
Round 2, reviewed head `df5740ac` with spec, code, test and security reviewers. M1-M4 are closed (the three round-1 shapes reproduce masked on this head; 21 mutants over the new clauses each redden at least one case; the two `splitLogTwins` clauses and the two bags are pinned independently). No blocker. Three small items remain, all required before merge — each is a one-line change or a doc line, so I expect this to be the last round; once it lands I run the three integs on my side and merge.

**Required before merge**

- **M5 — changelog residual list is stale.** `changelog.d/entries/2026-09-14-3100-sub-floor-join-sub-log-mask.md` still ends with `Residuals: #3113 (another lane's fixture scan) and #3114 (...)`, but this PR now closes go-to-k/cdkd#3113. Drop #3113 from the residuals (196 bytes of headroom under the 2000 cap).
- **M6 — object KEYS bypass the position mask, while the docs claim symmetry.** `maskValueLeaves` at `out[this.maskSecretsForLog(key, context)] = walk(child)` masks keys with the needle mask only; the `logTextOfLeaf` doc says it is shared by every caller of `maskValueLeaves`, and the `maskValueLeaves` doc says an unmasked key discloses exactly as much as an unmasked value. Measured: after `Fn::Sub port:${PIN_REF}` registers `port:q7 -> port:***`, `Fn::Select` over `[{ "port:q7": "v" }]` logs `{"port:q7":"v"}`. Fix: `out[this.maskSecretsForLog(this.logTextOfLeaf(key, context), context)]`, with one case. (Reachability with a real secret is low — keys are template literals or AWS attribute-map keys — but the code should match the claim, and the change is one line.)
- **M7 — `splitLogTwins` doc and code disagree on a delimiter inside the mask.** The doc says "a delimiter that occurs in the mask itself — no pairing can be trusted, so every piece is masked whole", but the code pairs by position whenever the COUNTS coincide: recorded secret `p*q*r*s` split by `*` gives 4 value pieces and `'***'.split('*')` gives 4 empty twins, so the line prints `["","","",""]`, each piece is registered `-> ''`, and an outer `Fn::Join` prints `Resolved Fn::Join: ---`. No disclosure (it prints less), but the registry now carries empty twins and the line misrepresents the shape. Either make the code match the doc — `const aligned = twinPieces.length === pieces.length && !SECRET_MASK.includes(delimiter);` — or make the doc match the code. Since you declined o1 on the grounds that it only over-masks, the doc route is acceptable; pick one.

**Optional (take or leave)**

- o4 — `tests/integration/nested-stack-secret/verify.sh` `grep -qF 'Resolved Fn::Join: port:***'` is a substring match; `grep -qxF` would make it the whole-line assertion the unit cases use.
- o5 — the `twinPieces[index] ?? SECRET_MASK` default is unreachable under `aligned`; harmless.

**Out of scope, recorded elsewhere:** the security reviewer confirmed no twin reaches a return value, state, a provider call or an SDK argument, and that no `src/` caller substring-matches a `maskValueLeaves`-routed throw. Its remaining findings are all pre-existing sites the position mask does not reach (`Resolved Ref to parameter:`, the `Parameter X: using ... value` lines, `Resolved Fn::GetAtt from attributes:`, and name-argument sites such as the FindInMap key throw / `Resolving Fn::GetStackOutput: StackName=` / the GetAZs region throw); I am adding them to go-to-k/cdkd#3114, together with the note that `nested-stack-provider.ts`'s `inheritedSecrets` is the per-resource bag, not the parent's pass bag, so route 1's first proposal (search the registry under the inherited bag) cannot find the parent's registration and the carrier design is the one that works.

nix-tkobayashi:
Thanks. `b0f886a6` addresses M5-M7.

- **M5:** the changelog entry no longer lists go-to-k/cdkd#3113 as a residual. It now says the PR also closes it, and keeps go-to-k/cdkd#3114 as the residual.
- **M6:** `maskValueLeaves` masks object keys through `logTextOfLeaf` too, so a key and a value take the same position mask. Case: your shape. After `Fn::Sub port:${PIN_REF}`, `Fn::Select` over `[{ "port:q7": "v" }]` logs `index 0 -> {"port:***":"v"}`.
- **M7:** I took the code route. `splitLogTwins` pairs pieces only when the counts match and the delimiter does not occur in `SECRET_MASK`, which is what its doc already said. Case: your `p*q*r*s` split by `*` under a Join by `-`, which now logs `["***","***","***","***"]` and `***-***-***-***`.
- **o4:** left as is. The logger prefixes every verbose line with a timestamp and level (`ConsoleLogger.formatMessage`), so `grep -qxF` would never match. The substring check only proves the masked line exists; the full-text scan above it fails on the plaintext wherever it appears.
- **o5:** left as is. The `?? SECRET_MASK` default is unreachable under `aligned`, as you say, but `twinPieces[index]` is typed `string | undefined` under `noUncheckedIndexedAccess`, and the default keeps the type honest without a cast.

Mutation probes on the two new clauses (author-reported): one mutation each, byte-exact restore, both killed. Reverting the key to the needle mask reddens the M6 case, and dropping the delimiter test reddens the M7 case.

Nothing in `tests/integration/**` changed in this push, so the three integ runs from the previous round still cover the fixtures.

go-to-k:
Merged as 057a72f5 — thank you for three tight rounds. The log-twin registry keyed by the pass bag is a good primitive: round 2's leaf-mask consolidation turned five separate leak sites into one helper, and it is what makes go-to-k/cdkd#3119 and go-to-k/cdkd#3114 tractable. I pushed one ledger-only commit on top with my own runs of lambda / secrets-dynamic-ref / nested-stack-secret.

Issue #241, opened 2026-05-10
chore: backlog of deferred minor/nit items from #224 follow-up PRs


go-to-k:
## Background

The 224 follow-up PRs (236, 237, 238, 239) each had a fix-back loop that
addressed must-fix + should-fix items but explicitly deferred a handful of
minor / nit items per the "in-scope polish" rule. They are tracked here as
a single backlog issue so they don't get lost.

### From PR 237 (cdkd local start-api authorizers + VPC simulation)

- IAM Resource glob is too permissive. Wildcard `*` is replaced with `.*` and
  matches across `:` / `/` boundaries, which is broader than AWS's
  per-segment glob semantics. Acceptable for a dev tool; document the
  intentional permissiveness.
- Bearer regex captures embedded internal whitespace. `/^\s*Bearer\s+(.+)\s*$/i`
  on `Bearer foo bar` returns `"foo bar"` (after trim). `parseJwt`'s 3-part
  split rejects garbage downstream, but a stricter `[A-Za-z0-9._\-]+` token
  class avoids feeding garbage into the parser.
- Cognito `audience` field on `CognitoUserPoolAuthorizer` is documented as
  "App client id when declared via IdentitySource references" but no code
  path ever populates it. Either remove the field or wire it from the
  IdentitySource scan.

### From PR 238 (start-api hot reload + CORS + stage variables)

- HTTP API v2 `--stage` override does not rewrite `route.stage` (stays
  `$default`). `stageVariables` IS injected correctly. Defensible per
  AWS-side ambiguity, but worth re-verifying against newer AWS docs.

### From PR 239 (cdkd local invoke Lambda Layers)

- `Fn::GetAtt` string-form (`'LayerLogicalId.Ref'`) parsing — CFn JSON never
  emits this YAML-only shorthand. Harmless but the resolver accepts it
  without consistent treatment.
- Symlinks / executable bits in `cpSync` — AWS Lambda layers commonly
  include shebang scripts at `/opt/bin/*`. `cpSync` defaults preserve
  +x bits and copy symlinks as symlinks (verified locally). No code change
  needed; add a one-line code comment pinning the contract.

### From the parallel-PR session as a whole

- Worktree lock cleanup. Many `.claude/worktrees/agent-*` directories are
  locked from completed sub-agent runs. Not blocking, but accumulate over
  long sessions. A periodic prune workflow would help.

## Plan

These do not need to ship together. Pick them up opportunistically when
[...]

Issue #2287, opened 2026-08-26
test(integ): 16 local-* fixtures declare no Lambda architecture, so their containers run under amd64 emulation on an arm64 host and the Go RIE segfaults


go-to-k:
## Summary

cdk-local pins `docker --platform` to each Lambda's declared `Architectures` (`pullImage` / `architectureToPlatform`), which is correct: a `provided.*` bootstrap compiled for one arch must get the matching base image. But a CDK `lambda.Function` that declares no `architecture` defaults to `X86_64`, so on an **arm64 host** its container runs `linux/amd64` under CPU emulation -- and the Go RIE inside `public.ecr.aws/lambda/*` faults there, at a different assertion on every run.

Root cause and the fix pattern are go-to-k/cdk-local#560 / go-to-k/cdk-local#567. This issue is the cdkd half.

## Evidence

Measured 2026-08-27 on an arm64 host, against `758f1c6b`:

```
=== local-start-api RC=1 DUR=623s ===
fatal error: qemu: uncaught target signal 11 (Segmentation fault) - core dumped
ERROR: RIE invoke failed for ItemsHandlerFB09CCF4: ... timed out after 60000ms
FAIL: GET /items/42 did not match "id":"42". Last response: <empty>

=== local-invoke RC=1 DUR=35s ===   (dies the same way at step 6/6)
```

Both are now fixed in go-to-k/cdkd#2278, which declares the HOST architecture in those two fixtures and fences it with `tests/unit/scripts/integ-fixture-host-architecture.test.ts`. This issue covers everything that PR deliberately did not touch.

## The remaining population, measured

30 `local-*` fixtures. 12 declare no Lambda at all (the ECS `run-task` / `start-service` / `start-alb` / `agentcore` families run containers, not RIE Lambdas) and are out of scope. Of the 18 that do, go-to-k/cdkd#2278 fixes 2, leaving **16 fixtures / 27 Lambda constructs**:

`local-invoke-buildkit` `local-invoke-container` `local-invoke-dotnet` `local-invoke-from-cfn-stack` `local-invoke-from-cfn-stack-multi-stack` `local-invoke-from-state` `local-invoke-java` `local-invoke-layers` `local-invoke-provided` `local-invoke-python` `local-invoke-ruby` `local-start-api-container` `local-start-api-rest-v1-non-proxy` `local-start-api-websocket` `local-start-cloudfront`, plus any fixture added since.

## Why this is NOT a mechanical sweep

[...]

PR #3118, opened 2026-09-14
fix(local): copy a layer's symlinks verbatim, last-wins, so a relative link resolves inside the container

### LINKED ISSUE #3106: fix(local): a relative symlink inside a Lambda Layer is dangling in the container — cpSync rewrites it to an absolute host path (verbatimSymlinks defaults to false)
## What breaks

A RELATIVE symlink inside a Lambda Layer asset (`bin/rel-link -> real.sh`, the shape a layer build tool emits for a shared binary) is dangling inside the local container. Both layer-materialising `fs.cpSync` calls — `src/cli/commands/local-invoke.ts` (`materializeLambdaLayers`, `cpSync(layer.assetPath, tmpDir, { recursive: true, force: true })`) and `src/cli/commands/local-start-api.ts` (the same shape for `start-api`'s container pool) — copy the link as a link (that is `dereference: false`, the default), but `verbatimSymlinks` ALSO defaults to `false` in every Node release, and a non-verbatim copy REWRITES a relative target to the ABSOLUTE path of the source: the copied link points at `<cdk.out>/asset.<hash>/bin/real.sh` on the host. Only the merged tmpdir is bind-mounted at `/opt`, so `/opt/bin/rel-link` resolves to a path that does not exist in the container.

Measured on Node v24.21.0 (and v22.23.2, by the go-to-k/cdkd#3105 code review) with a two-file layer tree:

```text
cpSync(src, dst, { recursive: true, force: true })                         -> dst/bin/rel-link -> /abs/path/to/src/bin/real.sh   (dangling in /opt)
cpSync(src, dst, { recursive: true, force: true, verbatimSymlinks: true }) -> dst/bin/rel-link -> real.sh                       (resolves in /opt)
```

The comment above each call has pinned the opposite since the go-to-k/cdkd#241 backlog ("copy symlinks as symlinks (verified locally)" — verified that the copy IS a link, never where it points), and go-to-k/cdkd#3105 rewrote the comment to state the facts and point here. No integ fixture carries a symlink in a layer, which is why the arm was never exercised.

## Fix

`verbatimSymlinks: true` on both calls; a unit test that materialises a layer holding a relative symlink and asserts the copied link's `readlink` is still relative (red without the option); and a symlink in the `local-invoke-layers` fixture's layer whose handler execs through it, so the integ proves the link resolves inside `/opt`. Absolute symlinks are unaffected either way (they were already host paths). cdk-local carries the same two call sites (`src/cli/commands/local-invoke.ts`, `src/cli/commands/local-start-api.ts`), tracked in its own issue; cdkd's `local-invoke.ts` copy is its own code, not a re-export, so both repos need the change.

Dup-check: searched go-to-k/cdkd issues (open and closed) for `verbatimSymlinks`, `symlink layer cpSync`, and the body window for `verbatimSymlinks|cpSync` -- only the closed go-to-k/cdkd#241 backlog row, which pinned the wrong contract; no open issue covers this root cause
Session-fit: now -- found by this session's review of go-to-k/cdkd#3105 with the probe in hand; both call sites are open in this session's worktree; the fix lands as its own PR after the floor PR merges (it changes `src/cli/commands/local-*.ts` behaviour, so it carries its own `integ-local` run)
Severity: medium -- any layer that ships a relative symlink (a build-tool-emitted `bin/` shim is the common case) fails at invoke time inside `cdkd local invoke` / `local start-api` with a "No such file or directory" that names a path the user never wrote
Effort: small (S) -- two one-token option additions, one unit test with a symlink fixture, one symlink added to an existing integ layer; drags the `integ-local` cycle (`/run-integ local-invoke-layers`)
Estimate: ~45 min -- most of it the integ run and the fixture arm; the option change itself is minutes

go-to-k:
## Summary

Both layer merges (`materializeLambdaLayers` in `src/cli/commands/local-invoke.ts` and the `start-api` twin) copied each layer's asset directory into one tmpdir with a bare `fs.cpSync`, whose `verbatimSymlinks` defaults to `false` in every Node release. A non-verbatim copy rewrites a RELATIVE link target — `bin/rel-link -> real.sh`, the shape a layer build tool emits for a shared binary — to the ABSOLUTE host path of the source asset, so `/opt/bin/rel-link` pointed at a path that does not exist inside the container and a handler executing through it failed with `spawnSync /opt/bin/rel-link ENOENT`. The comment above each call had claimed the opposite since the go-to-k/cdkd#241 backlog; the review of go-to-k/cdkd#3105 probed it.

The obvious fix — `verbatimSymlinks: true` — was measured wrong in the other direction, and so was the next one:

- with `verbatimSymlinks: true`, Node's C++ `cpSync` (every release after 22.12) throws `EEXIST` when a later layer carries a symlink at a path an earlier layer already placed one (`nodejs/node_modules/.bin/<tool>` in two layers built from the same dependency) — a merge that used to succeed would refuse;
- a `cpSync` + `readdirSync({ recursive: true })` pair DESCENDS into directory symlinks on the same releases, so a cyclic link hung the merge and an absolute link to a host directory made it write INSIDE the host directory; `cpSync`'s `force` also wrote THROUGH a destination that was already a directory symlink.

Both merges now go through one explicit walk, `src/local/layer-tree-copy.ts` (`copyLayerTreeLastWins`): recurse only on a real directory (`Dirent` `d_type`); per entry, "last layer wins" for every kind — whatever sits at the destination that is not of the same kind is removed first (a link, a file, a whole directory), then the entry lands; a file goes through a single-file `cpSync` so `+x` survives; a link is recreated with the source's own target string and nothing is ever written through one; a merge that throws mid-loop removes the tmpdir it allocated. Only the multi-layer merge is affected — a single layer is bind-mounted directly. The two `local-*.ts` comments, and every sibling comment that described the merge as `cpSync({force})`, now name the helper; `.claude/rules/layout-local.md` lists the module.

## Fences

- `tests/unit/cli/layer-symlink-verbatim-3106.test.ts` (12 cases): the relative link through BOTH entry points (`readlinkSync` of the copy is `real.sh`; red with the fix removed — the absolute `/private/var/.../bin/real.sh` target comes back); the helper on same-link-twice, link↔file both ways, link-over-dir, dir-over-link with the link's target untouched, file-over-dir / file-over-dangling-link / dir-over-file, a cyclic link + an absolute host link with the host directory unchanged, a root reached through a symlink; and the tmpdir-leak guard at both entry points (hermetic via a per-test `TMPDIR`). Three of them are red against the round-1 helper, two against the guard-less loop, one against each of the three removed-guard mutations.
- The `local-invoke-layers` fixture's counters layer ships `bin/real.sh` and `bin/rel-link -> real.sh` (tracked as mode `120000`); the handler execs `/opt/bin/rel-link`, `verify.sh` asserts `linkOutput: real-via-link`. Two fixture defects found on the way and fixed here: it SWALLOWED failures (`$(... 2>/dev/null | tail -1)` under `set -e` aborted with no diagnostic — it now carries the `capture` helper the cdk-local twin had), and it declared no `architecture`, so on an arm64 host the new spawn-in-handler ran under emulation and hung one run in three; every function now declares `HOST_ARCHITECTURE` and the fixture joins `integ-fixture-host-architecture.test.ts`'s list (3; go-to-k/cdkd#2287's remaining set is 15).

## Live test

- Pre-fix binary (fix removed, rebuilt): the invoke returns `spawnSync /opt/bin/rel-link ENOENT` and the fixture fails.
- Fixed binary via `/run-integ local-invoke-layers`: 4/4 passed, `"linkOutput":"real-via-link"`, 0 Docker orphans; four consecutive native runs green at 18-28 s after the architecture pin; `integ-local` marker set, ledger row included.

## Review

Code axis: four rounds (b8b5ce7c → 34271325 → 9811e4c9 → 5989daea), each finding closed in the next commit; the round that mattered found the EEXIST regression of the naive fix, the next the recursive-readdir descent. Spec + test axes on 12caddf8 found one unfenced arm of the helper's stated contract (file-over-directory, file-over-dangling-link and directory-over-file — the unguarded `cpSync` throws `ERR_FS_CP_NON_DIR_TO_DIR`, ABORTS the process on the dangling link, and `mkdirSync` throws `EEXIST`) plus the round-1 wording surviving in the changelog entry and `layout-local.md`; 888f07c2 closes both (one case, three mutations red). Recorded won't-dos: a FIFO / socket inside a layer makes the copy throw immediately (CDK staging cannot produce one — in the helper's doc); ARN-downloaded layers never carry symlinks (`layer-arn-materializer` refuses the entry), so the symlink walk is a no-op for them.

cdk-local carries the same fix: go-to-k/cdk-local#731 (issue go-to-k/cdk-local#727).

Closes #3106


go-to-k:
Merged — thanks. Four measured rounds turned the one-line `verbatimSymlinks: true` fix into the explicit walk (EEXIST on a duplicated link, then the recursive readdir descending into directory symlinks), and the final spec + test pass closed the last contract arm the cases had not reached. The cdk-local twin (go-to-k/cdk-local#731) landed first.

PR #3122, opened 2026-09-14
fix(secret-redaction): read a number or boolean Fn::Join part as literal text when positioning a secret reference

### LINKED ISSUE #3055: secret-redaction: a numeric or boolean Fn::Join part is read as unknowable by the skeleton parser, though the resolver joins it as String(part)
## Summary

`intrinsicSkeletonSegments` (`src/deployment/secret-redaction.ts`) reads an `Fn::Join` part as a literal only when `typeof part === 'string'`; any other scalar — a number or a boolean the template carries verbatim, `["port", 8080, ":{{resolve:...}}"]` — becomes an `UNKNOWN_PART`, i.e. a wildcard for the skeleton arm and a rendering placeholder for the frame arm. The resolver joins such a part as JavaScript does (`String(part)`), so its text IS knowable, and treating it as unknowable costs positioning: the skeleton pattern loses that literal's specificity, and the frame arm refuses outright when the scalar sits outside the token (a placeholder in the prefix or suffix is the nonliteral-frame refusal).

Pre-existing: `origin/main`'s parser carries the same `typeof part === 'string'` test (it predates go-to-k/cdkd#3052, which extracted the parser for a second reader without changing the test). Surfaced by the maintainer's review of go-to-k/cdkd#3052 as out of that PR's scope.

## Proposed fix

In `intrinsicSkeletonSegments`, treat a `number` / `boolean` part as the literal `String(part)`; keep `null`, objects and arrays unknowable. One unit case per reader: the skeleton arm positions `{"Fn::Join": ["", ["{{resolve:secretsmanager:", 8080, ":SecretString:pw}}"]]}`-shaped leaf on a pinned candidate that spells `8080` (today it wildcards the number and can match a sibling that differs there); the frame arm positions `["port", 8080, ":{{resolve:...}}"]` on a marked bag (today it refuses: the placeholder lands in the prefix). Mirror the resolver's coercion exactly — if `resolveJoin` does anything other than `String(part)` for a scalar, that is the rule to copy, not `String`.

## Verification plan

- Unit: the two cases above, each red on `origin/main`'s parser and green after; a mutation dropping the coercion reds both.
- The `secrets-dynamic-ref` fixture needs no new arm unless a real CDK construct emits a numeric `Fn::Join` part — check with a synth probe first; if none does, say so in the PR rather than adding a hand-written shape.

Dup-check: `gh issue list --state all --search "Fn::Join non-string part skeleton"` and `"scalar Fn::Join part unknowable"` — no match (2026-09-13).
Session-fit: now (do it in this session) — the file is loaded: go-to-k/cdkd#3115 worked `recordNestedStackParameterExpressions` and the frame arm that reads this parser; the earlier `next` reason (go-to-k/cdkd#3052's lane still open) ended when that PR merged.
Severity: low — a positioning refusal falls to the value scan (a survivor's spelling at 4+ characters, plaintext below the floor only on the frame arm's leaves), and no CDK L2 is known to emit a numeric `Fn::Join` part.
Effort: small (S) — one coercion in the parser plus two unit cases; no fixture arm unless the synth probe finds a shape.
Estimate: ~1-2 h — the synth probe and the mutation probes, not the coercion.

nix-tkobayashi:
## Summary

A number or boolean `Fn::Join` part is now literal text to every reader of the shared intrinsic parser, `intrinsicSkeletonSegments` in `src/deployment/secret-redaction.ts`.

The resolver joins every part as `String(await resolveValue(part))`, and `resolveValue` returns a number or boolean unchanged. So `8080` contributes `8080` and `true` contributes `true`. The parser read only a string part as literal and made the others `UNKNOWN_PART`, which cost each of its three readers:

- **Skeleton arm.** The position became a wildcard. Two recorded references differing only there both matched, the uniqueness rule refused, and the value scan wrote the collapsed map's surviving reference instead of the leaf's own.
- **Frame arm.** The position became a placeholder, so the rendered frame no longer equalled the leaf's text around the middle and the arm refused. A 1-3 character middle stayed in plaintext.
- **Nested-stack parameter recorder.** `frameSpellingOf` carries a framed parameter to the child only when the frame arm rewrote it, so such a parameter got no whole-value entry and no per-parameter association. This is a downstream effect of the frame arm's reading; the recorder's own rendering of a part outside the token does not decide it.

The new `joinPartLiteralText` helper returns `String(part)` for a string, number or boolean part, so all three now position or carry the frame. A `null`, array or object part stays unknowable, and a number delimiter is still refused.

Two residuals are stated in the helper's docstring. A `null` wildcard keeps the skeleton arm's existing wrong-reference residual. And the literal reading reaches that same residual in a few more places: a source that left two candidates matching can now leave one, which the arm accepts even when the leaf's own reference was never recorded. The arm's own residual paragraph now names both shapes that reach it, an unclassifiable or a public `ssm` reference. It previously named only the first. A test case measures the second shape.

The sibling docstrings on `SKELETON_WILDCARD` and `joinSkeletonSegments` now describe the new reading. `.claude/rules/layout-deployment-secrets.md` now says most L2-reached secrets are joins, not all: `Secret.fromSecretCompleteArn(...).secretValueFromJson(...)` renders a literal string.

## Reachability

A synth probe against this repo's aws-cdk-lib found no L2 construct emitting a non-string part. A TypeScript number interpolated, or `Token.asString(8080)`, is folded into the literal string before the template is written. A non-string part reaches a template only through a hand-written `Lazy.any` inside `Fn.join`, or a raw template. Per the issue's verification plan, the coverage is therefore unit-only and no integ fixture arm was added.

## Verification

New file `tests/unit/deployment/secret-redaction-intrinsic-scalar-part.test.ts`, 16 cases:

| Case | Reader | Pins |
| --- | --- | --- |
| number part `8080` | skeleton | leaf written as its own reference, not the collapsed map's survivor |
| boolean part `true` | skeleton | same, against a `false` survivor |
| string part `'8080'` | skeleton | control, unchanged by the fix |
| `null` part with a reference spelling `null` | skeleton | `null` stays unknowable; the leaf falls to the value scan |
| number delimiter `8080` | skeleton | the delimiter refusal: the leaf keeps the value scan's survivor |
| public `ssm` leaf, port as a wildcard or as `8080` | skeleton | the widened residual, measured: survivor `/sec/9090` before, lone match `/sec/8080` after, never the leaf's own `/pub/8080` (2 cases) |
| parts `8080`, `0`, `true`, `false` | frame | sub-floor middle written as its token (4 cases) |
| parts `null`, `['8080']`, `{}` | frame | refusal kept, middle stays plaintext (3 cases) |
| number part `8080` | nested recorder | whole-value entry and per-parameter association written |
| `null` part | nested recorder | nothing recorded |

Each positive case asserts its premise first: the value scan alone does not produce the expected result. The skeleton positioning cases record no pass-local pair and assert that too, through the embedded-span arm, so the frame arm cannot answer the same leaf. The frame refusal leaves are each the text `String(part)` would give, so reading the part as that text would position them.

Mutation probes on that file, one mutation per probe, source and test file restored byte-exact after each:

| Mutation | Result |
| --- | --- |
| pre-fix source | 8 red: skeleton number and boolean, the residual's literal row, the four frame positives, nested recorder number |
| skeleton arm call removed | 4 red: the three skeleton positioning cases, control included, and the residual's literal row |
| helper drops the number branch | 5 red: skeleton number, the residual's literal row, frame `8080` and `0`, nested recorder number |
| helper drops the boolean branch | 3 red: skeleton boolean, frame `true` and `false` |
| helper returns `'null'` for `null` | 3 red: the three `null` cases |
| truthiness test in place of the type test | 2 red: frame `0` and `false` |
| helper stringifies objects and arrays | 3 red: frame array and object refusals, and the residual's literal row (its `Prefix` Ref would read as literal text) |
| skeleton fixture records pairs | 3 red: the no-pair premise in each skeleton positioning case |
| delimiter guard accepts a number | 1 red: the number-delimiter case |

Full unit suite, typecheck, lint, build, generated-artifact freshness, the rules-payload budget and the changelog size cap pass locally.

Closes #3055


go-to-k:
Reviewed head `592ddd0` (merge-base `a386124b`). Three read-only reviewers (code / test / security) plus my own read; the resolver parity claim holds (`resolveJoin` does `String(await resolveValue(v))` and `resolveValue` returns a primitive unchanged), all four mutation probes I re-ran match the PR body, no existing test pinned the old wildcard reading, and the security trace found no new plaintext path and no regex-injection surface. No blocker. Three items before I merge, labeled by what I need from you:

**M1 (required before merge) — two sibling docstrings in `src/deployment/secret-redaction.ts` now contradict the parser.**
- `joinSkeletonSegments`' doc (≈L2924-2925 on your head): "an `Fn::Join` with an EMPTY delimiter and consecutive **non-string** parts" produce adjacent wildcards. After this PR consecutive number / boolean parts do not; only consecutive intrinsic (or `null`) parts do. "non-string" → "intrinsic".
- `SKELETON_WILDCARD`'s doc (≈L2851-2853): "an `Fn::Join` element that is itself an intrinsic". You updated the sibling `UNKNOWN_PART` doc to "or any other non-scalar or `null` part" but left this one; mirror the wording so the two constants agree.

**M2 (required before merge) — fence the number-delimiter refusal.** The new comment above the delimiter guard records a decision ("a number delimiter would be knowable ... but refusing it only costs the positioners a leaf"), and the suite cannot see it: relaxing the guard to `typeof delimiter === 'number' ? String(delimiter) : delimiter` leaves all 529 cases under `tests/unit/deployment/secret-redaction*` green (the existing refusal row in `secret-redaction-intrinsic-source.test.ts:404` uses `{ Ref: 'Sep' }` only). One case in your new file pins it: `collapsed('8080', '9090')` with source `{ 'Fn::Join': [8080, ['{{resolve:secretsmanager:', ':SecretString:pw}}']] }` → expect the survivor `9090` (the relaxed guard flips it to `8080`).

**m3 (optional, take-or-leave) — pin the widened residual as a case.** The `joinPartLiteralText` docstring states that the literal reading can now leave ONE matching candidate that is not the leaf's own reference (own reference never recorded, e.g. a public `ssm` parameter) and the arm accepts it. That is bounded — condition 1 already required the bag to be a recorded plaintext, so the same leaf previously took the value scan's survivor, also a wrong reference — but it is prose only today. A single case (number-part source, own reference unrecorded, one other candidate matching → assert what is written) would make it a measured fact.

Push M1 + M2 (and m3 if you take it) and I will re-run the round on the new head.

nix-tkobayashi:
Thanks for the round on `592ddd0`. All three items are in `94bc4fa`, including m3.

**M1.** Both sibling docstrings in `src/deployment/secret-redaction.ts` now match the parser.
- `joinSkeletonSegments`: "consecutive non-string parts" now reads "consecutive intrinsic (or `null`) parts".
- `SKELETON_WILDCARD`: now uses the same wording as `UNKNOWN_PART` ("an intrinsic or any other non-scalar or `null` part").

**M2.** The number-delimiter refusal is pinned in `tests/unit/deployment/secret-redaction-intrinsic-scalar-part.test.ts`, using your shape: `collapsed('8080', '9090')` with source `{ 'Fn::Join': [8080, ['{{resolve:secretsmanager:', ':SecretString:pw}}']] }`, expecting the survivor `9090`. A premise assertion first checks that `String(8080)` joins the parts into the `8080` reference exactly. With your relaxed guard, only that case goes red.

**m3.** The widened residual is now a measured case with two rows. The leaf's own reference is a public ssm parameter (`/pub/8080` through a `Prefix` Ref) that was never recorded. Two recorded SecureStrings, `/sec/8080` and `/sec/9090`, share its value. The source is `{{resolve:ssm:` + `{Ref: Prefix}` + `/` + port + `}}`.
- Port as a wildcard (`{Ref: Port}`, the reading before go-to-k/cdkd#3055): both candidates match, the arm refuses, and the value scan writes the survivor `/sec/9090`.
- Port as the literal `8080`: one candidate matches and the arm writes `/sec/8080`.

Both rows also assert the written reference is not `/pub/8080`. On the pre-fix source the literal row goes red and the wildcard row stays green.

Changes from the reviews of `a1a19322` that you have not seen yet: two comment wordings (the object/array sentence and the delimiter comment) and one commit-message clause. None of them changes behavior.

Mutation table on the new head (one mutation per probe, restored byte-exact), PR body updated to match:

| Mutation | Result |
| --- | --- |
| pre-fix source | 8 red: skeleton number and boolean, the residual's literal row, the four frame positives, nested recorder number |
| skeleton arm call removed | 4 red: the three skeleton positioning cases and the residual's literal row |
| helper drops the number branch | 5 red |
| helper drops the boolean branch | 3 red |
| helper returns `'null'` for `null` | 3 red: the three `null` cases |
| truthiness test in place of the type test | 2 red: frame `0` and `false` |
| helper stringifies objects and arrays | 3 red: frame array and object refusals, and the residual's literal row |
| skeleton fixture records pairs | 3 red: the no-pair premise |
| delimiter guard accepts a number (your relaxation) | 1 red: the number-delimiter case |

Local runs on this head: full unit suite 1005 files / 22459 tests, plus typecheck, lint, build and generated-artifact freshness.

go-to-k:
Round 2 on head `94bc4fa`: M1 / M2 / m3 all verified (code + test reviewers re-ran your mutation table, 5/5 match; the residual case exercises the arm the docstring names). Three prose / test-precision nits came out of it and I filed them as #3143 for myself — nothing for you to do there.

One new item, and it is a merge blocker that CI cannot see:

**M4 (required before merge) — the `.claude/rules` payload fence fails once this head is merged into CURRENT `origin/main`.** `tests/unit/scripts/rule-file-payload.test.ts` projects each budgeted path's rule payload against `origin/main`, and main moved after you branched: other lanes spent `src/cli/commands/scrub.ts`'s budget to exactly the 119000 B cap, and your `layout-deployment-secrets.md` rewording ("so every / L2-reached secret is a join" → "so an / L2-reached secret is usually a join") is a net +5 B on that path. Your PR CI passed because it merged against the older main; the run on this head against today's `origin/main` fails:

```
`src/cli/commands/scrub.ts`: this branch adds 5 B to what that path loads. Against origin/main's current 119000 B that projects to 119005 B, over the 119000 B cap
```

A net-zero rewording keeps the softened claim and passes (I measured it green locally against today's main): replace

```
renders the ARN as a `Ref`, so an
    L2-reached secret is usually a join)
```

with

```
renders the ARN as a `Ref`, so most
    L2-reached secrets are joins)
```

(33 characters either way). Verify with `git fetch origin && vp test run tests/unit/scripts/rule-file-payload.test.ts` — the test reads `origin/main`, so the fetch is what makes it see the number the merge will produce; no rebase needed. Push that and I will merge on green.

nix-tkobayashi:
M4 is in `6f81688`, with your wording verbatim. The line now reads "so most / L2-reached secrets are joins".

After `git fetch origin`, `tests/unit/scripts/rule-file-payload.test.ts` passes against today's `origin/main` (`35adf0f6`): 333 tests. The PR body sentence about that rules line now uses the same wording. Nothing else changed.

Thanks for filing go-to-k/cdkd#3143.

go-to-k:
Merged as 9bd36c8 — thank you. Three rounds and every one of them came back with the requested item plus a measured mutation table; the m3 residual case in particular turned a docstring claim into a fenced fact, which is the shape this module needs.

Issue #3121, opened 2026-09-14
fix(drift-normalize): the comparison-side canonicalizers rebuild onto {} literals -- an own __proto__ key is dropped on both sides and a Date readback is flattened to {} before the comparator


go-to-k:
## The defect

The comparison-side canonicalizers `cdkd drift` runs over BOTH bags before `calculateResourceDrift` compares them rebuild every object node onto a `{}` literal with a plain assignment: `canonicalizeTagListsDeep` and `canonicalizeIdArraysDeep` in `src/analyzer/drift-normalize.ts` (both `if (v && typeof v === 'object') { const out = {}; ... out[k] = ... }` arms), plus the same shape at `canonicalizeUnorderedArraysAtPaths`' object arm in that file, `src/analyzer/drift-principal-normalize.ts` (two `out` rebuilds) and `src/analyzer/cc-api-strip.ts:142`. Two consequences, both measured on `main` at `a386124b` by calling the functions directly (no AWS):

1. An OWN `__proto__` key (what `JSON.parse` yields from `state.json`, and what a readback may carry) is ASSIGNED onto the literal, so it becomes the rebuilt node's prototype and is dropped as a key: `canonicalizeTagListsDeep(JSON.parse('{"__proto__":{"polluted":"base"},"A":"a"}'))` returns a node with `Object.hasOwn(out, '__proto__') === false` and `out.polluted === 'base'`. Both sides lose it symmetrically, so `calculateResourceDrift(bag, live)` returns `[]` for two bags whose `__proto__` members DIFFER -- a hidden drift, never a wrong write. This is the reason the `__proto__` pin for go-to-k/cdkd#2899 in `tests/unit/cli/drift-secret-redaction.test.ts` has to place the key one level DOWN beside a drifting sibling: a top-level `__proto__` cannot be reported today.
2. A NON-PLAIN object in the readback is flattened: `canonicalizeTagListsDeep({ D: new Date(...) })` returns `{ D: {} }` (own enumerable keys of a `Date` are `[]`). An AWS SDK readback routinely carries `Date` members (`LastModified`, `CreationDate`), and the walk runs on the COMPARISON copy of the AWS bag, so against an ISO-string baseline the leaf reports as drifted with `awsValue: {}`, and `cdkd drift --accept` then persists `{}` into the baseline -- the sibling of go-to-k/cdkd#2427 (the value scan's flattening) one layer earlier, on the comparator's own input.

[...]

Issue #2776, opened 2026-09-07
fix(resolver): sweep the two residuals of #2767 — the downstream __proto__ walks and resolveSub's own variable-map test


go-to-k:
## What is wrong

Two residuals of go-to-k/cdkd#2767's prototype-chain sweep. Both were found by the code review of that PR and are outside its diff.

### 1. Two downstream property walks undo the `__proto__` fix

go-to-k/cdkd#2767 made `IntrinsicFunctionResolver.resolveValue` preserve a `__proto__` property (an OWN key after `JSON.parse`) by shadowing it with `defineProperty`. Two later walks over the same bags rebuild them with a plain `{}` and a plain assignment, so the key is dropped again on those paths:

- `src/deployment/rollback-executor.ts:1154` — inside `resolveReplayProps`, which re-resolves the rollback journal against AWS. A property the user authored is silently absent from the bag the replay sends.
- `src/cli/commands/drift.ts:1778` — the same walk on the drift path, so `drift --revert` pushes a baseline missing the key.

Both are the identical mechanism go-to-k/cdkd#2767 fixed one layer up, and the remedy is the same three lines. The reason they were not swept with it: `rollback-executor.ts` is in the `integ-destroy` gate's scope, so the fix requires a real-AWS destroy run that PR did not otherwise need.

### 2. `resolveSub`'s own variable-map `in` test

`src/deployment/intrinsic-function-resolver.ts` — `resolveSub` tests `varNameStr in variables` against the CALLER's plain object, so `Fn::Sub: "arn:${constructor}"` renders `arn:function Object() { [native code] }` into a live property and short-circuits the pseudo-parameter / `Ref` / `GetAtt` arms it should have reached. This is the SAME class as the sites go-to-k/cdkd#2767 fixed, and on the one-argument form it needs nothing else.

It was left out because go-to-k/cdkd#2764 rewrites exactly that block (issue go-to-k/cdkd#2739), so touching it would hand an external contributor a conflict in the region their PR edits.

Three test cases and one guard were deferred with it, and belong in the same follow-up:
- `${constructor}` through `Fn::Sub` rendering the placeholder rather than function source, on BOTH forms;
- a resource genuinely named `constructor` reached through `Fn::Sub`;
- `Fn::Sub`'s FIRST element gaining the non-string refusal its second element has, so a non-string template names its refusal instead of dying at `template.matchAll is not a function`.

## Why it matters

[...]

Issue #2427, opened 2026-09-02
fix(secret-redaction): the value scan flattens a Date in observedProperties to {}, corrupting the drift baseline


go-to-k:
Dup-check: searched open issues for an existing item naming this root cause (`Date observedProperties redaction flatten LastModified`) -- no hits. New root cause.

## What happens

`redactSecretsForState`'s value scan rebuilds objects through `hasPlainPrototype`, so any leaf whose prototype is not `Object.prototype` is replaced by a plain `{}`. An AWS SDK readback routinely carries `Date` instances (`LastModified`, `CreationDate`, `LastUpdatedTime`, ...), and `observedProperties` is exactly where those land.

Measured on `main` (`2f491105`), no AWS needed -- the module is a leaf, so `node` can import it directly:

```ts
import {
  redactSecretsForState,
  TEMPLATE_SOURCED_RULES,
  STATE_SOURCED_CROSS_GENERATION_RULES,
} from './src/deployment/secret-redaction.ts';

const EXPR = '{{resolve:secretsmanager:app/db:SecretString:password}}';
const secrets = new Map([['hunter2secret', EXPR]]);
const bag = { Password: 'hunter2secret', LastModified: new Date('2026-01-01T00:00:00Z') };

redactSecretsForState(bag, secrets);
// -> {"Password":"{{resolve:...}}","LastModified":{}}
redactSecretsForState(bag, secrets, { Password: EXPR }, TEMPLATE_SOURCED_RULES);
// -> same
redactSecretsForState(bag, secrets, { Password: EXPR }, STATE_SOURCED_CROSS_GENERATION_RULES);
// -> same
```

All three arms drop the `Date`. `hasPlainPrototype`'s own doc comment already says such values really arrive there, so this is not a hypothetical shape.

## Why it matters

`observedProperties` is the drift baseline. A `LastModified` that persists as `{}` instead of an ISO string means:

- `cdkd drift` compares `{}` against whatever AWS reports now, so the field is a permanent phantom difference on every resource whose readback carries a timestamp AND whose record has at least one recorded secret;
- `cdkd drift --revert` pushes the BASELINE to the live resource, so a baseline holding `{}` at a field the provider actually sends is the fabricated-baseline class this module refuses elsewhere.

The trigger is the SECRETS MAP being non-empty, so it only bites resources in a stack that resolved at least one dynamic reference -- which is why it has gone unnoticed.

## Scope note

[...]

PR #3124, opened 2026-09-14
fix(drift): own __proto__ keys survive the baseline walk, the token arm copies only a string, a token identity never pairs, non-plain values never compare equal, and the two state writers select the fail-closed rules by destination

### LINKED ISSUE #2897: fix(drift): the revert preserve walks rebuild a non-plain object (Date) as {}, and deepEqualUnordered calls two differing Dates equal
## Scope after go-to-k/cdkd#2882 round 5 (body rewritten; the original is in the edit history)

As filed, this issue had two halves and a deferral rationale of "reachability
disproven — no provider returns a raw `Date` from `readCurrentState`". Round 5
of go-to-k/cdkd#2882 falsified that rationale and FIXED the first half there,
so this body now carries only what remains.

**Fixed in go-to-k/cdkd#2882 (round 5), no longer this issue's scope:** both
revert preserve walks (`preserveLiveValuesAtMaskedLeaves` and
`preserveLiveValuesAtUnresolvedTokens`, `src/cli/commands/drift.ts`) rebuilt
every object node with `Object.entries`, fabricating `{}` (or a plain index
map) from a `Date` / `Uint8Array` / class instance and shipping it via
`provider.update`. Both object arms now return a non-plain value BY IDENTITY
(`hasPlainPrototype`), pinned by identity-assertion cases and mutation probes.

**Why the original deferral rationale was wrong** (recorded so the next
reader does not resurrect it): the send bag is NOT JSON-parsed state.
`drift.ts` builds `overlaid = buildRevertNewProperties(changes,
desiredProperties, outcome.awsProperties, ...)`, and `outcome.awsProperties`
is `provider.readCurrentState`'s RAW SDK return with no JSON round-trip — so
a non-plain SDK value reaches the walks directly through the non-drifted
subtrees. The per-provider "7 providers normalise via `toISOString`"
measurement describes today's providers, not the interface; and the binary
half was already conceded by the redaction side's `ArrayBuffer.isView` arm,
which exists precisely because AWS returns a `Uint8Array`.

## What remains: `deepEqualUnordered`'s non-plain equality at its two surviving callers

`deepEqualUnordered` (`src/cli/commands/drift.ts`) reports any two `Date`s
(or other zero-own-key non-plain objects) equal: `a === b` fails, both are
objects, key counts `0 === 0`. Its corroboration caller was guarded in
go-to-k/cdkd#2882 round 4 (`corroboratedLeafCount` refuses a non-plain object
on either side first, both sides pinned), but its OTHER callers still compare
through it. (An earlier revision of this section named the first one
`computeEffectiveDelta`, a symbol that does not exist — corrected per the
round-6 review so a grep actually lands.)

- `collectNarrowedTopLevelKeys` (`src/cli/commands/drift.ts`, the exported
  helper that diffs the SENT bag against a provider's `effectiveProperties`
  echo for the issue-1644 narrowing write; the `deepEqualUnordered` call sits
  in its per-key loop, and `runRevert` consumes its delta). Two differing
  non-plain values reading as equal HIDES a narrowing: the changed value is
  not persisted into `observedProperties`, leaving the baseline stale at that
  position.
- The `--accept` post-write check (the `redactedBaseline` verification loop
  in `runAccept`, same file): `deepEqualUnordered(getAtPath(redactedBaseline,
  change.path), change.awsValue)` decides whether an accepted change was
  really recorded, and a false-equal SUPPRESSES the "was NOT recorded"
  warning — the user is told nothing while the baseline still disagrees.
  Milder than the narrowing case (nothing wrong is written; a warning is
  lost), same remedy.

## Remedy direction

Give `deepEqualUnordered` the same `hasPlainPrototype` guard
`deepEqualJsonValue` carries in `secret-redaction.ts` (non-plain on either
side compares UNEQUAL unless identical), or record at
the two call sites why the looser reading is deliberate. Note the
direction of the change: making non-plain values compare UNEQUAL widens the
narrowing delta, which WRITES more — walk the (#1644) consumers before
shipping. Unit cases either way.

## Classification

Session-fit: next (not this session) — the surviving half is a
baseline-staleness / lost-warning question in the (#1644) narrowing path and the `--accept` post-write check, separate from the
write-path corruption go-to-k/cdkd#2882 fixed; changing an equality
predicate that feeds a WRITE needs its own consumer walk. Verification:
`vp test run tests/unit/cli/drift-masked-leaf-preserve.test.ts
tests/unit/cli/drift-secret-redaction.test.ts` plus new cases at
`collectNarrowedTopLevelKeys` and the `--accept` post-write check.
Severity: low — a hidden narrowing of a non-plain value leaves
`observedProperties` stale at that position (phantom-drift direction, no
disclosure and no live-resource write); it needs a provider echoing a
non-plain `effectiveProperties` member that CHANGED across the update.
Effort: small (S) — one prototype guard in one comparator plus unit cases
and the (#1644) consumer walk; drags the unit cycle only.
Estimate: ~1 h — most of it the consumer walk deciding whether the wider
delta is safe for every `collectNarrowedTopLevelKeys` reader.
Dup-check: searched issues on 2026-09-10 for `preserveLiveValues Date`,
`revert Date flattened`, `deepEqualUnordered` and `drift non-plain object` —
no hits beyond this issue. (#2869) is the secret-redaction readback walk;
(#2427) is the value scan's flattening; (#2855) is the revert-side
write-hazard family. No duplicate.

### LINKED ISSUE #2899: fix(drift): resolveStateSecretExpressions rebuilds objects onto a {} literal, so an own __proto__ key in a state baseline is dropped and the prototype clobbered
## The defect

`resolveStateSecretExpressions` (`src/cli/commands/drift.ts`, the async walk
inside the function defined at line 1683) rebuilds every object node with

```ts
const out: Record<string, unknown> = {};
for (const [k, val] of Object.entries(v)) {
  out[k] = await walk(val, ...);
}
```

Its input is the JSON-parsed `state.json` baseline — exactly the producer of
an OWN `__proto__` key (`JSON.parse` yields it as an ordinary own key) — so
at that key the assignment onto a `{}` literal SETS the rebuilt node's
prototype and silently DROPS the key. The output becomes
`desiredProperties` in `runRevert`, flows through
`buildRevertNewProperties` into the bag `provider.update` writes to live
AWS. Two consequences chain:

- the `__proto__`-keyed member is silently dropped from the revert payload
  (a baseline property is not restored, with no report);
- the rebuilt node carries a non-null custom prototype, so it fails
  `hasPlainPrototype` and cascades into the non-plain arms go-to-k/cdkd#2882
  added — `mergeUntemplatedValue`'s mismatch fallback and the preserve
  walks' identity return — turning an ordinary plain-object subtree into
  one the merge machinery refuses to walk.

Pre-existing (not introduced by go-to-k/cdkd#2882, whose rounds 6-7 fixed
the sibling sites in the same file), and it needs an exotic key in a
persisted baseline to trigger.

## Precedent

The repo fixes this class site-by-site with `Object.create(null)` rebuild
targets: `secret-redaction.ts` carries five such sites (the issue (#1943)
class comment at its `redactByPath` rebuild), `intrinsic-function-resolver.ts`
two (issues (#2767), (#2802)), and go-to-k/cdkd#2882 rounds 6-7 added three
in `drift.ts` (both preserve walks and `mergeUntemplatedValue`), each pinned
by an own-`__proto__` case. This site is the fourth in `drift.ts` and the
one those rounds deliberately did not take (found by the round-7 review;
fixing it there would have added production surface to a closing round).

## Remedy

`Object.create(null)` as the rebuild target, with the one-line comment the
sibling sites carry, plus an own-`__proto__` pin through
`resolveStateSecretExpressions` (and a probe reverting the target to `{}`).
Check the ARRAY arm's `out[i]` sibling in the same walk while there.

## Classification

Session-fit: next (not this session) — a production change filed during
go-to-k/cdkd#2882's final, non-production round; the class is already
contained there and this site needs its own red-without-fix pin.
Verification: `vp test run tests/unit/cli/drift-masked-leaf-preserve.test.ts
tests/unit/cli/drift-secret-redaction.test.ts` plus the new case.
Severity: low — requires a literal `__proto__` property key in a persisted
baseline (no known producer writes one); the failure is a silently dropped
revert member plus an over-refusing merge, no disclosure and no wrong value
written.
Effort: small (S) — one rebuild target, one comment, one pin, one probe;
drags the unit cycle only.
Estimate: ~30 min — the mechanism and the test shape are already
established by the three sibling sites in the same file.
Dup-check: searched issues on 2026-09-10 for `__proto__ own key`,
`Object.create(null) rebuild`, `prototype pollution drift` and
`resolveStateSecretExpressions` — no open hits; (#1943) / (#2767) / (#2802)
are the closed per-site precedents in other modules. No duplicate.

### LINKED ISSUE #2919: fix(drift): pairedLiveItems' MASK walk pairs on a corr identity that is a whole token, donating live leaves on a literal echo
## The defect

`pairedLiveItems` (`src/cli/commands/drift.ts`) refuses a WILDCARD identity value in its identity arm since PR 2912: an element whose identity field holds the caller's own marker (the mask for the mask walk; a whole token or the mask for the token walk) never pairs, because equality pairing on a marker is a guess and a live element literally carrying the marker text is producible.

The guard is keyed on the CALLER's `wildcardLeaf`, so one asymmetric shape survives in the MASK walk: a corroboration-bag identity value that is a whole `{{resolve:...}}` TOKEN is not a mask, so `isMaskLeaf` does not wildcard it — and a live element literally named by that token text (an echo of a shipped literal: cdkd deploys the token text itself for a spelling it resolves for nobody) pairs by string equality and donates its live leaves to the masked slots of the send element at that identity.

## Why it is bounded

Reaching it needs all of: a keyed list whose identity FIELD holds a surviving unresolvable token in the record, AWS echoing that literal back as the element's identity, AND a mask elsewhere in the same element (a (#2852) fail-closed mask or a NoEcho value). The donation is then from the element AWS itself names by that token text — likely the genuine counterpart, since cdkd deployed the literal — so the copy is usually correct; the hazard is the uncorroborated GUESS, not a proven wrong write. Found by the pr-security-reviewer of PR 2912 (round 1, nit) and confirmed by the parent review round; deferred out of that PR because widening the mask walk's wildcard changes mask-walk refusal behavior and needs its own fixtures.

## Remedy direction

In the identity arm, refuse an identity value that is a whole dynamic-reference token for BOTH callers (the token walk already does via `isTokenOrMaskLeaf`; the mask walk would add the token test beside `isMaskLeaf`), or parameterise an `identityRefusal` predicate separate from the corroboration wildcard. Pin with a mask-walk fixture whose corr identity is a token and whose live list carries the literal echo.

## Classification

Session-fit: next (not this session) — found at PR 2912's merge-ready point by its review rounds; the PR's own scope is settled and this changes MASK-walk behavior, which needs its own fixtures and a `drift-revert-arrays` round. Next session verification: `vp test run tests/unit/cli/drift-masked-leaf-preserve.test.ts` plus `/run-integ drift-revert-arrays`.
Severity: low — a wrong donation needs a token-named identity echoed literally by AWS beside a masked sibling leaf, and the paired element is usually the genuine counterpart; the residual on the correct fix is a refusal (the mask walk's safe drop).
Effort: small (S) — one predicate widening plus two unit fixtures; drags the unit cycle only.
Estimate: ~1 h — most of it deciding whether the widening is one-sided (mask walk only) or a shared `identityRefusal` parameter, and writing the two fixtures.
Dup-check: searched open and closed issues on 2026-09-10 for `pairedLiveItems`, `identity token wildcard mask walk` (0 hits each) and re-read (#2884) / (#2893) / (#2881): (#2884) is the positional-pairing parent this guard descends from and does not name the identity-value shape; no duplicate.

### LINKED ISSUE #2920: fix(drift): the token preserve arm copies a NON-STRING live value into a string-typed token position
## The defect

`preserveLiveValuesAtUnresolvedTokens` (`src/cli/commands/drift.ts`) copies the LIVE value into a leaf whose whole value is a surviving `{{resolve:...}}` token without checking the live value's TYPE, so a live value that is an OBJECT or ARRAY is copied into a position the baseline types as a STRING. `provider.update` then ships a container where the resource's schema expects a scalar — the same wrong-shape-to-the-wire class as (#2855), arriving through the preservation arm instead of the baseline.

Reachable before PR 2912 too (any `length >= 2` array whose frame corroborates, and any non-array token leaf, already copied `live` untyped), so it is not a regression — but that PR's `acceptForcedSingleton` widens the reachable set: a 1-vs-1 array now copies even when nothing corroborates, so `send: ['{{resolve:x}}']` against `live: [{Nested: 'v'}]` ships `[{Nested: 'v'}]` where the record held a string. Flagged by the pr-test-reviewer of PR 2912 round 2; deliberately not fixed there (the PR was at merge-ready and the fix needs its own decision — see below).

## Remedy directions

- Refuse the copy when `typeof live !== 'string'` at a whole-token STRING leaf (keep the token — the pass's residual), mirroring the mask walk's stated non-string residual in the opposite direction; or
- allow it only where the container SHAPE matches (both sides string), which is the same rule stated per-type.

Note the mask walk documents the opposite choice for its own non-string live values ("it is still copied, because sending the mask is worse") — the token walk's residual (keep the token) is cheaper than the mask walk's (drop the resource), so the trade may legitimately land differently here. Decide with fixtures for both directions.

## Classification

Session-fit: next (not this session) — flagged at PR 2912's final round with the PR at merge-ready; the remedy changes what the preserve arm ships and needs its own fixtures plus a `drift-revert-arrays` round. Next session verification: `vp test run tests/unit/cli/drift-masked-leaf-preserve.test.ts` plus `/run-integ drift-revert-arrays`.
Severity: low — needs a record whose token position AWS now holds a container at (an out-of-band restructure of a migrated position); the write is position-neutral data AWS already holds, so the corruption is the TYPE at that property, and the pre-existing `length >= 2` route has been shippable since issue 1914 with no field report.
Effort: small (S) — one type test at the copy site plus fixtures for both the string and container live shapes; drags the unit cycle only.
Estimate: ~1 h — most of it deciding the residual direction against the mask walk's documented opposite and writing the two fixtures.
Dup-check: searched open and closed issues on 2026-09-10 for `preserveLiveValuesAtUnresolvedTokens type`, `token live value object` and `non-string live` (0 relevant hits); nearest neighbours are (#2893) (pairing, closed by PR 2912), (#2919) (identity-value wildcard in the MASK walk), and (#2102) (mask by span). No duplicate.

### LINKED ISSUE #2939: fix(drift): --accept passes the non-failing redaction rules even when its destination IS the drift baseline
## The defect

`cdkd drift --accept` re-redacts the baseline it is about to persist, and passes
`STATE_SOURCED_READBACK_RULES` unconditionally — the constant that does NOT fail
closed. But that same call site has already computed where the result will land:

```ts
// src/cli/commands/drift.ts
const hasObserved = existing.observedProperties !== undefined;   // :3036
...
const redactedBaseline = redactSecretsForState(                  // :3108
  newBaseline,
  outcome.secrets,
  existing.properties ?? {},
  STATE_SOURCED_READBACK_RULES                                   // unconditional
);
...
resources[outcome.logicalId] = hasObserved                       // :3128
  ? { ...existing, observedProperties: redactedBaseline }
  : { ...existing, properties: redactedBaseline };
```

The whole reason `failClosedOnUncertifiedPositions` is DECLARED by the caller
rather than derived inside `secret-redaction.ts` is that the module cannot see
the destination — and `--accept` is the writer that argument is always made
about. That argument does not hold at THIS call site: the destination is
derivable, two lines away, from a variable already in scope.

## Consequence

On the `hasObserved` branch the destination IS `observedProperties`, i.e. a
drift baseline, so the fail-closed treatment applies by the same reasoning
(#2852) used for `cdkd state refresh-observed` and (#2885) for `cdkd import`.
Under the readback constant an uncertifiable position keeps whatever
`newBaseline` holds there instead of being masked.

The reachable scenario is narrow but real: a record whose `observedProperties`
carries a pre-GHSA plaintext at a position the walk cannot pair. The
re-redaction at `:3108` exists precisely to clean a stored plaintext the value
map cannot recognise (its own comment says so — "after a rotation the map holds
today's secret and the stored one is last week's, so only position can name
it"), and at an unpairable position the readback constant leaves it.

On the `else` branch — no baseline yet, destination `properties` — the current
constant stays correct, and that is the half worth keeping: a mask in
`properties` blocks `cdkd export` and the rollback replay rather than
protecting anything.

## Remedy direction

```ts
hasObserved ? STATE_SOURCED_BASELINE_RULES : STATE_SOURCED_READBACK_RULES
```

Worth checking the sibling `--revert` narrowing write in the same file
(`:5234`) against the same question rather than assuming it is the same answer:
its destination is also a baseline, but its bag is a merge of an AWS snapshot
with resolved desired subtrees, so the cost of a mask there is not obviously the
same.

## Not fixed in the (#2885) PR

Deliberately, and for the reason that PR gives for the same file elsewhere:
`src/cli/commands/drift.ts` is in the `integ-broad` gate scope while that PR's
files (`src/cli/commands/import.ts`, `src/deployment/secret-redaction.ts`) are
in neither `integ-broad` nor `integ-destroy`. Touching it would widen the
real-AWS fixtures that PR must pay for, for a defect it did not introduce and
does not worsen. Pre-existing since (#2852).

## Classification

Session-fit: next (not this session) — it is a `drift.ts` change, so it needs a
broad-set real-AWS round (`drift-revert` is the natural one) that the (#2885)
lane was scoped out of, plus the sibling `--revert` question above answered
rather than assumed. Next session starts with: `git switch -c <lane>
origin/main`, then read `runAccept`'s call site at `src/cli/commands/drift.ts`
around `:3036`-`:3128`.
Severity: medium — a stored plaintext that `--accept` was invoked to clean can
survive at an unpairable position in `observedProperties`. It needs a
pre-existing plaintext baseline (a pre-GHSA record, or one written before
(#2852)) AND an unpairable position AND the user to run `--accept`, so it is a
narrower population than (#2885)'s, and it never makes a CLEAN record worse.
Effort: small (S) — a ternary at one call site plus unit cases pinning both
arms; the cost is the verification cycle, not the edit.
Estimate: ~2-3 h — most of it the broad-set integ round and settling the
`--revert` sibling, not the change.
Dup-check: searched open issues on 2026-09-11 for `drift accept
STATE_SOURCED_READBACK_RULES`, `runAccept baseline fail closed`, `accept
observedProperties uncertified` and `drift accept redactSecretsForState
constant`, and grepped `src/cli/commands/drift.ts` for the call site. (#2881) is
the `--revert` mask refusal's MESSAGE wording (different call site, different
defect class — a message, not a constant); (#2885) is `cdkd import`'s capture;
(#2852) shipped the fail-closed mechanism and moved two writers onto it without
touching this one. None owns this call site. No duplicate.

go-to-k:
## Summary

Five small, file-disjoint-from-everyone-else defects in `src/cli/commands/drift.ts`, fixed together in one lane. All five live-verify through the same fixture, so one PR pays the integ once.

Closes #2899
Closes #2920
Closes #2919
Closes #2897
Closes #2939

### go-to-k/cdkd#2899 — `resolveStateSecretExpressions` rebuilt onto a `{}` literal (own `__proto__` key dropped, prototype set)

- **Fix**: the object arm rebuilds onto `Object.create(null)`, matching the preserve walks. The array arm needs no twin (numeric keys).
- **Same-class sweep of the file** (precondition minus remedy: a rebuild target or an overlay ASSIGNMENT keyed by a `JSON.parse`d bag):
  - literal rebuild targets — `grep -c 'Record<string, unknown> = {};' src/cli/commands/drift.ts`: **3 → 0** (the walk; `setAtPath`'s intermediate; `collectNarrowedTopLevelKeys`'s `delta`).
  - plain overlay assignments onto the spread AWS bag — `grep -c 'result\[topLevelKey\] ='`: **2 → 0** (`buildRevertNewProperties`, both arms → `defineOwnKey`).
  - `setAtPath`'s final write → `defineOwnKey`; `getAtPath` reads own keys only (a bare read answered `Object.prototype` for an absent `__proto__` segment). Both helpers are exported as test seams because a `__proto__` path cannot reach them through the command today — see the residual below.
- **Tests**: `tests/unit/cli/drift-secret-redaction.test.ts` — `go-to-k/cdkd#2899: --revert ships the own __proto__ member of a rebuilt node, on an ordinary prototype` (pins both consequences the issue names: the member ships, and the node is plain again so the token walk descends it). `tests/unit/cli/drift-masked-leaf-preserve.test.ts` — `buildRevertNewProperties defines a top-level __proto__ overlay as an OWN key` (2 cases; the AWS bag LACKS the key, the only shape that discriminates) and `the --accept path helpers and the narrowing delta treat __proto__ as an OWN key` (6 cases).
- **Probes** (each: production mutated, targeted file run, restored byte-exact, `git diff` empty): walk → `{}` RED (1 failed / 99); untemplated overlay → assignment RED (1/86); plain overlay → assignment RED (1/86); `setAtPath` final write → assignment RED; intermediate → `{}` + assignment RED; `getAtPath` guard deleted RED; `delta` → `{}` RED.
- **Residual, filed as go-to-k/cdkd#3121**: the comparison chain's own canonicalizers (`src/analyzer/drift-normalize.ts` and siblings, outside this lane's file scope) rebuild BOTH sides onto `{}` literals, so a top-level `__proto__` member cannot be REPORTED as drift today (a symmetric miss, never a wrong write) — which is why the CLI pin places the key one level down beside a drifting sibling. go-to-k/cdkd#2776 item 1 names this same `drift.ts` walk; that half of go-to-k/cdkd#2776 is closed here, its `rollback-executor.ts` half and item 2 are not.

### go-to-k/cdkd#2920 — the token preserve arm copied a NON-STRING live value into a string-typed position

- **Fix**: `return typeof live === 'string' ? live : value` — only a string is copied; otherwise the token is KEPT (the pass's own residual). The sibling mask walk keeps its documented opposite choice (its refusal drops the whole resource). Sweep: leaf-copy arms `return live;` **2 → 1**, the survivor being that mask walk. The two comments that said the token pass "CAN produce" a container mismatch in the mask walk's evidence source are reworded (the refusal there is kept as fail-closed).
- **Tests**: `drift-masked-leaf-preserve.test.ts` — `preserveLiveValuesAtUnresolvedTokens copies only a STRING live value (#2920)`: object / array / number+null / string control / forced 1-vs-1 singleton / keyed route / corroborated-frame route (7 cases).
- **Probe**: `?: value` → `live` RED (6 failed / 81).
- **Docs**: `docs/cli-drift.md` names the non-string case among the "token written literally" conditions.

### go-to-k/cdkd#2919 — `pairedLiveItems`' MASK walk paired on a whole-token identity

- **Fix**: the identity-arm refusal is `isTokenOrMaskLeaf(identity)` for BOTH callers instead of the caller's `wildcardLeaf` (which for the mask walk is the mask alone). Sweep: `grep -c 'wildcardLeaf(identity)'` **1 → 0**; the corroboration wildcard is untouched.
- **Tests**: `drift-masked-leaf-preserve.test.ts` — `pairedLiveItems refuses a whole-TOKEN identity in the MASK walk too (#2919)`: the token-named element is refused (`unpreservablePaths: ['Env[0].Value']`, nothing donated, nothing registered) + a positive control where a literal identity beside it still pairs.
- **Probe**: refusal back to `wildcardLeaf` RED (1/86).

### go-to-k/cdkd#2897 — `deepEqualUnordered` called two Dates equal

- **Fix**: `if (!hasPlainPrototype(a) || !hasPlainPrototype(b)) return false;` after the array arm — the guard `secret-redaction.ts`'s `deepEqualJsonValue` carries. Consumer walk: `collectNarrowedTopLevelKeys` (records the provider's echo — the truthful answer; only a key the baseline already declares is written, and only against an observed baseline), `corroboratedLeafCount` (already guarded upstream), and `runAccept`'s post-write check (its warning is reworded to name both causes).
- **Tests**: `drift-masked-leaf-preserve.test.ts` — `deepEqualUnordered treats a non-plain object as UNEQUAL unless identical (#2897)`: differing Dates, Date vs `{}`, equal-bytes `Uint8Array`s, same-object control, plain-object control (5 cases).
- **Probe**: guard deleted RED (3 failed / 84).
- **Why the `--accept` check has no CLI-level pin**: through the command a `Date` readback is flattened to `{}` by the drift normalisers BEFORE the comparator (measured: `canonicalizeTagListsDeep({D: new Date()})` → `{D: {}}`), so `awsValue` is already `{}` when the check runs — the go-to-k/cdkd#2427 class one layer earlier, recorded in go-to-k/cdkd#3121. The predicate is pinned at the function level, and the accept check calls the same function.

### go-to-k/cdkd#2939 — `--accept` passed the non-failing rules constant even when writing `observedProperties`

- **Fix**: `hasObserved ? STATE_SOURCED_BASELINE_RULES : STATE_SOURCED_READBACK_RULES` at the `--accept` re-redaction, exactly the issue's remedy. Sweep: unconditional `STATE_SOURCED_READBACK_RULES` arguments in the file **2 → 0**.
- **The `--revert` sibling the issue asked to be answered rather than assumed**: same answer, applied. A VALUE from the narrowing delta is persisted ONLY against an observed baseline (`if (!hasObserved) continue;` in the write loop — the `properties` arm takes drops alone), so wherever a redacted value lands it lands in a drift baseline. The mask costs there what it costs for `cdkd import` / `refresh-observed` (phantom drift at the position, `--accept` refuses, `--revert` preserves live, until a deploy rewrites the baseline); the alternative at an uncertifiable position is persisting the provider's echo of a decrypted value. The merge provenance of the bag changes how OFTEN a position is uncertifiable, never what a mask costs, and the refusal masks only where the source spells a reference. The `properties` arm of the ternary there is inert (values discarded) and kept for symmetry — which is why it has no case.
- **Tests**: `drift-secret-redaction.test.ts` — `go-to-k/cdkd#2939: on the observedProperties arm, a stored plaintext at an UNPAIRABLE position is MASKED` (a pre-go-to-k/cdkd#2852 record holding a rotated-away plaintext at a scalar-vs-container position; `properties` untouched), `go-to-k/cdkd#2939: on the properties arm (no observed baseline), an uncertifiable position is NOT masked` (the control wanting the other direction: a public `ssm` reference's accepted container value survives, no mask, no "NOT recorded"), `go-to-k/cdkd#2939 sibling: the --revert narrowing write masks an UNPAIRABLE echo on the observed arm`.
- **Probes**: accept → unconditional READBACK RED (the observed-arm case); accept → unconditional BASELINE RED (the properties-arm control); revert narrowing → unconditional READBACK RED.
- **Docs**: `docs/cli-drift.md` (what `--accept` writes at a position it cannot certify, per destination; both writers added to the list of paths that produce the mask), `.claude/rules/layout-drift.md`, `.claude/rules/layout-deployment-secrets.md` (net +8 B — the rules payload cap for `secret-redaction.ts` had 11 B of headroom).

## Verification

- `/check`: typecheck + lint + Prettier, `typecheck:test`, build, generators, full suite — 1004 files, 22469 passed, 1 skipped.
- `/check-docs`: the three satellites claiming `drift.ts` re-read; one stale residual sentence in `layout-drift.md` corrected.
- `/verify-pr`: markers `check` / `docs` / `verify-pr` bound to HEAD. `pr-review` is NOT set (the orchestrator's round). `drift.ts` is in neither the `integ-destroy` include list nor `integ-broad`'s cross-cutting regex.
- Five changelog entries under `changelog.d/entries/2026-09-15-<issue>-*.md`.

## Live verification (parent ran both, on `1f76665f` — the tree whose `drift.ts` is byte-identical to this HEAD)

- `/run-integ drift-revert-arrays` — PASS, 199 s, 19 created / 19 deleted, 0 errors, orphan scan clean (state prefix holds only `deployments/`; IAM / VPC / S3 / SQS by stack name all empty). The fixture the issues go-to-k/cdkd#2920 / go-to-k/cdkd#2919 / go-to-k/cdkd#2939 name.
- `/run-integ drift-revert` (broad set) — PASS, 183 s, 13 created / 13 deleted, 0 errors, orphan scan clean; all seven steps including the `--revert` narrowing write and the #1515 / #1626 arms. Ledger rows in `docs/_generated/integ-last-run.tsv`; `integ-destroy` + `integ-broad` markers set from this worktree.

## Review round 1 (lane-dispatched `pr-code-reviewer` + `pr-test-reviewer`, both against the merge base)

Applied, each with a case and a RED probe (`f296ebf8`):
- code minor 1 — `collectNarrowedTopLevelKeys` still tested membership with `in`, so a `__proto__` the plain provider echo does not own was recorded as `Object.prototype` instead of a drop; own-key membership now, and the narrowing write loop defines its key. Case: `collectNarrowedTopLevelKeys records a __proto__ the echo does not OWN as a DROP`.
- code nit 3 — the preserve walks read `liveObject?.[k]` (inherited); own-key reads now (`ownValue`). Case: `the mask walk does not copy an inherited read at a __proto__ position`.
- test nit 3 (demonstrated with a replica) — `mergeUntemplatedValue` and its mirror `collectMissingPaths` tested `key in bag`, so an AWS key named `constructor` / `toString` (a legal env-var name) merged against `Object`'s function and vanished from the revert payload; own-key membership at both loops and in the two plan reporters. Cases: `mergeUntemplatedValue keeps an AWS key named `constructor``, `...keeps a BASELINE key named `constructor` that AWS lacks`, `...the plan names that key`.
- test nits 1 + 2 — the two unfenced arms of `buildRevertNewProperties` (own-key `awsValue` read; own-key desired membership on the defensive arm) now have cases.
- Probes for this round: R1 / R1b / R2 / R3 / R4 / R5 / R6 all RED (1 failed each on the targeted file), restore byte-identical to the snapshot.

Deferred by the lane, then taken by the parent round (comment-only, `0a99e175`):
- code minor 2 — the write-site table at `src/deployment/secret-redaction.ts` and `STATE_SOURCED_BASELINE_RULES`'s docstring read stale after this PR. The two `cdkd drift` rows now say `BASELINE / READBACK by destination ***`, a `***` footnote names the caller-side selector, and the docstring's last paragraph names both drift writers with the #2939 link.

## Review round 2 (delta-scoped, both reviewers): Clean at blocker / minor level

Nits applied as a final polish batch (`068c8ffd`), no further round: the `hasPlainPrototype` guard comment in `mergeUntemplatedValue` still described the loops as `key in desiredValue` (reworded); the two plan reporters read `awsProperties[key]` bare (own-key now, symmetric with the merge); the three inline `Object.prototype.hasOwnProperty.call` copies use the new helpers; one discriminating case for `findRevertUnbaselinedAwsKeys`' top-level own-key membership (probe R7 RED). Both reviewers agreed the token walk's own-key `liveObject` read and the mask walk's `corrObject` read have no constructible discriminating case (no inherited value is a string or an array), and that `findRevertPreservedTagKeys`' switch and the narrowing loop's `defineOwnKey` are unobservable — left unfenced by design.

## Parent round (independent, 3-axis + security, all against merge-base `a386124b` on `068c8ffd`)

- `pr-spec-reviewer` (spec = the five issue bodies): Clean, every decision cited at a line; the two deviations (#2939's `--revert` sibling ternary, #2897's accept-warning rewording) confirmed as stated and sound.
- `pr-code-reviewer`: Clean. `in`-membership on bags 0 remaining; the one surviving `{}` rebuild (`siblings`, keyed by CFn logical ids) correctly exempt; null-proto bags reach no consumer that assumes a prototype (`hasPlainPrototype` accepts `null` on both sides; `JSON.stringify` / spread / `Object.entries` read own keys only; no `.constructor ===` or bare `.hasOwnProperty(` call in `src/`); `observedProperties: {}` takes the BASELINE arm at both writers and the two predicates read the same record. Two nits, both won't-do: the two remaining `Object.prototype.hasOwnProperty.call` sites (`corroboratedLeafCount`, `deepEqualUnordered`) are functionally identical to the helper — left as-is; a `typeof live !== 'object'` spelling for #2920 was considered and declined because the issue's direction A (string-only copy) is the one the position's TEMPLATE type licenses, and no reachable input differs after #2482.
- `pr-test-reviewer`: Clean at blocker level; re-ran six mutation probes independently (each RED at the named assertion, positive controls green, restore byte-exact). One minor applied (`1f76665f`): the `drift-secret-redaction.test.ts` describe block labelled itself as pinning #2897's `--accept` post-write check while holding no such case — retitled, with the reason the case cannot be built (a `Date` readback is flattened to `{}` by the normalisers before the check; go-to-k/cdkd#3121). One nit applied (`0a99e175`): a positive control pinning `hasPlainPrototype`'s null-prototype acceptance (`a null-prototype nested bag equals its plain twin — no narrowing`); probe dropping `proto === null` RED at that control plus six revert cases. One nit recorded as unfenceable by construction (three own-key sites with no constructible discriminating input), already named under round 2.
- `pr-security-reviewer` (fable): Clean. Write→reader trace for both #2939 arms: `failClosed` is consumed only at `refuseAgainstSource` and `orphanLiterals`, both in the "mask more" direction, so the `properties` arm is byte-identical to before and the `observedProperties` arm strictly stricter; `observedProperties: {}` and `observedBaselineRefused` cases reasoned through; #2920's returned `value` is always a `{{resolve:...}}` token (gated by `isWholeDynamicReference`), never plaintext; null-proto bags skip no masking arm.

## Sweep residual not fixed here

- go-to-k/cdkd#3121 — the analyzer-side canonicalizers (`drift-normalize.ts`, `drift-principal-normalize.ts`, `cc-api-strip.ts`), the same rebuild-onto-`{}` class on the comparison path (symmetric `__proto__` miss; `Date` flattened to `{}` before the comparator). Outside this lane's `drift.ts` scope; measured in the issue body.


go-to-k:
Merged — thanks. Five defects in one file, each with its own RED probe and positive control, and the two `drift` writers now agree with the write-site table; the parent round (spec / code / test / security) and both real-AWS fixtures came back clean on the first pass.

Issue #2740, opened 2026-09-07
fix(diff): an Output that never resolved at deploy is a phantom ADD in cdkd diff --fail on the unchanged stack when its failure happens inside a secret lookup the diff skips


nix-tkobayashi:
## What is wrong

An Output that has NEVER resolved at deploy — one whose resolution fails inside the resolver on every deploy under the default (non-`--strict-getatt`) arm — and whose failure `cdkd diff` does not reproduce, is a phantom `ADD` row in `cdkd diff --fail` on the unchanged stack, on every run.

The deploy side (`DeployEngine.handleOutputResolutionFailure`, `src/deployment/deploy-engine.ts`) warns, stores `undefined` for the key and skips it, so `state.outputs` never gets the key (an output that resolved on an EARLIER deploy keeps its stored value: the no-change path preserves the whole previous bag when any output fails, so a key that once resolved is not this shape). The diff side (`src/analyzer/outputs-diff.ts`, `computeOutputsDiff` over `resolveTemplateOutputs`) resolves outputs with `skipDynamicReferences: true`: a SECRET reference (`secretsmanager`, `ssm-secure`, an SSM `SecureString`) is left as its token rather than looked up, so a value whose deploy-time failure happens INSIDE that lookup — a JSON key that does not exist, a reference assembled from another secret's value — assembles at diff time without throwing (a plain `ssm` reference is still fetched with `WithDecryption: false` and can still throw, so it is not this shape either). `resolutionFailed` stays `false`, the state lacks the key, and the diff pushes `ADD <key>`.

Measured (2026-09-07, by the code reviewer on the go-to-k/cdkd#2728 PR, through the real resolver + `resolveTemplateOutputs` + `computeOutputsDiff` on that PR's `OutputFailureLeak` output): `resolutionFailed: false`, `[{ name: 'OutputFailureLeak', changeType: 'ADD' }]`. That PR gates the output behind `CDKD_TEST_OUTPUT_LEAK=true` so the `secrets-dynamic-ref` fixture's unchanged-stack `diff --fail` guard (Guard 6) never sees it — the gate is the fixture's workaround for this class, not a fix.

## Why it matters

`cdkd diff --fail` is the gate for "nothing to deploy". For this shape it reports an `ADD` the deploy will never perform, and stays red until the template changes — repairing the referenced secret / key, or the output's expression, is the way out, and the deploy's own warn already says which output. The diff's row is a second signal that names the wrong thing.

## Proposed fix

[...]

Issue #1838, opened 2026-08-13
fix(deployment): ResourceUpdateNotSupportedError interpolates the logical id into a message the retry classifier substring-matches, so a resource named *DependencyViolation* burns the full backoff


go-to-k:
Found by the review of PR #1827's non-retryable-marker mechanism. Unlike the
case that PR closes, **this one is live today**.

## The defect

`src/deployment/retryable-errors.ts`'s classifiers match by SUBSTRING against
the error MESSAGE. `ResourceUpdateNotSupportedError`
(`src/utils/error-handler.ts:308-315`) interpolates the resource's **logical
id** into its message, and is thrown from roughly 20 providers from INSIDE the
`update()` call that `deploy-engine.ts:3384` wraps in `withRetry`.

`DependencyViolation` is the only whitespace-free entry in
`RETRYABLE_ERROR_MESSAGE_PATTERNS`, and the match is a substring rather than an
equality, so an ordinary composite CDK logical id is enough:

- logical id `MyDependencyViolationSub` -> `isRetryableTransientError` returns
  **true** (measured)
- the deterministic "this type does not support UPDATE" refusal is therefore
  retried through the full generic schedule (8 retries, ~47s of sleep)
- only after that does the deploy reach the `--replace` DELETE+CREATE fallback
  the error exists to trigger

So the cost is ~47s of pure latency per affected resource on a path that was
never going to succeed, plus a misleading run.

## Why it was not fixed in PR #1827

That PR added the mechanism which fixes this shape — `markNonRetryable` /
`isMarkedNonRetryable`, a `Symbol.for('cdkd.nonRetryable')` marker walked down
the `.cause` chain and consulted by `isRetryableTransientError` before anything
else. Marking this error is a one-line change in its constructor.

But `src/utils/error-handler.ts` was owned by the concurrent #1710 lane for the
whole of that work, and the repo's one-lane-per-file rule applies.

## Suggested direction

Call `markNonRetryable` in the `ResourceUpdateNotSupportedError` constructor.
It is a deliberate, deterministic cdkd-authored refusal — the deploy engine
matches it BY CLASS to choose the replacement fallback
(`deploy-engine.ts:3406-3411`) — so it is never worth retrying, independently of
what its message happens to contain.

While there, audit the sibling refusal classes in the same file for the same
shape: any error that (a) interpolates user-controlled text and (b) is thrown
inside a retried call is a candidate.

## Verification bar

[...]

PR #3125, opened 2026-09-14
fix(resolver): refuse a live-read attribute the record omits instead of returning the physical id

### LINKED ISSUE #3096: fix(resolver): the EC2 Instance live-lookup arm returns the instance id for a still-unassigned IP / DNS / AZ attribute instead of refusing

go-to-k:
## Summary

Closes #3096

When the state record OMITS an attribute (go-to-k/cdkd#3077 made an unassigned read-back absent rather than `''`) and the resolver's live re-read finds it unassigned or fails, the resolver used to hand back the PHYSICAL ID — an `i-...` where an address belongs, a `vpc-...` in a security-group slot — loud only where AWS rejects the shape, silent in an Output or an export. It now refuses.

Design decisions, as landed (`src/deployment/intrinsic-function-resolver.ts` at this HEAD):

1. **`AWS::EC2::Instance` live arm** (L5547-5653): `PrivateIp` / `PublicIp` / `PrivateDnsName` / `PublicDnsName` / `AvailabilityZone` — a `DescribeInstances` that finds the member unassigned, or throws, raises `IntrinsicResolutionRefusalError` via the new `refuseUnservedAttribute` helper (L5877), UNMARKED (a `pending` instance settles; a failed describe can succeed on retry). Nothing is cached on a refusal. `InstanceId` still answers the physical id. **One refinement of decision 1, flagged for the parent**: a SETTLED instance (any reported state but `pending`) whose `PublicIp` / `PublicDnsName` is unassigned resolves to the known `''` (L5621-5626), cached — the same three-state rule `EC2Provider.describedInstanceAttributes` records by, now ONE shared predicate `isSettledInstanceState` (`src/provisioning/ec2-instance-state.ts`, a leaf, source-shape-fenced in both readers). Evidence: that provider's own doc comment (`ec2-provider.ts` L388-394 on `origin/main`) tracked exactly this residual on go-to-k/cdkd#3096 ("a `--no-wait` create of an instance in a PRIVATE subnet omits the public pair, and nothing later writes it ... degrades to the instance id ... until the next `update()`"); a literal refuse-when-unassigned would turn that residual into a PERMANENT refusal of `Fn::GetAtt [Inst, PublicIp]` on every later no-change deploy of such a stack, where CloudFormation answers `''`. Pending / no-state / a settled instance missing a PRIVATE member (terminated) still refuse.
2. **`AWS::EC2::VPC` `DefaultSecurityGroup`** (L4792-4864): the arm no longer reads `attributes.DefaultSecurityGroup || physicalId` (the flat lookup already served any present key, so the arm was always the VPC id). It re-reads live with the same `vpc-id` + `group-name=default` filtered `DescribeSecurityGroups` `createVpc` makes, caches per physical id (`cachedVpcDefaultSecurityGroups`, L1411, cleared in `resetAccountInfoCache`), and refuses (unmarked) on an empty result or a failed read.
3. **The three `guardedPhysicalIdFallback` members from the issue comment**: `AWS::CloudFront::Distribution` `DomainName` **re-reads live** (L5718-5785; `AwsClients` already exposes `cloudFront`, so only `GetDistributionCommand` is newly imported; cached in `cachedCloudFrontDomainNames`; `Id` answers the physical id). `AWS::RDS::DBProxy` / `DBProxyEndpoint` `VpcId` **refuse only** (L5793-5817): `AwsClients` has no RDS client and the resolver imports no RDS package, so it is a value-typed refusal decided from the omitted key alone — hence MARKED non-retryable (no retry rewrites the record; the message interpolates the logical id, go-to-k/cdkd#1838).
4. **`tests/integration/ec2-instance/verify.sh` Phase 0 (c)**: widened to the two legal arms — the live IPv4 address (matched against `^[0-9]+(\.[0-9]+){3}$`), or ABSENT from `.outputs` with `PublicIp` recorded in `.skippedOutputs`; `''` / `null` and any `i-*` value FAIL (the discriminating negative). Phase 1's heal assertion is unchanged: on the no-change second deploy `resolveOutputs` runs again (`deploy-engine.ts` L3131-3143), the failed-key merge only applies when something failed (L3159-3175), `outputsChanged` (L3218) triggers the save, and `skippedOutputs` is recomputed from THIS pass (L8566) — the (a-refusal) case in `deploy-engine-skipped-outputs.test.ts` pins that heal against the mocked resolver.
5. **Docs / rules**: `docs/cli-deploy.md` (the `--no-wait` second-deploy bullet: refusal replaces "instance ID with a warning"; the no-change Outputs bullet now describes the go-to-k/cdkd#2771 merge instead of the retired all-or-nothing keep), `docs/cli-deploy-safety.md` (a fourth rule: live-read attributes never fall back to the physical id), `src/utils/error-handler.ts` class JSDoc (enumeration + the "exactly one time-dependent site" claim), `.claude/rules/architecture.md` / `layout-deployment.md` (pointers only — both files sit at their payload caps), `ec2-provider.ts` comments that cited go-to-k/cdkd#3096.

### Message shape

`Cannot resolve Fn::GetAtt [MyInstance, PublicIp] for AWS::EC2::Instance: DescribeInstances reports no PublicIp yet (state pending). The physical id "i-0123..." is not a usable PublicIp, so cdkd refuses to substitute it. Deploy without --no-wait so the instance is running before its attributes are read, or reference the attribute from a later deploy once it is.`

A failed read names the error CLASS, plus the HTTP status when the SDK attached an integer one (`DescribeInstances failed (UnauthorizedOperation, HTTP 403); re-run with --verbose for the AWS error text`); AWS's raw message goes to debug via `describeFailureObserved` (L5904) and is NOT threaded as `cause` — `formatError` prints `Caused by: <cause.message>` at default verbosity, and a denied describe quotes the caller's account / role / session.

### What the refusal does at each consumer (traced)

- **Resource property**: `provisionResourceBody` resolves properties (`deploy-engine.ts` L5398) inside `withResourceDeadline` (L4738-4750), outside `withRetry`; the throw reaches the per-resource catch (L4848) → `RESOURCE_FAILED` (L4883) → rollback. Previously AWS rejected the `i-...` at most consumers; a free-text sink accepted it.
- **Stack Output**: caught per output by `handleOutputResolutionFailure` (L8317 → L8038); the default arm warns and sets the key `undefined` (L8091), `collectSkippedOutputs` records it (L8566, go-to-k/cdkd#2740), the deploy exits 0, and the next deploy re-resolves it (Phase 1). Under `--strict-getatt` it fails the deploy (L8083) with `cause` threaded.
- **`Fn::Sub`**: `resolveSub` re-raises the class (`intrinsic-function-resolver.ts` L6407, the go-to-k/cdkd#1740 rule) — pinned by a unit case.
- **`Fn::If` / `Conditions`**: `evaluateConditions` still absorbs any failure to `false` (L3334) — unchanged, stated in `refuseUnservedAttribute`'s doc.

### Sweep receipts

- `grep -n "|| physicalId\|?? physicalId" src/deployment/intrinsic-function-resolver.ts` — before: 2 (the VPC arm + `queueName` from an SQS URL); after: 1 (the SQS one, a URL-segment fallback, not this class).
- `grep -c "returning physical ID" src/deployment/intrinsic-function-resolver.ts` — before: 2; after: 1 (`guardedPhysicalIdFallback`'s generic warn, deliberately kept per its doc: an `Alias` / `Endpoint` is shape-indistinguishable from a name).
- `git grep -n 3096 -- src tests docs .claude`: the `ec2-provider.ts` docstring + `createVpc` comment and `verify.sh` (c) were the three live citations; all three rewritten. The go-to-k/cdkd#3077 changelog entry's parenthetical is left as-is (it describes the behaviour at that release).

### Mutation receipts (unit; tree restored byte-exact between probes, `git diff --quiet HEAD` after each)

| Probe | Mutation | Result |
| --- | --- | --- |
| P1 | EC2 unassigned arm `return physicalId` before the refusal | RED: 3 cases (pending refusal, terminated private member, `Fn::Sub`) |
| P2b | EC2 `catch` returns `physicalId` | RED: 1 (describe-throws refusal) |
| P3 | settled known-`''` branch disabled | RED: 1 (CloudFormation-parity case) |
| P4 | VPC arm restored to `attributes.X \|\| physicalId` | RED: 3 (live read, empty refusal, failed-read refusal) |
| P5 / P5b | VPC empty result returns id / failed read throws a plain `Error` | RED: 1 each |
| P6 / P6b | CloudFront arm disabled / empty `DomainName` returns id | RED: 2 / 1 |
| P7 | DBProxy refusal unmarked | RED: 3 (both type cases + the refusal-site census) |
| P8 | DBProxy arm disabled | RED: 2 |
| P9b | `refuseUnservedAttribute` marked non-retryable | RED: 6 (five `isMarkedNonRetryable` false assertions + census) |
| P10 | raw AWS text returned in the `observed` clause | RED: 3 (the `not.toContain('assumed-role')` assertions) |
| P11 | refusal cached | RED: 1 (call-count assertion) |
| P12 | `handleOutputResolutionFailure` re-raises the refusal class | RED: 1 (`deploy-engine-skipped-outputs` a-refusal) |

(P2 as first written placed the mutation after a `never`-returning call and stayed GREEN — a dead-code probe, voided and redone as P2b. P9 as first written did not compile — "no tests", voided and redone as P9b.)

Every probe ran 258 tests (the two resolver suites) or 28 (the engine harness); no load errors except the voided P9.

### Review rounds

Round 1 (3-axis + security, HEAD 19a12f0c): spec clean; code M1 + 3 nits; test 2 minors + 2 nits; security 2 nits. Round 2 (delta 19a12f0c..7d85dfed): the `observed:` source pin was prefix-keyed (a leading quote admitted `'literal ' + physicalId`, a leading masker call admitted `${mask(x) + physicalId}` — both measured passing the pin and the mask-coverage checker) → anchored on whole expressions; `typeof status === 'number'` vs `!== undefined` was indistinguishable → a string-status fixture; three cosmetic nits in `describeFailureObserved`. Round 3 (delta 7d85dfed..f0ca5f91): the anchored pin still admitted `.concat(...) || ``` / a ternary between two backtick-bounded ends (measured) → pinned BY CONSTRUCTION (holes removed, the remainder must be exactly one template literal); `Number.isInteger(status)` on `unknown` had dropped the type narrowing (the repo's only `restrict-template-expressions` warning) → `typeof status === 'number' && Number.isInteger(status)`. Closing probe (delta f0ca5f91..5cf790b9, twenty shapes): no admitted unmasked shape on a tree that passes `vp check`; two low notes applied — the masked hole is now pinned to `attributeName` by name (a masked `physicalId` had satisfied "masked" while contradicting the `not-in-class(observed)` note), and the pin's comment states its reach honestly (it reads the `observed:` LINE; a continuation led onto the next line is folded back by the format gate, a re-binding of `observedState` trips lint). The branch was then flattened to one commit. Declined, recorded here: extending `scripts/check-resolver-mask-coverage.ts`'s population to `refuseUnservedAttribute({ observed })` arguments (security round 1's "better" option) — the by-construction source pin fences the same claim, and widening the AST checker is a change to a security instrument that deserves its own PR rather than a review-nit batch.

**M1, code round 1 — declined as a design change, fixed as prose.** The reviewer showed that leaving the refusal unmarked buys no actual retry: `isRetryableTransientError` reads throttle names / statuses off the `cause` chain (not threaded here, per the per-reader rule) and everything else by message substring, and this message matches nothing on its own — so a pending / throttled refusal is retried only when the interpolated `logicalId` (or error class) happens to carry `DependencyViolation` / `QueueDeletedRecently` / `StateMachineDeleting`. Verified by running. The design stays unmarked (the parent's decision; the marker is a DECLARATION that no retry can succeed, and the fabricated-account guard sits in exactly the same place); the four prose sites that read as a mechanism claim were rewritten to say what the classifiers see. The alternative — a sanitized SDK error threaded as `cause` so a throttle classifies transient — was considered and left out: `formatError` prints `Caused by: <cause.message>` for every `CdkdError` at default verbosity with no secret masking (confirmed by the security round), so the clone would need its own fence, and the pending arm gains nothing from it.

Round-2 fix probes (each RED at the intended case, tally 263; the first attempt passed the test files as one quoted argument, ran NO tests and was voided): R1 settled arm widened to `attributeName !== 'PrivateIp'` → 1; R2 `'no instance state'` text → 1; R3 the no-state remedy dropped → 1; R4 the `--verbose` debug line deleted → 1; R5 HTTP status dropped → 1; R6 / R7 one new cache not cleared by `resetAccountInfoCache` → 3 / 2; R8 `${physicalId}` added to an `observed:` clause → 1 (the new pin). Round-3 probes: S-A `'literal ' + physicalId` → pin red; S-B `${mask(x) + physicalId}` → pin red (+2); S-C `const observedState = physicalId` → pin red (+4); S-D `status !== undefined` gate → 1. Round-4 probes against the by-construction pin: `.concat(...) || ``` / ternary / `+` / `??` / `'literal' + physicalId` / parenthesised template → all red at the pin; `${this.maskSecretsForLog(physicalId, context)}` → red at the pin since the hole is pinned to `attributeName`.

### Parent's independent round (HEAD 3019d2e2 → f633a24d, one flattened commit, rebased onto 52e08724)

Zero blockers; all eight items landed. MUST-LAND: (1) pending × `''` member fenced — mutant dropping `&& value !== ''` → 1 red; (2) same-template heal case in `deploy-engine-skipped-outputs.test.ts` (resolver mock toggled by a hoisted flag, digest unchanged) — conceptual mutant "skip resolving a key whose digest still matches" in the no-change path → 3 red (the new case, (b), and the add-beside case); (3) `isSettledInstanceState` leaf (`src/provisioning/ec2-instance-state.ts`) read by both `describedInstanceAttributes` and the resolver, with a table test + a source-shape fence — mutants: table flipped (`pending` settled) → 4 red; either reader re-inlined → 1 red each; (4) every interpolated physical id at the refusal sites goes through `maskThenStripThenMask` — **the sweep reached the two pre-existing `guardedPhysicalIdFallback` throws (L5988 / L6006) as well**, and the mask-coverage BL1 probe's premise moved to that spelling; (5) the VPC arm refuses (marked) a physical id that is not `vpc-<hex>` BEFORE the cache read and the filtered read (L4810) — mutant loosening the regex to `^vpc-` → 1 red. **Won't-do, recorded here**: `ec2-provider.ts`'s own identical filtered reads (`createVpc` ~L972, `getVpcAttribute` ~L1169) are the same class and are left as-is — a state writer can write attributes directly, so the resolver-side guard is not an escalation boundary; the provider reads take the id from its own `CreateVpc` response, not from a record. TAKE: (6) a non-`Error` throw names `Error` (mutant restoring `typeof err` → 1 red); (7) verify.sh Phase 1 asserts `.skippedOutputs.PublicIp` is absent after the heal (jq executed both polarities); (8) DBProxy `Endpoint` stays on the generic fallback (mutants widening / always-true `VpcId` condition → 1 red each). Mask census 133/158/106 → 134/159/107 (the VPC shape refusal: one mask, one note).

### Delta round (HEAD f633a24d → 70839926, one flattened commit on 52e08724; the first push, 259307f2, failed the English-only CI scan on a CJK error-name fixture — replaced by escaped Greek letters, same receipt)

No blockers; one real defect from the round-1 caches, fixed: the cache reads sat ABOVE the shape guards and the caches were plain `{}`, so a state physical id of `constructor` / `__proto__` read a FUNCTION out of `Object.prototype` and served it with zero AWS calls (the reviewer ran it). All three caches are now `Object.create(null)` (L1385-1411, still cleared by `delete`), the VPC guard sits above its cache read (L4810), and CloudFront gained a shape guard `/^[A-Z0-9]+$/` (L5726, marked). **Receipts** (tally 267 / 276): the `constructor` / `__proto__` behavioural cases → refusal, no AWS call; D1a2 revert one cache to `{}` → red at the new source pin (`intrinsic-refusal-non-retryable.test.ts`); D1d / D1e guard moved back below the cache read → **green** on their own — with a null-prototype cache an early read of an unknown key is harmless, so the two mechanisms are mutually redundant, which is the point of having both; D1f plain cache AND guard below → red at the behavioural `constructor` case. Item 2: `remedy: settled ? … : …` (D2 inverted → 3 red). Item 3: the settled-predicate fence scans comment- and string-stripped code, requires `= isSettledInstanceState(stateName)` in both readers, refuses any `stateName` ↔ `'pending'` comparison in either operand order (D3a re-inlined with a decoy comment → red; D3b swapped-operand re-inline in the provider → red), and the leaf fence refuses `export … from` / `import(`. Item 4: `displaySafe(this.maskThenStripThenMask(physicalId, context))` at all six sites (the two pre-existing `guardedPhysicalIdFallback` throws included); a U+2028 physical id case (D4 `displaySafe` dropped at `refuseUnservedAttribute` → red). Item 5: BL1 probe relabelled to the first strip-and-mask site (DBProxy `VpcId`). Item 6: a non-ASCII `err.name` case (D6a `|| 'Error'` dropped → red) and the same-template heal case now asserts the second template's digest equals the stored row. Mask census 134 / 159 / 107 → 135 / 160 / 108, and the `BANDS` floors moved by the same delta (131 / 155 / 98): the `resolveFindInMap` block-comment injection swallows 7 statements and 135 − 7 = 128 had landed exactly on the old floor, so the band instrument had gone inert (its own liveness case caught it).

## Test plan

- [x] `vp run check`, `vp run typecheck:test`, `vp run build`, generators regenerated (no drift), `vp test run`: 1005 files / 22528 tests green, rooted in this worktree
- [x] Unit: `tests/unit/deployment/intrinsic-functions.test.ts` (EC2 refusal / describe-throws / `InstanceId` / running negative control / settled `''` parity / terminated private member / `Fn::Sub`; VPC `DefaultSecurityGroup` served / live / empty / failed; CloudFront `DomainName` live+cache+`Id` / empty+failed; DBProxy + DBProxyEndpoint served / refused-marked), `deploy-engine-skipped-outputs.test.ts` (a refusal-class Output is skipped, recorded, healed on the next deploy), `intrinsic-refusal-non-retryable.test.ts` census (two deliberate unmarked sites), `resolver-mask-coverage.test.ts` pins (132/155/100 → 133/158/106)
- [x] `/run-integ ec2-instance` — PASS, 117 s (2026-09-14T21:13Z, us-east-1): Phase 0 pending-path omission observed on BOTH instances (no `''`); the Outputs pass ran after EC2 assigned the address, so (c) took the *assigned* branch (`3.235.129.129`) — the *skipped-output* branch was not entered live this run (its `''` / `i-*` negatives held; the refusal → `skippedOutputs` → same-template re-resolve chain is pinned by the resolver + `deploy-engine-skipped-outputs` unit cases); Phase 1 healed to the live address and `skippedOutputs.PublicIp` cleared; destroy 0 errors, 0 orphans
- [x] `/run-integ lambda` (broad set) — PASS, 110 s, 9 deleted 0 errors 0 orphans; `/run-integ drift-revert-vpc` (broad set, VPC) — PASS, 288 s, 21 deleted 0 errors 0 orphans → `integ-broad` set
- [x] `/run-integ s3-cloudfront` (CloudFront `DomainName` consumer) — PASS, 252 s, 5 deleted 0 errors 0 orphans; `/run-integ ecs-schedule-targets` (the fixture that resolves `Fn::GetAtt [Vpc, DefaultSecurityGroup]`) — PASS, 83 s, 21 deleted 0 errors 0 orphans. `ec2-vpc` has no `verify.sh` (maintainer-only from a session) and was not run. The VPC / CloudFront live arms fire only on an OMITTED key, which no fixture manufactures; these runs are the consumer regression nets, and the arms are pinned by unit cases. Account-wide sweep after the runs: 0 `state.json`, no fixture VPCs / functions / distributions / clusters / buckets

Maintainer-side review: 4 reviewers (spec / code / test / security, Fable) at 3019d2e2 — no blockers; delta round (code / test / security) at f633a24d — no blockers, one real defect found and fixed in the last touch (plain-object per-deploy caches read before the shape guard resolved a state `physicalId` of `constructor` / `__proto__` to a function object; caches are now null-prototype and both guards precede their cache read, with cases + receipts).


go-to-k:
Merged — thanks. The shared `isSettledInstanceState` leaf and the null-prototype caches with the guards ahead of the reads were the two things the rounds bought; the refusal is now the single answer for a value the physical id can never stand in for.

Issue #3126, opened 2026-09-14
test(integ): nine local-* fixtures still capture the CLI with `$(... 2>/dev/null | tail -1)` under pipefail, so a non-zero exit aborts the script with no diagnostic


go-to-k:
Nine `local-*` fixtures still read the cdkd CLI's last stdout line with the shape

```bash
RESULT_1=$(${CDKD} local invoke <Fn> --no-pull 2>/dev/null | tail -1)
```

Every one of them runs under `set -euo pipefail`, so when the CLI exits non-zero the pipeline fails, the command substitution fails, and `set -e` kills the script AT THE ASSIGNMENT — before the assertion that would print `FAIL: ... got: ${RESULT_1}`, and with the CLI's stderr already discarded by `2>/dev/null`. The run's log then ends at `[2/4] Invoking ...` with no error text at all, which is how a transient during go-to-k/cdkd#3106's verification read as an unexplained abort. PR go-to-k/cdkd#3118 replaced the shape in `tests/integration/local-invoke-layers/verify.sh` with a `capture` helper (exit status captured explicitly, the captured stderr's tail printed on a non-zero exit, the last stdout line still emitted so the assertion runs and FAILS with its own diagnostic — the shape cdk-local's twin fixture carries) and left the siblings alone.

Measured on `origin/main` at 66cb629d (`git grep -E '\$\(.*2>/dev/null *\| *tail' -- 'tests/integration/*/verify.sh'`): 33 sites in 9 fixtures — `local-invoke` (7), `local-invoke-agentcore` (10), `local-invoke-container` (3), `local-invoke-dotnet` (2), `local-invoke-java` (2), `local-invoke-provided` (2), `local-invoke-python` (4), `local-invoke-ruby` (4) — plus 3 more inside an `if out=$(...)` condition in `local-invoke-from-state`, `local-invoke-from-cfn-stack` and `local-invoke-from-cfn-stack-multi-stack`, where `set -e` is suspended so the retry loop does run, but the CLI's stderr is still lost on every attempt.

Two things to land together:

1. Adopt `capture` (from `local-invoke-layers/verify.sh`) at the 33 abort-shaped sites, and give the 3 `if`-guarded sites a stderr file so a failed attempt's error text reaches the log.
[...]

PR #3127, opened 2026-09-14
chore(work-issues): bind the pr-review sentinel only when the PR head equals local HEAD, and print verify.sh's rc from the watchdog block

### LINKED ISSUE #3037: chore(deps): decide whether to raise the published Node floor past 20 (EOL 2026-04-30), which is what currently pins commander at 14

go-to-k:
## Summary

`/work-issues` stage 10 (retro) for the run that closed go-to-k/cdkd#3037 (PR go-to-k/cdkd#3105) and go-to-k/cdkd#3106 (PR go-to-k/cdkd#3118). Two flow lessons, each amended in the step where it fired; no new bullets.

- **`/review-pr` step 6** — the sentinel snippet read `gh pr view --json headRefOid` straight into `.markgate-pr-review-sha`. Right after a push GitHub can still answer the PREVIOUS head, so the marker binds to a stale sha and `pr-review-gate` refuses the merge with `bound to <old> (mismatch)`. Memory recorded this on go-to-k/cdkd#879 with the equality remedy; it hit twice more in this run — a written rule violated anyway, so the check now lives in the snippet: bind only when the PR head equals `git rev-parse HEAD`, run after `gh pr checks --watch`, never in the push's own call, and `${SHA:?}` refuses an empty answer (a code review measured that two empties compare equal and would bind a one-byte sentinel). Paid for inside the file: the `## Important` section restated the intro and CLAUDE.md's "when in doubt go UP", and the go-to-k/cdkd#404 parenthetical restated the dry-run row. 22,965 B under the 23,000 B cap (the `NEVER set the marker` parenthetical restating the hook, which CLAUDE.md carries, paid for the review fix).
- **`/run-integ` step 5** — the watchdog block captured `RC` and never printed it, so the tool output read by the next call carried the grep count and no verdict. In this run a FAILED `local-invoke-layers` run had `integ-local` set and a `PASS` ledger row written in the same call as the run (undone by a clean re-run). The block now echoes `verify.sh rc=$RC`, and the prose says steps 6-13 are LATER calls that read it.

## Evaluated and refused

A PreToolUse hook refusing `verify.sh` + `markgate set integ-*` / a `PASS` row in one Bash call was measured against `.claude/rules/hooks.md`'s criterion and does not qualify: the marker is per-worktree and unset-able, the ledger row is an uncommitted local edit, so the harm neither completes at the action nor lands on a third party's artifact. One instance; the text amendment above is the minimal remedy.

The user's denial of a glob `rm -rf` in the OS tmp root (a probe of code calling `os.tmpdir()`) was a single correction, and the flow text already names the scratch root; it went to memory (`feedback_probe_tmpdir_in_scratch_root.md`), not to a stage file.

## Filed

- go-to-k/cdkd#3126 — nine sibling `local-*` fixtures still carry the `$(... 2>/dev/null | tail -1)` capture shape PR go-to-k/cdkd#3118 replaced in `local-invoke-layers`; a fixture-class defect, not a flow lesson.

## Mirrors

- cdk-local: its `/review-pr` carries the same sentinel line — mirrored in go-to-k/cdk-local#732 under that repo's flow.
- cdk-real-drift: no `/review-pr` skill, no watchdog block — not applicable, measured.

## Verification

- `vp check --fix`, `vp run check`, `vp run typecheck:test`, `vp run build`, `vp run gen:all-matrices && vp run audit:coverage:check && vp run format`: all rc=0.
- `vp test run`: 1006 files / 22529 passed, 1 skipped, rooted at the lane tree (single `RUN` header asserted).
- `tests/unit/scripts/skill-file-payload.test.ts` green at the new sizes.


go-to-k:
Merged, thank you — both review rounds caught real gaps (the two-empties bind and the bare-abort message), and the mirror is go-to-k/cdk-local#732.

Issue #1120, opened 2026-07-20
test(integ): close the remaining pattern-2 lint gaps (capture-form probes + function-wrapper detection)


go-to-k:
Follow-up to issue #1097 pattern 2 (shipped in PR #1110). #1097 was closed by PR #1116 (the CLI-flag lint, its last named item), but two lint-coverage gaps from the #1110 3-axis review were left as follow-ups on that issue and now have no open tracking. Both let the exact bug class #1110 eliminated survive in spellings the lint does not scan.

## 1. Capture-form fallback probes: `$(aws ... 2>/dev/null || echo 0)`

A count/list capture with an error-swallowing fallback reads a throttle as "0 remaining", the same silent-pass defect as a blind gone-probe, in statement positions the classifier never scans (it only handles `if`/`elif`/`while`/`until` and `&&`/`||` list-operator conditions). The one leak-assert instance (`backup/verify.sh` `REMAINING_PLANS`) was fixed in #1110; roughly 8 similar sites remain, mostly live-phase checks rather than post-destroy leak asserts. Work:

- Sweep the remaining sites onto strict captures (capture `2>&1`, hard-FAIL on probe error, then evaluate).
- Extend `scripts/check-integ-probe-not-found.ts` with a capture-form category: a `$( aws <read-verb> ... 2>/dev/null ... )` or `|| echo <literal>` fallback inside a command substitution is a violation. Same read-verb restriction as the existing categories to avoid mutation false positives.

## 2. Generic function-wrapper detection

A wrapper like the old `ssm_exists()` (silenced read probe defined once in a function, used as a gone/leak assert at call sites) is invisible to the line-oriented classifier. The known instances (destroy-interrupt, conditions-and-if, conditions-update-2, replacement-immutable-name, update-policy-mutations, glue-update-hardening) were fixed by hand in #1110, but nothing prevents the next fixture from reintroducing the shape. Work: teach the classifier to flag a function body containing a silenced `aws <read-verb>` probe when the function's return value is consumed in a condition position, or (cheaper) flag ANY silenced read probe inside a function body and allowlist the legitimate best-effort-cleanup wrappers.

## Related test debt (from the PR #1110 review, fine to bundle here or do opportunistically)

- Not-found-skip branch tests are missing for ~10 providers' probes (only S3's `NoSuchTagSet` class is exercised).
- S3 `AccessDenied` / HTTP 301 skip classes have zero coverage.
[...]

PR #3133, opened 2026-09-14
test(integ): replace every `$(cmd 2>/dev/null | tail -1)` capture with the `capture` helper, and fence the shape

### LINKED ISSUE #3126: test(integ): nine local-* fixtures still capture the CLI with `$(... 2>/dev/null | tail -1)` under pipefail, so a non-zero exit aborts the script with no diagnostic

go-to-k:
## Summary

Under `set -euo pipefail`, `RESULT=$(${CDKD} local invoke Fn 2>/dev/null | tail -1)` aborts the script AT the assignment when the CLI exits non-zero — before the assertion that would print the response, with the CLI's stderr already discarded — so a failing run's log ended at the previous `==> [2/4] ...` banner with no error text. PR go-to-k/cdkd#3118 replaced the shape in one fixture with a `capture` helper and left the siblings; this sweeps the rest and fences the shape.

- **Eight `local-*` fixtures, 35 sites** (measured by the scanner on the pre-sweep tree; one site a three-line env-prefixed call; `local-invoke-layers` had none — its trap-held helper is re-shaped to the same block; the two `grep`-piped captures and the two non-`local-*` sites below are counted separately) go through one byte-identical `capture` helper: exit status taken explicitly; on a non-zero exit it logs the status, the last stdout line and the stderr tail and emits NOTHING, so the assertion runs and fails with its own text — a response that happened to look right never passes a failed invoke, the one property the old shape had; on success it emits the last stdout line. The stderr file is per call and removed inside the helper, so the fixtures' cumulative `trap 'rm -f ...' EXIT` chains needed no entry; `local-invoke-layers` moves from its trap-held file to the same block. An env prefix (`AWS_REGION=x capture ...`) reaches the CLI — bash exports it for the function call.
- **Three `invoke_with_retry` loops** keep stderr in a file and print its tail on every failed attempt and on the final failure, instead of running a fourth invoke to see it.
- **Three sites the fence found beyond the issue's count**: `local-invoke-agentcore-from-state` (two `grep '"env":'` captures — the JSON line is deliberately not the last there, so they keep stderr in a file and fail through `|| { ...; exit 1; }` with its tail), `cc-to-sdk-reroute` (an `ls <literal> | head -1` whose `[ -n ]` FAIL line was unreachable — now `[ -f ]`), `intrinsics-torture` (an `aws ec2 describe-availability-zones` probe silenced then piped — silence dropped; a plain strict capture is the go-to-k/cdkd#1120 form).

## Fence

`scripts/check-integ-capture-shape.ts` + `tests/unit/scripts/integ-verify-capture-shape.test.ts`:

- refuses the shape in any `verify.sh` that sets pipefail (every one of the 257 does, pinned so none can slip out of scope; `set -e -o pipefail` and `-o errexit -o pipefail` count); comment lines and heredoc bodies are data (a here-string is not a heredoc, and an unterminated one is not skipped — both used to blank the rest of the file); one statement is joined out of backslash continuations, lines ending in `|` / `&&` / `||`, and a `$(` still open at a line end; nested substitutions are masked so an inner capture's redirections never read as the outer's; `... | head -1 || true` stays legal (the fallback hands the caller an explicit empty value — go-to-k/cdkd#1120's class, judged there);
- exports `CANONICAL_CAPTURE_BLOCK`; every `capture` copy must equal it byte-for-byte, the defining set is pinned by name (9), and a `capture` CALL is recognised only in command position (the word in an `echo` banner is prose — two fixtures say it);
- real-code probes re-introduce the shape into `local-invoke/verify.sh` (single-line and wrapped without a backslash) and the retry-loop form into `local-invoke-from-state/verify.sh` and require the flag at that line;
- a bash run against a failing stub proves the CONVENTION, not just the scanner: the banned shape dies with no diagnostic and never reaches the assertion; `capture` reaches it with `[verify] command exited 7` + last stdout line + stderr tail while emitting nothing (a `return 1` behind a good-looking JSON line must NOT pass — pinned), and leaves no temp file — checked through a `mktemp` shim on `PATH`, because macOS's `mktemp` ignores `TMPDIR` and the plain assertion was vacuous locally (a CONTROL case with the helper's `rm -f` removed goes red); an env prefix reaches the command.

## Docs

`.claude/rules/abort-capture.md` is a new satellite (`paths: tests/integration/*/verify.sh`) because `testing.md` sat 38 B under its 52,000 B path cap; `testing.md` keeps a one-clause pointer inside the go-to-k/cdkd#1120 sentence this change made stale (the payload fence's own remedy — no sibling entry trimmed). `docs/integ-fixture-conventions.md` gains the user-facing bullet, `layout-scripts.md` the checker's entry, and the run-integ skill's stderr note now says the shape is fenced. `rule-file-payload.test.ts`: corpus count 55 → 56, a reach floor for the satellite, a budget row for the `verify.sh` path family.

## Live test

All 15 touched fixtures run green (ledger rows in the commit; 9 Docker-only, 6 against real AWS; 0 orphans, Docker sweep empty); the nine Docker fixtures re-run green on the review-round helper. `local-invoke-dotnet` / `local-invoke-java` hit the pre-existing emulation fault on one run each (no `HOST_ARCHITECTURE`, go-to-k/cdkd#2287 — `fatal error: wirep: already in go` / `qemu: uncaught target signal 11`), which is the exact diagnostic the old shape swallowed; both passed on re-run, and direct invokes with stderr attached passed twice, so the fault is not a `capture` artefact. One `local-invoke-from-cfn-stack-multi-stack` run was killed by the host mid-deploy (memory pressure); its two CloudFormation stacks were deleted by hand, the re-run passed.

## Review

Three axes on 371ad6af (code / test / spec), no blocker: the code axis found `capture` passing a failed invoke behind a good-looking line, the wrapped-`$(` blind spot and the macOS `mktemp` vacuity; the test axis the here-string blanking and two unfenced arms; the spec axis the wrong site count — all closed in the second commit. Code + test on the delta (91d16bf7) found that the new join swallowed a heredoc opened inside an open `$(` and blanked 53 real lines of `dynamodb-globaltable/verify.sh` to the next same-named terminator, plus two arms without a discriminating case — closed in the third commit (the opener is recognised per physical line; a real-code probe on that fixture pins it). A round-3 pass on `codeLines` diffed the blanked-line sets old vs new over all 257 files (no real line lost, verdicts identical) and asked for one stated assumption to be fenced — no heredoc opener line ends in a continuation — done in the fourth commit.

## Won't-do (recorded here)

Rolling `local-invoke-dotnet` / `local-invoke-java` into `HOST_ARCHITECTURE` is go-to-k/cdkd#2287's per-fixture process and needs the host-architecture fence taught the `new lambda.CfnFunction(` spelling those stacks also use — a separate lane, not this PR's subject.

Closes #3126


go-to-k:
Merged — thanks. Three review rounds each turned up a way the fence could go silently inert (a failed invoke passing behind a good-looking line, a wrapped `$(`, a here-string, a heredoc inside an open `$(`), and each landed with the case that reds without it; the macOS `mktemp` finding is the one I would not have caught alone.

PR #3134, opened 2026-09-14
test: replace wall-clock unit-test verdicts with fake-clock arithmetic, a causal streaming fixture and declared sweep bounds across ten test files

### LINKED ISSUE #2444: test(local): rie-client-streaming's incremental-chunk assertion is a wall-clock race and flakes under full-suite load
Dup-check: searched open issues for an existing item naming this root cause (`rie-client-streaming flaky timing chunk`, `streams chunks incrementally`) -- no hits. New root cause.

## What happens

`tests/unit/local/rie-client-streaming.test.ts:198`, in `invokeRieStreaming > streams chunks incrementally -- body Readable emits as RIE writes`:

```ts
expect(lastArrival - firstArrival).toBeGreaterThan(80);
```

The fake RIE writes 3 chunks with `await new Promise(r => setTimeout(r, 50))` between them, so the SPAN between the first and last arrival is nominally ~100 ms and the assertion allows 20 ms of slack. That slack is not enough under a loaded full-suite run.

Observed 2026-09-02 on a `vp test run` of the whole suite (857 files):

```text
AssertionError: expected 52 to be greater than 80
 ❯ tests/unit/local/rie-client-streaming.test.ts:198:40
```

The same file passes 30/30 when run alone, immediately afterwards, on the same tree. Nothing in the diff being tested that day touched `src/local/`.

## Why it flakes in that direction

The measurement is a DIFFERENCE of arrival timestamps, so it shrinks when the reader is starved: if the consumer loop is descheduled while the server is still writing, several `res.write` chunks land in the socket buffer and are read back-to-back, collapsing the span. A loaded machine makes the consumer slower, not the producer -- which is exactly the direction that makes this assertion fail while the behaviour under test (real streaming, not buffer-then-emit) is still correct.

The sibling assertion (`chunkArrivalTimes.length >= 2`) already anticipates coalescing; the span assertion does not.

## Shape of the fix

The property worth pinning is "the body emits BEFORE the response ends", not "the arrivals are >80 ms apart". Options, cheapest first:

1. Assert ordering against a server-side event rather than a duration -- e.g. record when the server wrote its LAST chunk and assert the FIRST arrival precedes it. That is the actual claim (not buffered) and it has no timing slack at all.
2. If a duration must stay, drive the server's writes from a signal the test controls (write chunk N+1 only after the consumer has seen chunk N) so the span is causal rather than wall-clock.
3. Weakest: widen the slack. This only moves the flake rate; do not stop here.

Same file, same test -- no fixture or Docker work.

Session-fit: next (not this session) -- it is in `src/local/`'s test surface, unrelated to the lane that found it (a `src/deployment/secret-redaction.ts` change), and bundling it would mix an unrelated subsystem into a security PR.
Severity: low -- a false red in CI and in local full-suite runs; no shipped behaviour is wrong, and the test passes in isolation, so it costs re-runs and trust rather than correctness.
Effort: small (S) -- edit plus running that one file; it rides verification any lane in `src/local/` already pays.
Estimate: ~30 min -- most of it deciding between the causal-signal and server-event forms rather than writing either.

Verification the next session will run: `vp test run tests/unit/local/rie-client-streaming.test.ts`, plus one full `vp test run` to confirm it survives load. A fresh session can run both; nothing here is bound to this host, and no Docker or AWS is involved (the test uses a local fake server).

Found while running the full suite for PR #2415.

### LINKED ISSUE #2741: test(scripts): source-control-bytes' tracked-file floor has no declared timeout and flakes at 5000ms under load
`tests/unit/scripts/source-control-bytes.test.ts:271` — "actually READ a realistic number of TRACKED files, so a broken sweep cannot pass vacuously" — declares no timeout, so it runs under Vitest's default 5000 ms while reading ~3344 tracked files off disk. Under load it exceeds that and fails the whole suite.

Measured 2026-09-07 on this machine: it FAILED with `Test timed out in 5000ms` during a full `vp test run` executed while a second repository's suite was running concurrently, and PASSED in isolation seconds later (32/32, same tree, same commit). Nothing about the tree changed between the two runs.

The failure mode is the one that reads as someone else's bug: the timeout names the coverage-floor case, so the message points at the sweep's own vacuity guard rather than at scheduling, and the reflex is to go looking at the code under test.

Why this is not merely a flaky test: the repo's own convention in `.claude/rules/testing.md` ("A test that SPAWNS a subprocess must declare its own timeout") exists because the 5 s default is an IN-PROCESS bound that a loaded machine blows through. This case does not spawn, but it is the same class — thousands of synchronous file reads whose wall-clock cost is a function of machine load, not of correctness. A bound's job there is to stop a HANG, not to police latency, so it should be declared generously rather than tuned.

Scope check before fixing: the sibling cases in the same file walk the same population, so whatever bound this one takes probably belongs on them too. `grep -n "it(" tests/unit/scripts/source-control-bytes.test.ts` and decide per case rather than patching only the one that happened to cross.

Pre-existing on `main` — the file arrived with go-to-k/cdkd#2704 and this session only observed the flake while running two suites at once; nothing in go-to-k/cdkd#2738 touches it.

Dup-check: searched open issues by title for `timeout`, `flake`, `source-control-bytes`, `5000ms`, and by body for `source-control-bytes` / `testTimeout` / `Test timed out`. The nearest three are different root causes — go-to-k/cdkd#2444 is a wall-clock race in a streaming assertion, go-to-k/cdkd#2120 is real-AWS read lag, and go-to-k/cdkd#2701 is about that file's tracked-only population rather than its bound.

Session-fit: next (not this session) — the fix belongs to the file's own bounds and rides no file go-to-k/cdkd#2738 leaves open; the evidence is a one-line reproduction rather than something only this session holds.
Severity: low — a red full-suite run on a loaded machine, with a misleading message; no user-facing behavior and no coverage lost.
Effort: small (S) — a declared timeout per affected case; unit tests only, no integ.
Estimate: ~20-30 min — most of it deciding which sibling cases share the bound rather than adding the one.

Next session's verification: `vp test run tests/unit/scripts/source-control-bytes.test.ts` for the green path, plus a full `vp test run` with a second suite running concurrently to reproduce the original condition.

### LINKED ISSUE #2907: test(deployment): the intrinsic-resolver wrapped-budget assertion is a wall-clock RATIO and flakes under full-suite load

### LINKED ISSUE #3038: test(scripts): gen-sdk-attr-coverage's real-repo floor has no declared timeout and flakes at 5000ms under load
## What happened

`tests/unit/scripts/gen-sdk-attr-coverage.test.ts:192` failed once during a full-suite run on 2026-09-12:

```
× classifies a substantial number of SDK-backed types with an Arn/Url attribute
Error: Test timed out in 5000ms.
If this is a long-running test, pass a timeout value as the last argument or configure it globally with "testTimeout".
```

The case loads every fixture under `tests/fixtures/cfn-schemas` and walks `src/provisioning/providers` — real, growing I/O against Vitest's default 5 s in-process bound, with no third argument to `it`.

## It is a flake, not a regression

- The failing run was one of three full suites executing concurrently across three worktrees of this repo on one machine (24 vitest workers).
- The file passes **17/17 in 3.04 s wall-clock, 1.37 s of test time**, run alone immediately afterwards.
- The branch that observed it changes only `tests/unit/cli/local-*.test.ts` and `tests/unit/cli/options.test.ts` — nothing this test reads.

## Why it is worth a declared bound rather than a re-run

`.claude/rules/testing.md` already states the rule this case is missing — a test whose cost scales with repo content needs its own bound, generously sized, because its job is to stop a HANG, not to police latency. The measured headroom is only ~3.6x (1.37 s against 5 s), and it shrinks every time a schema fixture or a provider is added, so the flake rate rises silently with normal repo growth.

It also fails in the direction that misattributes itself: a contributor whose PR happens to run while the machine is loaded gets a red on an SDK-attribute coverage test they never touched.

Two sibling cases in the same file take the same input and should get the same treatment in one pass: check every `it` in `real repo coverage (regression floor)` for a declared bound.

Dup-check: searched go-to-k/cdkd issues (open and closed) for `gen-sdk-attr-coverage`, `test timed out 5000ms`, and `timeout under load flake`. Hits: go-to-k/cdkd#2741 (`source-control-bytes`' tracked-file floor, same defect, different file — still OPEN), go-to-k/cdkd#2907 (the intrinsic-resolver wall-clock RATIO, same family, different mechanism), go-to-k/cdkd#2795 (CLOSED, the websocket bounded-drain ceiling). Same FAMILY as #2741, but it names a different file and a different assertion, and fixing #2741 does nothing for this one. The `gen-sdk-attr-coverage` hits (#1844, #1824, #2833) are about the coverage DATA, not the test's bound. No issue covers this.

Session-fit: next (not this session) -- the observing session was handling three dependency-bump lanes; this file shares none of their scope, and the fix wants a look at all of that describe block's cases rather than a single-line bump made in passing.
Severity: low -- test-only. No shipped behavior is affected; the cost is a red CI run that misattributes itself to whatever PR was in flight.
Effort: small (S) -- one file, add a declared bound per case, one unit run to confirm. No integ.
Estimate: ~20-30 min -- most of it measuring each case in the block so the bound is sized against real content growth rather than guessed.

### LINKED ISSUE #3071: test(resolver): the two timed concurrent-drain cases exceed their bounds by ~1.5 s on the first run in a fresh vitest process
## Summary

Two timing cases in `tests/unit/deployment/intrinsic-resolver-concurrent-drain.test.ts` (the file and both cases came in with go-to-k/cdkd#2797; go-to-k/cdkd#3044 later modified the file) fail intermittently on a Linux/WSL2 host, each time on the FIRST run of the file in a fresh vitest process, with a measured duration around 1.9 s where the case's bound is a few hundred milliseconds. Later runs of the same file in the same tree pass.

## Measured, not inferred

All on bash 5.2 / Node 24.15.0 / WSL2 (Linux 5.15), `mise exec -- vp test run <file>`, 2026-09-13:

| tree | run | result |
| --- | --- | --- |
| branch of go-to-k/cdkd#3070 (one commit over `17b8e9d1`; no change to the resolver or to this file) | full suite, first run after a rebase | `a drain arming while another WAITS gets what is left of the open window`: `expected 1898 to be less than 280` |
| same branch | the file alone, run 1 | the same case, `expected 1898 to be less than 280` |
| same branch | the file alone, runs 2-4 | 54 passed |
| pristine `origin/main` at `17b8e9d1` in a fresh worktree, fresh `pnpm install` | the file alone, run 1 | `nested drains waiting CONCURRENTLY charge the budget once, not once per level`: `and it WAS charged -- this is not a fresh budget: expected 1937 to be less than 480` |
| same worktree | the file alone, runs 2-3 | 54 passed |
| branch of go-to-k/cdkd#3070, later | full suite | the `WAITS` case again, `expected ... to be less than 280`; a separate 5 s vitest timeout in a `tests/unit/scripts/` file failed in the same run while a Codex process shared the machine |
| same branch, idle machine | full suite, three times | 999 files passed each time |

What the numbers say: on a first run in a fresh process, one of the two cases measures about 1.9 s (1898 ms against a 280 ms bound; 1937 ms against a 480 ms bound) and then passes on every later run. What they do not say: the mechanism. Not established whether the first-run cost is module compilation landing inside the measured window, a timer clamp, or something in the drain budget itself; nothing was instrumented, and no claim is made about which case fails when.

## Where the bounds live

- `tests/unit/deployment/intrinsic-resolver-concurrent-drain.test.ts:730`: `expect(total, ...).toBeLessThan(280)`, over `Date.now()` around a `withSharedDrainBudget` call.
- Lines 622-624 of the nested case (which starts at line 553): `toBeLessThan(600 - spent * 0.6)`, a bound already expressed relative to a measured `spent`; the reported 480 is that expression evaluated at the run's `spent` of 200 ms.

Both compare a wall-clock duration against a bound that a cold first run on this host exceeds by roughly 1.5 s.

## Proposed fix

Not decided; the mechanism has to be measured first (instrument the first run: time from process start to the case, time inside the `withSharedDrainBudget` call, and whether the excess sits before the first `await` or inside the waits). Two candidates to evaluate against that measurement, neither asserted to work: a warm-up drain before the timed one, so a one-off first-run cost is paid outside the window; or timing each phase against a control timer measured in the same case rather than against a constant. Whatever replaces the bound must still fail for the two regressions the cases exist to catch -- a second drain granted a FRESH budget, and nested drains charged once per level -- which is the acceptance test below.

## Verification plan

- Reproduce first: a fresh worktree, `pnpm install`, then the file alone; the first run failed on this host in 2 of 2 trees.
- After the change: the file alone in 5 consecutive fresh processes, all green; AND the two regressions re-introduced by hand (reset the budget for the second drain; charge the nested level again) each red under the replaced bound -- a bound that only stops flaking is not the fix.

Session-fit: next (not this session) -- the file belongs to the drain-budget work (go-to-k/cdkd#2797 / go-to-k/cdkd#3044) and the lane that hit it (go-to-k/cdkd#3070) does not touch the resolver.
Severity: low -- a unit-suite flake on a first run; nothing shipped changes.
Effort: small (S) -- one test file, plus the instrumentation run that decides the change.
Estimate: ~1-2 h -- most of it in measuring where the first-run cost lands.
Dup-check: searched open and closed issues for `concurrent-drain`, `drain budget flaky`, `toBeLessThan(280)`, `intrinsic-resolver-concurrent-drain` right before filing -- go-to-k/cdkd#2797 added the cases and go-to-k/cdkd#3044 modified the file; neither mentions a first-run flake.

go-to-k:
## Summary

Five test-only flakes, one branch, no `src/**` change (so no changelog entry — the reasoning is in the commit message and each test's own comment). Every fix replaces a wall-clock verdict with an invariant, or declares the bound a repo-walking case was missing.

Closes #2907
Closes #3071
Closes #2741
Closes #3038
Closes #2444

## go-to-k/cdkd#2907 + go-to-k/cdkd#3071 — `tests/unit/deployment/intrinsic-resolver-concurrent-drain.test.ts`

**Fix.** Every case that reads a duration now runs on vitest's fake clock (`vi.useFakeTimers()`, restored in the file-level `afterEach`). The cap is a `setTimeout`, the fake client's delays are `setTimeout`s and `Date.now()` is what the budget arithmetic reads, so with all three faked the clock moves only when a timer fires and the fake time at which a resolution settles is the timer that released it — for a drain, the cap wait it paid. Each verdict is therefore `toBe(cap)` / `toBe(cap - spent)` / `toBe(3 * cap)` over the fixture, not a ratio of two host durations (`wrapped < bare / 2`) or an absolute ceiling (`< 280`). Two helpers: `runClockUntilSettled` (steps `vi.advanceTimersToNextTimerAsync()` ONE timer at a time until the promise settles, bounded — so the reading is the settlement instant, not the last timer anyone scheduled) and `fakeElapsed`. Eleven timed cases converted (`SLOW but successful` keeps real timers: it asserts a value, not a duration); `settleTurn` is clock-aware.

**Why not a wider bound.** A per-level charge and a shared one differ by exactly one cap (400 vs 200 vs 600 in the nested-concurrent case), so no tolerance covers the measured 1.9 s first-run excess without losing the regression; the fake clock removes the noise instead of absorbing it.

**Measured.** Before: the timed cases took 44–2202 ms each (file 6.17 s wall). After: 2–11 ms each (file 3.57 s, dominated by imports); 54/54 green on 3 consecutive runs at load 24–44. The `SLOW but successful` case keeps real timers (it asserts a value, not a duration).

**Probes** (each mutation applied to the real `src/` file, anchor count 1, `git diff --stat HEAD` non-empty as the landing receipt, restored from a byte-exact copy with `git diff HEAD -- <file>` empty afterwards):

| probe | mutation | result |
|---|---|---|
| P1 | `resolve` opens a FRESH store (`drainDeadlines.run({...})` instead of `withSharedDrainBudget`) — a budget per iteration | RED 6/54: wrapped LOOP, ordinary-resolution, nested CONCURRENT, stagger, re-entrant, reports-once |
| P2 | charge on EVERY drain exit, not only the outermost (`finally` charges per level) | RED 1/54: nested drains waiting CONCURRENTLY |
| P3 | `wait = shared.remaining` (no open-window subtraction) | RED 1/54: a drain arming while another WAITS |
| P4 | `shared.remaining = budget` (fresh budget per drain, `??=` → `=`) | RED 7/54 incl. both go-to-k/cdkd#2907 assertions and the CONDITION cases |
| P5 | module-level singleton budget in `withSharedDrainBudget` | RED 22/54 incl. `the cap timer is cleared ... budget is released` (its exact `toBe(40)` vs a leaked 400) |
| P6 | cap timer fires at `wait * 3` | RED 11/54 — every fake-clock case |
| P8 (reviewer's) | `clearTimeout(capTimer)` deleted in the `finally` | RED 1/54: `the cap timer is cleared on the way out` (`vi.getTimerCount()` reads 1 where 0 is asserted). Under the first cut's run-all helper this mutant red `two resolutions sharing one context` instead, on a reading of 140 for a resolution released at 90 — the finding that produced the stepping helper |

## go-to-k/cdkd#2444 — `tests/unit/local/rie-client-streaming.test.ts`

**Fix** (issue option 2, the cheapest one that still discriminates): the fake RIE writes chunk N+1 only after the consumer has acked chunk N (a promise per chunk, opened by the consumer when it reads the `cN|` marker), and ends only after the last ack. The verdict is the ORDER `write:0, acked:0, write:1, acked:1, write:2, acked:2, end`. A buffer-then-emit client can never ack while the response is open, so each server wait is bounded (3 s, a hang stop, not a latency police) and records `starved:N`, and the order assertion fails by name. The `chunks.length >= 2` sibling is kept; the case declares `30_000`. Acks are matched over the joined body (review round 1 nit), so a marker split across two reads is acked when its second half lands instead of starving the server.

**Measured.** 3 ms (was ~150 ms of real sleeps); 30/30 green.

**Probe P7** — buffer-then-emit mutant of `invokeRieStreaming`'s drain loop (collect every chunk, push all at `done`; compiling — a first attempt referenced an undeclared identifier and red 7 cases at 1–3 ms, which is a load error, not discrimination, so it was voided and redone): RED exactly the streaming case at 6014 ms with received order `write:0, acked:0, write:1, starved:1, write:2, starved:2, end` (chunk 0 rides in with the prelude's own network chunk via `bodyTail`, which is why the fixture writes three), 29/30 green. Restored, `git diff HEAD` empty.

## go-to-k/cdkd#2741 — `tests/unit/scripts/source-control-bytes.test.ts`

All **nine** cases that call `sweepTree()` (~3300 tracked files read each; the tree-wide verdict, the tracked floor, the four per-extension floors, and the three untracked-probe cases) declare `SWEEP_TIMEOUT_MS = 60_000`. Measured 3 runs at load 25–45 (2026-09-15): max per case 1053 ms (run 1: 1045 / 942 / 1053 / 950 / 1041 / 605 / 578 / 602 / 573; runs 2–3: 573–624 ms), so the bound is >= 57x. The constant's comment records the regime it is sized for (one suite on a loaded CI runner) and that N-way local contention is a different regime (the 778 s observation on go-to-k/cdkd#2741).

## go-to-k/cdkd#3038 — `tests/unit/scripts/gen-sdk-attr-coverage.test.ts`

The ONE case in `real repo coverage (regression floor)` that loads every schema fixture and parses every provider declares `REAL_REPO_TIMEOUT_MS = 60_000` (measured 624 / 722 / 621 ms, 3 runs; >= 83x). The issue's "two sibling cases take the same input" did not survive re-reading the file: the other three cases read only the `SDK_ATTR_ALLOW_LIST` constant (0 ms measured) and keep the default, stated in the constant's comment.

## Class sweep (`tests/unit/**`)

**Wall-clock verdicts.** Scanner: every `expect(<clock-derived expr>).toBe{Less,Greater}Than[OrEqual](...)` where the expression is `Date.now()`/`performance.now()` or an identifier assigned from one (multi-line `expect` handled). Control: run over `origin/main`'s copy of the drain file it reports 23 sites; over the branch, 2 sites tree-wide (`websocket-server.test.ts:890`, a LOWER bound on a real drain the file itself justifies; `custom-resource-provider-response-body-log.test.ts:204`, fake clock). Because that scanner cannot see a value derived through an array (the go-to-k/cdkd#2444 shape), a second pass listed every numeric-bound assertion in the 27 files that read a clock and I read them by hand. **Reported by the lane; taken by the parent round in this PR (`883cabec`):**
- `tests/unit/deployment/dag-executor.test.ts` — the no-level-barrier case compared two `Date.now()` readings with a ~90 ms margin; now `vi.useFakeTimers()` + `advanceTimersByTimeAsync` and the verdict is the ORDER of the recorded events (`X:start` after `A:end`, before `B:end` / `C:end`, all 8 events present). Probe: gate the dispatcher on `active === 0` (a level barrier) → RED at that case only (1 failed / 67 passed across the two files), restored byte-exact.
- `tests/unit/state/lock-manager.test.ts` — the two TTL cases allowed ±1000 ms between the test's clock and `acquireLock`'s; now `vi.setSystemTime` and an exact `expiresAt`. Probe: default TTL 30 → 31 → RED at `should use default TTL of 30 minutes`, restored byte-exact.
- Reviewed and fine: `efs-access-point-idempotency-token.test.ts:352` (`setSystemTime`), `dag-executor.test.ts:190-191` (a logical tick, not a clock), `elapsed-budget.test.ts:144`, `dry-run.test.ts:654`.

**Undeclared bounds on repo-walking cases.** Population: 23 files matching `loadAllFixtures(|git ls-files|readdirSync(<repo dir>)|globSync(`; all 21 not already covered by this PR were run with `--reporter=verbose` (1246 cases, 212 s wall at load 19–37) and every case >= 500 ms was checked for a declared bound. `gen-nested-key-coverage.test.ts` (file-level 60 s) and `diagnose-schema-refresh.test.ts` (per-case) are covered. **Reported by the lane; taken by the parent round in this PR (`883cabec`) — each now declares a 60 s bound (a named constant with the measurement beside it), the five harness spawns take the file's own `slowIt`:**
- `tests/unit/scripts/rule-file-payload.test.ts` — the dead-glob walk over every tracked file (1045 ms) → `REPO_WALK_TIMEOUT_MS`. The 2114 ms merge projection (`git show` per rule file) ALREADY carried a 120 s bound — the lane misread which `it` that closing line belonged to, and the parent's first fold-in repeated the misread onto a pure in-memory case; both corrected on the reviewer's finding.
- `tests/unit/local/docker-argv-redaction-fence.test.ts` — the `src/**` walk (1855 ms) → `SRC_WALK_TIMEOUT_MS`.
- `tests/unit/scripts/check-scope-checker-inputs.test.ts` — all six cases that run the uncached `extractBareTargets()` population walk (slowest 896 ms) → `POPULATION_WALK_TIMEOUT_MS`; bounding only the one that crossed would leave five siblings to reproduce the issue on the next loaded run.
- `tests/unit/scripts/integ-s3-versions-harness.test.ts` — every case spawns bash (single-page ones 503–750 ms), so the file takes one `vi.setConfig({ testTimeout: 60_000 })` (precedent: `pr-title-prefix-scope.test.ts`) under its existing 120 s `slowIt`, per `.claude/rules/testing.md`'s spawn rule.

## Review round 1 (`pr-test-reviewer`, read-only, against the merge base)

No blockers. Two minors and two nits, all applied in this PR: (1) the elapsed helper counted every scheduled timer, not the settlement instant — now steps one timer at a time (P8 above is its receipt); (2) the cleared-timer case's fake-clock `toBe(0)` was unconditionally true — kept as a statement about the clock, with `vi.getTimerCount() === 0` added as the direct leak assertion beside the spy; (3) "twelve" converted cases was eleven; (4) per-read acks in the streaming fixture — now over the joined body. All P1–P6 re-run on the final tree with identical red sets.

## Test plan

- `vp test run` — 1007 files / 22613 tests green, rooted in this tree (twice: before and after the last comment edit).
- Each touched file alone 3x; the seven probes above.

## Parent round

Independent `pr-test-reviewer` (1-reviewer tier: tests-only, but six fences changed what they ACCEPT) against merge-base `52e08724` on `883cabec`: **Clean**. Re-ran six production-side mutation probes itself, each RED at exactly the intended case (TTL 30→31; a level barrier in `DagExecutor`'s dispatcher; `invokeRieStreaming` buffer-then-emit → `starved:1`, `starved:2`; budget charged per level; `openWindow = 0`; per-call cap → 8 cases) with the exact `toBe` values matching the mutation amounts. Fake-timer hygiene confirmed (every `useFakeTimers` paired in `finally` / `afterEach`; `setSystemTime` leaves the S3 mock's promises resolving). Two minors, both applied in `980d574b`: the `rule-file-payload` bound had landed on an in-memory case (above); the same-shape siblings in `check-scope-checker-inputs` and the harness were unbounded (above). Two nits: the `runClockUntilSettled` comment described the macrotask yield in the wrong order (reworded); `acks[Number(n)]!` in the rie fixture would throw on an unexpected marker — left, the fixture writes exactly three. Full suite on the final tree: 1007 files / 22613 tests green.


go-to-k:
Merged — thanks. Every rewritten verdict came with a production-side mutation probe that went RED at exactly one case, and the class sweep turned a five-issue fix into a ten-file one without a second review round.

Issue #3136, opened 2026-09-14
fix(providers): 41 pasteable aws commands in provider messages still interpolate a template- or response-borne id raw, three hand-quoted


go-to-k:
## The class, after go-to-k/cdkd#2669

go-to-k/cdkd#2610 gave the six go-to-k/cdkd#2579-shaped protected-arm remedies ONE rendering for the resource id they paste (`renderDisableCommand` in `src/provisioning/replacement-protection-advice.ts`: `displaySafe(asciiOnly)`, then `shellQuote`, then SUPPRESS the whole command when sanitizing changed the id). go-to-k/cdkd#2669 routed the last of the six, the log-group site, through it. That closes the protected-arm FAMILY. It does not close the CLASS.

The lane that fixed go-to-k/cdkd#2669 swept the class with

```
grep -rnE "aws [a-z0-9-]+ [a-z0-9-]+ [^\`]*\$\{" src/provisioning/providers/ | grep -vE "^\S+:\s*(//|\*)"
```

which returns **42** lines: every provider message that renders an `aws <service> <verb> ...` command with an interpolated value. Exactly ONE of them already routes the value through `shellQuote` (`scheduler-schedule-provider.ts:372`, the go-to-k/cdkd#2610 shape). The rest interpolate raw. Three hand-QUOTE the value with single quotes, the exact go-to-k/cdkd#2669 shape (a `'` in the value breaks out of the quoting; a control byte forges terminal lines and `deployments/*.jsonl` lines):

- `src/provisioning/providers/s3-bucket-provider.ts:6520` and `:6547` — `aws s3api delete-bucket --bucket '${bucketName}'` (template-chosen name)
- `src/provisioning/providers/ssm-parameter-provider.ts:783` — `aws ssm get-parameter --name '${explicit}'` (the `--resource` override / template name)

[...]

Issue #3135, opened 2026-09-14
test(provisioning): the DynamoDB capacity forwarders still read Integer properties through Number(), unmeasured against CloudFormation


go-to-k:
## What is unverified

Issue go-to-k/cdkd#2698 measured CloudFormation's Integer parsing on ONE resource handler (`AWS::Logs::LogGroup.RetentionInDays`, us-east-1, 2026-09-14): the accepted string set is `optional sign + decimal digits`, surrounding whitespace trimmed; hex / exponent / decimal-point spellings are REJECTED by the handler's model validation. cdkd's log-group provider now reads the property through `toCfnInteger` (`src/provisioning/dynamodb-warm-throughput.ts`), which matches that measurement.

The DynamoDB providers still FORWARD numbers read through the wider `toFiniteNumber` (`Number()` filtered to finite values), so a template spelling `ReadCapacityUnits: '0x5'` or `'1e1'` reaches `CreateTable` / `UpdateTable` as a number CloudFormation would very likely have rejected. Callers that forward, enumerated with `grep -rn "toFiniteNumber(" src/provisioning/providers/dynamodb-*.ts` on the go-to-k/cdkd#2669 lane:

- `src/provisioning/dynamodb-warm-throughput.ts` `coerceWarmThroughput` (`WarmThroughput.{Read,Write}UnitsPerSecond`, both DynamoDB types)
- `src/provisioning/providers/dynamodb-globaltable-provider.ts`:
  - L6868 (`mergeExplicitThroughputBlock` numeric members)
  - L7064 / L7084 (`deriveWriteCapacityUnits` / `deriveReadCapacityUnits` literal arm)
  - L7125-L7127 (auto-scaling `MinCapacity` / `SeedCapacity` as the derived capacity)
  - L7412-L7413 (`OnDemandThroughput` max read / write)
  - L8291 (per-index `MaxReadRequestUnits`)

The remaining call sites in both files are GUARDS (already-matches / decrease / zero tests, live-readback compares, diagnostics) and are safe on the wider set: over-acceptance can only skip fewer calls or produce more refusals.

## Why it was not switched on the go-to-k/cdkd#2669 lane

[...]

PR #3137, opened 2026-09-14
fix(providers,export): render the log-group remedy id through the shared sanitizer, read RetentionInDays as a CloudFormation Integer, refuse the rejected falsy family, and keep a redaction mask out of the export identifier

### LINKED ISSUE #2669: fix(providers): the log-group protected-arm remedy pastes an unsanitized state-borne physical id
`src/provisioning/providers/logs-loggroup-provider.ts` builds the go-to-k/cdkd#2579 protected-arm remedy as `aws logs put-log-group-deletion-protection --log-group-identifier '${physicalId}' --no-deletion-protection-enabled`, inside a backticked span the user is told to paste. `physicalId` is hand-quoted with single quotes and otherwise raw. go-to-k/cdkd#2610's first PR gave the same shape at five other providers the treatment `src/state/lock-contention-message.ts` established for a stack name -- `displaySafe(..., { asciiOnly: true })`, then that module's `shellQuote`, then SUPPRESS the whole command when sanitizing CHANGES the id, because a command naming a sanitized id acts on a DIFFERENT resource -- and centralised it in `src/provisioning/replacement-protection-advice.ts`'s `renderDisableCommand`. This site was left out.

The provenance argument that would excuse it is the one that PR retracted: the id is not an AWS-minted literal. The deploy engine passes `currentResource.physicalId` straight off `state.json`, and this repo's own `lock-manager.ts` treats state-bucket values as attacker-influenced for anyone who can write the bucket. A `'` breaks out of the hand-quoting; a control byte forges lines on the terminal and inside the persisted `deployments/*.jsonl`.

Why it was not fixed in that PR, which is the whole of the deferral reason: the exact command string is asserted in TWO places, and one of them is an integ fixture. `tests/unit/provisioning/logs-loggroup-provider-class-guard.test.ts` pins the literal, and `tests/integration/loggroup-class-guard/verify.sh` REBUILDS the same string to grep cdkd's output. Changing the rendering silently blinds that grep -- the fixture keeps passing while asserting nothing -- and `.claude/rules/testing.md` requires re-running a fixture after any edit to a string it greps. That is a real-AWS run, which is the verification this work owns and the reason it is not a same-PR change.

Evidence: `renderDisableCommand` in `src/provisioning/replacement-protection-advice.ts`; the five migrated call sites; `logs-loggroup-provider.ts`'s hand-built string; `tests/integration/loggroup-class-guard/verify.sh` line ~427.

Dup-check: searched open issues for "pasteable recovery command physical id shell quoting" and every open body matching `shellQuote|pasteable|shell-quote` -- both windows returned nothing; go-to-k/cdkd#2610 covers the messages' PRECONDITIONS, not the rendering of the id inside them
Session-fit: next (not this session) -- the fix must be landed together with a re-run of `loggroup-class-guard`, whose `verify.sh` rebuilds the exact command string to grep for it; a rendering change that is not re-run leaves that fixture green and blind, which is worse than the defect
Severity: medium -- a pasteable command carrying an unsanitized state-borne id, on the path whose whole purpose is to be pasted; bounded by the fact that a real CloudWatch log-group name cannot contain a quote, so it needs a hand-edited or attacker-written state record to fire
Effort: small (S) -- route the one call site through `renderDisableCommand` (build the command separately; no `suffix` parameter is needed) and update the unit literal, then re-run the fixture
Estimate: ~30 min of edit plus one `loggroup-class-guard` real-AWS run, which dominates

### LINKED ISSUE #2698: test(provisioning): cdkd's RetentionInDays coercion is not known to match CloudFormation's Integer parsing
## What is unverified

`toFiniteNumber` (`src/provisioning/dynamodb-warm-throughput.ts`) is the repo's
one answer to "is this stringly-typed CloudFormation value a number?", and it
is `Number()` filtered to finite values. Measured against the shipped helper:

```
"30"      -> 30         "0x1e"     -> 30         "0o36"   -> 30
"0b11110" -> 30         "1e3"      -> 1000       " 30 "   -> 30
"30.5"    -> 30.5       "+30"      -> 30         "\n30\t" -> 30
"1_000"   -> undefined  "Infinity" -> undefined  "30abc"  -> undefined
```

Whether CloudFormation accepts those spellings for an `Integer`-typed property
such as `AWS::Logs::LogGroup.RetentionInDays` is **not known here**. Nobody has
measured it, and the docs do not settle it. `docs/cli-deploy-safety.md`
previously said the value is "COERCED the way CloudFormation coerces it"; that
sentence was a claim about an external system with nothing behind it, and it
has been narrowed to describe what cdkd does, with a pointer here.

## Why it matters in one direction only

In the stateful GUARD the wider accepted set is safe: more strings coerce to a
positive number, so more log groups answer `has-retention`, so MORE
replacements are refused. A guard that over-refuses is the side to err on.

`LogsLogGroupProvider` is the half that is not obviously safe, because it
FORWARDS the coerced number. A template carrying `RetentionInDays: '0x1e'`
reaches `logs:PutRetentionPolicy` as `30`, and `'30.5'` reaches it as `30.5` --
on a template CloudFormation might have rejected outright, and with a value
CloudWatch Logs will reject in the second case (its retention is an integer
from a fixed enum, which cdkd does not validate against either -- a plain `31`
is forwarded today too).

## What settling it requires

A live CloudFormation A/B, not a reading: deploy an `AWS::Logs::LogGroup` with
`RetentionInDays` spelled `'0x1e'`, `'1e3'`, `' 30 '` and `'30.5'` through
CloudFormation and record which it accepts, which it coerces, and to what.
Then decide between:

- narrowing `toFiniteNumber`'s string arm to the spellings CloudFormation
  takes -- but it is SHARED with the DynamoDB capacity readers, so narrowing it
  changes those too and needs its own pass; or
- leaving the shared helper alone and validating at the provider, which is
  where the value is forwarded; or
- accepting the divergence and saying so, which is what the page does today.

Do not narrow anything on a reading of the CloudFormation docs alone. The whole
reason this issue exists is that a parity sentence was written without a
measurement behind it.

Dup-check: searched open issues for "RetentionInDays coercion CloudFormation parity Number" -- nothing comes back; no open issue covers this root cause
Session-fit: next (not this session) -- settling it needs a live CloudFormation
stack, which the lane that found it may not spend, and the guard direction is
already safe. The false parity sentence, which was the shippable half, is
already corrected.
Severity: low -- no known user impact: the guard errs toward refusing, and the
provider's exposure needs a template spelling an integer property in hex,
scientific or decimal-fraction notation. What is undone is that cdkd's accepted
set and CloudFormation's are not known to match.
Effort: small (S) -- one throwaway CloudFormation stack with four variants,
then a decision recorded on this issue; the code change, if any, is a few lines.
Estimate: ~1-2 h -- almost all of it the A/B deploys and their teardown.

### LINKED ISSUE #2699: fix(provisioning): an empty-string or false RetentionInDays silently disables a live log-group retention policy
## The behaviour

`LogsLogGroupProvider.applyUpdate` treats every FALSY `RetentionInDays` as a
request to REMOVE the retention policy: absent, `0`, `''`, `false` and `null`
all fall into the arm that issues `logs:DeleteRetentionPolicy`. For an absent
property that is correct and matches CloudFormation -- an unset retention is
CloudWatch Logs' never-expire. For `''` / `false` / `null` it is a guess, and
the guess disables a live retention policy silently.

This is PRE-EXISTING: the pre-coercion code tested
`if (properties['RetentionInDays'])` and routed every falsy value to the same
`else`. Issue (#2521) left the
falsy family exactly as it found it and says so at the call site; the rows are
pinned in that PR's `UPDATE_MATRIX`, so the behaviour is recorded rather than
accidental.

## Why it is worth revisiting

It is data-PRESERVING -- removing a retention policy makes a log group keep its
events forever -- so it is not a data-loss bug and does not belong to the
stateful guard. What it defeats is a COMPLIANCE retention: an organisation that
sets 30 days deliberately, to bound how long log data is held, gets an
unbounded log group from a template typo, and nothing warns.

Reachable without a hand-written L1: a `Type: String` parameter left at an
empty default, or an `Fn::If` whose false branch yields `''`, both put an empty
string in the resolved bag.

## Options, none obviously right

- **Refuse** `''` / `false` / `null` the way `'abc'` and `true` are already
  refused. Consistent with the round-4 argument that a malformed value is not a
  spelling of "no retention" -- but `null` in particular is how several template
  paths spell "absent", so refusing it would break templates that work today.
- **Warn and skip**, leaving the live retention alone. Preserves the compliance
  setting, but a skip is the silent-drop class this provider's `LogGroupClass`
  guard exists to prevent.
- **Leave it and document it** on `docs/cli-deploy-safety.md`.

Settle which spelling CloudFormation itself treats as "remove the retention"
before choosing -- the same measurement
issue (#2698) needs, and worth
doing in one pass with it.

Dup-check: searched open issues for "DeleteRetentionPolicy falsy retention disable" -- nothing comes back; no open issue covers this root cause
Session-fit: next (not this session) -- pre-existing, unchanged by the PR that
found it, and choosing between the three options needs the same live
CloudFormation measurement as (#2698). Taking it inside a lane scoped to the
stateful guard would have widened that PR's behaviour change on an unmeasured
premise.
Severity: low -- data-preserving in every case; what breaks is a deliberate
retention bound, silently, and only for a template that spells the property
falsy-but-present.
Effort: small (S) -- the code change is one condition; the cost is the decision
and the CloudFormation A/B behind it.
Estimate: ~1-2 h -- shareable with the measurement in (#2698), which is most of it.

### LINKED ISSUE #2932: fix(export): the mask blocker tests properties only, so a Cloud-Control-imported record's masked attributes pass it
## The residue

`cdkd export` blocks a resource whose state holds the redaction mask, at
`src/cli/commands/export.ts:3894`:

```ts
if (carriesSecretMask(stateEntry.properties)) {
```

It tests `properties` and NOT `attributes`. That was complete when the only
producer of a mask was the mask-only channel (#2274), which writes into
`properties`. It is no longer complete: since go-to-k/cdkd#2924 (issue
go-to-k/cdkd#2847), `CloudControlProvider.import` MASKS every Cloud Control
model key it cannot certify is a schema-declared `readOnlyProperties` attribute,
and those masks land in `attributes`. A Cloud-Control-imported record therefore
routinely carries masks in a bag this blocker does not look at.

## Why it is not currently a disclosure or a wrong write

The export path reads `attributes` at `src/cli/commands/export.ts:3984`, through
`resolveResourceIdentifier`. Both of its resolvers SHAPE-VALIDATE what they read
-- `recordedArnIdentifier` requires an `arn:` prefix with a service segment, and
`recordedRuleIdIdentifier` requires an `sgr-` prefix -- so the literal `***` is
refused rather than emitted as a CloudFormation import identifier.

That is the whole guarantee, and it is accidental. It holds because two
resolvers happen to test a prefix, not because anything states that a masked
`attributes` value must never reach the exported template. A third resolver
added for a type whose identifier has no distinguishing prefix would emit `***`
as the resource's import identifier, which CloudFormation either refuses at
import time or writes onto the live resource at the next update -- exactly the
outcome the `properties` blocker exists to prevent.

## Why go-to-k/cdkd#2924 did not just widen the test

Widening it to `carriesSecretMask(stateEntry.attributes)` is one line, and it is
NOT a safe drive-by: after that PR every Cloud-Control-imported record has
masked attributes whenever `cloudformation:DescribeType` was unavailable (the
fail-closed arm masks the whole model), and many have them even with the grant,
since a writable model key is masked by design. So the widening would newly
BLOCK a population that exports fine today. Whether those records SHOULD be
blocked is a `cdkd export` decision -- it may want to block only when the mask
is at a position the export actually reads -- and it needs its own review and
its own `export` integ run.

## Remedy sketch

Decide between:

1. Widen the blocker to `attributes` and accept that CC-imported records become
   unexportable until re-imported with the grant. Simplest, most conservative,
   biggest behaviour change.
2. Block only when a mask sits at a position `resolveResourceIdentifier` would
   read for that resource type. Narrow, and keeps today's exports working, but
   couples the blocker to the resolver's key selection.
3. Keep the shape fences as the guarantee and make them explicit: state at each
   resolver that it must refuse `SECRET_MASK`, and fence that with a test, so a
   third resolver cannot be added without meeting it.

Whichever is chosen, `tests/integration/export/` is the integ to re-run.

## Classification

Session-fit: next (not this session) -- the fix is a `cdkd export` behaviour
decision, not a defect in go-to-k/cdkd#2924. Option 1 newly blocks a population
that exports today, so it needs the `export` integ and its own review round;
folding it into a PR whose subject is `cdkd import` would ship an unreviewed
behaviour change to a different command. Next session starts with
`gh issue view <this issue>` then `/run-integ export` after the chosen option
lands.
Severity: medium -- no disclosure and no wrong write today; the mask is refused
by two prefix tests in `resolveResourceIdentifier`. What is broken is the
GUARANTEE: the blocker no longer covers a bag that now routinely holds masks, so
the protection rests on an accident that the next resolver added to that
function can remove silently.
Effort: small (S) -- a one-line predicate change plus its rationale for option
1, or a per-resolver fence for option 3. It drags the unit cycle and one
real-AWS `export` integ.
Estimate: ~2-3 h -- most of it deciding between the three options against the
real record population, and the `export` integ run.
Dup-check: searched open and closed issues on 2026-09-11 for `export blocker
attributes carriesSecretMask` and `export mask blocked record`, and grepped
`src/cli/commands/export.ts` for `carriesSecretMask`. (#2274) is the mask-only
channel that introduced the `properties` blocker (closed); (#2881) is `cdkd
drift --revert`'s refusal naming only one of the two mask causes, a different
command and a different function; (#2925) is the Cloud Control create/update
path rewriting the whole model into `attributes`. None names this predicate. No
duplicate.

go-to-k:
Closes #2669
Closes #2698
Closes #2699
Closes #2932

Four measured fixes in one lane (`/work-issues`), each with a unit test that fails without it (mutation-probed, one mutation per probe, tree restored byte-exact from a snapshot between probes).

## Issue 2669 — the log-group remedy renders its id through the shared sanitizer

**Fix.** `logs-loggroup-provider.ts`'s `LogGroupClass` protected-arm remedy hand-quoted the state-borne `physicalId` (`--log-group-identifier '${physicalId}'`). It now renders the command through `replacement-protection-advice.ts`'s `renderDisableCommand` (exported for this one caller): `displaySafe(asciiOnly)` -> `shellQuote` -> the WHOLE command suppressed when sanitizing changed the id, with `UNNAMEABLE_ID_CLAUSE` in its place. The log-group SENTENCE is unchanged (the header's "decision about the TEXT" stands). A clean id renders BARE (`shellQuote` leaves `[A-Za-z0-9._/@:+-]` unquoted) — `tests/integration/loggroup-class-guard/verify.sh`'s needle is now that bare form, so the fixture discriminates against the old hand-quoting. The same-file create-cleanup `aws logs delete-log-group` command was the same shape and took the same rendering.

**Tests.** `logs-loggroup-provider-class-guard.test.ts`: the pinned literal is now bare, plus a `describe` with a `'`-carrying id (shell-quoted), a control-byte id (suppressed, clause present), a non-ASCII id (suppressed), and the fixture's clean name (bare). `logs-loggroup-provider-partial-create-cleanup.test.ts`: whole rendered command pinned + the two hostile names.

**Probes.** Revert the remedy to hand-quoting -> 5 red (the literal + all four hostile-id cases). Revert the cleanup command -> 2 red.

**Sweep.** `grep -rnE "aws [a-z0-9-]+ [a-z0-9-]+ [^\`]*\$\{" src/provisioning/providers/ | grep -vE "^\S+:\s*(//|\*)"` -> 42 lines; 1 already via `shellQuote`, 2 fixed here (both in the log-group file), the remaining 39 (three hand-quoted: `s3-bucket-provider.ts:6520` / `:6547`, `ssm-parameter-provider.ts:783`) are outside this lane's file set and filed as (#3136) with the per-site provenance question.

## Issue 2698 — `RetentionInDays` is read as a CloudFormation Integer

**Fix.** New `toCfnInteger` beside `toFiniteNumber` in `dynamodb-warm-throughput.ts`, carrying the live A/B table from the issue in its doc comment. It is `config-shape.ts`'s `coerceCfnInteger` (measured digits grammar) plus a trim — the ONE measured difference between the two properties (`" 12 "` rejected on a nested `PasswordLength`, `" 30 "` accepted on `RetentionInDays`), so the two helpers cannot drift on the digits and each carries its own padding measurement. Both log-group forwarding sites switched. `'0x1e'` / `'1e3'` / `'30.5'` / `'30.0'` / a non-integer number are now REFUSED by name; `'+30'` and `' 30 '` deploy as 30 and compare equal to a recorded 30.

**`toFiniteNumber` caller audit** (`grep -rn "toFiniteNumber(" src/`):

| caller | role | verdict |
| --- | --- | --- |
| `logs-loggroup-provider.ts` create + update retention read | FORWARDS to `PutRetentionPolicy` | switched to `toCfnInteger` |
| `stateful-types.ts:619` `has-retention` | GUARD | keep (over-acceptance = more refusals) |
| `dynamodb-warm-throughput.ts` `coerceWarmThroughput` | FORWARDS (`WarmThroughput` Long fields) | keep — unmeasured on DynamoDB, (#3135) |
| `dynamodb-warm-throughput.ts` `isWarmThroughputDecrease` | GUARD | keep |
| `dynamodb-table-provider.ts:247-250, :530-531, :3896-3897` | GUARD (already-matches / zero tests) | keep |
| `dynamodb-globaltable-provider.ts:6868, :7064, :7084, :7125-7127, :7412-7413, :8291` | FORWARD (capacity / on-demand / warm blocks) | keep — unmeasured on DynamoDB, (#3135) |
| `dynamodb-globaltable-provider.ts:6913, :7322, :7993/7997, :8113-8127, :8253, :8370/8374` | GUARD / live-readback compare / diagnostics | keep |

The DynamoDB forwarders were NOT switched on a Logs-only measurement (#2698's own rule); (#3135) carries the A/B they need and the site list.

**Docs.** `docs/cli-deploy-safety.md`'s `Number()` parity paragraph now states what cdkd does and cites the measurement (issue links removed per the docs-site fence).

**Tests / probes.** `dynamodb-warm-throughput.test.ts` gains `toCfnInteger` rows (every rejected spelling also asserted accepted by `toFiniteNumber`, so the row discriminates; the "coerceCfnInteger plus a trim" relation pinned in both directions). Probe: `toCfnInteger` -> `toFiniteNumber` body -> 12 red; the update-site wiring reverted to a `Number()` reader -> 8 red.

## Issue 2699 — the falsy family follows the measurement

**Fix.** `isAbsentRetention` names CloudFormation's measured "no retention" spellings: absent, `''`, whitespace-only (whitespace-only was REFUSED before and now takes the delete arm, matching CloudFormation). `false`, `null`, `0`, `'0'` are refused by `assertUsableRetention` like `'abc'` (five arms, each with its own `detail`). `create()` / `update()` now take their contexts (the update-context fence's `OMITS_UPDATE_CONTEXT` ratchet moved accordingly; `assertRegionMatch` on the update path stays the job of (#2613)).

**The `null` decision.** Refused. `grep -n "NoValue" src/deployment/intrinsic-function-resolver.ts` -> lines 3333 (array elements filtered), 3485-3499 (`Property ${key} resolved to AWS::NoValue, omitting from object`), 8427: cdkd's resolver DROPS the key for `AWS::NoValue` (bare or as an `Fn::If` branch) and never manufactures `null`; CloudFormation rejects `null` at `update-stack`. So a `null` reaching the provider is a template literal, not cdkd's spelling of absent.

**The `0` decision, per caller.** A template `0` is refused (CloudFormation: `Invalid retention value`). But `readCurrentState` writes `0` for a never-expiring log group, that lands in `observedProperties`, and `cdkd drift --revert` hands it back as the DESIRED bag with `desiredFromAwsReadback: true` — reverting a console-added retention is desired `0` over live `30` and must DELETE. `isAbsentRetention` therefore admits numeric `0` on that flag alone (its documented contract: every value in the bag reports what AWS held). `'0'` is never cdkd-produced and stays refused everywhere. Accepted residual, recorded at the update site in the second review round: the rollback executor's revert arms pass no flag (and `UpdateContext` has no `replayingState`), so a `properties` bag carrying `0` / `'0'` / `false` / `null` (a pre-fix template cdkd accepted and recorded, or a `drift --accept` on a record with no `observedProperties`) is now REFUSED on a rollback where it used to take the delete arm — a knowing regression on that arm, since the update path cannot tell the replay from a template deploy and admitting `0` on every bag is the silent-removal defect being closed. `'abc'` was already refused on that arm; the zero family joins it.

**Create-path replay.** Under `CreateContext.replayingState` every refusal downgrades to warn-and-SKIP (never forward): a record can carry a value the pre-fix create accepted, and re-creating from a spelling CloudFormation rejects is the wrong restoration.

**Tests / probes.** `CREATE_MATRIX` (25 rows) / `UPDATE_MATRIX` (61 rows, `readback` column) re-pinned to the measurement with literal tallies and named witnesses per arm; `logs-loggroup-provider-roundtrip.test.ts`'s `30 -> 0` case now drives the readback flag it models. Probes: readback exemption removed -> 2 red; `0` absent on every bag -> 5 red; `null` arm removed -> 2 red; whitespace-only back to refuse -> 5 red; `false` back to delete -> 5 red; unknown-previous whitespace exclusion removed -> 2 red; create replay downgrade removed -> 1 red.

## Issue 2932 — a redaction mask never reaches the export identifier

**Decision (deviates from the brief's option 1, with the reason).** Widening the `properties` blocker to the whole `attributes` bag would make every Cloud-Control-imported record PERMANENTLY unexportable: `CloudControlProvider.import` masks each writable model key BY DESIGN (whole model without `cloudformation:DescribeType`), so a re-import does not clear it, and that population exports fine today. The export reads `attributes` at exactly one position — the identifier attribute of a `COMPOSITE_PHYSICAL_ID_IDENTIFIERS` type — so the guarantee is stated where the identifier is built: after `resolveResourceIdentifier`, a `resourceIdentifier` or overlay carrying `***` blocks the resource with a mask-specific reason, whatever produced it (masked attribute, masked physical id, splitter, backfill, a future resolver with no shape test). `resolveIdentifierValue` additionally replaces the state-only shape error with the mask message when `attributes.<field>` is masked — the old remedy ("re-deploy so cdkd records it") sends a masked record in a circle. `docs/cli-export.md` gains the row and the paragraph.

**Tests / probes** (`export-composite-identifier.test.ts`): masked `TableARN` + composite physicalId -> blocked, reason names `attributes.TableARN` and the DescribeType remedy, NOT the re-deploy one; control with a real ARN -> exports; POPULATION CONTROL: `AWS::S3::Bucket` with masked `Arn` / `DomainName` / `WebsiteURL` and clean properties -> exports; masked attribute with a bare-ARN physicalId -> exports (nothing masked ships); masked physicalId -> blocked at the choke point. Probes: choke point removed -> 1 red; mask message removed -> 1 red; option 1 applied (whole-bag test) -> 3 red, the population control among them.

## Sweeps (§5-b)

- `grep -rn "carriesSecretMask(" src/` -> 14 sites; the export one is the only one reading a single bag of a state record for an outbound artifact. `rollback-executor.ts:1230` and `nested-stack-provider.ts:684` test `properties` alone by design (the fail-closed mask never lands there — `drift.ts` header); no change.
- `grep -rn "renderDisableCommand(\|protectedReplacementAdvice(" src/` -> 7 files, the six protected-arm sites now all routed.

## Live verification (parent ran both on `18843a89`; the src tree is byte-identical to this HEAD, which adds only the ledger rows)

- `/run-integ loggroup-class-guard` — PASS, 97 s, all 7 phases: phase 5 grepped the re-rendered bare-form command and phase 6 EXECUTED it against real AWS (`aws logs put-log-group-deletion-protection --log-group-identifier /cdkd-integ/loggroup-class-guard/class --no-deletion-protection-enabled`), then the replace succeeded; destroy 1 deleted / 0 errors; the fixture's fixed log-group name is gone and the state prefix holds only `deployments/`.
- `/run-integ export` (broad set) — PASS, 253 s: deploy → `export --include-non-importable` → CFn stack `UPDATE_COMPLETE` with every composite identifier adopted (the #2932 choke point sits after those resolvers, on the path) → cdkd state gone → post-export `cdk diff` proposes no replacement → CFn stack deleted; no IAM / Lambda / VPC / bucket left.
- Ledger rows committed; `integ-destroy` + `integ-broad` set from this worktree.

## Review round 2 (applied)

`pr-code-reviewer` and `pr-test-reviewer` returned no blockers. Applied: `renderDisableCommand` is exported GENERIC over the constrained `ProtectedReplacementDisableCommand` (so `CdkdAuthoredLiteral` binds the direct entry too; `replacement-remedy-cli-facts.test.ts` gains a `@ts-expect-error` row) and `ResolvedDisableCommand` stays unexported; the update-site residual above is now a comment at the site; the `#`-in-a-name comment corrected (`shellQuote` quotes it, correctly); the cleanup-warn comment no longer claims the prose interpolation is sanitized; the replay-downgrade test asserts the WARNING (a silent skip had passed — probe 15 reds it now); the overlay disjunct of the export choke point is marked defence-only with the reason; the `previousRetentionUnknown` witness message names the `false` row. Noted, not changed: the masked-`TableARN` + composite-physicalId fixture has no cdkd writer today (CC import records the bare ARN as the physicalId, which is the exporting case also pinned); it exercises the message arm as defence. The delta round (code reviewer, final) found one mis-citation in a new comment (the prose half of the cleanup warning is held by no issue; (#3136) covers commands only) — corrected, comment-only.

## Parent round (independent 3-axis + security, merge-base `f3148d88`, on `baf27169`)

- `pr-spec-reviewer` (spec = the four issue bodies + the CloudFormation A/B comment): Clean. Both deliberate deviations judged consistent with the issues' own mechanism — the #2932 identifier-level choke point is the issue's option 2 strengthened to the VALUE (a whole-bag test would newly block every Cloud-Control-imported record), and the readback-only `0` admission is licensed by `UpdateContext`'s contract. One wording nit (a JSON `30.5` is refused with the "type number" detail) — left.
- `pr-code-reviewer`: Clean. Every caller of the widened `create`/`update` signatures still type-checks (the only direct construction is `register-providers.ts`); the replay downgrade is double-gated on `replayingState` + `ProvisioningError` and only the rollback executor sets the flag; `desiredFromAwsReadback` has exactly one producer (`drift.ts`); the `previousRetentionUnknown` arm was hand-traced for whitespace/`''`/`'0x1e'` previous values (no spurious update, no missed delete); the generic `renderDisableCommand` keeps `CdkdAuthoredLiteral` on every existing caller; `carriesSecretMask` is whole-leaf and no writer embeds `***` inside a composite; the `.ts` import extension has precedent (`stateful-types.ts`) and `rewriteRelativeImportExtensions`. Three nits, all left (a noisy replay warn for a `drift --accept`-written `0`; `trim()` also strips Unicode whitespace beyond the ASCII measurement; the create-path refusal runs after `CreateLogGroup`, pre-existing since #2521).
- `pr-test-reviewer`: Clean. Ten production-side mutation probes re-run independently, each RED at the intended cases only — including the #2932 POPULATION CONTROL going RED under the whole-bag alternative, and the readback exemption widened to any context reddening the template-bag refusals. `verify.sh` phase 5's bare needle measured against both renderings (OLD hand-quoted: no match; NEW: match).
- `pr-security-reviewer`: Clean on exposure / re-application for all three traced values (state-borne id → pasteable command, mask → export identifier, replay downgrade). Executed the copied sanitizer chain: `'` → POSIX-escaped; newline / NUL / ESC / NEL / U+202E / non-ASCII / leading space / empty → whole command SUPPRESSED. Two minors: (1) **applied in `18843a89`** — the create-path replay used to SKIP a retention the record certifies the old binary applied (`'30.0'` / `'0x1e'` read by `Number()` and forwarded as 30), so a rollback re-creating the old log group silently dropped a compliance retention; under `replayingState` a positive-integer legacy reading is now RESTORED with a warning naming the spelling, every other refused member stays warn-and-skip (case + negative control; probe dropping the restore: RED at that case only). (2) **filed as go-to-k/cdkd#3141** — the update-path twin (a rollback revert of a record spelling the retention falsy is refused where it took the delete arm) needs `replayingState` through `UpdateContext` and the executor's revert arms, held by open PR go-to-k/cdkd#3104; the in-code residual points at the issue. One pre-existing nit (state-borne id in two default-verbosity warn PROSE lines, unsanitized) — the prose class, outside (#3136)'s command scope, left.

## Residuals filed

- (#3135) — DynamoDB capacity forwarders still on `toFiniteNumber`, needs its own CloudFormation A/B.
- (#3136) — the other 39 pasteable `aws ...` commands in provider messages (three hand-quoted).
- (#3141) — the rollback update arm's refusal of a falsy `RetentionInDays` record (above).


go-to-k:
Merged — thanks. Four issues on one file pair, the CloudFormation A/B settled two of them before a line was written, and both real-AWS fixtures exercised the changed path (phase 6 pasted the sanitized command back into AWS). The rollback-arm residual is #3141.

PR #3138, opened 2026-09-14
test(integ): pin the capture-shape scanner's pipe-join arm and the continuation-line heredoc opener

go-to-k:
## Summary

Follow-up to go-to-k/cdkd#3133 (issue go-to-k/cdkd#3126). The cdk-local port's review (go-to-k/cdk-local#734) found two arms of the scanner's `codeLines` with no discriminating case: the `|` / `||` join (only `&&` was pinned, and a wrapped pipe is already joined by the open-`$(` arm, so removing `\|` from `continues` stayed green) and the `skipHeredoc` call INSIDE the join loop (the shipped first-line-opener case never reached it, so deleting that call stayed green). Both cases added; both mutations measured red on a scratch copy. The opener-continuation assumption fence now reuses the exported `HEREDOC_OPENER` instead of a copy of the regex, so the two cannot drift.

Tests only (`scripts/check-integ-capture-shape.ts` gains one `export` keyword); no fixture, no `src/`. Tree-wide verdicts unchanged (1009 files green).


go-to-k:
Merged — thanks. The two arms came out of the cdk-local port's review reading the same scanner with fresh eyes; both cases red without their arm.

PR #3139, opened 2026-09-14
fix(ec2): read a security group's VpcId back from AWS instead of copying the template, and resolve it live on an absent record

### LINKED ISSUE #3097: fix(ec2): a SecurityGroup declared without VpcId records the attribute as '' and the resolver arm answers undefined instead of reading the group's VPC

go-to-k:
## Summary

`EC2Provider`'s two `AWS::EC2::SecurityGroup` result maps recorded `VpcId: (properties['VpcId'] as string) ?? ''` -- the template's own property -- so a group declared without `VpcId` (a hand-written L1 that lands in the account's default VPC) recorded `''`, and the resolver arm behind it answered `undefined` (`// Would need API call`), which `Fn::Join` / `Fn::Sub` render as the literal `'undefined'`. CloudFormation answers the default VPC's id. These were the two sites issue go-to-k/cdkd#3077 allow-listed.

Both halves now read the group back:

- **Provider** (`src/provisioning/providers/ec2-provider.ts`): `readSecurityGroupVpcId` issues `DescribeSecurityGroups({ GroupIds: [groupId] })` after create (after the wiring settled; never on the wiring-failure cleanup path) and after update, recording the response's `VpcId` through `definedAttributes`. A not-found answer (EC2's read-after-write lag) is retried up to `SG_READBACK_NOT_FOUND_ATTEMPTS` (5 attempts, 4 retries) `SG_READBACK_RETRY_DELAY_MS` (1000) apart via the `securityGroupReadbackDelays.sleep` seam (fallback: the provider's own `sleep`); every other failure, an empty answer, and a rejecting sleep omit the key at once (never `''`), naming the error class + HTTP status at the head of a debug line (AWS text behind it, same line, debug only). `import()` records the same two members off the describe it already issues; `getAttribute` shares the helper. There was NO existing describe on the create / update path -- `CreateSecurityGroup` returns only the group id -- so this is one new call per create / update.
- **Legacy `''` records** (`src/deployment/intrinsic-function-resolver.ts`, `isImpossibleEmptyStoredAttribute`): a record written by a pre-#3097 binary for a group declared without `VpcId` holds `VpcId: ''`, and a no-change deploy never rewrites it (the provider records only on `update()`), so the flat lookup would serve `''` forever. For exactly `AWS::EC2::SecurityGroup` / `VpcId` a stored `''` is read as ABSENT (no security group has no VPC; deliberately NOT a general rule -- go-to-k/cdkd#3077's known-empty EC2 `PublicIp` keeps being served from state, pinned by a control case). The record itself keeps `''` until the next update; only the resolution changes (said in the changelog and `docs/cli-deploy-safety.md`). **`cdkd diff` behaviour change on such a record**: `outputs-diff` now issues one `DescribeSecurityGroups` and shows a one-time `'' -> vpc-...` Output delta (so `cdkd diff --fail` exits 1 once; the next deploy's Outputs persist heals it); on a refusal (no `ec2:DescribeSecurityGroups`) the key lands in `failedKeys` and the Outputs section is suppressed with a warning. `cdkd orphan --force`'s cache fallback (`src/analyzer/orphan-rewriter.ts`) reads the same predicate, so a legacy `''` is never spliced in as a VpcId there either.
- **Resolver** (`src/deployment/intrinsic-function-resolver.ts`): the `AWS::EC2::SecurityGroup` `VpcId` arm guards the persisted physical id with `^sg-[0-9a-f]+$` (marked non-retryable, no read; the id keys the cache), reads `DescribeSecurityGroups({ GroupIds })` through `clientsForRegion(this.explicitRegion)`, caches the value per deploy in a null-prototype `cachedSecurityGroupVpcIds` cleared by `resetAccountInfoCache`, and refuses through `refuseUnservedAttribute` (unmarked, retryable) on an empty or failed read. `GroupId` is unchanged.
- **Fences**: `tests/unit/provisioning/attribute-map.test.ts` -- allow-list now EMPTY (both entries retired; the equality-consumption probe keeps a synthetic two-entry list, and asserts every copy is unexpected against the real list). `tests/unit/scripts/resolver-mask-coverage.test.ts` -- census 135 / 160 / 108 -> 136 / 161 / 109 (measured: one throw statement, one masked expression, one `not-in-class` note; the two `refuseUnservedAttribute` calls share the existing throw site). `tests/unit/deployment/intrinsic-refusal-non-retryable.test.ts` -- fourth null-prototype cache pinned; `unmarked` stays exactly 2.
- **handled-property-wiring**: `EC2Provider#VpcId` lost the update-path seeds (`applyUpdate` / `update` / `updateSecurityGroup`) -- accepted deliberately with `--accept-evidence-loss`. That read was the attribute copy, not delivery: `VpcId` is create-only on a security group, and the create seams (`createSecurityGroup`'s `CreateSecurityGroupCommand`) still read it, so the property stays `wired`.
- **Docs**: `.claude/rules/layout-provisioning.md` (attribute-map entry no longer names two allow-listed sites), `docs/provider-development.md` (the fence carries no allow-list), `docs/cli-deploy-safety.md` (the live-read arm list gains the security group). `changelog.d/entries/2026-09-15-3097-sg-vpcid-live-read.md`.
- **Fixture** (`tests/integration/ec2-instance`): a `CfnSecurityGroup` with NO `vpcId` (explicit `GroupName: CdkdEc2InstanceIntegDefaultVpcSg`, the name `verify.sh` sweeps by) plus a `DefaultVpcSgVpcId` Output. `verify.sh` reads the default VPC live (`describe-vpcs --filters Name=is-default`; FAILS up front if the region has none), then `assert_default_vpc_sg` runs after BOTH the `--no-wait` deploy and the default deploy: the record's `attributes.VpcId` equals the default VPC (FAIL on `<absent>` / `''` / `undefined`), the Output equals it (FAIL on `''` / `null` / `undefined` / an `sg-` id), the negative control `.outputs.VpcId` still equals the stack's own VPC which is NOT the default one, the record is SDK-routed, and a live describe agrees. After destroy: the group is gone (`assert_gone`), the default VPC is still describable.

## Mutation probes (unit; each restored byte-exact, `git diff` empty after)

| # | mutation | red at |
|---|---|---|
| P1 | create map back to `(properties['VpcId'] as string) ?? ''` | 5 create cases in `ec2-sg-vpcid-readback` + 2 attribute-map fence cases |
| P2 | update map back to the same | 2 update cases + 2 fence cases |
| P3 | `readSecurityGroupVpcId` rethrows instead of catching | both "OMITS the key when the describe fails" cases |
| P4 | `import()` back to `attributes: {}` | the import case |
| P5 | resolver cache write dropped | "caches it per deploy" (2 sends, expected 1) |
| P6 | `sg-<hex>` guard dropped | all 6 shape cases |
| P7 | resolver arm back to `return undefined` | 11 cases incl. the `Fn::Join` `'undefined'` case |
| P8 | `resetAccountInfoCache` SG clear dropped | reset case + both "nothing cached" refusal cases |
| P9 | cache declared `{}` instead of null-prototype | the non-retryable census case |
| R1 | retry branch disabled | lag-then-success + gives-up cases |
| R2 | retry on EVERY error (not-found predicate dropped) | the denial-not-retried case |
| R3 | attempt bound moved by one | gives-up case (sleeps 5 vs 4) |
| M14 | resolver cache keyed by `logicalId` | the physical-id key case (second stack served the first's VPC) |
| H3 (reviewer) | cache keyed `${logicalId}#${physicalId}` | the same case, third assertion |
| H4 | production `sleep` resolves at once (seam deleted) | the fake-timers real-delay case |
| H6 | empty answer retried like a not-found | "OMITS the key when the describe answers no group / no VpcId" (describes 1, sleeps []) |
| S1 | rejecting-sleep guard removed | the rejecting-seam case |
| L1 | `isImpossibleEmptyStoredAttribute` returns false (carve-out dropped) | the legacy-`''` case (served `''`, no describe) |
| L2 | carve-out widened to every type (`storedValue === ''`) | 3 red: the new EC2 `PublicIp` control (`intrinsic-functions.test.ts:3811`), the pre-existing EC2 "serves a stored empty string WITHOUT re-describing" control (`:3049`), and the Route 53 empty-`NameServers` normalization (`:2522`) |
| C1 | resolver writes the live id back into the record after the cache write | the legacy-`''` case (same context instance for both resolves and the assertion) |
| E1 | raw-AWS-message debug line promoted to `error` | the same 4 failure cases as W1 (the wiring-failed case never reaches that line) |
| O1 | `orphan-rewriter.ts` fallback bypasses `isImpossibleEmptyStoredAttribute` | the new `--force` legacy-`''` case |
| A1 | `allowListVerdict` answers `unconsumed: []` | the synthetic-list probe (third entry `never-present` must be unconsumed) |
| W1 | raw-AWS-message debug line promoted to `warn` | 4 failure cases (`expectNothingAboveDebug`) |

## Review rounds (all four reviewers per round, `fable`)

- **Round 1** (35af8ad4): spec clean; security clean (2 nits); code 1 minor (no not-found retry after create -> Phase 0 assertion could red on EC2 lag) + 2 nits; test 1 minor (cache KEY undiscriminated -- a logical-id-keyed mutant survived). Both minors fixed.
- **Round 2** (delta 684cb9d3): spec no drift (1 nit: docstring overstated "only not-found" / omitted `getAttribute`); security clean (1 nit: a rejecting sleep seam escaped the best-effort boundary); code 1 minor (docstring's update-path premise -- rule/tag calls do not always precede the read) + 2 nits (duplicate sleep indirection; "5 retries" vs 5 attempts); test 2 minors (`defaultSleep` unfenced; empty-answer no-retry unpinned) + 1 nit. All fixed in the final commit.
- **Round 4** (parent's delta round on 94c222d0): two BLOCKING test findings -- the legacy-`''` case inspected a fresh context literal (fixed: one instance, C1 red), and `expectNothingAboveDebug` did not spy `error` (fixed, E1 red); plus the `cdkd diff` sentence above, the orphan-rewriter fallback (O1), and the L2 recount (3 red, file:lines named).
- **Round 3** (parent's independent round on b78749c2, zero blockers): legacy `''` shadowing (code minor + test medium) -> `isImpossibleEmptyStoredAttribute`; attribute-map `unconsumed` direction unfenced (test medium) -> third synthetic entry; stale `outputs-diff.ts` / `layout-analyzer.md` example list (nit) -> dropped; warn/info negatives on the failure cases (security nit) -> `expectNothingAboveDebug`; cache-key "unique across regions" comment (nit) -> reworded to the region-pinned read. All in this single commit.
- **Declined (won't-do, recorded here)**: (e) round-3 -- the orphan rewriter's live `getAttribute` fetch pays the same bounded (<= 4 s) not-found wait on a group that is gone: accepted, the same helper with one caller and a fixed bound; (a) round-1 security nit -- the provider's failed-describe line stays at DEBUG rather than a WARN with the class: it matches the VPC `DefaultSecurityGroup` sibling, and the resolver's refusal (class + status) reaches default verbosity when the attribute is actually referenced; (b) round-1 security nit -- state-borne `groupId` / AWS text in that debug line do not pass `displaySafe`: the provider-wide existing class for debug lines, not widened here; (c) round-1 code nit -- an update whose describe fails DROPS a previously recorded `VpcId`: by design under the go-to-k/cdkd#3077 rule (absent -> the resolver live-reads; inheriting the previous map would also inherit a pre-fix `''`); (d) round-2 test nit -- the literals 5 / 1000 are asserted through the exported constants, not pinned: they are tunables, the changelog states the shipped values.

## Maintainer-side review

4 reviewers (spec / code / test / security, Fable) at b78749c2 — no blockers; delta round (code + test) at 94c222d0 — two blocking test gaps (a vacuous "record untouched" assertion through a context factory; the above-debug negatives missed `error`) fixed at 619cfe6f with receipts, plus the `cdkd diff` behaviour on a legacy record stated in the docs and the `orphan-rewriter --force` fallback routed through the same predicate.

## Test plan

- [x] `vp run check`, `vp run typecheck:test`, `vp run build`, `vp test run` (1012 files, 22865 passed, 1 skipped) on the final rebased tree (619cfe6f, on top of go-to-k/cdkd#3145, 0.290.4)
- [x] `vp run gen:all-matrices && vp run format && vp run audit:coverage:check` clean
- [x] `bash -n tests/integration/ec2-instance/verify.sh`; fixture synthesized (`DefaultVpcSg` with only `GroupDescription` + `GroupName`, Output `Fn::GetAtt [DefaultVpcSg, VpcId]`)
- [x] `/run-integ ec2-instance` — PASS, 165 s (2026-09-15, us-east-1): `DefaultVpcSg.attributes.VpcId == vpc-08cae0035717deb9d` and `.outputs.DefaultVpcSgVpcId == vpc-08cae0035717deb9d` on the Phase 0 (`--no-wait`) record AND after the Phase 1 heal; the stack's own `VpcId` Output ≠ the default VPC (negative control); the default-VPC group is gone after destroy and the default VPC untouched; destroy 0 errors, 0 orphans. The RESOLVER half is not reachable by this fixture (the provider records the key, so the live arm never runs) — unit-covered (P5-P8, L1/L2, O1).
- [x] `/run-integ lambda` (broad set) — PASS, 101 s, 9 deleted 0 errors 0 orphans → `integ-broad` set for this lane. Account-wide sweep after both runs: 0 `state.json`, no fixture VPCs / functions / the default-VPC group.

Closes #3097


go-to-k:
Merged — thanks. Reading `VpcId` back from the group instead of the template, plus the legacy-`''` carve-out at the flat lookup, closes the last of the #3077 family: the fence's allow-list is empty and every stored attribute is now a value AWS reported.

PR #3142, opened 2026-09-14
test(integ): pin local-invoke-dotnet and local-invoke-java to the host architecture, and teach the fence the CfnFunction spelling

go-to-k:
## Summary

Two more fixtures join `integ-fixture-host-architecture.test.ts` under go-to-k/cdkd#2287 (5 covered; 11 Lambda-bearing `local-*` fixtures remain, measured): `local-invoke-dotnet` and `local-invoke-java`. Both faulted under emulation on one run each during go-to-k/cdkd#3133's runs on this arm64 host (`fatal error: wirep: already in go` / `qemu: uncaught target signal 11`) — the first time the cause was readable, since that PR's `capture` helper prints the stderr. Both assets are portable bytecode (framework-dependent `dotnet publish`, `javac`) and the base images are multi-arch, so the pin is mechanical: `HOST_ARCHITECTURE` derived from `process.arch`, `architecture: HOST_ARCHITECTURE` on the `lambda.Function`.

Each stack also carries a `lambda.CfnFunction` for its inline-rejection arm, which the fence refused as an unknown spelling. It now accepts two constructors, each with the property it must carry — `lambda.Function` with `architecture: HOST_ARCHITECTURE`, and the L1 `lambda.CfnFunction` with `architectures: [HOST_ARCHITECTURE.name]` (a hardcoded `['arm64']` fails the same way a hardcoded enum does). Probed on real copies: the `architectures` line removed, and hardcoded, both red naming `InlineHandler`. `FIXTURE_STACKS` count 3 → 5; `.claude/rules/testing.md` and the local-invoke stack comment carry the new counts (`testing.md` stays under its 52,000 B path cap by tightening the sentence this change made stale).

## Live test

The synthesized template carries `Architectures: ['arm64']` on both functions here. 3/3 consecutive native runs green per fixture, 20–35 s each (java 24–95 s with 1 fault in 3, dotnet 29–102 s with 2 faults in 4 under emulation today); Docker sweep empty; `integ-local` marker set; ledger rows updated.

## Review

One code reviewer on 0fa79fe2 (the fence gains a pattern future fixtures copy): all six requested mutations red on real copies, `Architecture.ARM_64.name === 'arm64'` and the `string[]` prop confirmed against the installed `aws-cdk-lib`, the inline-rejection path traced to `resolveRuntimeFileExtension` before any image pull (the property does not change routing). Two minors closed in the second commit: the "13 remain" figure was a subtraction from an already-stale 15 (the fence's own comment said 16; measured 11), and the java ledger note had dotnet's timings copied in. Nits (stale two-fixture prose, a dead default argument, a single-quote-only id extractor) taken.

## Scope

Tests and agent-instruction prose only. `provided.*` and `DockerImageFunction` fixtures stay out per the issue's note (their architecture is the binary's / the image's).


go-to-k:
Merged — thanks. The fault had been on the go-to-k/cdkd#2287 list since June; it took the `capture` diagnostic printing the stderr to make the roll-in an afternoon rather than a diagnosis.

Issue #3147, opened 2026-09-15
fix(providers): the AWS::DynamoDB::Table capacity forwarders read ProvisionedThroughput through a bare Number(), the sibling of the toFiniteNumber population outside its grep


go-to-k:
## The defect

Issue go-to-k/cdkd#3135 switched every `toFiniteNumber`-reading DynamoDB forwarder (`coerceWarmThroughput`, the `AWS::DynamoDB::GlobalTable` capacity / ceiling readers) to `coerceCfnInteger`, CloudFormation's measured DynamoDB Integer grammar (`/^[+-]?\d+$/`, no trim). The sweep that produced that issue's list was `grep -n toFiniteNumber src/provisioning/providers/dynamodb-*.ts`, and the `AWS::DynamoDB::Table` provider's capacity forwarders never appeared on it because they do not go through that helper at all -- they read through a bare `Number()`, which is WIDER still:

- `src/provisioning/providers/dynamodb-table-provider.ts` `create()`: `ReadCapacityUnits: Number(pt?.['ReadCapacityUnits'] ?? 5)` / `WriteCapacityUnits: Number(pt?.['WriteCapacityUnits'] ?? 5)` on `CreateTable`.
- the same spelling on the `update()` billing-flip arm (`updateInput.ProvisionedThroughput = { ReadCapacityUnits: Number(pt['ReadCapacityUnits'] ?? 5), ... }`).
- `readCapacityNumber` (module-level), which reads a per-GSI `ProvisionedThroughput` member for the flip's `GlobalSecondaryIndexUpdates[].Update.ProvisionedThroughput` -- `Number(raw)` filtered to finite, `''` / `null` read as absent.
- `hasUsableTableCapacity`, the pre-flip look-ahead that MIRRORS the flip's `Number()` arithmetic by design (its doc says so), and `hasUsableDeclaredCapacity`, which mirrors `readCapacityNumber`.

[...]

PR #3148, opened 2026-09-15
fix(analyzer,providers): the drift canonicalizers keep an own __proto__ key and a non-plain readback, and the DynamoDB capacity forwarders read the measured CloudFormation Integer grammar

### LINKED ISSUE #3121: fix(drift-normalize): the comparison-side canonicalizers rebuild onto {} literals -- an own __proto__ key is dropped on both sides and a Date readback is flattened to {} before the comparator

### LINKED ISSUE #3135: test(provisioning): the DynamoDB capacity forwarders still read Integer properties through Number(), unmeasured against CloudFormation

go-to-k:
Closes #3121
Closes #3135

## Issue 3121 -- the drift canonicalizers keep an own `__proto__` key and return a non-plain readback by identity

**Fix.** The six comparison-side canonicalizers -- `canonicalizeTagListsDeep` / `canonicalizeIdArraysDeep` / `walkUnordered` in `src/analyzer/drift-normalize.ts`, `rewritePrincipalUniqueIds` / `rewritePrincipalValue` in `src/analyzer/drift-principal-normalize.ts`, `stripWalk` in `src/analyzer/cc-api-strip.ts` -- rebuilt every object node onto a `{}` literal on BOTH comparison copies. Each object arm now (a) returns a non-plain object (`Date`, `Uint8Array`) BY IDENTITY and (b) rebuilds a plain one onto a null-prototype record, the pattern go-to-k/cdkd#3124 applied inside `drift.ts`. The four helpers `drift.ts` spelled file-locally (`hasOwnKey` / `ownValue` / `defineOwnKey` / `hasPlainPrototype`) moved to a new LEAF module `src/utils/own-keys.ts` (plus `nullPrototypeRecord`), and `drift.ts` imports them -- its only edits are the import, the removed local definitions, and the `getAtPath` doc that cited this issue as the reason a top-level `__proto__` could not reach the command. `secret-redaction.ts` (held by another lane) keeps its own `hasPlainPrototype` copy; the module doc records why (it is `vi.mock`ed by several drift suites).

**Net effect.** A top-level or nested `__proto__` drift is detected, reported (`--json` path `__proto__.polluted`) and accepted like any other key; a `Date` readback reaches `calculateResourceDrift` as the Date, so it compares as itself and -- on a resource with NO recorded secret -- `--accept` persists its ISO string rather than `{}`. **Scope (review M1):** on a resource that records a secret, `--accept`'s redaction VALUE walk (`secret-redaction.ts`, held by another lane) still flattens the Date to `{}` one layer later; that is open issue go-to-k/cdkd#2427, cited from the changelog entry, the module doc, `layout-drift.md` and the test comments rather than claimed closed here.

**Sweep (command + counts, before -> after).**
- `grep -rn "Record<string, unknown> = {}" src/analyzer/` : 9 -> 3. The three left are `orphan-rewriter.ts:527,759` and `diff-calculator.ts:792` -- not on the drift comparison path (the issue names them as the resolver / replay-walk class, go-to-k/cdkd#2776).
- `key in bag` membership tests in the three files (code lines, `for (const` excluded): 0 -> 0.

**Tests.** `tests/unit/analyzer/drift-normalize.test.ts` (one own-`__proto__` case per function, one Date-identity case over all three, and both through `calculateResourceDrift` with a declared unordered path so every site is on the chain), `drift-principal-normalize.test.ts` (one case per rewrite walk, Date identity at both arms), `cc-api-strip.test.ts` (own key kept while stripping beside it; Date identity), `tests/unit/cli/drift-secret-redaction.test.ts` (two command-level cases: a top-level `__proto__` drift reported through `--json`, and a Date readback under `--accept`). Assertions read JSON text / `Object.getOwnPropertyNames`, never an object-literal `{__proto__: ...}` expectation.

**Mutation probes** (`scratchpad/laneE-private/probe.sh`, line-addressed, restored byte-exact, receipts read per probe): 14 probes over the six sites -- at each, (1) the rebuild target reverted to a `{}` literal and (2) the `hasPlainPrototype` guard deleted -- plus the two command-level cases against the first site on the chain. 13 RED on the first run; `rewritePrincipalValue`'s guard was GREEN because every Date in the fixture sat beside a plain `Principal` object, so a case with the Date AS the Principal value was added (commit `e7914104`) and that probe went RED. Final: 14/14 RED.

## Issue 3135 -- the DynamoDB capacity forwarders read the measured CloudFormation Integer grammar

**Measurement** (the issue's table, us-east-1, 2026-09-14, `AWS::DynamoDB::Table.ProvisionedThroughput`): accepted grammar `/^[+-]?\d+$/` with NO trim -- `" 7 "`, `"7 "`, `""` rejected at properties validation (before the handler), `"0x9"` / `"1e1"` / `"6.0"` / `"6.5"` rejected by the handler, `"+8"` / `"010"` accepted. `config-shape.ts`'s `coerceCfnInteger` is exactly that grammar; `toCfnInteger` (the Logs helper, trim measured on that handler) is unchanged, and the DynamoDB table now sits beside the Logs one on its doc comment.

**Per-site table** (`toFiniteNumber(` call sites: warm-throughput 4 -> 3, globaltable 20 -> 4, table 8 -> 8; `coerceCfnInteger(` now 2 + 16):

| site | role | reader now | what a rejected spelling becomes |
|---|---|---|---|
| `dynamodb-warm-throughput.ts` `coerceWarmThroughput` | FORWARD (both types) | `coerceCfnInteger` | dropped and NAMED in `droppedMembers` (the unresolved-intrinsic arm); never absent, never `Number()`'s reading |
| `dynamodb-warm-throughput.ts` `isWarmThroughputDecrease` | GUARD (coerced spec vs live) | `toFiniteNumber` | unchanged |
| globaltable `mergeExplicitThroughputBlock` (L6899) | FORWARD | `coerceCfnInteger` | "present but did not resolve" diagnostic; derived value stands in (go-to-k/cdkd#1428 warn-and-fall-back) |
| globaltable `reportUnresolvedRawMember` (L6949) | DIAGNOSTIC predicate paired with the on-demand forwarders | `coerceCfnInteger` | reported (on the wider reader `" 7 "` was "usable", the forwarder dropped it, nothing said so) |
| globaltable `deriveWrite/ReadCapacityUnits` literal arm (L7102 / L7122) | FORWARD | `coerceCfnInteger` | falls through to the auto-scaling block, then the announced 5/5 fallback (go-to-k/cdkd#1511) |
| globaltable `pickAutoScalingCapacity` (L7163-7165) | FORWARD | `coerceCfnInteger` | same |
| globaltable `explicitCovers` (L7364) | MIRROR of the merge | `coerceCfnInteger` | the go-to-k/cdkd#1511 fallback diagnostic fires instead of being suppressed |
| globaltable per-GSI on-demand ceilings (L7454-7455) | FORWARD | `coerceCfnInteger` | suppressed and reported |
| globaltable `collectUncomparableCapacityGsiNames` explicit block (L8038 / L8042) | MIRROR of the translation (live-baseline guard) | `coerceCfnInteger` | index counted NOT template-determined -- otherwise the 5/5 fallback read as a scale-down against a live baseline |
| globaltable `buildLiveRecoveryGsiBaseline` live compares (L8158-8172) | GUARD (SDK numbers) | `toFiniteNumber` | unchanged |
| globaltable replica override suppression test (L8298) | MIRROR of the merge | `coerceCfnInteger` | suppression diagnostic fires |
| globaltable replica on-demand override (L8336) | FORWARD | `coerceCfnInteger` | suppressed and reported |
| globaltable `collectTableOnDemandCeilings` `RawCeiling.value` (L8415 / L8419) | FORWARD (table-level `OnDemandThroughput` on create / update) | `coerceCfnInteger` | `declared: true, value: undefined` -- no `-1` reset, nothing forwarded, reported |
| table-provider `liveCapacityAlreadyMatches`, `warmThroughputAlreadyMatches`, `skipZeroCapacityIndexUpdate` (L247-250, L530-531, L3897-3898) | GUARD | `toFiniteNumber` | unchanged |
| globaltable `applyAutoScalingDiff` `MinCapacity` / `MaxCapacity` -> `RegisterScalableTarget` (L3817-3818; review M2, was a bare `Number()`) | FORWARD | `coerceCfnInteger` | the block's existing warn-and-skip arm, message names the grammar |
| globaltable `applyAutoScalingDiff` `ScaleInCooldown` / `ScaleOutCooldown` -> `PutScalingPolicy` (L3911; was `Number()`) | FORWARD | `coerceCfnInteger` | announced and OMITTED (service default) -- never `Number()`'s reading or `NaN` |
| globaltable `applyAutoScalingDiff` `TargetValue` (L3878) | FORWARD of a CFn **Double** | `Number()` | unchanged -- the Double grammar is unmeasured |

REFUSE vs skip: no site throws. The GlobalTable translators are pure collectors whose warn-and-fall-back is a thrice-relitigated decision (go-to-k/cdkd#1428: they also run on `previousProperties`, and the replica path runs after `CreateTable` committed), so a rejected spelling takes the arm each site already had for an unresolved intrinsic -- announced fallback on a PROVISIONED capacity (AWS requires one), announced suppression on an on-demand ceiling, dropped-and-named on `WarmThroughput`. Every diagnostic on these paths now ends with the grammar (`CFN_INTEGER_SPELLING_NOTE`), so a user whose value IS a number learns why it was still dropped; the Table provider's warm-throughput hint says the same.
Precedent divergence, stated: unlike the `RetentionInDays` refusal (go-to-k/cdkd#2699 / go-to-k/cdkd#3137), which is create-path-only and template-borne, every DynamoDB site takes its existing unresolved-intrinsic arm because the translators are pure collectors that also run on `previousProperties` and the replica path runs post-`CreateTable` (go-to-k/cdkd#1428).

**Sibling filed, not widened:** the `AWS::DynamoDB::Table` provider's `ProvisionedThroughput` forwarders read through a bare `Number(x ?? 5)` with a default arithmetic `hasUsableTableCapacity` mirrors by design -- the same root cause outside the issue's `grep toFiniteNumber` population, needing a per-site REFUSE-vs-skip decision and its own `dynamodb-table` integ: go-to-k/cdkd#3147 (`Session-fit: now`).

**Tests.** `tests/unit/provisioning/dynamodb-warm-throughput.test.ts` (the four rejected spellings plus `"7 "` / `"6.0"` / a fractional number dropped and named, with the wider reader's acceptance asserted as the premise; `"+8"` / `"010"` accepted), new `tests/unit/provisioning/dynamodb-globaltable-provider-cfn-integer.test.ts` (52 cases: the derivation arms, the merge, both mirror predicates, both on-demand ceilings, the replica overrides, `collectTableOnDemandCeilings`, `derivePerCallProvisionedThroughput`, each asserting the wire value AND the diagnostic), `dynamodb-table-provider-warm-throughput-decrease.test.ts` (a padded member through the Table provider's own `update()` send site).

**Review round 1** (both reviewers' findings applied in commits `c39eb960` and the one after it): test review found THREE forwarder halves unprobed and inert under revert -- the GSI on-demand READ ceiling (both its spellings), `collectTableOnDemandCeilings`' read `value`, and the write arm of `collectUncomparableCapacityGsiNames` -- each now has a case per rejected spelling and its own RED probe; code review M2 found the auto-scaling registration reading `MinCapacity` / `MaxCapacity` through a bare `Number()` while `pickAutoScalingCapacity` read the same member through `coerceCfnInteger` (the paired-reader divergence this PR closes elsewhere) -- folded in (same file, same root cause, an existing announced skip arm; stated here per the widening tripwire: +25 src lines, no design decision), with the cooldown Integers alongside; two cases in the index-autoscaling suite, 7/7 round-1 probes RED (the `MaxCapacity` read was inert until the case spelled it hex too).

**Mutation probes:** 13 probes, one per forwarder / predicate reverted to `toFiniteNumber` (`coerceWarmThroughput`, the merge, `reportUnresolvedRawMember`, both literal arms, both auto-scaling arms, `explicitCovers`, the GSI ceiling, the uncomparable-names mirror, the override suppression predicate, the on-demand override, `RawCeiling.value`). 12 RED on the first run; the override suppression predicate (L8298) was GREEN because no case carried an explicit override beside an unresolvable CFn block -- added (commit `49157442`), RED. Final: 13/13 RED. Total 27/27, re-run on the final tree after round 1 (line numbers relocated from anchors): 27/27 RED.

## Parent round (3-axis + security on the flattened `e8c8fa1f`), applied in one amend

- spec: Clean.
- code: 3 minors. (1) `deepEqual` / `diffAt` compared the persisted ISO string against run 2's `Date` by `typeof` and reported drift forever -- `drift-calculator.ts` now folds a non-plain `toJSON` value (a `Date`) to its JSON form before comparing and descending (pushed values stay the originals); run 2 = clean pinned in `drift-secret-redaction.test.ts` (`#3121: run 2 after --accept ...`), probe: fold removed -> that case RED. (2) On an explicit-only shape the merge diagnostic said only "ignored" while 5 went to AWS -- `mergeExplicitThroughputBlock` now names the stand-in (`it was ignored and 5 was sent instead`); the cfn-integer cases assert it; probe: announcement removed -> 4 cases RED. (3) `RecoveryPeriodInDays` through bare `Number()` -> added to go-to-k/cdkd#3147's site list (comment), not fixed here.
- test: 2 gaps applied -- the accepting direction at the auto-scaling register (`'+5'` / `'010'` -> `{MinCapacity: 5, MaxCapacity: 10}`; probe: refuse every string -> RED) and `reportUnresolvedRawMember`'s accepting arm (the `'+41'` / `'010'` control now passes `diagnostics` and asserts `[]`; probe: accept JSON numbers only -> RED) plus a `'+8'` `ProvisionedThroughputOverride` control (probe: refuse strings at the literal arm -> RED). Nits: the "template it accepts" comment corrected; `tests/unit/utils/own-keys.test.ts` fences the module as import-free (probe: an added import -> RED) and pins the helpers' own contract.
- security: Clean. **Won't-do** (recorded here): a pre-flight `configIntegerRefusal` on the PROVISIONED billing-flip arm (`update()` ~L1955-1963) -- the flip's translators run on `previousProperties` and on the replica path after `CreateTable`, so a refusal there strands a replay (the divergence line above); with item (2) landed the substituted value is ANNOUNCED on every arm, which is the property a refusal would have bought.

## Other

- `.claude/rules` payload fence: `layout-utils.md` sat 118 B under the `aws-client-defaults.ts` path cap, so the own-key rule got its own satellite `.claude/rules/own-keys.md` (four literal paths) with a one-line pointer; `CORPUS_FILE_COUNT` 56 -> 57, a `REACH_FLOORS` row (EXACT 4) and a budget row (`src/utils/own-keys.ts`, measured 49,437 B) in `tests/unit/scripts/rule-file-payload.test.ts`.
- `.claude/rules/layout-provisioning.md` / `layout-drift.md` updated; two `changelog.d/entries/2026-09-15-{3121,3135}-*.md`.
- Gate liveness probed by shape (`commit-msg-heredoc-gate` blocked the `$(cat <<'EOF')` form) before the first commit.

## Delta round (code + test on `0378bfe8`, applied in `a394770c`)

- Code: NO new runtime defect from the fix round. One minor — `jsonForm`'s docstring claimed a non-plain object WITHOUT `toJSON` "compares equal to nothing but itself"; the comparator key-walks it like a plain object (a keyless `Map` equals `{}` — pre-existing), and a bare prototype guard would break the typed-array convergence the fold exists for, so the docstring now states the mechanism. Two nits left (the fold runs in both `diffAt` and `deepEqual`, idempotent; a `derived?.[member]` double read).
- Test: both earlier gaps CLOSED (probes RED at the new register-accepting case and the `diagnostics: []` control). One new medium gap, applied: a `Date` INSIDE AN ARRAY reaches only `deepEqual`'s fold (arrays are never descended by `diffAt`), and deleting that fold left the whole suite green — a run-2 case with a fresh `Date` instance in an array element now pins it (probe: RED at that case only, 1/104). The leaf fence catches `^\s*import\b` only; a `getBuiltinModule` escape reaches builtins alone — left.

## Live verification (parent ran both on `a394770c`; this HEAD adds only the ledger rows)

- `/run-integ drift-revert-arrays` — PASS, 198 s, all arms (clean deploy drift-free, repeated drift stable, benign tag reorder not a false positive, real drift detected, `--revert`, properties-fallback baseline), 19 created / 19 deleted, 0 errors, orphan scan clean.
- `/run-integ lambda` (broad set) — PASS, 104 s, `RecursiveLoop` + `ReservedConcurrentExecutions` backfills verified end-to-end, 9 deleted / 0 errors, no IAM / Lambda / DynamoDB / SQS left; the fixture's `AWS::DynamoDB::Table` + `GlobalTable` capacity reads went through `coerceCfnInteger`.
- Ledger rows committed; `integ-destroy` + `integ-broad` set from this worktree.


go-to-k:
Merged — thanks. The parent round's Date-phantom-drift finding was the one that mattered, and the fix round landed it with the array-element twin pinned; both fixtures exercised the changed comparison and capacity paths.

Issue #3149, opened 2026-09-15
fix(state): validate the nested record shapes both parsers hand to the executor (previousState, attemptedProperties, provisionedBy, oldResourceRetained, and state.json's per-record fields)


go-to-k:
## Summary

go-to-k/cdkd#3140 makes `parseRollbackJournal` refuse an operation whose `logicalId` / `resourceType` / `changeType` / `physicalId` is not the string the executor keys its lookups on. The NESTED records inside an operation are still unchecked casts, and so is every per-record field of `state.json` in `parseStateBody` (`src/state/s3-state-backend.ts:1372`). Both files sit in the same bucket, writable by anyone with `s3:PutObject`. Found by the security review of the go-to-k/cdkd#3140 branch (reviewer trace, file:line as of `2d9f5de1`):

- `op.previousState` as a string / number / null: the classifier does not throw (`rollback-executor.ts:766-767` uses `?.`, the `!` sites are reached only past a `physicalId !== undefined` guard), but the `revert` arm hands `provider.update(..., undefined)` when `.properties` is missing (`:2913-2938`), and `properties: "abc"` passes a STRING bag to the provider; `stateResources[op.logicalId] = prev` (`:2257`) and `recordAfterRollbackUpdate(previousState, ...)` (`:3050`) then PERSIST the unvalidated record into `state.json`.
- `failedOperations[].attemptedProperties`: `provider.delete(..., op.attemptedProperties)` (`:3345-3349`), `resolveReplayProps(op.attemptedProperties ?? current.properties)` (`:3409`) -- same class.
- `op.provisionedBy`: a ROUTE key. `getProviderFor` tests only `=== 'cc-api'`, and `...(op.provisionedBy && { provisionedBy })` at `:2286` / `:2888` / `:3089` / `:3474` writes ANY truthy value into state / events; `effectiveProvisionedBy` (`record?.provisionedBy ?? fallback`) inherits it.
- `op.oldResourceRetained` (`:890`): read with `??`, so a truthy non-boolean (`"no"`) takes the readopt arm -- no re-create, state re-pointed at the old physical id.
- `parseStateBody` validates the ROOT shape only; per-record `physicalId` / `resourceType` / `properties` / `provisionedBy` / `deletionPolicy` are trusted. The journal is not a weaker twin: every hazard above also exists for a hand-edited or planted `state.json`.

[...]

Issue #1198, opened 2026-07-24
feat(rollback): optionally revert the failed in-flight resource (follow-up to #1183)


go-to-k:
## Summary

Follow-up to #1183 (standalone `cdkd rollback`). In v1, `cdkd rollback` reverts every **completed** operation but leaves the resource whose operation **FAILED** as-is. CloudFormation additionally reverts the failed resource; cdkd does not, because only completed ops are journaled and doing so would mean force-applying `previousState` to a resource whose remote state is unknown.

This is a documented v1 limitation (see `docs/cli-reference.md#cdkd-rollback` "Known limitations" and the #1183 design "Known limitations" section).

## Groundwork already in place

- The `RESOURCE_FAILED` deployment event (`src/types/deployment-events.ts`) identifies the candidate resource.
- The rollback executor (`src/deployment/rollback-executor.ts`) already routes per-op through `provisionedBy` (#614), so a failed resource's revert could reuse the same provider dispatch.

## Proposed scope

Record the failed resource's pre-op `previousState` in the journal segment (or a sibling field) so `cdkd rollback` can optionally attempt to revert it, gated behind an explicit flag (e.g. `--revert-failed`) given the remote state is unknown. Off by default to preserve the safe v1 behavior.

## Non-goals

- Making it the default (the failed resource's remote state is genuinely unknown; forcing `previousState` could corrupt it).

go-to-k:
Working on this in branch feat/1198-1199-rollback-revert-failed-and-replacement (bundled with #1199 — same files). Touching src/deployment/rollback-executor.ts, src/deployment/deploy-engine.ts (journal segment for failed resource previousState), src/cli/commands/rollback*.ts, src/types/*, tests. Claiming to avoid collision with parallel agents.

PR #3153, opened 2026-09-15
fix(rollback): refuse a journal operation whose logicalId is not a string at the parser

### LINKED ISSUE #3140: fix(rollback): parseRollbackJournal accepts a non-string per-op logicalId, so a planted journal can crash cdkd rollback at the first lookup
## Summary

`parseRollbackJournal` (`src/types/rollback-journal.ts`) validates `journalVersion` and that `segments` is an array, and NOTHING per operation: `operations[].logicalId` / `resourceType` / `changeType` / `physicalId` are trusted casts. The journal is a sibling of `state.json` in the same bucket, writable by anyone with `s3:PutObject`.

go-to-k/cdkd#3104 made every RENDER of those fields safe for a non-string (`displayIdent` coerces through `displaySafe`'s throw-safe path; the pasted `--orphan` remedy is gated on `typeof === 'string'`). What is still open is the USE side, found by the round-5 security review of that PR:

- a `logicalId` of `{"toString": null}` (the go-to-k/cdkd#2947 shape) throws `TypeError: Cannot convert object to primitive value` at the first bracket lookup -- `classifyRollbackOp`'s `stateResources[op.logicalId]` -- so `cdkd rollback` dies with a raw stack trace after taking the lock (released by the `finally`); a numeric / `null` / `undefined` `logicalId` coerces at every lookup (`'123'`, `'null'`, `'undefined'`) and can silently match a state record of that name.
- `changeType` is routed on by literal; an unknown one already lands on the UPDATE arm by design (pinned by a case in `tests/unit/deployment/rollback-executor-log-injection.test.ts`), so that field needs no change.

## Proposed fix

Refuse at the parser, once, so every downstream lookup inherits it: in `parseRollbackJournal`, for each `segments[].operations[]` and `failedOperations[]`, require `typeof op.logicalId === 'string' && op.logicalId.length > 0` (and `resourceType` a string), throwing the existing `RollbackJournalParseError` shape with the segment index and op index -- rendered through the parser's `safeJournalText`, never the raw value. A per-op `physicalId`, when present, must be a string too. `parseRollbackJournal`'s doc comment states the "unchecked cast" gap; retire that sentence with the fix.

Tests: one case per refused shape (`{"toString":null}`, number, `null`, missing) asserting the parse error names the index and carries no raw value; the executor suite's existing NON-STRING case then documents defense-in-depth rather than the only guard (keep it).

Dup-check: searched open and closed issues for `parseRollbackJournal per-op logicalId string`, `rollback journal validation operations`, `toString null journal`; no hits. go-to-k/cdkd#3092 / go-to-k/cdkd#3104 cover the RENDER side only and say so.

Session-fit: now (do it in this session) -- `src/types/rollback-journal.ts` and its test file were edited this session (go-to-k/cdkd#3072) and the executor's lookups were read for go-to-k/cdkd#3104; a fresh session would re-read both.
Severity: low -- a planted journal can crash `cdkd rollback` with a raw TypeError (no exposure, no forged line, lock released) or steer an op at a same-named state record; the render side is already closed.
Effort: small (S) -- one validation loop in the parser, ~4 unit cases, no integ scope (`src/types/**` is outside every integ gate).
Estimate: ~45 min -- most of it the review round.

go-to-k:
Closes #3140

## Summary

`parseRollbackJournal` validated `journalVersion` and that `segments` was an array — and nothing per operation. The executor keys every lookup on `op.logicalId` (`stateResources[..]`, `orphanLogicalIds.has(..)`, the events it records), so a non-string id planted in `rollback-journal.json` (writable by anyone with `s3:PutObject`) either coerced at each lookup (`123` found the record named `'123'`) or threw a raw `TypeError` at the first one (`{"toString": null}`, the go-to-k/cdkd#2947 shape) after `cdkd rollback` had taken the lock. Found by the round-5 security review of go-to-k/cdkd#3104, which closed the RENDER side of these fields.

**Refused at the parser, once**, so no downstream reader has to remember: a segment that is not an object; an `operations` (or a present `failedOperations`) that is not an array; an operation that is not an object; an operation whose `logicalId` is not a non-empty string, whose `resourceType` / `changeType` is not a string, or whose present `physicalId` is not a string. Only the fields a lookup or a route is keyed on — `previousState` and the rest stay as they were (a bad value there fails the op it belongs to, not the parse; go-to-k/cdkd#3149 takes those). An unknown `changeType` still lands on the executor's UPDATE arm by design (pinned there).

**The refusal names the INDEX and the TYPE, never the value** — `segments[0].operations[1].logicalId must be a non-empty string (got number). Remove the stack's rollback-journal.json (next to its state.json) to discard it.` Absent rather than sanitized: there is no rendering step to get wrong. A case pins that a planted value never reaches the message; another pins the recovery hint on every refusal shape. The same text reaches `cdkd deploy`'s `Failed to write rollback journal` warning, since appending a segment parses the existing journal first.

**Every journal cdkd itself writes passes unchanged** (the round-trip case, plus a control with `physicalId` present and absent, an UPDATE carrying `previousState` / `provisionedBy` / `oldResourceRetained`, and a `failedOperations` entry with `attemptedProperties`).

The executor keeps its own `typeof` guard on the pasted `--orphan` remedy as defense-in-depth: the deploy engine's in-process rollback reaches it with ops it built itself, without this parser. Both comments now say so.

## Test plan

- [x] **Unit tests pass** — 1012 files / 22,879 passed / 1 skipped, root asserted to this worktree, exit 0 (rebased onto current main).
- [x] `tests/unit/types/rollback-journal.test.ts` +15 cases: one per refused shape (`it.each` over number / null / `{toString:null}` / array / empty string; missing id; non-object op; non-object segment; non-array `operations` / `failedOperations`), the value-absent pin, the recovery-hint pin, the exact body the review planted, and the extended acceptance control.
- [x] **Review round** (code + test + security on the branch): no blocker. Applied: `asserts` → `void` (no consumer); `(got <type>)` on the segment-level refusals; the recovery hint on every refusal, as `UnknownRollbackJournalVersionError` names `Upgrade cdkd`. **Recorded, not changed**: a planted malformed journal now makes the NEXT failed deploy's segment append fail (the engine warns, forwarding this message and its hint) where coercion used to let it through — the same class as the pre-existing invalid-JSON behaviour. **Follow-up filed**: go-to-k/cdkd#3149 — the nested record shapes (`previousState`, `attemptedProperties`, `provisionedBy`, `oldResourceRetained`) and `state.json`'s per-record fields, in both parsers.
- [x] **Mutation table** — measured on the final tree, one mutation per probe over the whole `tests/unit/`, hash-verified restore, every row's count equal to its named-case count (checked mechanically):

  | mutation | cases red |
  | --- | --- |
  | logicalId check dropped | 10 |
  | logicalId: empty-string arm dropped | 1 |
  | logicalId: typeof replaced by a COERCING test (String(v).length) | 7 |
  | resourceType check dropped | 1 |
  | changeType check dropped | 1 |
  | physicalId check dropped | 1 |
  | physicalId: absent no longer allowed (regression on every legitimate journal) | 3 |
  | op-is-object check dropped | 1 |
  | failedOperations never checked | 1 |
  | segment-is-object check dropped | 1 |
  | operations-is-array check dropped | 2 |
  | message carries the VALUE (kind -> String) | 8 |
  | recovery hint dropped from the refusal | 2 |

  - **logicalId check dropped** (10): `checks failedOperations with the same rule, at their own index`; `is refused BEFORE the executor could key a lookup on it -- the same body a round-5 review planted`; `names the way out in every refusal`; `never carries the planted value into the message`; `refuses a MISSING logicalId, saying undefined`; `refuses a logicalId that is a number, naming the index and the type`; `refuses a logicalId that is an array, naming the index and the type`; `refuses a logicalId that is an empty string, naming the index and the type`; `refuses a logicalId that is an object whose toString is not callable, naming the index and the type`; `refuses a logicalId that is null, naming the index and the type`
  - **logicalId: empty-string arm dropped** (1): `refuses a logicalId that is an empty string, naming the index and the type`
  - **logicalId: typeof replaced by a COERCING test (String(v).length)** (7): `checks failedOperations with the same rule, at their own index`; `is refused BEFORE the executor could key a lookup on it -- the same body a round-5 review planted`; `names the way out in every refusal`; `never carries the planted value into the message`; `refuses a logicalId that is a number, naming the index and the type`; `refuses a logicalId that is an array, naming the index and the type`; `refuses a logicalId that is an object whose toString is not callable, naming the index and the type`
  - **resourceType check dropped** (1): `refuses a non-string resourceType, changeType, and a present non-string physicalId`
  - **changeType check dropped** (1): `refuses a non-string resourceType, changeType, and a present non-string physicalId`
  - **physicalId check dropped** (1): `refuses a non-string resourceType, changeType, and a present non-string physicalId`
  - **physicalId: absent no longer allowed (regression on every legitimate journal)** (3): `accepts every legitimate operation shape, physicalId present or absent`; `setRollbackJournalFailedOperations persists a PARTIAL remaining list (per-op strip)`; `setRollbackJournalFailedOperations([]) strips the field from the NEWEST segment only (go-to-k/cdkd#1198)`
  - **op-is-object check dropped** (1): `refuses an operation that is not an object, naming its type`
  - **failedOperations never checked** (1): `checks failedOperations with the same rule, at their own index`
  - **segment-is-object check dropped** (1): `refuses a segment that is not an object and an operations that is not an array`
  - **operations-is-array check dropped** (2): `names the way out in every refusal`; `refuses a segment that is not an object and an operations that is not an array`
  - **message carries the VALUE (kind -> String)** (8): `checks failedOperations with the same rule, at their own index`; `is refused BEFORE the executor could key a lookup on it -- the same body a round-5 review planted`; `never carries the planted value into the message`; `refuses a logicalId that is a number, naming the index and the type`; `refuses a logicalId that is an array, naming the index and the type`; `refuses a logicalId that is an empty string, naming the index and the type`; `refuses a logicalId that is an object whose toString is not callable, naming the index and the type`; `refuses a non-string resourceType, changeType, and a present non-string physicalId`
  - **recovery hint dropped from the refusal** (2): `names the way out in every refusal`; `never carries the planted value into the message`
  </details>

  The op-is-object row measured 0 until a case planted a `null` / `7` / `['B']` operation (with the check gone, a `null` op threw a raw TypeError at the first property read).
- [x] **Live-tested against the BUILT bundle and the real state bucket**: journal with `logicalId: {"toString": null}` planted at `cdkd/Cdkd3140Probe/us-east-1/` (empty `resources`). main's binary: `TypeError: Cannot convert object to primitive value`. This branch: `Error: Rollback journal for 'Cdkd3140Probe' is malformed: segments[0].operations[0].logicalId must be a non-empty string (got object).` No lock left behind; probe swept with `s3_purge_prefix_versions ... all`, `s3_assert_versions_swept` = 0.
- [x] **Integration**: `rollback-executor.ts` (comment-only edit) is in `integ-destroy`'s scope → `/run-integ rollback-command`: PASS, 176 s, destroy 3 deleted / 0 errors, 0 surviving versions under both stack prefixes; ledger row included.
- [x] **Documentation** — `changelog.d/` entry (1704 chars); `.claude/rules/layout-state-types.md` clause on the parser.

https://claude.ai/code/session_01HbJf9kM7x3JWZT8F1yG4oV


go-to-k:
Merged as 7c3c611b — thanks. The reviewer-planted `{"toString":null}` body became the test that pins it, and the recovery hint on every refusal came straight out of the security round.

PR #3152, opened 2026-09-15
chore(work-issues): a lane forbidden real-AWS runs still writes the fixture arm; a 429-killed lane resumes with SendMessage

go-to-k:
## Summary

Retro of the `/work-issues` run that shipped go-to-k/cdkd#3077 (go-to-k/cdkd#3103), go-to-k/cdkd#3096 (go-to-k/cdkd#3125) and go-to-k/cdkd#3097 (go-to-k/cdkd#3139). Two flow lessons, each with the run's evidence, amended into the stage file where they fire (`references/retro.md` 10-b / 10-c: amend, do not append; pay for what you add):

- **`verify.md` 8-c** — a lane forbidden real-AWS RUNS still WRITES the fixture arm; the parent runs it. The go-to-k/cdkd#3103 lane read "no real-AWS integ" as "no fixture arm either", filed the `--no-wait` phase as `Session-fit: next` (go-to-k/cdkd#3096's original reason), and the parent wrote it (~120 lines of `tests/integration/ec2-instance/verify.sh`) and reddened its mutant within the hour. Paid for in the same file: 8-i's scratch-copy paragraph restated what `.claude/agents/pr-*-reviewer.md` already carries and is now a pointer plus the one carve-out. `verify.md` lands at exactly 30,000 B (the cap).
- **`gotchas.md`** — a lane killed by the account rate limit (HTTP 429 mid-turn) keeps its context: `SendMessage` it, never re-dispatch, and read the TREE first. Three kills on 2026-09-14 across the go-to-k/cdkd#3103 / go-to-k/cdkd#3139 lanes; one had the PR already open with `verify-pr` bound to a superseded sha.

`tests/unit/scripts/skill-file-payload.test.ts`: `MEASURED` re-recorded (corpus 187,940 → 188,312; the two leaders swapped — `verify.md` 30,000 largest, `implement.md` 29,992 runner-up) and `MIN_REFERENCE_CORPUS_BYTES` re-derived upward 158_300 → 158_670 (binding threshold 158,320, cleared by 350 B, inside the band the derivations carry). Re-derived, not raised to buy room — the per-file caps are untouched.

Retro §10-0 for the run: closed 3 / filed 2 (new 2 — go-to-k/cdkd#3096, go-to-k/cdkd#3097, both filed by the go-to-k/cdkd#3103 lane as `next` and both promoted to `now` at the promotion check and shipped in the same session / folded 1 — one row on go-to-k/cdkd#1852).

## Test plan

[...]

Issue #3150, opened 2026-09-15
fix(resolver): name arguments still print a sub-floor secret on resolver log lines and messages after the log twin


nix-tkobayashi:
## Summary

The resolver masks the NAME arguments of five intrinsics with `maskSecretsForLog` alone. That mask matches a recorded secret shorter than `MIN_NEEDLE_LENGTH` (4) only when it is the WHOLE text. So a 1-3 character secret that an `Fn::Join` or `Fn::Sub` assembles into a longer name argument is printed in the clear, even though the same pass registered a log twin for that string (go-to-k/cdkd#3100). The value lines already use the twin: `maskValueLeaves` runs every string leaf through `logTextOfLeaf` first, and go-to-k/cdkd#3114 routed the parameter, `Ref` and `Fn::GetAtt` value lines the same way. The name arguments were left out of that change. The second addendum on go-to-k/cdkd#3114 lists them as the same class with lower priority.

## Sites (`src/deployment/intrinsic-function-resolver.ts`)

The CLASS is: any log line, warning or thrown message that interpolates a name argument which came back from `resolveValue`, through `maskSecretsForLog` (or `maskThenStripThenMask`) alone. The list below was built by hand plus a grep, and three review rounds each found more of it, so treat it as a starting point rather than the set. The fix derives the set from the code before editing, for example with `grep -nE "maskSecretsForLog\((exportName|outputName|stackName|region|mapName|topLevelKey|secondLevelKey|attributeName|origin)|maskThenStripThenMask\(" src/deployment/intrinsic-function-resolver.ts`, plus every `logged*` binding those feed and every caller that ASSEMBLES a description from the raw names, and lists each site's disposition in the PR.

- `Fn::FindInMap`: the three "not found" throws and the `Resolved Fn::FindInMap:` success line print `mapName`, `topLevelKey` and `secondLevelKey` through `maskSecretsForLog` (`resolveFindInMap`).
- `Fn::ImportValue`: `loggedExportName` and the export-name throws print `exportName` through `maskSecretsForLog`, and so do the `logger.info` success lines of the exports-index, state-scan and CloudFormation-fallback paths, which reach default verbosity.
- `lookupCfnStackOutputs`: the warning printed when the CloudFormation fallback itself fails masks `stackName` and `region` with `maskSecretsForLog`.
[...]

Issue #3119, opened 2026-09-14
fix(secret-redaction): a sub-floor secret framed by Fn::Join / Fn::Sub under Fn::Base64 is persisted to state.json in its base64 form


go-to-k:
## Summary

`resolveBase64` (`src/deployment/intrinsic-function-resolver.ts`, the `Resolved Fn::Base64:` site) records the encoded result as a mask-only value ONLY when `maskSecretsForLog(resolvedValue) !== resolvedValue`, i.e. when the NEEDLE mask changes the decoded string. A 1-3 character secret embedded in a longer string (`port:` + `secretValueFromJson('pin')`, the CDK UserData shape `Fn::Base64` over `Fn::Join`) is below `MIN_NEEDLE_LENGTH` (4), so the needle mask leaves `port:q7` unchanged, `recordMaskOnlyValue` is skipped, and the base64 form of the framed plaintext (`cG9ydDpxNw==`, reversible) is persisted to `state.json` `properties` / `observedProperties` and reaches every reader of the bag (rollback journal, `cdkd diff`, events). The debug line prints both the decoded plaintext and the encoding.

Measured on go-to-k/cdkd#3116's head (`ba8ba0b6`) with the fixture harness of `tests/unit/deployment/intrinsic-resolver-sub-floor-log-mask.test.ts`, `Fn::Base64` over `Fn::Join ['', ['port:', PIN_REF]]`:

```
value = cG9ydDpxNw==
Resolved Fn::Join: port:***
Resolved Fn::Base64: port:q7 -> cG9ydDpxNw==
bag keys = ["q7"]
```

Same behavior on `origin/main` (which also prints the Join line in the clear), so it is not a regression of go-to-k/cdkd#3116. It is the sub-floor gap go-to-k/cdkd#2759 left: that fix keyed the detector on the needle mask, which by design cannot see a secret shorter than the floor.

## Proposed fix

go-to-k/cdkd#3116's log-twin registry (`LOG_TWINS_BY_PASS`) holds exactly the position information the detector lacks. At the `Fn::Base64` site, treat the input as secret-bearing when EITHER the needle mask changes it OR the registry holds a masked twin for it (`logTwinOfProduct({result, twin: result}, context).twin !== result`), and in that case `recordMaskOnlyValue(context.recordedSecretValues, result)` as today. The log line is the log-side half and belongs to go-to-k/cdkd#3116 (asked there); this issue is the STATE side, which needs its own state assertion.

## Verification plan

[...]

PR #3154, opened 2026-09-15
fix(resolver): mask a sub-floor secret on a nested child's parameter lines and on a two-stage Join part

### LINKED ISSUE #3114: fix(resolver): two routes still print a sub-floor secret on the Resolved Fn::Join / Fn::Sub lines after the log twin

nix-tkobayashi:
Closes #3114

## Summary

After go-to-k/cdkd#3100 the resolver prints a masked LOG TWIN of each value a write put a recorded secret into, so a 1-3 character secret (below `MIN_NEEDLE_LENGTH`) stays masked on the `--verbose` lines. go-to-k/cdkd#3114 found two routes that still reached those lines with no twin. This PR closes both, plus the extra lines the issue's two addenda named.

### Route 1: a nested-stack child

The parent registers the twin of a parameter value it built (`port:q7` -> `port:***`) under the resolver context's `recordedSecretValues`. The child looked the registry up only under its OWN bag. The lookup (`registeredLogTwin`, used by `logTwinOfProduct` and `splitLogTwins`) now reads the child's own bag, then its inherited bag. Two different registered twins for one string mask the whole string. Registrations are unchanged: they still go to the context's own bag.

**On the second addendum's objection to this design.** The addendum says `NestedStackProvider` passes the PER-RESOURCE bag, a different object from the one `LOG_TWINS_BY_PASS` is keyed by, so the carrier design would be needed. In the code the per-resource bag IS the resolver context's bag:

- `deploy-engine.ts` CREATE arm: `const createSecrets = context.recordedSecretValues ?? new Map()` after `this.resolver.resolve(desiredProps, context)`, then `withCurrentResourceSecrets(createSecrets, ...)`. The UPDATE arm binds `context.recordedSecretValues` the same way.
- `rollback-executor.ts` `resolveReplayProps` resolves with `recordedSecretValues: secrets` and binds that same `secrets`.
- `NestedStackProvider` forwards `getCurrentResourceSecrets()` unchanged as `inheritedSecrets`, and the child engine puts it on every resolver context.

`tests/unit/deployment/deploy-engine-resource-secrets-binding.test.ts` now pins that identity at all six CREATE / UPDATE sites plus the top-level scope control (each bound bag must be an object that resolved the resource's `{ Ref: <Param> }` leaf). Binding a copy at the first CREATE site reds two of its cases while every existing content assertion stays green. The rollback half is established by reading the code; no test pins its identity.

The route-1 lines the addenda named now take the twin before their encoder: `Parameter X: using user-provided value` (from the inherited bag, through a minimal context in `resolveParameters`), `Resolved Ref to parameter`, and the four `Fn::GetAtt` value lines (from attributes, from nested attributes, the legacy `NameServers` normalization, and the physical-id fallback). `Fn::Base64`'s input side already used `logTextOfLeaf` and is covered by the lookup. The outer `maskSecretsForLog` / `maskInherited` wrappers stay, so `check-resolver-mask-coverage.ts` sees each site as before. The SSM-resolved and `Default` parameter lines are unchanged: their values come from the child's own template or lookup, never from the parent.

### Route 2: a string resolved in two stages

`resolveJoin` re-resolves a string part that still spells `{{resolve:` (a resolved value that is itself reference text). That second stage now starts from the part's registered twin instead of its plaintext. `resolveSub`'s string-variable site is deliberately NOT seeded: a Sub variable's string is template text, and its one substitution resolves every token it spells, so it cannot be two-stage.

### One state effect

`resolveBase64`'s mask-only detector (go-to-k/cdkd#3119) asks the position mask, so in a child an `Fn::Base64` over a parent-built value now registers its encoding mask-only, and the child's state stores `***` for it, as the parent's already does. A rollback replay of that child resource therefore refuses the masked baseline (`refuseMaskedReplayBaseline`), as it does the parent's. No other caller of the lookup decides persistence.

### Residuals

- ONE LEVEL ONLY. The bag a child hands its own nested stack is the child's OWN resource bag, so no twin registered ABOVE the child reaches a grandchild. Where the middle passes the value THROUGH (`{ Ref: Param }` straight into the grandchild's `Parameters`) nothing ever registers a twin in that bag, so the grandchild prints the value whether or not the bag holds pairs; where the middle RE-WRAPS it the middle registers its own twin and the hand-off still drops a bag holding no pairs. The root cause is the carry refusing these frames, which also leaves the value in plaintext in state wherever a child RESOURCE consumes it: go-to-k/cdkd#3156.
- Not done: the masked walk now runs on the seven debug lines even when debug logging is off. The resolver's other `maskValueLeaves` lines already build their messages the same way; skipping the walk when the encoder would redact is a separate change.

## Tests

- `tests/unit/deployment/intrinsic-resolver-nested-child-log-twin.test.ts` (new, 17 cases), on the real resolver with the parent context's bag handed to the child as `inheritedSecrets`: the child `Fn::Join` / `Ref` lines, the user-provided parameter line, the `Fn::Base64` input line and its mask-only registration (with a registers-nothing control), the four `Fn::GetAtt` value lines, a child `Fn::Split` over the parent-built value, same-mask and different-mask registrations in both bags, a value the parent never masked (printed), a whole-value inherited entry (masked), an `ssm` SecureString frame (the one the carry refuses), a pin that the child registers only in its own bag, and route 2 with a direct-reference control. Route 1's Join case and route 2's case fail on `origin/main`.
- `tests/unit/scripts/resolver-mask-coverage.test.ts`: the JSON-encoding mutation probe replaced the first `this.maskValueLeaves(value, context)` in the file, which the new `Ref` site now precedes; it targets its intended `stringifyValue(this.maskValueLeaves(value, context))` site again.
- Mutation probes, one mutation at a time with the file restored between them, each redding at least one case: dropping the inherited lookup, the own-bag-only lookup, a `splitLogTwins`-only revert, dropping the different-mask and equal-mask rules, dropping the Join seed, and removing the leaf mask at each of the seven changed log sites or the inherited bag from the parameter-line context.

## Integration

`nested-stack-secret` gains a `SubFloorPinSsm` arm: a parent parameter spelled as an `Fn::Join` around an `ssm` SecureString token (a 2-character value created out of band), a frame `recordNestedStackParameterExpressions` refuses to carry. The child consumes it only in a `PinSsmNeverMatches` condition, so nothing persists it. `verify.sh` asserts the child's `Parameter SubFloorPinSsm: using user-provided value port:***`, `Resolved Ref to parameter: SubFloorPinSsm -> port:***` and `Resolved Fn::Join: x-port:***` lines. The exact `port:***` also proves the carry refused the frame, since a whole-value entry would print `***`. The framed value joins every plaintext scan. The premises are pinned: the synthesized shape, condition-only consumption (exactly one reference anywhere in the child template), and the parent's own row holding the framed expression.

- `nested-stack-secret`: PASS, `verify.sh` rc=0 in 394 s on the final tree (after the go-to-k/cdkd#3154 review round), destroy 0 errors. The new arm ran: the SecureString premise, the phase 1 and phase 2c `--verbose` scans (the three child lines present and masked), the three template and parent-row premises, and cleanup of the new out-of-band parameter. Not established by this run: that the arm FAILS on a pre-fix binary, which was not run; the unit cases carry that direction (route 1 and route 2 both fail on `origin/main`).
- `lambda` (broad set): PASS, `verify.sh` rc=0 in 119 s on the same final tree, destroy 0 errors; the S3 state versions of all three stacks swept to `_index/` only.

## Verification

- `/check`: lint, CI check, test typecheck, build, matrices all pass; `vp test run` 1015 files / 23007 tests passed (1 skipped) on the final tree; `/check-docs`: the `.claude/rules/layout-deployment-secrets.md` sentence on the needle floor now names the log twin (same byte count, the payload fence has no headroom on that path).
- Codex delta review: 1 round, NO ISSUES. Maintainer-checklist proxy pass: 3 rounds -- r1 five findings (the child `Fn::Split` lookup and the equal-mask rule were unpinned, the condition-only premise checked one spelling, the registry-write docstring overstated, three comments and a rules sentence were stale), r2 one (two registry-scope comments), r3 NO ISSUES.
- Rebased onto `origin/main` after go-to-k/cdkd#3139, go-to-k/cdkd#3145, go-to-k/cdkd#3122, go-to-k/cdkd#3142, go-to-k/cdkd#3148, go-to-k/cdkd#3153 and go-to-k/cdkd#3152 merged: no conflicts (the only shared file is the integ ledger); the fixture files are byte-identical to the commit the integ runs used.
- Reviewers (security, code, test): no blockers. Addressed in this PR: a pin that a child never registers into its parent's bag, an `ssm` SecureString frame case (the frame production routes through the lookup), the `Resolved Fn::Split:` line and a whole-log negative, the identity assertion narrowed to the `{ Ref }` resolution, and the `resolveJoin` / `registeredLogTwin` comments, changelog and rules sentence.

## Out of scope

The name-argument sites the second addendum lists (map keys, stack / output / export names, regions, attribute names, including transformed and composed names) are go-to-k/cdkd#3150.


go-to-k:
Reviewed at head `11743e2d`, four reviewers in parallel (spec, code, test, security). **No blocker.**

The second addendum's design objection is refuted, and I confirmed it independently rather than taking the PR body's word: the per-resource bag IS the resolver context's bag at every CREATE / UPDATE site (`deploy-engine.ts:5415/5459/5587/6024/6082/6477/6811`, with `buildResolverContext:1455` always minting one so the `?? new Map()` fallback is unreachable) and on the rollback path (`rollback-executor.ts:1215-1223`, one hoisted `secrets` local used at `:2330/:2499/:2653`). No carrier design is needed. The security trace of every reader also holds: no mask escapes into a resolved or persisted value, and "no caller other than `resolveBase64` decides persistence" is accurate — every other lookup caller sits inside a `logger.*` or `throw` template literal.

One item to fix before merge, one to file, the rest take-or-leave.

## M0 — required before merge: the stated grandchild bound is narrower than the residual

`intrinsic-function-resolver.ts:9134-9138` and the changelog entry both give the reason as *"the hand-off passes `inheritedSecrets` only when the bag's `size` is nonzero"* — the middle's bag holds a twin but no pairs. That is one case, not the bound.

`rememberLogTwin` has exactly four call sites (`:6255` `Fn::Join`, `:6644` `Fn::Sub`, `:9063` `splitLogTwins`, `:9750` dynamic references). Nothing on the inherited-parameter path registers a twin: `recordInheritedParameterSecrets` never calls it, and `resolveRef`'s parameter arm registers nothing. So the two grandchild spellings behave differently:

- **The middle RE-WRAPS the value** (an `Fn::Join` around `{Ref: MiddleParam}`): the middle's own pass registers a twin, and the `size > 0` pair gate is indeed what drops it. The stated reason holds here.
- **The middle passes the value THROUGH** (`{Ref: MiddleParam}` straight into the grandchild's `Parameters`): no twin is ever registered in the middle's bag, so the grandchild prints the plaintext **whether or not that bag holds pairs**. The stated reason does not reach this case at all.

This is not only wording. #3156 point 2 carries the same bound, and its *fallback* fix ("hand the bag down when it holds pairs OR registered log twins") closes only the first spelling — in the second there is no twin to hand down. Its primary fix (carrying the pair at `recordNestedStackParameterExpressions`) does close both, since a whole-value pair masks with no `MIN_NEEDLE_LENGTH` floor.

Please (a) restate the bound in the docstring and in the changelog entry as *the handed-down bag is the middle's own, so no twin registered above the middle is reachable*, and (b) add a line to #3156 point 2 recording that the fallback does not close the pass-through spelling.

## m1 — file separately, not for this PR: two error paths are now quieter than the lines this PR fixed

`intrinsic-function-resolver.ts:3387` (`Failed to evaluate condition ...`, `logger.warn`) and `:8780` (the `Fn::GetAZs` region refusal, `throw`) interpolate a resolved value through the needle mask only, with no twin lookup. `:3384-3386` already documents the sub-floor residual for the first.

After this PR a child's parameter lines print `port:***` at `--verbose`, while on those two paths the value prints in the clear **at default verbosity**. At `:3387` any sub-floor value leaks, since it is always embedded in the sentence and the substring arm has the floor; at `:8780` the raw value is masked first so the whole-value arm applies, and only an embedded sub-floor leaks. The integ's own `PinSsmNeverMatches` arm consumes the value in exactly such a condition — it simply never throws.

Both are pre-existing and this PR does not widen them, so they do not block. Please file this (or fold it into #3150) before merge so the asymmetry is not lost.

## m2 — take-or-leave: say at the `Fn::Base64` guard that it now crosses into the child's state

`:8710-8716`'s comment argues the log case only. Now that the lookup reaches the parent's registry, the guard's mask-only decision writes `recordMaskOnlyValue` into the **child's** bag, so `***` lands in the child's `state.json`. The PR body and the docstring at `:837-841` both say this; the guard itself does not, and the guard is what someone changing this reads.

Related, for the doc rather than the code: `splitLogTwins`'s unaligned arm (`:9037-9063`) registers every piece by value, public ones included, so a child `Fn::Base64` over a string exactly equal to such a piece persists `***` for a non-secret property, unrecoverably. Same hazard as within one pass since #3119 — recording it in `registeredLogTwin`'s doc is enough.

## m3 — take-or-leave: two test hardenings

- The rollback half of the identity claim is unpinned. It holds by construction today (one `secrets` local in both expressions), but nothing reds if a later edit interposes `new Map(secrets)`. One `expectBoundBagsAreResolverBags`-shaped case in the rollback replay suite closes it.
- `resolver-mask-coverage.test.ts`'s re-anchored probe is stable only by uniqueness of `stringifyValue(this.maskValueLeaves(value, context))`. `base.split(pattern).length === 2` would make a second occurrence fail loudly instead of silently re-targeting.

Everything else came back clean and I will not repeat it: all seven changed log sites have a failing-direction case, the sub-floor premise is real (`q7` / `po` / `m8` / `pb` are all 2 characters, so no outer masker makes a case vacuous), the identity assertions compare live objects rather than content, both over-masking and wrong-bag negative controls exist, `registeredLogTwin` cannot produce a partial mask, and the integ arm's cleanup covers every exit path.

## Before merge, on the gates

`docs/_generated/integ-last-run.tsv`'s `lambda` row reads `2026-09-15T04:26:44Z / 104 s / PR #3148` — main's row, not this PR's. The flatten-rebase onto #3148 dropped this branch's own row, so the committed ledger holds no record of the broad-set run the body describes. `integ-broad` is sentinel-bound and will not notice, but the ledger is the record. Please restore the row.

Separately, `integ-destroy` is `hash: diff` against the merge base, and the body says the review fixes changed comments in `src/` after both runs — so that marker has re-staled and `nested-stack-secret` needs a re-run on the final tree. Please budget for it before the merge attempt.

nix-tkobayashi:
Thanks for the review. Pushed as `7aa41692`. Item by item:

**M0 (grandchild bound).** Agreed; the size gate was one case, not the bound. `registeredLogTwin`'s docstring now states the bound as "the bag a child hands its nested stack is the child's own, so no twin registered above the child reaches a grandchild", and describes both spellings: the pass-through one (no twin is ever registered on the inherited-parameter path, so the grandchild prints the value whether or not the bag holds pairs) and the re-wrap one (the middle registers its own twin and the `size > 0` gate drops the bag). The changelog residual now carries the same bound and links go-to-k/cdkd#3156. go-to-k/cdkd#3156 point 2 now records the pass-through spelling and that its fallback fix closes only the re-wrap spelling, while the carry fix closes both.

**m1.** Folded into go-to-k/cdkd#3150 as an addendum comment: the `Failed to evaluate condition` warning and the `Fn::GetAZs` region refusals, with your analysis of which one leaks any sub-floor value and which only an embedded one.

**m2.** Taken. The `Fn::Base64` guard's comment now says the decision persists and, since this PR, crosses into the child's `state.json` (with the rollback replay refusal). The `splitLogTwins` unaligned-arm hazard (a public piece registered by value, so a child `Fn::Base64` over an equal string persists `***`) is recorded in `registeredLogTwin`'s doc.

**m3.** Both taken.
- `rollback-executor-nested-stack-secret-scope.test.ts` gains an identity case: it spies on `resolveDynamicReferences` and asserts the bag bound around `provider.update` is an object the replay resolved with. Replacing the binding at the retried update branch with `new Map(secrets)` reds it.
- The re-anchored mask-coverage probe now asserts its pattern occurs exactly once. Scoped to that row: the other two rows mutate the first of several occurrences on purpose, and a blanket check reds them. Appending a second occurrence of the pattern to the resolver reds the new assertion.

**Gates.** You were right about the ledger: the normalizer keeps the newest row per test, so main's `lambda` row replaced this branch's during the rebase. Rather than hand-restoring an older row the normalizer would drop again, both fixtures were re-run on the final tree:
- `nested-stack-secret`: PASS, `verify.sh` rc=0 in 394 s (07:21:26Z), destroy 0 errors, the `SubFloorPinSsm` arm's premises and masked-line assertions all ran.
- `lambda`: PASS, `verify.sh` rc=0 in 119 s (07:23:41Z), destroy 0 errors; state versions of all three stacks swept.

Both rows are now this PR's. On `integ-destroy`: `markgate set integ-destroy` on this branch reported an empty in-scope delta (the branch touches none of its `include:` paths, and `intrinsic-function-resolver.ts` is not among them), so the marker did not re-stale. The re-run was worth doing anyway, since the tree changed after the first runs.

`vp test run`: 1015 files / 23007 tests passed (1 skipped) on the final tree.

go-to-k:
Merged as 2eb2c7e6. Thanks — this was a careful round, and two things in particular:

The M0 fix went past what I asked for. I asked you to restate the bound; you separated the two grandchild spellings (pass-through vs re-wrap) in `registeredLogTwin`'s docstring, carried the same bound into the changelog, and then went and corrected go-to-k/cdkd#3156 point 2 — including the part that matters most, that the fallback fix closes only the re-wrap spelling while the carry fix closes both. A follow-up issue aimed at the wrong mechanism is the expensive kind of wrong, and you caught that yourself.

And the m3 claims were stated so they could be checked rather than believed. I ran both: replacing the retried update binding with `new Map(secrets)` reds the new identity case (and, as it happens, `revert a FAILED UPDATE` too — so the mutation is doubly caught), and appending a second occurrence of the pattern to the resolver reds the new premise with `expected 3 to be 2`. Both restored, tree clean.

Two disclosures.

**I was wrong about `integ-destroy`.** You were right that the branch touches none of its `include:` paths — `markgate set integ-destroy` said so in as many words on my run: *"this branch changes nothing the gate covers"*. `intrinsic-function-resolver.ts` is not in that list; I read the `hash: diff` semantics correctly and the scope wrong. The ledger half of that item was the real one, and re-running both fixtures rather than hand-restoring a row the normalizer would drop again was the better call.

**I edited the PR body before merging.** The Residuals bullet still carried the pre-M0 bound ("the hand-off passes `inheritedSecrets` only when the bag holds pairs, and the middle's bag holds only a twin") — the one place the correction had not reached, and with a squash merge that paragraph becomes the permanent commit message. I replaced it with the same ONE LEVEL ONLY wording you put in the docstring. Nothing else in the body was touched.

Gate runs on the final tree, for the record: `lambda` rc=0 in 89 s (9 deleted, 0 errors), `nested-stack-secret` rc=0 in 176 s (9+4 deleted, 0 errors, 0 surviving state versions on both prefixes), suite 1015 files / 23007 tests. I did not write ledger rows for those two runs, since recording them would mean pushing to your branch; your rows already document equivalent runs on the same tree.

Issue #3155, opened 2026-09-15
fix(state): cdkd state list --tree prints nothing for a parent chain thousands of records deep


nix-tkobayashi:
## What happens

`cdkd state list --tree` and `--tree --json` stop with a `RangeError` when the state bucket holds a long enough parent chain, and print no rows at all.

Four steps in the `--tree` path go one stack frame per tree level, and none of them sits inside `renderTreeMode`'s per-stack guard, which only covers the state READ:

- `buildStackTree`'s `sortRecursive` in `src/cli/commands/state-list-tree.ts`, which sorts every child list;
- `renderChildren`, behind `renderStackTreeAscii` (the text view);
- `stackTreeToJson` (the `--json` view);
- the `JSON.stringify(stackTreeToJson(roots), null, 2)` call in `renderTreeMode` itself (`src/cli/commands/state.ts`), which recurses into the nested object however `stackTreeToJson` builds it.

So one deep chain empties the view for every stack in the bucket, not only the chain's own rows.

Measured by the security review of go-to-k/cdkd#3069, calling the module directly with a synthetic chain of `N` entries (each naming the previous one as its parent), and by a follow-up check on `JSON.stringify` alone:

| Step | Depth at which it threw |
| --- | --- |
| `--tree --json` (`stackTreeToJson` / `JSON.stringify`) | about 5000 |
| `--tree` text (`renderChildren`) | about 10000 |
| inside `buildStackTree` (`sortRecursive`) | about 20000 |

These depths are not a fixed limit: V8's stack budget depends on frame size, runtime flags and what else is on the stack. A later run of `buildStackTree` alone on a 10000-entry chain threw `RangeError` inside `sortRecursive`, where an earlier run of the same input had returned. So the table shows the order of magnitude, not a threshold to design around.

## Why iterative walks alone would not fix it

Two of the four steps are bounded by the output itself, not by how it is built:

- `JSON.stringify` recurses into nested objects, so an iterative `stackTreeToJson` still hands it a structure too deep to serialize.
- The text view indents each level by four characters, so a chain `N` deep needs about `2 * N^2` characters of indentation alone. At `N = 50000` that is about 5 billion characters, far past V8's maximum string length, so no rewrite of `renderChildren` can produce byte-identical output there.

## Who can reach it

[...]

PR #3157, opened 2026-09-15
fix(state): keep cdkd state list's --long, --tree and plain views whole when one stack's state record or S3 key is unreadable or malformed

### LINKED ISSUE #3069: `cdkd state list --long` loses the whole listing to one stack it cannot read
## What happens

`cdkd state list --long` (and `--long --json`) reads every stack's record and
lock inside one `Promise.all` per row (`src/cli/commands/state.ts`, the
`refs.map` in `stateListCommand`). Neither read is guarded, so the first
rejection rejects the whole listing: the command logs the error and exits 1
without printing any listing rows, however many healthy stacks the bucket
holds.

**Scope, stated precisely:** only the `--long` modes are affected. Plain `cdkd
state list` and `state list --json` return from the reference listing before
these reads, and `--tree` already tolerates a record it cannot read
(`renderTreeMode`). So a user can still enumerate stack names and regions —
what they lose is the detail view, which is the one that shows which stack is
locked and when each was last written.

Three separate inputs reach it:

- **A record whose ROOT is not a JSON object.** Since go-to-k/cdkd#2947
  `parseStateBody` refuses `null`, an array and a primitive by name, and that
  refusal reaches this path. Only the falsy three (`0`, `false`, `""`) are a
  behaviour CHANGE — before #2947 they were read as falsy and rendered as a
  0-resource row. `null`, an array and a truthy primitive already aborted the
  same listing before #2947 (the exact shape varied by input: some threw
  inside `parseStateBody` and were wrapped by `getState`, others reached the
  resource-count walk in `stateListCommand`), so they are strictly better off
  for having a named reason.
- **The unsupported-version refusal**, which has aborted this listing since
  long before #2947 — same `Promise.all`, same `getState`, same throw. So the
  class is pre-existing and the root refusal is consistent with it, not a new
  hazard.
- **Any AWS read failure** — `AccessDenied`, a throttle, a transport error.
  `getState` wraps these in the same `StateError` as the parse refusals, so a
  fix cannot discriminate by error class and must tolerate all of them.

## Why a per-row fallback is not a one-liner

The tolerance was written and reverted during go-to-k/cdkd#2947 (PR
go-to-k/cdkd#3067) because it grew past that PR's scope. What the attempt
measured is worth keeping, so the next one does not rediscover it:

1. **Both halves must degrade, and report SEPARATELY.** The record read and
   the lock read share an S3 client, so a credential failure lands on
   `isLocked` too; guarding only `getState` leaves the listing aborting on the
   lock. And a single conflated field made the renderer withhold the resource
   count when only the LOCK had failed — the row must be able to say "counts
   real, lock unknown".
2. **The reason string can quote the state body.** The invalid-JSON refusal
   interpolates V8's `SyntaxError`, which carries bytes the record's author
   chose; anyone with `s3:PutObject` on the state bucket picks them. Sanitizing
   and truncating does not redact — a plaintext canary planted in a malformed
   body reached both `--long` and `--long --json` in the reverted attempt. So
   the row needs a FIXED, class-level reason (or an explicit redaction pass),
   not the producer's message copied through.
3. **`renderTreeMode` is the precedent to follow** — it already degrades a row
   it cannot resolve instead of failing the tree, so the listing has a
   house style for this.
4. **A legacy `version: 1` reference with no resolved region is not read by
   this view at all**, so it renders zero resources with no reason attached
   and a fix aimed at the read path does not reach it.

## Proposed fix

Guard each of the two reads independently in `stateListCommand`'s `refs.map`;
carry a per-row record-side and lock-side failure indicator through
`StackDetail`; render a class-level reason in `--long` and expose it in
`--long --json`. Decide the reason text from the FAILURE CLASS, not from the
caught message.

## Verification plan

- A listing with one unreadable stack and one healthy peer still prints the
  healthy row, asserted PER BLOCK (a whole-output negative cannot fail — the
  healthy row legitimately carries the same labels).
- A row whose RECORD failed still reports its lock; a row whose LOCK failed
  still reports real, nonzero counts (a case with an empty `resources` bag
  cannot tell a preserved count from a default).
- A malformed body carrying a synthetic plaintext canary does NOT put that
  canary in `--long` or `--long --json`.
- An AWS `AccessDenied` on each of the two reads separately.

Dup-check: searched open issues for `state list` / listing / `Promise.all`
tolerance — go-to-k/cdkd#3018 is the nearest and is a DIFFERENT property
(consumers dereferencing a malformed resources bag: `state destroy`, `state
orphan`, `state refresh-observed` and 18 further sites). This one is "one view
survives a peer stack's failure" and shares no site with it.

Session-fit: now (do it in this session) — re-classified on 2026-09-15: the
only reason for next was the go-to-k/cdkd#2947 PR scope, which expired when
go-to-k/cdkd#3067 merged; no open PR holds src/cli/commands/state.ts.
Severity: low — a single unreadable stack costs the whole `--long` detail
listing, but plain `state list`, `state list --json` and `--tree` still
enumerate every stack, `state show <stack>` still works per stack, and nothing
is corrupted.
Effort: medium (M) — two guarded reads, two carried fields, a renderer arm,
plus the redaction decision in (2), which is the part that needs a call rather
than code.
Estimate: ~2-4 h — most of it in the reason-text design and the per-block test
shapes, not the guards.

nix-tkobayashi:
## Summary

`cdkd state list --long` no longer loses the whole listing to one stack it cannot read. Review widened the same guarantee to `--tree` and to the plain listing, for records that are readable but malformed.

Each row read its state record and lock inside one unguarded `Promise.all`, so the first rejection rejected the whole listing: the command exited 1 without printing a row. Three inputs reach it:

- a record whose body cannot be parsed, or whose root is not a JSON object;
- the unsupported-version refusal;
- any AWS read failure, such as `AccessDenied` or a throttle.

`getState` wraps all three in the same `StateError`, so the fix tolerates every failure rather than discriminating by class.

## What changed

In `src/cli/commands/state.ts`, the two reads are guarded separately, following `renderTreeMode`'s existing per-row tolerance, and they degrade separately:

- **Record read fails.** The row renders `Resources: unknown (...)` and still reports its lock.
- **Lock read fails.** The row still reports its real resource count and last-modified time, and renders `Lock: unknown (...)`.

`--long --json` carries the same split as two new fields, `stateReadError` and `lockReadError`. It reports `resourceCount` or `locked` as `null` on the failed side instead of a made-up `0` or `false`.

The reason is a fixed, class-level string, never the caught message. The invalid-JSON refusal quotes bytes of the state body, and whoever can write the bucket chooses those bytes. The new `readOrFailure` helper drops the error, so no row can copy it. `cdkd state show` for that one stack reports the actual error.

A warning on stderr counts the degraded rows. The exit code stays 0.

Review found more shapes that break the same listing, and this PR closes each:

- **An out-of-range `lastModified`.** A number such as `1e300` made `toISOString()` throw after the read guard had already reported success, so one planted record could still reject the whole listing. The new `isoTimestampOrNull` renders it as `unknown`.
- **A `resources` that is not a JSON object.** The count was `Object.keys(state.resources ?? {}).length`, and `parseStateBody` validates nothing inside the root. A planted string was counted per character (`"abcdef"` showed `Resources: 6`) and a list per element, both made-up numbers. A string of millions of characters built an array of that many keys outside the read guard, which could exhaust memory and lose the whole listing. The new `resourceCountOrNull` counts a JSON object's keys, keeps an absent or `null` bag at 0 as before, and returns `null` for anything else. That row gets its own fixed `stateReadError` reason, so `resourceCount` stays `null` exactly when a reason is set: the text row shows `Resources: unknown (...)` rather than `Resources: null`, and the stderr warning counts it.
- **Unsanitized text rows.** The `--long` and `--tree` text views printed the stack name and region raw. Both come from an S3 key segment, which can carry a newline or ESC, so a planted key could forge a line inside another row. Both views, and the plain one-reference-per-line listing, now render through the file's existing `formatStackRefSafe` / `safe()`. The plain listing needs it most, because a `while read` loop consumes it: a planted key forged a second, fully formed reference line. `safe()` is an ASCII allowlist, so every legitimate row is byte-identical. `--json` is not sanitized at all: `JSON.stringify` escapes only C0, `"`, `\` and lone surrogates, tracked as go-to-k/cdkd#3163.
- **A non-string parent link in `--tree`.** `renderTreeMode` copied `parentStack` / `parentRegion` whenever they were defined, and `buildStackTree` interpolates both into a key outside the per-row `catch`. A record whose `parentStack` is `{"toString": null}` threw a `TypeError` and emptied the whole view. Only string parent fields are copied now. cdkd writes these fields as strings, so a record it wrote links as before. A hand-edited array value such as `["Parent"]`, which interpolation used to coerce into a working link, now lands at the root instead. The link is kept or dropped whole: a string `parentStack` beside a non-string `parentRegion` would otherwise key with an empty region and nest under an unrelated legacy record of the same name. An absent `parentRegion` is the one exception and still makes a valid link, because that is how a child names a legacy region-less parent.
- **A parent loop in `--tree`.** `buildStackTree` in `state-list-tree.ts` broke a self-link but not a longer loop. Two records naming each other as parent each became the other's child, so neither was a root and both vanished from the text and JSON views. Every node on a parent loop now lands at the root, and a node hanging off the loop stays under its parent. Loop membership is found in one linear pass over the parent links. Measured on this branch over three runs, `buildStackTree` takes 14-20 ms for a 10,000-record loop and 3-8 ms for a 5,000-record chain. A 10,000-record chain intermittently hits the recursion limit of the existing `sortRecursive`, which predates this PR; that is filed separately as go-to-k/cdkd#3155, with the other depth-bound steps of `--tree`.

A legacy row with no region gets a plain `lock could not be read`. `cdkd state show` refuses a region-less record before it reads the lock, so pointing there would name a command that never shows the error.

`docs/cli-state.md` describes the new behaviour, and its malformed-root paragraph now names the `--long` exception.

## Residual

A legacy `version: 1` reference with no resolved region is still not read by this view, so it renders zero resources with no reason attached. That is item 4 of the issue, which a fix on the read path does not reach.

## Verification

**Unit.** `tests/unit/cli/state-list.test.ts` gains a `--long degrades a row whose read fails` block with 19 cases plus a 36-row parent-link table and a check that the table covers all 36 pairs, and `tests/unit/cli/state-list-tree.test.ts` gains 5 parent-loop cases. The existing `--long --json` case now expects the two new fields.

| Case | Pins |
| --- | --- |
| healthy peer beside a record read that rejects | the healthy row prints in full, asserted on its own block; the failed row shows the fixed reason |
| record fails on one row, lock fails on another | the record-failed row keeps `Lock: locked`; the lock-failed row keeps its nonzero count |
| the same split under `--long --json` | `resourceCount` / `locked` are `null` on the failed side, and `stateReadError` / `lockReadError` carry the reason |
| a canary in both rejection messages and their `cause` | the canary reaches neither output mode nor any logger level, checked through a deep inspect, after a premise that the row really degraded |
| three stacks, one record failing, then all healthy | one warning naming `1 of 3`, and no warning when every read succeeds |
| a lock-only failure | the warning counts it |
| a legacy row with no region whose lock fails | the record is never read, only the lock degrades, and its reason names no follow-up command |
| `lastModified: 1e300` beside a healthy stack | `unknown` on that row; the healthy row keeps its timestamp |
| `lastModified` of `true`, `1e300` and `'2026-01-01'` | `unknown` in text and `null` in `--json` for all three, not a made-up date |
| `resources` of `"abcdef"`, `[1, 2, 3]`, `42`, `true`, `null` and absent, beside a healthy two-resource stack | the first four show the malformed reason in text and `resourceCount: null` with that reason in `--json`; `null` and absent stay `0`; the healthy row keeps `2`; the warning counts exactly the four malformed rows |
| a newline and ESC in the stack name, and a newline in the region, under `--long` | exactly 5 lines, one `Lock:` and one `Resources:` line, no ESC |
| a newline and ESC in the stack name under `--tree` | one line, and it is the sanitized label, with no ESC |
| a `--tree` record whose `parentStack` and `parentLogicalId` are `{"toString": null}` | both rows print in text and `--json`, with both fields `null` |
| a newline and ESC in the stack name under the default listing | one line, and it is the sanitized reference, with no ESC |
| a string `parentStack` beside a non-string `parentRegion`, with a legacy and a regional record of the parent's name | the stack lands at the root, not under the legacy record |
| a valid link whose `parentLogicalId` is not a string | the child still nests under its parent, with `parentLogicalId` `null` |
| a non-string `parentStack` beside a string `parentRegion` | `parentStack`, `parentLogicalId` and `parentRegion` are all `null` in `--json` |
| a `--tree` row whose record read resolves to no record, beside a healthy stack | both rows print in text and `--json`, so the link rule never dereferences the missing state |
| the parent-link table: six `parentStack` shapes against six `parentRegion` shapes, with a legacy and a regional `Parent` | where `Child` lands and the exact `parentStack` value `--tree --json` emits for it, written out per row |
| the table's own size | exactly 36 distinct rows, so a shrunk shape list cannot silently shrink the product |
| a child naming a legacy region-less parent, with no `parentRegion` | the child still nests under the legacy parent, so the whole-link rule does not over-tighten |
| two records naming each other as parent (tree unit) | both land at the root |
| two disjoint parent loops, one with a hanging child (tree unit) | all four loop members land at the root, and the child stays under its parent |
| a five-record parent loop (tree unit) | every member lands at the root, so a walk capped at a small step count cannot pass |
| a three-record loop with a hanging child (tree unit) | every stack visible once, and the child stays under its parent |
| the hanging child listed first, so the loop is found on a walk that starts from it (tree unit) | the child is not counted as a loop member and stays under its parent |

Each block assertion reads one row, because the healthy peer legitimately carries the same labels.

Mutation probes on the two test files (96 tests), one mutation per probe, source and test files restored byte-exact after each, measured on the tree of 0a90edbe:

| Mutation | Result |
| --- | --- |
| pre-fix source | 43 red: 20 rows of the parent-link table, the updated JSON-shape case, and every other new case except three that the pre-fix code also passes: the legacy-parent case, the `--tree` no-record case and the 36-pair size check |
| state read unguarded | 5 red |
| lock read unguarded | 5 red |
| caught message copied into the reason | 4 red, including the canary case |
| a record failure also marks the lock failed | 3 red |
| `resourceCount` 0 instead of `null` on failure | 1 red |
| `locked` `false` instead of `null` on failure | 1 red |
| warning never fires | 3 red |
| warning always fires | 1 red |
| warning filter ignores a lock-only failure | 1 red |
| lock reason back to its first wording | 3 red |
| a legacy row gets the regular lock reason | 1 red |
| `lastModified` converted with a bare `new Date(...).toISOString()`, no type or range check | 2 red |
| only the out-of-range date check dropped | 2 red |
| `resourceCountOrNull` without the `typeof` check | 1 red: the malformed-`resources` case |
| `resourceCountOrNull` without the `Array.isArray` check | 1 red: the malformed-`resources` case |
| the `null`-bag check dropped, so `Object.keys(null)` throws and aborts the listing | 1 red: the malformed-`resources` case |
| the absent-bag check dropped, so a record with no `resources` shows unknown | 1 red: the malformed-`resources` case |
| the old `Object.keys(state.resources ?? {}).length` restored | 1 red: the malformed-`resources` case |
| a malformed `resources` with no `stateReadError` | 1 red: the malformed-`resources` case |
| the generic record-read reason used for a malformed `resources` | 1 red: the malformed-`resources` case |
| the type and finiteness checks dropped, leaving only the invalid-date check | 1 red |
| default listing printed raw | 1 red |
| `--long` header printed raw | 1 red |
| `--long` `Region:` line printed raw | 1 red |
| `--tree` label printed raw | 1 red |
| `--tree` accepts a non-string `parentStack` | 8 red: 7 table rows |
| `--tree` keeps a link whose `parentRegion` is not a string | 7 red: 6 table rows |
| `--tree` requires a string `parentRegion` for any link | 3 red: 2 table rows and the legacy-parent case |
| `parentRegion == undefined` instead of `=== undefined` | 2 red, both table rows |
| a `null` `parentStack` accepted as a link | 1 red, a table row |
| a number `parentStack` accepted as a link | 3 red, all table rows |
| a number `parentRegion` accepted in a link | 2 red, both table rows |
| the string arm of the `parentRegion` check dropped | 8 red: 4 table rows |
| `--tree` copies a non-string `parentLogicalId` on a valid link | 1 red |
| `--tree` rewrites a retained `''` `parentStack` to `'FORGED'` | 3 red: the three table rows that keep a `''` link |
| the `''` region column dropped from the parent-link table | 1 red: the 36-pair size check |
| the `123` stack row dropped from the parent-link table | 1 red: the 36-pair size check |
| the link rule reads `state!.parentStack` instead of `state?.parentStack` | 1 red: the `--tree` no-record case |
| `--tree` copies `parentRegion` without a valid link | 1 red |
| `--tree` copies `parentLogicalId` without a valid link | 1 red |
| `parentRegion` type check inside a kept link loosened | 0 red (equivalent: the link check already requires a string region) |
| parent-loop detection off | 6 red, including the existing self-link case |
| a whole walk path counted as the loop | 1 red: the tail-first case |
| reaching an already-settled node counted as a loop | 15 red |
| settled nodes never recorded | 0 red (see below) |

One mutant of the loop pass is not caught by any unit test, by design: not recording settled nodes keeps the output identical and only makes the pass quadratic. A timing assertion would compare two moving things, so the linear cost is shown by the measurement above instead.

Full unit suite, typecheck, lint, build and generated-artifact freshness pass locally.

**Live, against the test account's state bucket.** I planted two records under throwaway stack names: a valid record with 2 resources, and a record whose body is not valid JSON and carries a canary string. I ran the built CLI, then deleted every version of both keys and confirmed zero remained.

| Binary | Command | Result |
| --- | --- | --- |
| this branch | `state list` | both rows |
| this branch | `state list --long` | exit 0; the valid row shows `Resources: 2`; the malformed row shows `Resources: unknown (state record could not be read; ...)` and `Lock: unlocked`; stderr warned `1 of 2 stack(s) could not be fully read; their rows show which read failed.`; the warning now reads `1 of 2 stack(s) could not be fully read or counted; their rows say why.`, a wording change made after this run |
| this branch | `state list --long --json` | exit 0; `resourceCount: null` and `stateReadError` set on the malformed row only |
| `main` at `9bd36c8d` | `state list --long` | exit 1, no rows printed, the healthy stack included |

The canary appeared in none of the branch's six captured streams. The pre-fix error on `main` did not quote it either, since V8 reported only a position for this body, so the live run does not demonstrate the quoting hazard; the unit canary case covers that.

A second live run covered the two shapes review added. It planted three records: a valid one, one with `lastModified: 1e300`, and one whose S3 key's stack-name segment holds a newline, `  Lock: unlocked` and an ESC sequence. All keys and versions were deleted afterwards, and zero remained.

| Binary | Command | Result |
| --- | --- | --- |
| this branch | `state list --long` | exit 0, 3 rows and exactly 3 `Lock:` lines; the `1e300` row shows `Last Modified: unknown`; the crafted name renders on one line with no ESC byte |
| this branch | `state list --long --json` | exit 0; `lastModified: null` on the `1e300` row; the crafted name is JSON-escaped |
| this branch | `state list --tree` | exit 0, 3 lines, no ESC byte |
| `main` at `9bd36c8d` | `state list --long` | exit 1 with `RangeError: Invalid time value`, no rows |

A third live run covered the two `--tree` shapes. It planted four records: a valid one, one whose `parentStack` is `{"toString": null}`, and two naming each other as parent. All probe versions were deleted afterwards, and zero remained.

| Binary | Command | Result |
| --- | --- | --- |
| this branch | `state list --tree` | exit 0, all 4 stacks as root rows |
| this branch | `state list --tree --json` | exit 0; the non-string parent is `null`; both loop members are roots with their `parentStack` kept |
| `main` at `9bd36c8d` | `state list --tree` | exit 1 with `TypeError: Cannot convert object to primitive value`, no rows |

The pre-fix run stopped at the non-string record, so it does not separately show the loop hiding its stacks; the tree unit case measures that.

Closes #3069


go-to-k:
Reviewed at `127e9588`, 3-axis plus a security pass. The tier came out 3-axis + security: base 1-reviewer by size (708 LOC / 6 files), up-biased because the last three commits on `src/cli/commands/state.ts` are all `fix(state):` (#3095, #3087, #3067) and #3067 is the same class this PR extends; the security reviewer is additive because the diff does escaping and error-message redaction over attacker-chosen bytes.

The core of the change holds up. I re-traced the two things the issue said were hard and both are right: `readOrFailure` never binds the error, so no caller can copy it, and I could not find a reader — text, `--json`, the stderr warning, any logger level, or a `cause` chain — that the body bytes reach. The loop pass in `buildStackTree` is correct on self-link, 2-cycle, N-cycle, a child hanging off a loop listed first and last, an absent parent, and empty input, and it really is linear: a node enters `path` at most once across all starts, and `settled` is written only after the loop verdict, so it cannot hide a membership. The mutation table matches the tests. CI is green.

Four things to fix before this merges, then I will merge it.

### M0 — required — a half-valid parent link binds to the wrong record instead of landing at the root

`src/cli/commands/state.ts:615-621` applies the three `typeof === 'string'` filters independently, so a record with a string `parentStack` and a non-string `parentRegion` keeps the link and drops only the region. `refKey` maps a missing region to `''` (`state-list-tree.ts:51`), which is the same key a legacy region-less record produces. With

```
{ stackName: 'Child',  region: 'us-east-1', parentStack: 'Parent' }  // parentRegion was {"toString": null}
{ stackName: 'Parent' }                                              // legacy, no region
{ stackName: 'Parent', region: 'us-east-1' }                         // the real parent
```

`Child` is filed under the **legacy** `Parent`. Before this PR a coercible non-string such as `123` produced key `Parent\0123`, matched nothing, and landed at the root — so for that input the PR moves a row from the root onto an unrelated parent, and the comment's own promise ("now lands at the root instead") does not hold for the mixed case.

Reject the whole parent link when `parentStack` is a string but `parentRegion` is present and not a string, and pin the three-record shape above.

### M1 — required — the plain listing stays raw on a rationale that is inverted

`state.ts:451` prints `formatStackRef(ref)` raw, and the new comment at `state.ts:523-527` records that as deliberate: "it is a record set a `while read` loop consumes". That is the reason to guard it, not to exempt it. `listStacks` applies no validation, so `PutObject cdkd/Decoy<LF>ProdStack (us-east-1)/us-east-1/state.json` makes `cdkd state list` — the **default** mode — emit a second, fully formed reference line for a stack that has no record, and a cleanup script acts on it. At a TTY a planted `ESC[1A` overwrites the preceding genuine row, hiding a real stack. `--long`'s worst case, which this PR did guard, is a cosmetic row; the default mode's is a forged record.

Two reviewers reached this independently and proposed different fixes (sanitize vs refuse the ref). Sanitize, for consistency: route line 451 through `formatStackRefSafe`, the same helper `--long` and `--tree` now use. `safe()` is `displaySafe(v, {asciiOnly: true})`, a positive allowlist over `[ -~]`, and CloudFormation stack names and AWS region codes are printable ASCII, so it is byte-identical on every legitimate row and fails closed on the planted one. The round-trip objection cuts the other way: today the forged line round-trips into `cdkd state show` as a *different* stack, which is the attack.

Then rewrite the comment to say why, and update the `docs/cli-state.md` sentence that currently reads "the plain listing and `--json` are not changed".

This is pre-existing behaviour, so I would normally file it. I am asking for it here because this PR is what put a written justification on the exemption, and shipping the justification without the fix is worse than the silence it replaced.

### M2 — required — "`--json` escapes by construction" is false

Same comment, `state.ts:527`. `src/utils/display-safe.ts:11-18` says the opposite in this repo's own words: `JSON.stringify` escapes C0, but emits DEL, `U+0085`, the C1 range (xterm reads `U+009B` as CSI) and `U+2028` verbatim. `cdkd state list --json` with a planted `Evil\u009B2J` segment puts a live CSI in the operator's terminal.

The `--json` surfaces are uniform across cdkd and not this PR's to fix. Just say "escapes C0" so the claim is true and the next reader can still find the gap.

### M3 — required — the new `docs/cli-state.md` paragraph breaks the page rules

`docs/cli-state.md:154-171` is one 19-line paragraph. `.claude/rules/docs-page-template.md:76` caps a paragraph at about six lines without a list, table or code block, and `:67` puts enumerable facts in a table rather than prose. The degradation matrix is exactly enumerable — which read failed, what the text row shows, what `--json` carries — so a four-column table plus a short note reads better and satisfies both rules. Drop "where before such a record could empty the view", which is history framing that `:86`'s Voice section bans.

### m4 — optional — three test strengthenings

Take or leave, none of them block:

- `tests/unit/cli/state-list.test.ts:508-512` — the "no warning when every read succeeds" arm asserts only `not.toHaveBeenCalled()` and discards the output, so an early bail-out that prints zero rows also passes. A `expect(out).toContain('  Resources: 0')` premise closes it.
- `:467-473` — the premise proves only the state side degraded, so the lock-message canary's absence is unguarded.
- `state-list-tree.test.ts` — `loopMembers` and `settled` are shared across starts and nothing exercises two **disjoint** cycles. I traced it by hand and it is correct, so this is coverage for a future edit rather than a suspected bug.

### m5 — out of scope, filed elsewhere

A parent chain thousands of records deep hanging off a loop now reaches `sortRecursive` where before `roots` was empty and `--tree` printed nothing at exit 0, so this PR makes that `RangeError` shape newly reachable. The depth bound is pre-existing and already filed as #3155, and rooting loops strictly reduces the reachable depth elsewhere, so nothing is owed here — I have added the shape to #3155.

The `["Parent"]` parent link that used to coerce into a working link and now lands at the root is a deliberate behaviour change and correctly described; no action.

Push M0-M3 and I will re-review and merge.

nix-tkobayashi:
Thanks for the round on `127e9588`. M0-M3 are in `3db7aba`, and I took all three m4 strengthenings.

**M0.** `renderTreeMode` now keeps or drops the parent link whole. `parentStack` is copied only when it is a string AND `parentRegion` is absent or a string; `parentLogicalId` and `parentRegion` are copied only when that link is kept. Your three-record shape is pinned in `tests/unit/cli/state-list.test.ts` ("drops a half-valid parent link WHOLE"): under `--tree --json`, `Child` is a root with `parentStack: null`, beside both `Parent` records. A probe then showed the opposite half-valid shape unpinned: a string `parentRegion` copied from a record whose `parentStack` is not a string. A second case pins it, so all three parent fields come out `null` there. A third pins `parentLogicalId` on a link that is otherwise valid: the child still nests under its parent, with `parentLogicalId: null`. The `parentRegion` type check inside the kept link is implied by the link check itself, so its loosened mutant behaves identically and no case can pin it; the PR body records that.

**M1.** The default listing now prints through `formatStackRefSafe`, the helper `--long` and `--tree` use. The comment above it now gives your reason: a `while read` consumer is why it must be guarded, and `safe()` keeps every legitimate row byte-identical. A new case plants `Decoy<LF>ProdStack (us-east-1)<ESC>[1A` as a stack name and asserts exactly one line and no ESC. The existing default-listing cases pass unchanged. The `docs/cli-state.md` sentence saying the plain listing is unchanged is gone.

**M2.** The comment now says `--json` is not covered, because `JSON.stringify` escapes C0 but emits DEL, the C1 range and `U+2028` verbatim, pointing at `display-safe.ts`. The docs page says the same.

**M3.** The paragraph is now two tables plus two short notes. One table covers which read failed, what the text row shows, and what `--long --json` carries. The other covers the malformed values that render instead of stopping the listing. The history framing is removed.

**m4.**
- The no-warning arm now asserts three `Resources: 0` lines before `not.toHaveBeenCalled()`.
- The canary case's premise now also asserts `Lock: unknown (...)` in text and `lockReadError` in `--json`.
- `state-list-tree.test.ts` gains a case with two disjoint loops, one carrying a hanging child.

Thanks for adding the deep-chain-off-a-loop shape to go-to-k/cdkd#3155.

Local runs on this head: full unit suite 1011 files / 22849 tests, plus typecheck, lint, build and generated-artifact freshness. The PR body's mutation table is remeasured on this head: 29 probes over the two test files (57 tests). All but two fail at least one case; the two that pass are the equivalent `parentRegion` check and the settled-set omission that only costs time, both explained there.

go-to-k:
Round 2, reviewed at `3db7aba5`, same 3-axis plus security pass as round 1.

**M0, M1, M2 and M3 are all fixed**, and I verified each rather than taking the reply for it. M0: I re-ran the three-record probe and `Child` is now a root with `parentStack: null` beside both `Parent` records, and all twelve parent-field shapes behave — including the one the whole-link rewrite could have broken, a string `parentStack` with an ABSENT `parentRegion`, which still links to a legacy region-less parent. M1: `formatStackRef` is gone with zero remaining references in `src`, `tests`, `scripts`, `docs` and `.claude`, and `safe()` is byte-identical on every legitimate name — `Parent~Child` included, since `~` is inside the allowlist — so the round-trip into `cdkd state show` is intact. M2 is measured true by code point. M3's prose-to-table conversion lost no clause; I walked the old paragraph against the new tables.

Your claim that the inner `parentRegion` check is now implied by `linkIsValid` and therefore unpinnable is **correct** — the only delta from loosening it is a `parentRegion: undefined` own key, and none of the three readers uses `in` or `Object.keys`. Keep the check; removing it would emit that key for a future `in`-based reader to trip on.

Two things to fix, then I will merge.

### M4 — required — the legitimate half of the M0 fix is unpinned, and one case that looks like it pins it cannot

The disjunct `state.parentRegion === undefined ||` at `src/cli/commands/state.ts:621-623` is what keeps a string `parentStack` with an absent `parentRegion` linking to a legacy region-less parent. **No test supplies that shape.** I grepped every `parentStack:` site in both test files: all fifteen pair a string `parentStack` with a `parentRegion`. So the obvious over-tightening of this very fix —

```ts
const linkIsValid =
  typeof state?.parentStack === 'string' && typeof state.parentRegion === 'string';
```

— survives all 57 tests while promoting a legitimately nested legacy child to the root. That is the silent-row-move class this issue exists to close, and the mutation table probes this clause only in the loosening direction.

Add the shape: a parent `{ stackName: 'Parent' }` with no region, and a child `{ stackName: 'Parent~Child', region: 'us-east-1' }` whose state carries `parentStack: 'Parent'` and no `parentRegion`; assert the child nests.

While you are there, `tests/unit/cli/state-list.test.ts:751-776` needs to go or be rewritten. Its comment — "a bad `parentStack` throws first, so a case planting both cannot tell whether the `parentRegion` check exists" — describes pre-fix behaviour; nothing throws now. And it is vacuous as a pin: its fixture has `Parent` in `us-east-1` but no legacy region-less `Parent`, so `refKey('Parent', '')` matches nothing and `Planted` lands at the root whether or not the second clause exists. The new three-record case at `:653` is what actually catches that mutant. The two changes are the same edit: give this one the legacy `Parent` record it is missing, or delete it and note it is subsumed.

### M5 — required — the new `--json` note under-states the gap it names

`src/cli/commands/state.ts:526-527` and `docs/cli-state.md:177-178` both say `--json` escapes C0 "but not DEL, the C1 range or `U+2028`". Measured pass-through is wider: DEL, `U+0085`, `U+009B`, `U+2028`, **`U+2029`**, **`U+202E`**, **`U+2066`**, `U+200B`, `U+FEFF`. A three-item list reads as the whole gap, so a reader hardening `--tree --json` against it still ships the other line terminator and the Trojan-Source overrides.

This is the same defect class as M2 — a security note the next reader acts on — which is why it is required rather than a nit. Say instead that `--json` is not sanitized at all: `JSON.stringify` escapes only C0, `"` and `\`. I have filed the underlying hole as #3163, so point both the comment and the docs note at it; a public page naming a gap with nothing tracking it was the other half of the problem.

### m7, m8 — out of scope, filed

- **#3163** — `cdkd state list --json` emits the unsanitized class the text views strip. This is the hole M5's note describes.
- **#3164** — `safe()` is an allowlist, not a boundary quoter, so an all-ASCII planted stack name `ProdStack (us-east-1)` prints byte-identical to the genuine `ProdStack` in `us-east-1`, and the `while read` consumer M1 exists to protect still cannot tell them apart. Pre-existing — it worked on this PR's base too — and `displayIdent` is the repo's own answer, so it is a follow-up and not yours. Filing it is the right outcome of M1, not a reason to reopen it.

Both are classified `Session-fit: next`, blocked on this PR merging, because they touch `src/cli/commands/state.ts` and `docs/cli-state.md`.

### m9 — optional

- `tests/unit/cli/state-list-tree.test.ts:164-165` — the comment says the case exercises `settled` sharing. It cannot; what it uniquely catches is an outer loop that stops after the first cycle, which no other case covers. Worth rewording since that is a real and different property.
- `docs/cli-state.md:173` says "control characters … replaced". The mechanism is an ASCII allowlist, so non-ASCII is replaced too, and a value with nothing renderable left becomes `<unrenderable>` — neither is visible in the table.
- The M0 comment says the whole-link rule prevents binding to a legacy record of the same name. An explicit `parentRegion: ""` is a valid string, keeps the link, and keys as `Parent\0` — the same legacy binding. Harmless, since the attacker owns their own link target either way, but the comment over-claims.
- `docs/cli-state.md:107-189` — the `cdkd state list` section is now 83 lines with no `###` subheadings, against `.claude/rules/docs-page-template.md:80`. Pre-existing shape that this PR grew.
- The residual — a legacy `version: 1` row rendering `Resources: 0`, indistinguishable from a real empty stack — is in the PR body and the changelog entry but not on the page. The new table is keyed on "read that failed", which a skipped read is not, so it no longer over-implies; a row saying so would close it.

Push M4 and M5 and I will merge.

nix-tkobayashi:
Thanks for round 2. M4 and M5 are in `e7ae16a`, along with every m9 item.

**M4.** A new case in `tests/unit/cli/state-list.test.ts` plants your shape. The parent is `{ stackName: 'Parent' }` with no region. The child is `Parent~Child` in `us-east-1`, with `parentStack: 'Parent'` and no `parentRegion`. The case asserts the child nests under the legacy parent. Your over-tightening, `typeof state.parentRegion === 'string'` in `linkIsValid`, now fails exactly that case. I deleted the vacuous `parentRegion`-alone case rather than rewrite it: with a legacy `Parent` added, it would duplicate the three-record case that already catches the loosened mutant. The PR body's case table records the swap.

**M5.** The `state.ts` comment now says `--json` is not sanitized at all: `JSON.stringify` escapes only C0, `"`, `\` and lone surrogates. It lists the classes that pass through and points at go-to-k/cdkd#3163. The `docs/cli-state.md` note says the same without a list. It does not link go-to-k/cdkd#3163, because `tests/unit/scripts/docs-site-links.test.ts` refuses issue links on a docs page.

**m9.**
- The disjoint-loop comment in `state-list-tree.test.ts` now names what the case catches: an outer walk that stops after the first loop.
- The docs table now says a character outside printable ASCII is replaced, and a value with nothing printable left shows as `<unrenderable>`.
- The M0 comment now says an absent `parentRegion` is a valid link on purpose. It also admits that an explicit `parentRegion: ""` keys the same way as a legacy record.
- The `cdkd state list` section now has four `###` subheadings.
- The page now states the legacy `version: 1` residual: such a row shows `Resources: 0` with no reason.

Local runs on this head: full unit suite 1011 files, 22849 passed and 1 skipped (22850 total), plus typecheck, lint, build and generated-artifact freshness. The PR body's mutation table is remeasured on `3cf15ae`, whose code this head keeps unchanged (`e7ae16a` only corrects the escape set in one comment and one docs sentence): 30 probes over the two test files (57 tests), including your over-tightening. All but two fail at least one case; the two that pass are the equivalent inner `parentRegion` check and the settled-set omission that only costs time.

go-to-k:
Round 3, reviewed at `e7ae16a`, same 3-axis plus security.

**M4 and M5 are fixed, and every m9 item landed.** I had these measured rather than reasoned. M4: your over-tightening mutant now reds exactly one case, the new `:684` legacy-parent case, and only it; re-adding the case you deleted produces a byte-identical failing set under all four mutants, so the deletion lost no coverage — your call to delete rather than rewrite was right. M5: `JSON.stringify` was probed across the whole code-point range on this repo's Node, and the escape set is exactly C0, `"`, `\` and lone surrogates — both the comment and the docs sentence are now precise in both directions, and the lone-surrogate clause is non-vacuous, since a planted record BODY can carry one where a key segment cannot. The src delta really is comment-only: an AST leaf-token compare across the two heads gives 11453 identical tokens, guard-the-guard confirmed against a one-token mutation.

Two things, then I merge.

### M6 — required — pin `linkIsValid` as a table, not as one more case

`state.parentRegion === undefined` mutated to `== undefined` **survives all 57 tests**, and `lint` does not flag it (no `eqeqeq` rule). That mutant admits `parentRegion: null` as a valid link, so the child binds to a legacy region-less parent of the same name — the exact hazard M0 closed. No test plants `null` as an INPUT for either parent field; the three `parentRegion: null` occurrences in the suite are all expected OUTPUT. The code is correct; the branch is simply unfenced.

I am not asking for one more case, because that is the third round in which mutation has found an unfenced branch in this same three-line expression — M0 was the source bug, M4 was the absent-`parentRegion` arm, and this is the `null` arm. Patching per instance keeps finding the next one. End the class instead:

Drive it from a table over the cross product of `parentStack` × `parentRegion` shapes — at minimum `'Parent'`, absent, `null`, `{toString: null}`, `123`, and `''` — with the expected verdict per pair, and assert the resulting tree. `null` is the shape that matters most and is the one missing today: it is what a JSON writer actually emits for an absent field, where `{toString: null}` is exotic by comparison. A table also makes the next reader add a row rather than reason about which arms are already covered.

If the cross product turns up a pair whose current verdict you think is wrong, say so rather than encoding it — that is the useful outcome of writing it as a table.

### M7 — required — the mutation table cites a commit nobody can fetch

The PR body says the table was remeasured on `3cf15ae` and that `e7ae16a` keeps that code unchanged. `3cf15ae` exists in no pushed ref — `git cat-file` fails on it against `refs/pull/3157/head` and `/merge`, and the PR carries one commit. A reviewer cannot check the provenance of a table the body leans on heavily. Cite `e7ae16a`, or whatever the measurement is re-run on after M6. One line.

### m10 — optional

- `src/cli/commands/state.ts:624-626` — "whoever writes that record chooses their own link target either way" covers the `parentRegion: ""` half but not the absent half: a planted legacy-layout key `cdkd/Prod/state.json` captures any record naming `parentStack: 'Prod'` with no `parentRegion`, and the attacker did not write the captured record. Pre-existing — the base copied `parentStack` with no check at all — and unreachable from cdkd-written state, since every writer sets a non-empty `parentRegion`, so the captured child is itself hand-written. Only the sentence over-generalizes.
- `docs/cli-state.md:183-185` — "the same as a stack that has no resources" over-states: the legacy row also shows `Last Modified: unknown`, which a genuinely empty stack does not.
- `docs/cli-state.md:160` — "Neither view …" now sits behind the new `###`, so its antecedent moved out of the paragraph. "Neither `--long` nor `--tree`" fixes it.
- `docs/cli-state.md:187-189` — the `--json` note describes the gap but gives its reader nothing to do; one clause telling a `--json` consumer to sanitize before rendering in a terminal would.
- `docs/cli-state.md:183-189` — both paragraphs sit under "Unreadable and malformed records", but one is a SKIPPED read and the other is sanitization; the heading is slightly wider than its title.
- `src/cli/commands/state.ts:527` — 92 columns in an otherwise 76-80 column block; the insert was not reflowed, and oxfmt does not reflow comments.
- `state.ts:627` — a dropped parent link is silent, while `--long` announces its degraded reads. `--tree` cannot distinguish "no parent" from "link refused". A design question, not a defect.
- `state.ts:468` — the `Promise.all` fan-out over every stack is unbounded, so at scale a throttle storm now produces many `unknown` rows where it used to produce one hard failure. Pre-existing, and the stderr count keeps it from being silent.

One process note, not a request: two reviewers ran mutation probes in the same tree, and one of them briefly READ the other's mutant through the working copy. They caught it and switched to `git show HEAD:` snapshots, and I confirmed the tree is clean and `state.ts` is byte-identical to `e7ae16a` — so nothing above is measured off a mutated file. Flagging it only because the measurements in this round are the load-bearing part.

Push M6 and M7 and I will merge.

nix-tkobayashi:
Thanks for round 3. M6 and M7 are in 9a598845, along with the m10 wording items. 9a598845 carries the same tree as 37006315: four CI jobs on 37006315 (`check-build-test`, `once-leak-detect`, `pr-content`, `prefix-scope`) were never started (each check run's annotation reads "The job was not started because it repeatedly failed to be acquired (5 attempts)."), and `POST .../actions/runs/34952460300/rerun-failed-jobs` from this account returned 403 "Must have admin rights to Repository.", so I re-committed it unchanged to start a fresh run.

**M6.** `tests/unit/cli/state-list.test.ts` now drives the parent-link rule from a table: six `parentStack` shapes (`'Parent'`, absent, `null`, `{toString: null}`, `123`, `''`) against six `parentRegion` shapes (`'us-east-1'`, absent, `null`, `{toString: null}`, `123`, `''`). Each of the 36 rows plants that pair on a `Child` beside a legacy `Parent` and a regional `Parent`, and asserts where `Child` lands and whether `--tree --json` keeps its `parentStack`. The expected verdicts are written out per row, not derived from `linkIsValid`. Your `== undefined` mutant now fails the two `parentRegion: null` rows with a string `parentStack`.

Writing the table also found one arm it could not see at first: a `parentStack: null` admitted as a link keys as the text `null` and matches no record, so the child lands at the root either way. The fixture therefore also carries stacks literally named `null` and `123` in `us-east-1`, and every row asserts both stay childless. With that, each of these mutants fails at least one row:

| Mutant of `linkIsValid` | Rows that fail |
| --- | --- |
| `parentRegion == undefined` | 2 |
| a `null` `parentStack` accepted | 1 |
| a number `parentStack` accepted | 3 |
| a number `parentRegion` accepted | 2 |
| the string arm of the `parentRegion` check dropped | 4 table rows, plus 4 existing cases |

Writing the table surfaced no verdict I think is wrong, but two rows encode behaviour worth naming rather than implying:
- `parentStack: 'Parent'` with `parentRegion: ''` keeps the link and nests under the LEGACY `Parent`, the shape the M0 comment already calls out.
- `parentStack: ''` with any string or absent region keeps a link that names no record, so `Child` lands at the root with `parentStack: ""` still in the JSON.

**M7.** The PR body now cites 9a598845 for the mutation table, and every row was remeasured on that tree.

**m10.**
- The M0 comment now also covers the absent-`parentRegion` capture by a planted legacy-layout key, and says neither shape is reachable from cdkd-written state.
- The docs legacy row now says it shows `Resources: 0` and `Last Modified: unknown`.
- "Neither view" is now "Neither `--long` nor `--tree`".
- The `--json` note now tells a consumer to sanitize before printing to a terminal.
- The subsection is retitled "Unreadable records and unsafe values".
- The `--json` comment in `state.ts` is reflowed to the block's width.
- Unchanged: the silent dropped link under `--tree`, and the unbounded `Promise.all` fan-out. Both are design questions rather than defects, as you say, and both predate this PR.

Local runs on this tree: the full unit suite (1011 files, 22885 passed, 1 skipped), plus typecheck, lint, build and generated-artifact freshness.

go-to-k:
Round 4, reviewed at `9a598845`.

**M6 and M7 are fixed, and every m10 item landed.** The table is the right shape: 36 distinct rows, an exact cross product with none collapsed, expected verdicts as literals rather than derived from `linkIsValid`, and the observed side classified from the emitted JSON. Your round-3 `== undefined` survivor now reds 2 rows, the round-2 over-tightening still reds 8, and your own mutant table reproduced 5 of 5 exactly. The `null` / `123` decoy stacks are a genuinely good idea — they make a mutant admitting a non-string `parentStack` observable through `refKey`'s interpolation, which no number of extra rows would have done, and I confirmed they are the sole observer of that mutant rather than something other rows lean on.

I also re-derived the body's pre-fix "42 red" against the 36 written verdicts and got exactly 20 table rows plus 22 cases, matching to the unit; confirmed `9a598845` and `37006315` carry the identical tree `08894786`, so the re-commit was a no-op; and checked the two new comment claims under a full enumeration — `withParentInfo` in `deploy-engine.ts` and `import.ts:3101-3103` are the only writers, both set all three fields together, and every region leg terminates non-empty, so "not reachable from cdkd-written state" is true rather than plausible.

Your judgement on the two behaviours the table surfaced was right on both, and I checked the second: `parentStack: ""` retained on a rooted child is consistent with how a Ghost parent is already emitted, so it is not a pinned bug.

Three things, then I merge. M8 is a source defect; M9 and M10 are fences in what this round added.

### M8 — required — a planted `resources` fabricates a count, and a large one kills the listing

`src/cli/commands/state.ts:491` is `Object.keys(state.resources ?? {}).length`. `parseStateBody` deliberately validates nothing inside the root, so `resources` may be any JSON value. Measured on this repo's Node:

| `resources` | `Resources:` shows |
| --- | --- |
| `"abcdef"` | `6` |
| `[1, 2, 3]` | `3` |
| `42` / `true` | `0` |
| `"x".repeat(5_000_000)` | 5,000,000 — 1489 ms, **319 MB RSS** |

The first two rows are a fabricated number presented as a real one, which is what this PR argued against when it chose `resourceCount: null` over a made-up `0`. The last row is the headline promise failing on its own terms: at a few tens of MB of planted body this is an OOM, it is outside `readOrFailure`'s `try`, and an OOM is not catchable anyway — so one planted record still loses the whole listing.

This is the same class you already closed two lines below. `isoTimestampOrNull` exists because `toISOString()` threw past the guard on a planted `lastModified`; `resources` is its sibling and was left unguarded. The docs table lists a malformed `lastModified` row and no `resources` row, so the page implies a completeness it does not have.

It went unseen for four rounds because the suite cannot express the input: `tests/unit/cli/state-list.test.ts:59` types the mock's `resources` as `Record<string, unknown>`. Loosening that type is part of the fix, not incidental to it.

Two things to decide rather than assume, since I would rather you pick than have me prescribe:

- **What the row shows.** `Resources:` prints `String(detail.resourceCount)` whenever `stateReadError` is `null`, so returning `null` without a reason renders the literal text `Resources: null`. A malformed `resources` is not a read failure either, so reusing `stateReadError` is not obviously right.
- **Whether `--json` reports it**, and under which field, given `resourceCount: null` already means "the record read failed".

Yes, the expression predates this PR. I am asking for it here on the reasoning you applied to `lastModified` in round 1: it sits inside the row this PR rewrote, it defeats the guarantee this PR makes, and the code is warm.

### M9 — required — the table's `kept` column is a presence check, so behaviour (b) is not actually pinned

In the new table, `kept` is `child[0]!.parentStack !== null`. That asserts a value is present, not which value. A mutant rewriting a retained `''` to `'FORGED'` survives all 93 tests — so the second behaviour your reply named explicitly, `parentStack: ''` surviving into `--tree --json` on a rooted child, is described but not fenced. Assert the retained VALUE per row.

While you are in it: nothing fences the product size. Shrinking `regionColumns` or the stack column silently shrinks the cross product and the suite still goes green, which is the failure mode a table is supposed to remove. An `expect(rows).toHaveLength(36)` beside the definition costs one line.

### M10 — required — no `--tree` row plants `getState` resolving `null`

`typeof state?.parentStack === 'string'` mutated to `state!.parentStack` survives all 93. It is reachable: a `--tree` row whose `getState` resolves `null` then throws OUTSIDE the per-row `try`, and `--tree --json` emits an empty string — precisely the whole-view collapse the comment at `state.ts:608-617` promises to prevent. The only `mockResolvedValue(null)` in the suite is the `--long` case at `tests/unit/cli/state-list.test.ts:262`; no `--tree` case has one.

The current code is correct and the base had the identical exposure, so this is a fence, not a bug. I am asking for it because the comment right above the expression makes that promise, and the promise is the one thing in this PR with no test behind it.

### m11 — optional

- `src/cli/commands/state-list-tree.ts:51` — `refKey`'s single-`\0` join is ambiguous: `name="a\0b", region="c"` and `name="a", region="b\0c"` collide, silently dropping one record from `--tree`. Two reviewers raised it independently. Pre-existing and barely reachable. I have recorded it on #3164 rather than asking for it here.
- The PR title names `--long` only, while the change also covers `--tree` and the default listing.
- The "What changed" bullet says the parent link is "kept or dropped whole" without the absent-`parentRegion` exception, which is load-bearing and is carried correctly in the case table and the code comment.
- `docs/cli-state.md:287-295` now runs nine lines with a nested em-dash clause, against the page template's ~six.
- Keep `drops a string parentRegion too when the parentStack beside it is not a string` even though the table subsumes several of its neighbours — it is the unique killer of both spread-line mutants.

Separately, not a request: #3163's body cites `:641` for the `--tree --json` sink, which is `:650` at this head. I will correct it on the issue.

Push M8, M9 and M10 and I will merge.

nix-tkobayashi:
Thanks for round 4. M8, M9 and M10 are in 92df86e4, along with three of the m11 items.

**M8.** I took both decisions this way:

- **What the row shows.** A `resources` that is present and not a JSON object now shows ``Resources: unknown (resources is not a JSON object; run `cdkd state show --json` to see the record)``. `resourceCountOrNull` counts a JSON object's keys and keeps an absent or `null` bag at `0`, which `docs/cli-state.md` already documented as tolerated. Anything else returns `null`, so `Object.keys` never walks a string or a list.
- **`--json`.** Such a row reports `resourceCount: null` with that reason in `stateReadError`. I reused the field rather than adding a second one. That keeps one rule for a consumer: `resourceCount` is `null` exactly when `stateReadError` is set, so a single field says why a count is missing. It also closes the `Resources: null` gap you pointed out without a new renderer branch, since the text row already prints the reason whenever `stateReadError` is set. A separate field would work too, but every consumer would then check two fields to explain one `null`. The cost is that `stateReadError` now also covers a record that was read but is malformed. Its doc comment and the docs table say so, and the stderr warning counts the row.

`42` and `true`, which used to show `0`, now show unknown too, since `0` was as made up as `6`. The mock's `resources` is typed `unknown`, and a new case plants `"abcdef"`, `[1, 2, 3]`, `42`, `true` and `null` beside a healthy two-resource stack, asserting text and `--json` per row. The docs malformed-values table has the new row, and the changelog entry names it.

**M9.** The table's second column is now the exact `parentStack` value `--tree --json` emits for `Child`: `'Parent'` or `''` on kept rows, `null` elsewhere. A new case beside the table asserts 36 rows and 36 distinct pairs.

**M10.** A new `--tree` case resolves `getState` to `null` for one stack beside a healthy one, and asserts both rows in text and in `--json`.

Each of these mutants now fails, one mutation per probe, with the source and test files restored byte-exact after each:

| Mutant | Fails |
| --- | --- |
| `resourceCountOrNull` without the `typeof` check | the new `resources` case |
| `resourceCountOrNull` without the `Array.isArray` check | the new `resources` case |
| the `null`-bag check dropped, so `Object.keys(null)` throws and aborts the listing | the new `resources` case |
| the old `Object.keys(state.resources ?? {}).length` restored | the new `resources` case |
| a malformed `resources` with no `stateReadError` | the new `resources` case |
| the generic record-read reason used for it | the new `resources` case |
| a retained `''` rewritten to `'FORGED'` (your M9 mutant) | 3 table rows, the three kept `''` rows |
| the `''` region column dropped from the table | the 6 x 6 size check |
| the `123` stack row dropped from the table | the 6 x 6 size check |
| `typeof state?.parentStack` changed to `typeof state!.parentStack` (your M10 mutant) | the new `--tree` no-record case |

**m11.**
- The PR title now names all three views.
- The "What changed" bullet now carries the absent-`parentRegion` exception.
- `docs/cli-state.md`'s root-malformed paragraph is split in two, and the `--long` exception moved out of the nested dash clause into its own sentence.
- Unchanged: `refKey`'s `\0` collision, which you recorded on go-to-k/cdkd#3164. I also kept `drops a string parentRegion too when the parentStack beside it is not a string`, as you asked.

The full mutation table in the PR body was remeasured on 92df86e4. Local runs on this tree: the full unit suite (1011 files, 22888 passed, 1 skipped), plus typecheck, lint, build and generated-artifact freshness.

nix-tkobayashi:
One follow-up to my reply above, pushed as 0a90edbe before you merge. A last Codex pass run against your review checklist found four gaps in what round 4 added, all small:

- **The warning text.** A malformed `resources` now joins the stderr count, but the warning still said "their rows show which read failed", and neither read failed. It now reads `N of M stack(s) could not be fully read or counted; their rows say why.` The code comment above it says the same, and the two existing warning cases expect the new text.
- **Two M8 arms had no test.** The `resources` case now also plants a record with no `resources` key at all, which shows `0`. It also asserts the warning counts exactly the four malformed rows (`4 of 7`). Dropping the absent-bag arm of `resourceCountOrNull` now fails that case.
- **Docs parity.** The malformed-values table said "not a JSON object", which reads as covering `null`. It now says "neither a JSON object nor `null`" and notes that an absent or `null` bag counts as `0`. The parent-link row now names the absent-`parentRegion` exception, and a new row covers a non-string `parentLogicalId` on a valid link. The reason paragraph notes that a legacy row's lock reason names no command. `.claude/rules/layout-cli.md` still described `buildStackTree` as a self-link defense only; it now names parent-loop root promotion, in one byte fewer, because the rule files loaded for `src/cli/commands/scrub.ts` already sit exactly at that path's 119000 B cap.
- **A test comment.** Your M9 size check fences a shrunk `regionColumns` list or a dropped stack row, but my comment also credited it with catching a short verdict row, which it does not. That row still yields six entries and fails its own assertion. The comment now says so.

No change to `linkIsValid` or the table's verdicts, and the full mutation table in the PR body was remeasured on 0a90edbe with the new absent-bag row. Local runs on this tree: the full unit suite (1011 files, 22888 passed, 1 skipped), plus typecheck, lint, build and generated-artifact freshness.

go-to-k:
Round 5, reviewed at `0a90edbe`. **M8, M9 and M10 are all fixed, and all four axes recommend merge.**

Your 10-row mutant table reproduced 10 of 10 on HEAD with no survivors, counts matching exactly. The M8 guard is measured closed rather than inferred: `resources: "x".repeat(5e6)` costs 708 ms and 586 MB through the old expression and returns `null` in 0.00 ms through `resourceCountOrNull`, and `Object.keys` is provably unreachable for a string or a list. The `resources` case asserts exact rendered text AND `[resourceCount, stateReadError]` per row, and pins that an absent or `null` bag is deliberately NOT counted. M9's second column is a real value check — the `FORGED` mutant kills 3 rows — and the `toHaveLength(36)` fence fires on a dropped column; the distinctness assertion is not trivially true, since duplicating a region column keeps the length at 36 and reds the `Set` assertion alone. M10's `--tree` case asserts the whole rendered array, which is stronger than containment and is what kills the `state!` mutant.

**On the `stateReadError` decision: your call stands, and I am closing it.** I asked you to pick and you picked with reasons. A reviewer argued the field name now lies for one of its two meanings, and that a consumer written as `if (row.stateReadError) retryRead()` would loop forever on a malformed row. That is a fair point about the name, but the two causes carry distinct fixed reason strings in both output modes, nothing inside cdkd branches on the field, and your "one field explains one `null`" invariant holds on all four branches. A `stateReadErrorKind` enum is the better long-term shape for machine consumers, and I have put it on #3163 rather than reopening a decision you justified.

**Provenance is settled, no action needed.** Two reviewers independently confirmed that `92df86e4..HEAD` changes only the warning string, its two assertions, the `Absent` column and comment prose — no predicate or branch — so no mutation row moves, and the body already cites `0a90edbe`. Your reply's `92df86e4` was the misleading part, not the measurement; your follow-up naming `0a90edbe` closes it, and everything below was reviewed on that head.

Your own Codex pass caught four of these before I did — the warning wording, the absent-bag arm, the `null` half of the docs row, and the M9 comment over-crediting the size check. That is the second round running where your self-review found a real gap rather than restating the change, and it is why the three items left are text rather than behaviour.

Three text corrections and I merge. All three are one-line edits, and this is the last set.

### M11 — required — the `--long --json` contract table's key column no longer covers what the field means

`docs/cli-state.md:165-168` is keyed **"Read that failed"**, and its state-record row says `resourceCount: null`, `stateReadError` set. After this round's widening, `stateReadError` set no longer implies a failed read. The malformed-`resources` case IS documented correctly, fifteen lines below at `:183` — but a consumer reads the contract table as the field's definition and stops there. Signpost it in the key column, something like "Read that failed, or a `resources` that could not be counted (see below)".

### M12 — required — the changelog entry is the one statement of four that omits the widening, and mis-states the tolerated shapes

`changelog.d/entries/2026-09-15-3069-state-list-long-row-tolerance.md` says a `resources` "that is not a JSON object shows an unknown count instead of one counted per character or element". Two problems: it is the `--json` side that is missing — the entry still describes `stateReadError` as being for "the failed side" and never says a malformed record sets it — and "not a JSON object" is the wording you tightened in `docs/cli-state.md` this round to "neither a JSON object nor `null`", because an absent or `null` bag still counts as `0`. As written the entry promises an unknown count for `null`, which the code does not do. You have roughly 180 characters of headroom under the 2000 cap.

### M13 — required — the live-run table quotes warning text the binary no longer emits

The PR body's first live-run table says stderr warns `1 of 2 stack(s) could not be fully read`. This round changed that string to `... could not be fully read or counted; their rows say why.`. The live runs were real and the rest of the row still holds, so quote the current text and say the wording changed after the run, rather than re-running.

### m14 — optional

- `docs/cli-state.md:285-286` — "shows an unknown count **there** instead": the nearest antecedent is `--show-nested`, which shows no count. Name `cdkd state list --long`.
- The PR title says "when one stack's **record** is unreadable or malformed", but the plain listing reads no record — its fix is a planted S3 key.
- `.claude/rules/layout-cli.md:19` — the user-facing `state list --tree` prose in the same file still names only orphan-child promotion, so the file now describes one behaviour twice, once incompletely. `--long` is absent from that file entirely, which predates this PR.
- `tests/unit/cli/state-list.test.ts:632` — the `--long --json` run's own `warn` call is unasserted, since the assertion sits before it.
- A record that vanishes between `listStacks` and `getState` renders `Resources: 0` with no reason and escapes the warning — the same made-up-number class M8 closed, for a different cause. Pre-existing.

### Filed, not yours

**#3172** — the same unguarded `resources` walk in three sibling sites this PR does not touch. The serious one is `src/cli/commands/destroy-runner.ts:480`: `Object.keys(42)` is `[]`, so a record whose `resources` is a number reads as an empty stack and `cdkd destroy` takes the fast path that, in that code's own words, "deletes state.json with no confirmation at all". `cdkd state resources` and `cdkd state show`'s TEXT path fabricate rows from a string, and `state show` costs 2.1 s and 1372 MB on the 5e6-char body — which matters because it is the command your new reason string sends the user to, though the `--json` form it names is the safe one. Filed at `Severity: high`, blocked on this PR because the guard to reuse is `resourceCountOrNull`, which lands here. Nothing owed in this PR.

Push M11, M12 and M13 and I will merge.

nix-tkobayashi:
Thanks for round 5. M11, M12 and M13 are in 9d0b3dc6, along with two of the m14 items. The delta from 0a90edbe is `docs/cli-state.md` and the changelog entry only, with no source or test file, so no mutation row moves and the PR body's table still cites 0a90edbe.

**M11.** The contract table's state-record row is now keyed "state record, or a `resources` that could not be counted (see below)", so a reader who stops at the table sees that `stateReadError` covers both.

**M12.** The changelog entry now says a `resources` that is "neither a JSON object nor `null`" shows an unknown count, and that under `--json` it sets `resourceCount: null` with its own `stateReadError` reason. The entry is now 1917 characters, under the 2000 cap.

**M13.** The first live-run row now quotes the warning as it was emitted then, `1 of 2 stack(s) could not be fully read; their rows show which read failed.`, followed by the current text, `1 of 2 stack(s) could not be fully read or counted; their rows say why.`, and says the wording changed after that run. I did not re-run it.

**m14.**
- `docs/cli-state.md`'s "shows an unknown count there" now names `cdkd state list --long`.
- The PR title now says "state record or S3 key", since the plain listing's fix is about a planted key rather than a record.
- Unchanged: `.claude/rules/layout-cli.md:19`. The rule files loaded for `src/cli/commands/scrub.ts` sit one byte under that path's 119000 B cap, so completing that sentence means moving text into a new satellite, which I would rather not fold into this PR.
- Unchanged: the unasserted `warn` of the `--long --json` run. Adding it is a test change that would move the mutation table, and the text run's warning is already pinned in the same case.
- Unchanged: a record that vanishes between `listStacks` and `getState`, which renders `Resources: 0` with no reason. It predates this PR, as you note.

Thanks for filing go-to-k/cdkd#3172 and for recording the `stateReadErrorKind` idea on go-to-k/cdkd#3163.

Local runs on this tree: the full unit suite (1011 files, 22888 passed, 1 skipped), plus typecheck, lint, build and generated-artifact freshness.

go-to-k:
Merged as 4dd43400. Thank you — six rounds and every one of them moved the PR forward.

Two things stood out. You took the review asks as classes rather than instances: when I pointed out that mutation kept finding one more unfenced branch in `linkIsValid`, you replaced the per-case pins with a 6x6 table whose verdicts are literals, and it killed every non-equivalent mutant on the first try. And your own Codex pass twice found real gaps before I did — the warning wording, the absent-bag arm, the `null` half of the docs row — which is why the last two rounds were text rather than behaviour.

The M8 guard is the one I am most glad we did here. Before the merge I live-tested the built binary against the state bucket with four planted records: the `1e300` row renders `Last Modified: unknown`, the malformed body's canary reached neither stdout nor stderr, and `resources: "abcdef"` shows `Resources: unknown (...)` rather than `6`, with the row keeping its real timestamp. Exit 0 with all four rows and `2 of 4 stack(s) could not be fully read or counted` on stderr. All planted keys and their versions were deleted afterwards, zero remaining.

Follow-ups from this review are filed and none are yours: go-to-k/cdkd#3163 (`--json` is unsanitized, plus the `stateReadErrorKind` discriminator), go-to-k/cdkd#3164 (the display-boundary spoof and `refKey`'s ambiguity), and go-to-k/cdkd#3172 (the same unguarded `resources` walk in `cdkd destroy`, `state show` and `state resources` — the destroy one deletes `state.json` unconfirmed, filed at high severity). All three were blocked on this PR holding `src/cli/commands/state.ts`, so they are actionable now.

Issue #2786, opened 2026-09-08
test(unit): three prose claims in check-scope-checker-inputs and .markgate.yml drift from the code beside them, and nothing fences the class


go-to-k:
## Summary

Three prose inaccuracies left behind by go-to-k/cdkd#2758, plus one fragility. None changes behaviour; all sit in comments beside the code they describe, and nothing in the suite watches any of them. They are filed together rather than fixed in that PR because they are the fifth consecutive round of the SAME class (see "Why this is one issue" below).

### 1. `carveOutFenceProblems`' JSDoc claims an order the loop does not follow

`tests/unit/scripts/check-scope-checker-inputs.test.ts:258` reads "Returns one message per violated rule, in the order the loop applies them:" and then lists reason / exists / ignored / optional-membership. Two things are off:

- Bullets 2 (exists) and 3 (ignored) are the two arms of one `if/else` whose condition tests `optional.has(rel)` FIRST, so bullet 3's statement precedes bullet 2's.
- The output is **entry-major, not rule-major**. Measured against a copy of the function with an optional entry inserted first: `['b/opt.json ... does not ignore it ...', 'a/gone.txt no longer exists']`. And "one message per violated rule" is really one per (entry, rule) pair — two reason-less entries yield two messages.

The reason rule IS applied first for every entry, unconditionally (no `continue`), so that half of the claim holds. Suggested wording: one message per violated entry-rule pair, entries in map order; within an entry the reason rule runs first, then either the optional-ignored or the exists rule; the optional-membership rule runs last over the whole set.

### 2. The same JSDoc states a universal that go-to-k/cdkd#2758 already retreated from in its changelog

`check-scope-checker-inputs.test.ts:219-221`: "The top-level file is the specific one demanded because it is **the only one inside** `.markgate.yml`'s `check` include". False as written — `check.include` carries `src/**`, `tests/**`, `scripts/**`, `docs/**`, `.claude/skills/**`, `.claude/rules/**`, `.claude/agents/**` and `.claude/hooks/**`, each of which would cover a nested `.gitignore` beneath it. It is true only for the sources that can win for `.claude/settings.local.json`, because no include entry covers a bare `.claude/`.

`gitIgnores` refuses ANY nested source, so a future optional entry under `docs/` would be refused with a rationale that does not hold for it. The changelog paragraph was scoped to `.claude/.gitignore` during that PR's round 5; this copy was not.

[...]

Issue #2779, opened 2026-09-08
chore(process): the changelog is the only file every PR must edit, and it conflicts on every parallel-lane rebase


nix-tkobayashi:
Filed as something to weigh, not as a proposal to adopt: both options below are workflow changes that are yours to make, and either may lose more than it saves. What I can contribute is the measurement and what the tree says about each option. Observed here are the conflicts, their automatic resolution and their downstream cost; the merge-behaviour claims below were checked against the tree rather than observed.

go-to-k/cdkd#1837 named this class ("the changelog conflicts on nearly every parallel-lane rebase") and explicitly proposes a per-entry file layout among its suggested directions. Its own open work is now one duplicate cluster needing a content merge, and the volume half went to go-to-k/cdkd#2552, which is closed. The layout decision has no issue of its own, which is why this is separate rather than a comment there.

## The conflict rate has not changed

Four rebases across two PRs in one day (go-to-k/cdkd#2742 and go-to-k/cdkd#2758), with main between one and eight commits ahead each time. All four conflicted, all four on this file alone. Nothing else in either branch's diff ever collided. Both branches were single flattened commits, so this is not the multi-commit replay shape the issue body describes: it is two lanes inserting a bullet at the same anchor.

## What it costs in a working session

The conflict itself did not need a human. In this session my local babysitter automatically resolved all four, preserving the incoming HEAD entries and placing this branch's entry first. That automation is mine, not this repo's, so treat it as an observation rather than as evidence about the repo. The cost is downstream of the rebase itself: a rebase brings in files covered by `check`, which invalidates its content hash, so the full unit suite has to run again before the marker can be re-set and the commit amended and pushed. Each push then starts a fresh CI cycle.

Two of the four rebases followed a `DIRTY` merge state, where CI does not run at all until it is resolved, so those were forced. The other two were my automation rebasing whenever main moved, which is my choice rather than a requirement: a branch merely behind main still merges. So the numbers below are rebase-associated cost as observed. The two conflict-forced rebases are the potentially avoidable portion: A addresses those insertion conflicts, while B helps only when an entry on one side or the other would be exempt. Which of the seven runs and how much of the elapsed time those two account for is not separated here, since the three repeats were caused by host load rather than by any particular rebase.

Measured across the two PRs in one session:

| what | count |
| --- | --- |
| rebases, all conflicting on this file alone | 4 |
| full-suite runs, including three host-load retries | 7 |
| minutes per full-suite run | 6 to 11 |

[...]

PR #3158, opened 2026-09-15
chore(hooks,tests): a failed cd no longer moves main-tree-edit-gate's base, and five agent-tooling fences read both directions

### LINKED ISSUE #2458: test(work-issues): both negative pins on the withdrawn branch condition are case- and wrap-blind, so neither can match
Dup-check: searched open issues for `work-issues-launch-mode`, `hard wrap regex pin` and `negative assertion vacuous`. No hits. Nearest neighbours are the carve-out/negation family recently worked in go-to-k/cdkd#2448 and go-to-k/cdkd#2451, which are about a checker's EXEMPTION being too wide; this is the opposite direction -- a negative pin that cannot match its subject at all, so it never had a chance to be too wide.

Session-fit: next (not this session) — the fix is a normalization pass shared by every prose pin in that suite, plus a decision about how the HISTORICAL quote stays legal while a prescriptive re-introduction is flagged; that is the same design question go-to-k/cdkd#2448 / go-to-k/cdkd#2451 just spent three rounds on, and it wants its own review rather than riding a retro. Verification for the next session: `vp test run tests/unit/scripts/work-issues-launch-mode.test.ts` plus a planted re-introduction in each of the three shapes below (wrapped, unwrapped, reworded) — runnable from any fresh checkout, no fixture and no AWS.
Severity: low — the contract still has one LIVE guard. The third assertion in the same test is POSITIVE (`toMatch(/git fetch origin && git switch -c <branch> origin\/main/)`) and does pin the unconditional recipe. What is missing is detection of a condition re-introduced BESIDE it, so the failure mode is a silent gap in a fence, not a wrong instruction shipped to users.
Effort: small (S) — one reading helper and three pinned shapes; no src, no fixture, no new suite.
Estimate: ~30-60 min — deciding the quote-versus-prescription rule is what eats it, not the regex.

## The two pins cannot match their subject

`tests/unit/scripts/work-issues-launch-mode.test.ts:457-481` guards contract point (c) — section 5 branches in place UNCONDITIONALLY — with two NEGATIVE assertions over `references/implement.md`. Both are defeated by properties the file has everywhere.

**Pin 1 (`:470`) is case-bound.** It greps `/If the branch here is detached, or its PR has already merged/` with a capital `I`. The sentence in `implement.md` is a QUOTE of the withdrawn condition and is lowercase, so the pattern has matched zero times since it was written.

**Both pins are WRAP-blind, which is the more general problem.** `implement.md` is hard-wrapped at ~78 columns, and the quote currently straddles a line break:

```
what the evidence bought: the rule was CONDITIONAL until go-to-k/cdkd#2417 ("if
the branch here is detached, or its PR has already merged") and the condition
```

Measured on the current file:

```
exact capital ........................  0 matches
exact lower ..........................  0 matches
whitespace-flexible, case-insensitive .  1 match
```

Pin 2 (`:478`) is case-INSENSITIVE, which is better, but its `[^.\n]{0,60}` gap explicitly excludes newlines, so a reworded condition that happens to wrap is equally invisible. Any re-introduction lands wrapped roughly one time in N by luck alone, because the file's own wrapping is what decides.

## Why it is worth fixing rather than deleting

The negative direction is the one the test was written for -- its own comment says "its survival is the actual regression". A pin that cannot fail reads as caution and is a gap, which is the shape go-to-k/cdkd#2451 named. Deleting the pins would leave only the positive recipe assertion, which passes just as happily with a condition sitting next to it.

## Direction

Normalize before matching -- read the file through a helper that collapses runs of whitespace (including newlines) to a single space, and run every PROSE pin in this suite against the normalized text while structural pins (the fenced-block equality checks) keep reading the raw file. Then pin all three shapes as planted-mutation cases so the fix is probed rather than asserted: the exact sentence unwrapped, the exact sentence wrapped, and one rewording.

The second half is the design question: `implement.md` deliberately QUOTES the withdrawn condition as history ("the rule was CONDITIONAL until go-to-k/cdkd#2417 (...)"), so a normalized pin would then fire on the file's own historical note. Whatever rule separates a quote from a prescription here should be the same one go-to-k/cdkd#2448 and go-to-k/cdkd#2451 settled on, not a third invention.

## Found by

A `pr-code-reviewer` pass on the 2026-09-02 work-issues retro flagged the case-bound half as pre-existing vacuity; the wrap half was measured while confirming it.

### LINKED ISSUE #2681: chore(hooks): pr-review-gate says markgate verify enforces the sha binding, but the digest cannot see an unrewritten sentinel
`.claude/hooks/pr-review-gate.sh:579-587` explains the sentinel comparison like
this:

```sh
# Also verify the sentinel file's content matches the PR's HEAD sha.
# markgate verify already enforces this via the digest, but reading
# the sentinel directly lets the error message name the mismatch
# explicitly ("marker bound to <other-sha>, PR is at <current-sha>")
# rather than the generic "(digest differs)" markgate emits.
```

The second sentence is wrong, and it matters because it demotes the ONLY check
that actually enforces the binding to a cosmetic one. `markgate verify` compares
a DIGEST of the gate's scope. The sentinel is in that scope, so REWRITING it
stales the marker — but a sentinel nobody rewrote keeps its digest, whatever the
branch or the PR moved to. So `verify` says `match` for a sentinel naming a
different commit entirely.

Measured 2026-09-05 on this repo, in an IN-PLACE `/work-issues` run where lanes
share one worktree (markgate state and the gitignored root sentinels both resolve
under `git rev-parse --git-dir`, so a later lane inherits the earlier lane's):

```
pr-review sentinel: 4c634eeeaf9aa988dcdc6f207647e15bd5382420   # a MERGED lane's tip
this branch HEAD:   c76f11dfcc63a866ca92ad7cafa9d13a0faa5f2e
PR 2680 headRefOid: c76f11dfcc63a866ca92ad7cafa9d13a0faa5f2e
markgate verify pr-review rc=0
```

The gate itself is CORRECT and blocks — line 589 requires
`[ "$recorded_sha" = "$head_sha" ]` in addition to `status -eq 0`. Nothing is
currently fail-open, which is why this is a comment defect rather than a bug.

The risk is a future edit. A reader who believes the comment sees the
`recorded_sha` comparison as duplicated work whose only value is a nicer error
string, and deleting it looks like a safe simplification. It is not: it is the
whole enforcement. This is the same shape as the sentinel-binding fail-open filed
as go-to-k/cdkd#2679 for `integ-broad`, reached from the other direction — there
the gate really is bound to nothing but the sentinel; here the gate is fine and
the EXPLANATION is what would mislead the next editor.

Suggested fix: replace the second sentence with what the measurement shows — the
digest cannot see a sentinel that was never rewritten, so this comparison is the
binding's actual enforcement and must not be removed; the explicit message is a
bonus, not the reason. Consider a case in
`.claude/hooks/pr-review-gate.test.sh` that pins it: a fresh marker plus a
sentinel holding a FOREIGN sha must still BLOCK. Grep for whether such a case
exists before adding one.

Dup-check: searched open issues for "pr-review sentinel", "markgate-pr-review-sha"
and "pr-review-gate comment" -- go-to-k/cdkd#2336 (the test harness mutating the
LIVE sentinel and its double EXIT trap) and go-to-k/cdkd#2638 (flattening erasing
the fix-back up-bias) are the nearest and both are CLOSED; go-to-k/cdkd#2679 is
the `integ-broad` binding, a different gate and the opposite polarity. Grepped
`.claude/hooks/pr-review-gate.sh` and `.markgate.yml` for the binding, and
`.claude/skills/work-issues/references/verify.md` (which already tells a run to
compare the sentinel to HEAD by hand -- that prose is what caught this).
Session-fit: next (not this session) -- nothing is fail-open today, and the
session that found it is a retro whose PR is under review with a diff that must
not change; adding a hook file to it would restale the markers and invalidate the
in-flight review.
Severity: low -- no current mis-verdict. The cost is entirely prospective: a
future simplification that trusts the comment turns the pr-review gate fail-open
for exactly the case it exists to catch (a marker set on a different sha).
Effort: small (S) -- one comment rewritten, plus one hook-harness case if the
grep shows none exists.
Estimate: ~30 m -- most of it checking whether the foreign-sha case is already
pinned in `pr-review-gate.test.sh`, and running the whole hook harness rather
than one suite, since a hooks edit re-triggers the path-filtered workflow.

### LINKED ISSUE #2683: test(work-issues): three residual gaps in the section 2 probe fence — prose continuations, invisible floor carriers, and semantic siblings
The `work-issues section 2 worktree probe` fence added in go-to-k/cdkd#2680
(`tests/unit/scripts/work-issues-launch-mode.test.ts`) pins that §2's
per-worktree probe ranges over `origin/main...HEAD` rather than reading one
commit. It went through six drafts, each of which was defeated by a shape the
previous round had not imagined, and the last two review rounds surfaced three
residual gaps that were deliberately NOT fixed there — the PR was already six
review rounds deep and each of these is a widening rather than a defect in what
it currently claims.

All three were MEASURED against the shipped fence, not reasoned about.

## 1. The continuation join is inert outside a fence

`commandUnits` joins `\`-terminated lines only across ADJACENT picked lines.
Inside a fence every line is picked, so a wrapped command joins correctly. In
PROSE the continuation line usually carries no `git` token, so it is never
picked and never joined:

```
git show \
  --stat HEAD
```

written as two prose lines ESCAPES the ban; the same thing inside a fence is
caught. The shipped comment says as much, so this is a stated bound rather than
a false claim — but it is still a hole.

## 2. Floor carriers that render invisibly

The floor requires `show --stat HEAD` to survive in PROSE — outside every fence,
not a command, not a comment, not indented, and (since go-to-k/cdkd#2680) not
inside an HTML comment. Four carriers still satisfy it while rendering
invisible or near-invisible, so the rationale paragraph could be deleted and the
floor would still pass:

- a fence INFO STRING (```` ```text show --stat HEAD ````)
- text on the same line after a closing fence
- an HTML attribute (`<div title="show --stat HEAD">`)
- an image alt or link-reference title

Same class as the `<!-- … -->` carrier that PR closed; the fix is presumably to
normalise the doc to rendered text before the floor reads it, rather than
enumerating carriers one at a time — which is the mistake that produced six
drafts.

## 3. Semantic siblings of the withdrawn command

`SINGLE_COMMIT` bans `git … show … --stat`-family reads. These read a single
commit just as wrongly and are not matched:

- `git log -1 --stat`
- `git show --raw`
- `git show --summary`
- `git show --compact-summary`

None is a plausible ACCIDENT — a lane would have to reach for an unusual
spelling — which is why it is low rather than medium.

## Not in scope

The fence's positive half (exactly one peer-worktree probe, and it ranges) is
unaffected by all three: a probe using any of these spellings still fails the
range assertion if it REPLACES the ranged one. These gaps matter only for a
probe ADDED beside it, or for the floor.

Dup-check: searched open issues for "work-issues probe fence", "show --stat"
and "launch-mode test" — no hits; go-to-k/cdkd#2679 (integ-broad sentinel
binding) and go-to-k/cdkd#2681 (pr-review-gate comment) are the other two
follow-ups from the same retro and are unrelated mechanisms. Grepped
`tests/unit/scripts/work-issues-launch-mode.test.ts` for the three predicates
and re-ran each shape against the shipped fence before filing.
Session-fit: next (not this session) -- go-to-k/cdkd#2680 is a retrospective PR
that already absorbed six review rounds on this one fence, and each item here is
a WIDENING of what it claims rather than a defect in it; item 2 in particular
wants a rendered-text normaliser instead of another carrier enumeration, which
is a design call rather than a patch.
Severity: low -- nothing user-facing, and nothing that lets the ORIGINAL defect
back in silently: replacing the ranged probe still reds the range assertion. The
exposure is a probe added alongside it in an unusual spelling, or a rationale
paragraph deleted while an invisible carrier keeps the floor green.
Effort: small (S) -- items 1 and 3 are a predicate widening each with their own
probes; item 2 is the one that could grow, since doing it properly means
normalising to rendered text rather than adding a fourth carrier rule.
Estimate: ~1-2 h -- most of it item 2's design call and re-running the full
evasion matrix, which is the part that has repeatedly found the next gap.

### LINKED ISSUE #2684: chore(hooks): main-tree-edit-gate advances its base on a cd that FAILS, so the write after it escapes the protected tree
`main-tree-edit-gate` resolves a write target against a running base that it
advances on every `cd` segment it can read. It does not ask whether the `cd`
would SUCCEED, so a `cd` into a directory that does not exist moves the base
anyway and the write that follows resolves outside the protected tree.

Bash does the opposite: a failed `cd` leaves the shell where it was, so the
write really does land on the tracked file.

## Reproduction

Measured on a fixture repo on `main`, payload cwd = the main checkout, with the
tracked file `docs/_generated/ledger.tsv` present:

```
cd /nonexistent 2>/dev/null ; echo POISON > docs/_generated/ledger.tsv
```

- real bash: cwd stays in the main tree, the tracked file **is overwritten**
- gate: **rc=0** (allowed)
- expected: rc=2

Two more spellings of the same defect, same verdict:

```
cd /nonexistent 2>/dev/null ; echo POISON | tee docs/_generated/ledger.tsv
cd /nonexistent 2>/dev/null ; echo POISON | tee -a docs/_generated/ledger.tsv
```

## Not a regression

`origin/main`'s hook answers rc=0 on all three as well, measured side by side
during go-to-k/cdkd#2650, where the differential oracle
(`.claude/hooks/main-tree-edit-oracle.test.sh`) reported the identical
fail-open count for both revisions. It is inherited, which is why that PR did
not fix it: its own tolerance is 0 and its corpus deliberately excludes these
shapes rather than buying them off with a nonzero number.

## Why it is worth fixing

The oracle makes the defect cheap to confirm and cheap to re-check, and the
shape is not exotic: `cd <dir that may not exist> ; <write>` is what a
scaffolding or cleanup command looks like when the directory has already been
removed. A gate that silently steps aside there is the failure mode this hook
exists to prevent.

## Sketch of a fix

The base should only advance for a path the resolver already treats as
readable, a notion `gate_target_dir_strict` carries. Applying the same
readability test before advancing the running base in the ordered walk would
close it, and the oracle corpus gains a `cd /nonexistent` lead row to fence it.
The risk to weigh is a legitimate `cd` into a directory created earlier in the
SAME command, which the walk cannot see; that case must keep working, so the
test has to cover both polarities.

Session-fit: next (not this session) — go-to-k/cdkd#2650 is already a five-round
review of this hook; adding an inherited-defect fix to it would mix a
regression-repair PR with a behaviour change and enlarge a diff that has
already needed four reviewers.
Severity: medium — a tracked file in the main checkout can be overwritten while
on `main`, but only via a `cd` that fails, so it is not on the common path.
Effort: small (S) — one readability test in the ordered walk plus corpus rows.
Estimate: ~1-2 h — most of it spent on the both-polarity test for a directory
created earlier in the same command.
Dup-check: searched open and closed issues for "cd nonexistent", "failed cd",
"main-tree base" on 2026-09-06; nothing filed.

### LINKED ISSUE #2732: test(scrub): the name-fallback sync fence is one-sided, so a third Export.Name warn would go ungrepped with nothing failing
## What is wrong

`tests/unit/scripts/secrets-dynamic-ref-name-fallback-sync.test.ts` (added by go-to-k/cdkd#2562) binds the two phrases `tests/integration/secrets-dynamic-ref/verify.sh`'s Guard 7b greps for to the `logger.warn` string literals in `src/cli/commands/scrub.ts`'s `Export.Name` loop, so a reword of a warning cannot silently make the fixture's absence-greps vacuous. It does that direction correctly and measurably. What it does NOT do is watch the inverse, and three smaller shapes get past it.

**1. One-sided (the main item).** The fence asserts `nameWarns.length >= 2` and that each of the fixture's two strings is owned by a distinct `Export.Name` warn. Nothing asserts that every literal-bearing name-fallback warn IS grepped. Add a THIRD fallback arm with its own `logger.warn` to the name loop and Guard 7b silently stops watching it while the fence stays green — the exact drift the file exists to catch, running in the other direction. This is the shape the repo's "pair every cap with a floor" rule names. Fix: assert the COUNT of literal-bearing `Export.Name` warns EQUALS `strings.length`. The two today are the only ones — the collision warning at `scrub.ts:3387` is built by a helper and carries no inline literal, so it is correctly not classified — which is why the equality is writable now and is the thing that would fail on a third arm.

**2. It binds the strings, not the guard's polarity.** The reader takes the fixture's `for NAME_FALLBACK in "..." "..."` line only. Measured: deleting Guard 7b's `grep -qF ... exit 1` body while leaving the loop line in place keeps the fence green — the strings are still "used", by a loop that no longer asserts anything.

**3. False positive on a legitimate refactor, with a misleading message.** Also measured: re-indenting the `for NAME_FALLBACK in` line fails the fence with `strings.length=0` (the reader anchors at `^`). And the JSDoc names `+`-of-two-literals as a refused spelling but omits the one with in-repo precedent — `scrub.ts:3523` already spells a scrub warning as `logger.warn(exportAliasCollisionScrubWarning(...))`. Extracting either name-fallback message to such a helper drops the Export.Name warn count to 1 and trips the floor, which is loud but points at the wrong thing. One sentence in the JSDoc, or accepting a helper call whose name is fenced separately.

**Nits in the same file**, worth folding into whichever round takes the above:

- `:137-147` — negative controls 1-5 are individually inert: deleting any one of them leaves the self-probe green (only control 6 is fenced, by `toHaveLength(2)`). A per-control `expect(exportNameWarnRuns(line)).toEqual([])` loop closes it. The reader's own clauses ARE fenced, so this is belt-not-braces.
- `:63-77` — `literalRuns` descends into nested call arguments, so `logger.warn(truncate('could not be resolved during scrub'))` satisfies the binding even though the emitted line may differ. Fail-open and narrow.
- `:113` — `toHaveLength(2)` hard-codes the fixture's arm count; a legitimate third grep reds the fence. Derive it from `strings.length` once item 1 is applied.
- `:103-104` — `readFileSync` at describe-body scope, so a renamed fixture or source surfaces as a raw `ENOENT` at collection rather than a named failure. Peers use the same idiom, so it is consistent; a message costs one line.

**Unrelated asymmetry found in the same review**, `tests/unit/cli/commands/scrub-export-name-collision.test.ts:591`: go-to-k/cdkd#2562's round 2 added a `Stale` state key to the cross-region pin case so its `not.toContain(MOVING_FIRST)` became load-bearing. The twin non-pin moved-value case at `:591` still has no such key, so ITS `not.toContain(MOVING_FIRST)` remains vacuous. Same fix, one file over.

## Why it matters

Nothing user-facing: this is a fence over a fence. The cost is the one this repo has paid before — a checker that reads as coverage while a whole direction goes unwatched, so the next session trusts Guard 7b to be watching every name-fallback arm when it is watching exactly the two that existed when it was written.

## Verification plan

- Apply the count equality and re-measure the seven kills the file already claims (they are reproducible: warn demoted to `debug`; reword with the old text in a standalone comment; the same in a trailing comment; the real warn split across a substitution; the fixture repeating one string; the fixture loop line renamed; both strings folded into one warn), plus the NEW kill the equality buys: a third `logger.warn` arm added to the name loop and not grepped.
- Add the per-control loop and re-measure that deleting any single negative control now fails the probe.
- No integ needed: the file is `tests/**` only. `/check` covers it.

Dup-check: searched open and closed issues for `name-fallback sync fence`, `secrets-dynamic-ref-name-fallback-sync`, `one-sided fence logger.warn`, `Guard 7b fallback grep` right before filing -- no hit; go-to-k/cdkd#2728 is the unmasked log site at a different file, and nothing covers this checker.
Session-fit: next (not this session) -- the file does not exist on `main` yet; it arrives with go-to-k/cdkd#2562, which is at approve, and this is a fence-over-a-fence refinement rather than anything that change needs.
Severity: low -- test-only. Nothing a user runs changes; what is lost is that a future third name-fallback arm would go unwatched by Guard 7b with no test failing to say so.
Effort: small (S) -- a count equality plus a per-control assertion loop in one test file, re-measuring the existing probe set.
Estimate: ~1 h -- re-measuring the eight kills against mutated copies is most of it; the edit itself is a few lines.

### LINKED ISSUE #2769: test(unit): the bare-parser floor's measured count is ~80 stale, so the floor is ~42% of its subject instead of the ~77% its rationale claims
## Summary

`tests/unit/scripts/check-scope-checker-inputs.test.ts`, in `parser floor: the BARE extraction sees what the JOIN parser cannot (issue go-to-k/cdkd#2381)`, documents its own population as:

```
// Measured 97 on the shipping tree (99 with the carve-out disabled), by
// appending a probe case to this file that prints
// `[...extractBareTargets().keys()].filter((k) => !joinKeys.has(k)).length`.
```

Measured now, with exactly that probe (writing the result to a file, because the stream fence in `tests/setup.ts` swallows `console.error` on a passing test):

| tree | carved | uncarved (carve-out disabled) |
| --- | --- | --- |
| `origin/main` @ `333f3f64` | **178** | **180** |
| `fix/2751-scope-critic-local-settings` @ `4d44ed14` (PR go-to-k/cdkd#2758) | **175** | **178** |

Both deltas are internally consistent -- 2 carve-out entries on `main`, 3 on that branch -- and the two trees' uncarved figures differ by 2 because the branch predates go-to-k/cdkd#2760 / go-to-k/cdkd#2763, which added literals. The documented pair is ~80 low on either tree.

The comment was written in go-to-k/cdkd#2383. Nothing has updated it since, and nothing reds when it drifts: the assertion beside it is `toBeGreaterThanOrEqual(75)`, so the comment is pure prose with no fence on it.

## Two things follow, and the second is the one worth deciding

1. The prose is wrong and should carry the re-measured figure plus the tree it was taken on (the comment already establishes that convention -- it names a `git archive` copy measuring 93 as a legitimately different tree).
2. **The floor no longer means what its own rationale says.** It reads `Floored at 75 (~77%) so ordinary churn is free while a narrowing is not`, and the same comment argues against a floor far under its subject: `The first revision floored this at 5, which a 95% collapse would have passed -- a floor far under its subject fences only total disappearance, which the named paths above already catch`. 75 was ~77% of the documented 97; against the real 178 it is ~42%. By the comment's own standard the floor is now in the shape it was rewritten to avoid, and a 57% collapse of the bare parser's population would pass it.

Raising the floor re-opens the churn question the comment settled once, so this wants a deliberate number rather than a mechanical `0.77 * 178`.

## Not a regression from go-to-k/cdkd#2758

Surfaced while reviewing PR go-to-k/cdkd#2758, which moves the carved figure by exactly 1 (one new carve-out entry). Two independent reviewers of that PR each proposed correcting the comment to `96`, reasoning by delta from the documented figure rather than measuring; the drift predates both that PR and go-to-k/cdkd#2738.

## Classification

Session-fit: next (not this session) -- the fix lands in `tests/unit/scripts/check-scope-checker-inputs.test.ts`, the one file PR go-to-k/cdkd#2758 holds open, and a second PR editing it would hand that contributor a conflict; unblocked the moment go-to-k/cdkd#2758 merges. The measurement is in this body, so nothing is lost by waiting. Next session's verification: `mise exec -- vp test run tests/unit/scripts/check-scope-checker-inputs.test.ts`, plus re-running the probe above on the post-merge tree.
Severity: low -- wrong text that does not execute; the named-path assertions above the floor still fence total disappearance of the bare parser, so no checker input goes unchecked. What stays broken is the floor's discriminating power and a rationale that misdescribes it.
Effort: small (S) -- a comment correction and one floor constant, riding the unit run already paid.
Estimate: ~30 min -- most of it is re-measuring on the tree left after go-to-k/cdkd#2758 merges and choosing the floor, not the edit.
Dup-check: searched issues for `check-scope-checker-inputs`, `extractBareTargets`, `parser floor`, `bareOnly`, `Measured 97 shipping tree`. Hits: go-to-k/cdkd#2751 / go-to-k/cdkd#2752 (this file, the `.claude/settings.local.json` carve-out -- a different assertion, and go-to-k/cdkd#2752 is closed as a duplicate of go-to-k/cdkd#2751); go-to-k/cdkd#2701 (open; tracked-only `git ls-files` populations in five other tests, explicitly excluding this scope checker); go-to-k/cdkd#2364 / go-to-k/cdkd#2381 (closed; the include-scope gaps this floor was added for); go-to-k/cdkd#2643 / go-to-k/cdkd#2555 (open; fences in unrelated files). None describes a stale measured count or an under-scaled floor in this test.

### LINKED ISSUE #2794: test(hooks): pin worktree-owner-gate's notebook_path fallback, which no case reaches
## What

`.claude/hooks/worktree-owner-gate.sh:55-57` reads its target through a
two-branch fallback:

```sh
file_path=$(printf '%s' "$input" | jq -r '
  .tool_input.file_path // .tool_input.notebook_path // ""
' 2>/dev/null || echo "")
```

The second branch is what makes the gate work for `NotebookEdit`, which is in
its registered matcher (`Edit|Write|NotebookEdit`) and which names its target
`notebook_path`, not `file_path`.

**Nothing exercises that branch.** Every payload in
`.claude/hooks/worktree-owner-gate.test.sh` is built by one helper at line 33:

```python
d={'tool_input':{'file_path':sys.argv[2]}, 'cwd':sys.argv[3]}
```

`file_path` always, `notebook_path` never, `tool_name` never set at all. So all
24 cases pass through the FIRST branch, and deleting `.tool_input.notebook_path
//` from the hook leaves the suite green.

## Why it matters

The hook is CORRECT today — this is a missing fence, not a live defect, which is
why the severity below is `low`. What it costs is that the branch can be removed
or mistyped in a refactor with no signal, and the consequence would be silent:
a `NotebookEdit` would resolve to an empty `file_path`, fall to the
`${file_path:-$hook_cwd}` default at line 66, and claim ownership against the
payload's cwd instead of the file's worktree. That is the multi-session
uncommitted-work guard going quietly wrong, which is the class the gate exists
for.

## How it surfaced

Found while widening the sibling `main-tree-edit-gate` to cover `NotebookEdit`
(go-to-k/cdkd#2717). That change read only `file_path` and was therefore INERT
for the tool it was added for; `worktree-owner-gate` is where the correct field
list was read from. Checking whether the sibling's own fallback was pinned
showed it is not.

## Fix

Add `NotebookEdit` cases to `worktree-owner-gate.test.sh` in both polarities —
a first write claiming the worktree, and a second session's write being refused
— with the payload carrying `notebook_path` and NO `file_path`. The measurement
that makes them worth having: delete `.tool_input.notebook_path //` from the
hook and require the new cases to go red.

Keep the existing `file_path` cases. A real payload may carry either, and
swapping one for the other trades one blind spot for another.

While there, the helper at line 33 never sets `tool_name`, so no case
distinguishes the three matched tools from each other. Worth one case per label
if it is cheap.

Verify with: `bash .claude/hooks/worktree-owner-gate.test.sh`, then the same
under `HOOK_BASH=/bin/bash` per the repo's two-bash rule.

Dup-check: searched open issues for `notebook_path`, `NotebookEdit`,
`worktree-owner-gate notebook` and `worktree-owner-gate coverage` — the only hit
is go-to-k/cdkd#2717, which is about the guard layer's tiering and does not
cover this. Also checked the code: no comment at or above
`worktree-owner-gate.sh:55-57` cites an existing issue number. Not a duplicate.

Session-fit: next (not this session) — the session that found it is shipping a
scoped change to a DIFFERENT hook, and folding a second hook's suite into that
diff widens a review already at three rounds. The evidence is a grep, not a
live measurement, so it does not expire with this session.
Severity: low — the hook's behaviour is correct; what is missing is the fence,
so nothing is broken for a user until someone edits that line.
Effort: small (S) — one suite, two to five cases, no source change; the
verification cycle is the hook suite under both bashes, about a minute.
Estimate: ~30 min — most of it is the delete-the-fallback measurement that
proves the new cases discriminate, not the cases themselves.

go-to-k:
## Summary

One lane, eight small agent-tooling / test-fence issues, one behaviour change in a hook (`main-tree-edit-gate`) and five fences made two-sided. No `src/**` change, so no changelog entry; `check` scope only (hooks, skills and tests), so `/check` + `/verify-pr`.

Closes #2684
Closes #2794
Closes #2681
Closes #2458
Closes #2683
Closes #2769
Closes #2732
Refs #2786

### `main-tree-edit-gate` advances its base on a `cd` that FAILS (#2684)

- **Fix** (`.claude/hooks/main-tree-edit-gate.sh`): the ordered walk advances `cur_base` only for a directory that EXISTS at hook time (`[[ -d ]]`, after `gate_expand_tilde` on the `cd` target — a leading `~/` used to yield a bogus `<base>/~/x` base that agreed with bash by accident and would have become a false block under the existence test). One exception, because the hook runs BEFORE the command: a directory an EARLIER segment of the same command NAMES (`mkdir -p X && cd X`, CLAUDE.md's `git worktree add X ... && cd X`) still advances — a MENTION test rather than a list of creating verbs, since every creator left off such a list is a false block and the list goes stale silently. `rm`, `rmdir`, `cd` and any segment carrying `worktree remove` note nothing; a noted token loses one trailing slash and is otherwise taken as written.
- **Two review rounds, FOUR blockers, all four in the false-block MITIGATION and none in the `[[ -d ]]` test — the mitigation is WITHDRAWN.** The merge base `1e4a75d7` has neither `__cd_named_earlier` nor `-d "$cdt"`, so the existence test is the whole of the go-to-k/cdkd#2684 fix (oracle: 10-15 fail-opens before, 0 now) and everything else was a mitigation for the false blocks it introduces. Each incarnation produced a fail-open or a regression:
  - **B1 (security).** Removers un-noting + dual raw/resolved noting + materialised ancestors: `${var#pat}` is quadratic and the note string held every word of every segment twice, so an ORDINARY `gh pr create --body "<3 KB>" && rm -f a b c && echo > README.md` took **28.2 s on bash 3.2** and was **killed at 30 s on 5.3**. It leaked anyway — adjacent duplicates, a differently-spelled remover, a removed parent.
  - **B2 (code).** Those three leaks, measured rc=0 each.
  - **B3 (security).** The replacement `__norm_mention` removes one slash per iteration and **rescans**, reaching the 10 s kill from **under** `GATE_EDIT_MAXBYTES`: 21.7 s for a 4042-CHARACTER command (`${#cmd}` counts characters, so multi-byte text buys more tokens per byte of cap). "Bounded by construction" was therefore FALSE; that comment is retracted — the cap bounds the INPUT, not the work.
  - **B4 (code, regression).** Basename removers + an ANCHORED `git … worktree remove`: the anchor matched `-C` but not its separate argument, so `git -C <path> worktree remove X ; cd X ; <write>` went **rc=2 → rc=0**, on the spelling this repo's own instructions prescribe.
- **What is left**: the `[[ -d "$cdt" ]] || __cd_named_earlier …` test, `gate_expand_tilde` on the `cd` side, ONE trailing-slash strip on the note side, and the merge base's exact remover spelling (`rm|rmdir|cd` plus an UNANCHORED `worktree remove`, whose cost is segment-scoped and refuses). No normaliser, no `__NORM`, no basename match, no `mv`, no note-side tilde.

**Timing after the withdrawal** (`mkdir <token> && cd /nx ; echo POISON > README.md`, both bashes, all rc=2):

`U+65E5` stands for a 3-byte UTF-8 character (this PR's body must stay ASCII): `${#cmd}` counts CHARACTERS, so a multibyte prefix packs more of them under `GATE_EDIT_MAXBYTES` and made the withdrawn loop worst-case.

| shape | chars | before | after |
| --- | --- | --- | --- |
| `a//`×1330 | 4032 | 9.5 / 8.5 s | 0.24 / 0.24 s |
| `x`×1000 + `a//`×1000 | 4042 | 8.9 / 13.1 s | 0.22 / 0.21 s |
| `U+65E5`×1000 + `a//`×1000 | 4042 | **21.7 / 12.6 s** | 0.22 / 0.23 s |
| `U+65E5`×1000 + `/`×3000 | 4042 | 7.4 / 8.3 s | 0.23 / 0.30 s |

**Residues, each measured, stated in the danger direction.** FAIL-OPEN (rc=0 — bash wrote the tracked file, the gate allowed it), all one class: a mention that does not survive to the `cd`, which no text scan separates from a creator it does not know — `echo X ; cd X`, `test -d X || cd X`, `ls X X X ; rm -rf X ; cd X`, `mkdir -p X ; rm -rf X ; cd X`, `mkdir -p X ; rm -rf <abs>/X ; cd X`, `mkdir -p X/sub ; rm -rf X ; cd X/sub` — in each the remover IS read; what survives is the earlier mention. Plus one INHERITED shape unchanged by go-to-k/cdkd#2684 (rc=0 on the merge base, both withdrawn revisions and here): `mkdir -p <main>/s && cd <main>/s && echo > ../README.md` — the base is credited to a directory that does not exist yet, so `..` cannot be `pwd -P`-resolved. Backstop for all of them: `main-tree-dirty-detector`, as the header says. FALSE BLOCK (rc=2 on a write that never touches the protected tree), the price of withdrawing the normalisation and the note-side tilde: `cd X//`, `cd .//x`, `cd ./x`, `mkdir -p ~/zz && cd ~/zz`; plus a creator sharing a segment with the text `worktree remove`. **`cd X/` is NOT among them any more** — a final round restored the two O(1) pieces that no blocker implicated: one trailing slash stripped on the `cd` side too (so the tab-completion spelling is rc=0, as on the merge base) and the remover verb read by BASENAME with `mv` (so `/bin/rm -rf X ; cd X` and `mv X Y ; cd X` are rc=2, both measured rc=0 without it). The ANCHORED `git worktree remove` regex stays withdrawn — it was the B4 regression — and a case now pins the `-C` spelling, so re-anchoring it cannot pass the suite. All residues are pinned as cases, so re-closing any is deliberate.
- **Fence**: `main-tree-edit-gate.test.sh` 194 cases (floor re-derived from the tree, 159 → 191); cases pinning withdrawn behaviour are DELETED rather than repointed. `main-tree-edit-oracle.test.sh` 237 rows — the trailing-slash row is back in the corpus now that `cd X/` passes, so no corpus edit is propping up its zero-false-block figure.
- **Receipts**: `main-tree-edit-gate.test.sh` **194/0 under bash 5.x AND 3.2**; origin/main's hook fails 11 of them. Oracle **237 rows, 0 fail-open / 0 false-block / 0 slow on both bashes**; `GATE_HOOK=<origin/main copy>` 10 fail-opens. The B4 spelling and its plain twin are rc=2 again on both bashes. 20/20 on the full shape matrix (B4 + the go-to-k/cdkd#2684 fix + every residue + the creator controls), both bashes.

### `worktree-owner-gate`'s `notebook_path` fallback has no case (#2794)

- **Fix** (`worktree-owner-gate.test.sh`, test-only): the payload helper takes a field selector (`file_path` / `notebook_path` / `none`, `tool_name` set to match) and a cwd; six cases: a NotebookEdit-shaped payload with NO `file_path` claims the FILE's worktree from a main-tree cwd (the cwd is deliberately not the worktree — with the fallback deleted the target collapses to the cwd, and a worktree cwd would have claimed and refused identically; the first draft sat green over the deleted branch for exactly that reason); a foreign session's NotebookEdit is refused naming the owner; neither key → judged against the payload cwd, both polarities. The suite now honours `HOOK_BASH` like its siblings.
- **Receipt**: `.tool_input.notebook_path //` deleted from the hook → `27 passed, 3 failed` (the three notebook cases); restored byte-exact (`git diff` empty). 30/30 under `bash` and under `HOOK_BASH=/bin/bash`. Review round: the cwd-fallback case now seeds the sentinel itself — a reviewer measured that it reddened under the mutation only because the previous case had failed to claim, which is state dependence rather than discrimination.

### `pr-review-gate`'s comment demotes the real enforcement (#2681)

- **Fix** (`pr-review-gate.sh`, comment only): the sentence claiming `markgate verify` "already enforces this via the digest" is replaced by the measurement — the digest cannot see a sentinel nobody rewrote, so the `recorded_sha = head_sha` comparison IS the binding's enforcement and must not be removed; the explicit message is a bonus. Cites suite cases 7 (`medium PR + fresh marker + sha mismatch → block`) and 10 (`large PR + fresh marker + sha mismatch → block`).
- **Receipt**: `&& [ "$recorded_sha" = "$head_sha" ]` deleted → exactly those two cases red (`Pass: 47 Fail: 2`); restored, `Pass: 49 Fail: 0`.

### Section-5 negative pins are case- and wrap-blind (#2458) + section-2 probe fence residuals (#2683)

- **Fix** (`tests/unit/scripts/work-issues-launch-mode.test.ts`): `proseOf()` — whitespace runs (newlines included) collapsed, case-folded — is the reader for every prose pin; the section-5 pins are a function of the text so three PLANTED shapes (exact unwrapped, exact wrapped, reworded wrapped) plus a control run against a doctored copy. The design question about the historical quote is MOOT: `grep -ci 'detached\|already merged' implement.md` = 0. For go-to-k/cdkd#2683: (item 1) the `\` join reads the RAW next line, so a two-line prose `git show \` + `--stat HEAD` is one unit, and the join never crosses a fence marker; (item 3) `SINGLE_COMMIT` widened to `--raw`, `--summary` / `--compact-summary`, and `git log` limited to one commit (`-1`, `-n 1`, `--max-count=1`, either flag order), with `-10`, a ranged `log --stat`, and `--show-current` as controls; (item 2) the floor reads RENDERED text through `marked` (already the repo's tool for the visibility question in `rule-file-payload.test.ts`) with `<pre>` and raw comments stripped and every tag removed, so a fence info string, text after a closing fence, an HTML attribute, an image alt, a link-reference title, an HTML comment, and fenced / indented code all fail to satisfy it — eight carriers pinned, plus the real doc and a stripped copy as the control pair. The lane brief said item 2 only if it fell out of the shared reader; it did not, but closing that issue needs it and the sanctioned tool made it ~20 lines — flagging the scope call.
- **Review round**, four findings, all fixed: the reword pin missed a CONDITIONAL sentence carrying none of its four verbs (`Branch in place only if the branch here is detached or its PR has already merged.` passed both pins), so a branching-verb-near-the-condition pin joins them and three more planted shapes cover `only if` / `unless` / a leading `When …`; `visibleProse` let `<span hidden>`, `display:none`, `<script>` and `<style>` satisfy the floor, so those are dropped with their content and pinned as four more carriers — the header now says this is a short carrier list rather than none; the rendered-line command exclusion dropped a legitimate explanation OPENING with a backticked `git show --stat HEAD`, so it now requires the line to carry no sentence punctuation, with both directions pinned; and the `log` arm's `-1` matched a ref named `feat-1`, now anchored on whitespace with three ref controls. **Declined**: widening `SINGLE_COMMIT` to `git diff --stat HEAD~1` — go-to-k/cdkd#2683 item 3 enumerates the `show` / `log` family it measured, and a `diff` over an explicit range is a different read; filing rather than guessing is the rule this fence exists to follow.
- **Receipts** (production docs mutated, restored byte-exact): a wrapped quote appended to `implement.md` → 2 cases red; `triage.md`'s rationale replaced by an info-string carrier → 2 red; a two-line `git log -1 --stat` appended to `triage.md` → 1 red; each of the reviewer's three bypass sentences appended to `implement.md` → the section-5 pin red. `origin/main`'s copy of the suite under the same mutations: 54/54 green. New suite 86/86.

### The bare-parser floor's "Measured 97" is ~100 stale (#2769) + `gitIgnores` JSDoc universal (#2786 item 2)

- **Fix** (`tests/unit/scripts/check-scope-checker-inputs.test.ts`): the prose figure becomes `BARE_ONLY_MEASURED = { count: 198, tree: 'origin/main @ 1e4a75d7, 2026-09-15' }` (measured with the comment's own probe: bare-only 198, uncarved-only 201, join 136); the floor is DERIVED from it at `BARE_ONLY_BAND = 0.25` and a CEILING at the same band forces the re-measure when the population outgrows the figure — the direction that went unwatched. The carved/uncarved gap is asserted as the RELATION the thread named (`NON_READ_LITERALS` minus join-covered entries) rather than a second number. The `gitIgnores` JSDoc no longer claims "the only one inside" the `check` include; it states what the code does (refuses every nested source), scopes the reason to the one optional entry, and names what a nested entry under an include-covered directory would need.
- **Receipts**: `count: 97` kept → red (ceiling); the bare walk narrowed by a third (`rel.length % 3 === 0` skip) → red; the carve-out skipping one entry → red (relation). 22/22.

### The name-fallback sync fence is one-sided (#2732)

- **Fix** (`tests/unit/scripts/secrets-dynamic-ref-name-fallback-sync.test.ts`): `nameWarns.length` must EQUAL `strings.length` (floor 2, exact count derived from the fixture); Guard 7b's loop BODY is bound to the grep's POSITIVE branch reaching `exit 1`, not only its `for` line — a reviewer showed the first spelling accepted `! grep …`, `grep … || exit 1` and `if ! grep …`, all three INVERTED, so eight controls now separate the directions and the flag word is read loosely (`-q`, `-qF`, `-q -F`) since a reflow of the flags is legitimate; the loop line may be indented; a literal handed to a nested CALL is not read as emitted text (the helper-call spelling the JSDoc now describes); the six negative controls are asserted one by one; a missing subject is a named failure. Sibling `tests/unit/cli/commands/scrub-export-name-collision.test.ts`: the moved-value case gains `Stale: MOVING_FIRST` and `expect(saved.outputs['Stale']).toBe(MOVING_EXPR)`, so its `not.toContain(MOVING_FIRST)` is load-bearing.
- **Receipts** (production mutated, restored byte-exact): a third literal-bearing `Export.Name` warn in `scrub.ts` → red (equality); Guard 7b's body deleted with the `for` line kept → red (polarity); the loop re-indented → green (control); one string repeated on the loop line → red. Sibling: `SharedEntriesSecrets.set`'s write-through deleted → the moved-value case red on the new file, GREEN on `HEAD`'s copy of the same file (the vacuity the issue described).

## `#2786` disposition

Item 2 closed here (`Refs`, not `Closes`). Item 1 was fixed by go-to-k/cdkd#2758; item 3 is HELD by open PR go-to-k/cdkd#3082 (`.markgate.yml`); item 4 is moot (`docs/changelog-cdkd.md` is gitignored and assembled since go-to-k/cdkd#2779). A comment on go-to-k/cdkd#2786 records this.

## Parent round

THREE rounds: this lane's own two reviewers, then the orchestrator's 3-axis + security round on `fa50d840`, then its delta round on `c4115ab4`. **Four blockers in total, every one in the false-block MITIGATION around the `[[ -d ]]` test and none in the test itself — so the outcome is WITHDRAWAL, not another repair.** B1 (security, 28.2 s / killed at 30 s), B2 (the un-note's three leaks), B3 (security, the replacement normaliser: 21.7 s from under the byte cap), B4 (regression: `git -C <path> worktree remove` rc=2 → rc=0). The merge base carries neither `__cd_named_earlier` nor `-d "$cdt"`, so the existence test is the whole of the go-to-k/cdkd#2684 fix and the mitigation was never load-bearing for it.

- **spec**: one nit, later WITHDRAWN by the reviewer on inspection — the suite does number its cases (`# 7.`, `# 10.`), so the original ordinals were right. They are restored and the false justification I had added is deleted.
- **code**: B2, then B4 in the replacement — the minors that round asked for (the `./`/`//` normalisation, the basename remover match, the anchored `worktree remove`) are exactly what produced B3 and B4, so all three are withdrawn and their false blocks are stated residues instead.
- **test**: adequate, 1 low gap — the new `declare -F gate_expand_tilde` clause was unfenced. Fixed with a second stub library; the clause's deletion now reds it. Its re-run receipts (merge-base hook 175/11, 15 executed oracle fail-opens on the 237-row corpus it ran; 10 on this one, which drops the created-then-removed rows) are what establish the `-d` check independently of this lane.
- **security**: B1, then B3 in the replacement — both are why the mitigation is withdrawn entirely rather than repaired a third time. The per-token `gate_expand_tilde` fork the second round flagged is gone with the note-side tilde (stated as a false-block residue instead).

**Also in this commit, from the delta round**: a botched edit had duplicated comment text inside one line of `main-tree-edit-gate.test.sh` (fixed). The pre-existing "19 cd targets x 2000 candidates" latency case had a 4 s budget against ~3 s isolated and 6-12 s under load on BOTH this revision and `fa50d840` — zero headroom, the go-to-k/cdkd#2741 / go-to-k/cdkd#3038 flake class — so it is 8 s now with the measurement in a comment. **My own two latency cases had the same defect** and one flaked live at load 27. Every figure is now recorded PER ENGINE, because they differ by up to 10x and CI's macOS runner is the slow one: the 19-cd case is ~3 s on bash 5.x but 2.75-6.18 s on 3.2 (so 8 s sat inside its own measured range — it is **9 s**), and the 3 KB-body case 0.21-0.38 s against 1.96-3.66 s (6 s stands). All still fail well before the 10 s kill — the withdrawn revisions measured 28.2 s and were killed at 30 s. Separately, the hook's "4096 → 0.58 s, roughly 17x of margin" note is **re-measured and dated**: on the worst shape fitting under the cap (4089 chars) it is 4.1 / 4.9 s on the merge base and 5.4 / 6.4 s here (5.x / 3.2, load ~25), so the cap buys 1.6-2x, most of it inherited.

The residues are listed with a measured rc in the `#2684` section above. Two more, in the `cd` parse rather than the mention rule and both inherited: the target is read up to the first whitespace, so `mkdir -p "/x/fresh q" ; cd "/x/fresh"` matches on the truncation though bash never entered it; and a relative `cd` reaching a directory created under a DIFFERENT base (`mkdir -p a/b && cd a && cd b`) is not credited — measured unchanged from the merge base.

## Verification

- Hook suites (`bash .claude/hooks/run-tests.sh`, both bashes): all green. Per-suite before/after: `main-tree-edit-gate` 159/0 → 194/0, `worktree-owner-gate` 24/0 → 30/0, `pr-review-gate` 49/0 → 49/0, oracle 207 → 237 rows, 0/0/0.
- `/check`: `vp check --fix` + `vp run check` 0 errors; `typecheck:test` 0; build; `gen:all-matrices` + `audit:coverage:check` no drift; `vp test run` 1014 files / 23022 tests passed, rooted in the lane tree.
- Reviewers: this lane's own `pr-code-reviewer` + `pr-test-reviewer` (no blockers; two majors and six minors fixed, two declines stated with reasons — the `git diff` sibling above, and the band ceiling's maintenance cost under go-to-k/cdkd#2769, which is the instrument's purpose), then the orchestrator's independent 3-axis + security round, whose two blockers are the section above.
- `pr-review` is NOT set by this lane. The marker in this worktree reads `match` but its sentinel names `16d335fc`, a merged lane's tip inherited in the shared IN-PLACE tree — stale by binding, not by digest.
- No real-AWS run (no `src/**`, no deletion logic, no integ scope touched).

## Parent round 4 - the withdrawal head (`e5e4dacc`), then the last five

Independent code + security against `e5e4dacc`: **withdrawal complete, no blocker.** Code: zero `rc=2 -> rc=0` across 31 shapes x 2 revisions x 2 bashes, `git -C <path> worktree remove` back to rc=2. Security: the four worst-case shapes measured **12.29 / 9.73 / 12.48 / 8.75 s on `c4115ab4` -> 0.21-0.33 s here** (merge base 0.10-0.30 s), zero `rc=2 -> rc=0` across 60 shapes x 2 bashes, and the latency bound now holds **structurally** rather than by the byte cap: every added construct is linear per segment, forks happen only on mark-0 segments, the newline sentinel limits glob restarts to token boundaries, and `cur_base` growth is bounded by the cap.

Both reviewers converged on the same two, applied in `e255a5dd`: the BASENAME remover match and `mv` restored (the regression was the ANCHORED `worktree remove`, not `${__verb##*/}`, which is O(1) and only widens the refusing direction) - `/bin/rm -rf X ; cd X` and `mv X Y ; cd X` back to **rc=2**; and the `cd X/` false block closed with a single O(1) trailing-slash strip on the `cd` side, which let the oracle row be **restored** rather than left as a corpus edit (237 rows again). Plus: a case pinning `git -C ... worktree remove` so re-anchoring that regex can no longer pass the suite; the 19-cd budget 8 -> 9 s with the bash-3.2 figure recorded beside the 5.x one (8 s sat inside its own measured load range, and CI's macOS runner is 3.2); and the stale "4096 -> 0.58 s, 17x" figure re-measured and dated.

**One last item, applied by the orchestrator in `4fdc8301`.** The oracle's own 5 s latency assertion is a WALL-CLOCK reading on a shared machine, and CI runs it (`hooks.yml` -> `run-tests.sh`): it reported **3 rows over budget at load 46** with six peer vitest processes and **0 rows on the very next run of the same tree** - the go-to-k/cdkd#2741 / #3038 class this PR already fixed three times elsewhere. A slow row is now RE-MEASURED once before it counts: a load spike is transient, a pathological command is slow every time, so the 5 s is neither raised nor weakened and the clean path costs nothing. Probed both ways - forcing every row through the retry keeps it at 0 slow; forcing the retry to COUNT reds the assertion at 52 rows.

Final tallies on `4fdc8301`: gate suite **194/0**, oracle **237 rows, 0 fail-open / 0 false-block / 0 slow**, `run-tests.sh` **all suites pass, 0 failures** - each under bash 5.x AND 3.2; unit **1014 files / 23022 tests**.


go-to-k:
Merged — thanks, and this one earned its rounds. The `-d` existence check was right from the first push (the executed oracle went from 10-15 fail-opens to 0), but every layer added to keep it from false-blocking produced a defect of its own: four blockers across three rounds, all in the mention machinery, two of them reaching the 10 s PreToolUse kill where a killed hook disarms every gate on the call. Taking the deletion rather than bounding it was the right call, and finding two defects inside your own fix by RUNNING it — the per-token subshell fork and the bash-3.2 escape divergence — is what kept the last round from being a fifth.

Issue #1948, opened 2026-08-18
fix(diff): a pre-GHSA output deleted from the template prints its stored plaintext as a REMOVE row


go-to-k:
## Context

Found by the final security review of the (#1919) lane. **Pre-existing on `main`**
— identical behavior before and after that branch — so it was left out of it
under a semantics freeze rather than bundled in.

## The gap

`src/analyzer/outputs-diff.ts` withholds a stored `oldValue` when the record
looks pre-GHSA, using `secretSourceKeys` as the signal. A comment there claims
the set also covers "one deleted from the template".

It cannot: `secretSourceKeys` is built ONLY from outputs DECLARED in today's
template. An output that existed in a pre-GHSA state record and has since been
REMOVED from the template contributes nothing to the set, so if it was the
record's only secret-bearing output the record is judged "not legacy" and its
stored value is rendered.

That value is pre-GHSA plaintext, and it prints as the `old:` side of a REMOVE
row — into the terminal and into CI logs for any diff-gated pipeline.

## Reachability

Needs all of: a state record written before the redaction work shipped; a
secret-bearing output in it; that output deleted from the template; and no OTHER
declared output carrying a secret reference (any one of those would set the
legacy flag and withhold correctly). Narrow, but it is exactly the shape a
cleanup commit produces — deleting an output is a normal refactor.

## Direction

The signal has to come from the STORED bag rather than only from the template,
since the whole question is whether the stored bag predates redaction. Candidate:
treat a record as legacy when any stored value is a plaintext that today's
template cannot account for, or carry an explicit marker written by the first
post-GHSA write.

Whichever is chosen, the comment must stop claiming coverage the set does not
have — that claim is what would let a future reader conclude the case is handled.

Two smaller siblings from the same review, worth folding in since they are the
same "last human-render site" class:

- `src/cli/commands/state.ts:729` prints output bag KEYS unstripped — a resolved
  export name passed no CloudFormation validator, so it can carry control
  characters.
- `src/cli/commands/diff-recursive.ts:1024-1039` has an orphaned JSDoc above
  `stripDisplayOnlyChars`.

## Verification

A unit case building a pre-GHSA state record whose only secret-bearing output is
absent from the template, asserting the diff withholds the stored value.
Mutation-probe it — the withholding is a negative behavior and those are easy to
write green.

Session-fit: next (not this session) -- pre-existing rather than a regression of
the (#1919) lane, and the fix needs a decision about where the legacy signal
[...]

Issue #2888, opened 2026-09-09
chore(deployment): decide the sanitiser class and null-handling for the 17 stripControlChars-alone call sites


go-to-k:
Split out of go-to-k/cdkd#2874, whose Scope section covered two things: the deletion-reconstitution hazard (fixed there) and a per-site null-handling decision across every remaining `stripControlChars`-alone call site. This issue is the second half. It is filed rather than left implied by go-to-k/cdkd#2874 staying open, because that issue is being closed by its PR.

## What remains

`src/deployment/outputs-export-alias.ts` now deletes ONE canonical class derived from `\p{Cc}` / `\p{Cf}` / `\p{Zl}` / `\p{Zp}` plus the variation selectors and Hangul fillers, and tests, masks and prints in that space. Every OTHER `stripControlChars` call site in `src/` still sanitises with the narrower class, so the tree now has two spellings of "make this string safe to print".

That divergence is deliberate and recorded (the two answer different questions — see `src/utils/display-safe.ts`'s header and `SECRET_SCAN_INVISIBLES`' doc comment), but each remaining site still owes the decision go-to-k/cdkd#2874 asked for: does it need the wider class, and what should it do with `undefined` / `null` (`displaySafe` normalises both to `''` and trims; `stripControlChars` takes a `string` and would change behaviour under either).

## The population, grepped

`git grep -o 'stripControlChars(' origin/main -- src/` outside `outputs-export-alias.ts` and the definition in `utils/regexp.ts`:

| file | call sites |
| --- | --- |
| `src/cli/commands/state.ts` | 8 |
| `src/analyzer/outputs-diff.ts` | 5 |
| `src/deployment/intrinsic-function-resolver.ts` | 3 |
| `src/cli/commands/diff-recursive.ts` | 1 |
| **total** | **17** |

**go-to-k/cdkd#2874's own enumeration of this set was wrong and should not be reused.** It listed `outputs-diff.ts` at 3 sites with line numbers (`:487`, `:495`, `:557`) that belong to some earlier base; the real calls are at `:464`, `:534`, `:542` (twice) and `:604` — 5 calls on 4 lines, which a line-oriented count misses. Its "23 call sites in five files" and the 12 implied by its per-file list are both wrong. The table above is pasted from grep output.

## What is NOT in scope

[...]

PR #3162, opened 2026-09-15
fix(diff): preview the no-change outputs merge beside a failing output, so cdkd diff --fail sees the sibling change the deploy writes

### LINKED ISSUE #3101: fix(diff): a no-change stack whose previously resolved output now fails suppresses an Outputs delta the partial persist writes, so cdkd diff --fail exits 0

nix-tkobayashi:
## Summary

`cdkd diff --fail` exited 0 for an Outputs change the next deploy writes. Since go-to-k/cdkd#2771, a deploy with no resource change persists the outputs that resolved and keeps a failed output's stored value. The diff still dropped the whole Outputs section whenever one output failed to resolve, so a sibling output added or changed beside a broken one never rendered a row.

`computeStackDiff` now previews the deploy's own merge in exactly that shape:

- the resource diff is empty (`diffCalculator.hasChanges`, the same predicate the deploy branches on), and
- every failure is one the deploy records too — the resolver threw, or returned `undefined` — named by output key, and
- its parameters bound and its conditions evaluated (evaluation is skipped when a condition depends on a secret-valued parameter), and no condition verdict can reach an output (`templateLetsConditionsReachOutputs`): no output carries its own `Condition`, and no `Fn::If` or `{Condition: ...}` appears outside `Resources` and `Conditions`. The diff evaluates conditions best-effort, so a verdict can differ from the deploy's. An output never reads a resource's template properties, so a verdict about a resource changes the resource set instead. The test is deliberately not "declares `Conditions`": every env-agnostic CDK app declares `CDKMetadataAvailable` for its metadata resource alone. With parameter binding failed, a `Ref` to the unbound parameter fails in the diff and resolves at deploy.

It calls `mergeNoChangeOutputs` itself rather than a copy, feeding it the preview bag with each failed output present as `undefined` (the deploy's shape; the diff resolver drops a failed key) and `importableOutputKeys(currentState)` as the previous export set. A `merged` verdict renders `computeOutputsDiff` against the merged bag and export set, and warns naming the failed outputs, split into those compared at their stored values and those with no stored value under their own names — a failure can still depend on something only the deploy has (a secret the diff never fetches, used as a mapping key, is one route; they are not enumerable), so the preview never claims such an output is unchanged; a `kept` verdict falls back to the existing suppression — which warns when a difference remains beyond the failed outputs, naming the merge's keep-whole reason — matching the deploy, which keeps the previous bag whole.

The deploy runs the mixed-generation refusal twice: inside the merge, which the preview shares, and again on the bag as its save redacts it. That second check is not reproduced, and it can answer differently: the preview's own bag never gains a late expression, but the deploy's can, when a secret recorded during its observed-capture drain has the same plaintext as a resolved literal output, which turns `merged` into `kept` (`deploy-engine-outputs-only-change.test.ts` pins that deploy behaviour). What the deploy then writes depends on the values: a row the preview showed may not be written, and a late needle matching a stored value makes the save redact that value in whichever bag it keeps, a rewrite no diff path previews, merge or not. `--strict-getatt` also departs from the preview: an output whose resolution throws aborts the deploy instead of reaching the merge (one that resolves to `undefined` still reaches it). The go-to-k/cdkd#1948 residual `computeOutputsDiff` documents (a secret-handling stack deleting its only secret-bearing output prints that output's stored value on a REMOVE row) now reaches a rendered section on the merge path, where the suppression used to hide it; a carried key never exonerates a legacy record. A carried key is compared at its stored value, which erases evidence only its resolved value held (an output resolving to a secret expression marks a pre-GHSA record), and where such an expression comes from is not enumerable. So whenever a carried key's stored value is anything but a secret expression, the record is treated as legacy and every stored value on a rendered row is withheld.

## The two decisions the issue asked for

1. **Output keys only, and failures the preview cannot name keep the suppression.** `resolveTemplateOutputs` gains `failedOutputKeys` (output names whose value failed; never the literal alias `failedKeys` also holds) and `failuresMirrorDeploy`, which turns false on every arm the merge cannot be handed: an `Export.Name` that threw, one that stayed intrinsic, and a literal one the preview cannot decide in a secret-bearing stack.
2. **`isUnresolvedValue`'s wider arms keep the suppression.** Only a top-level `undefined` and a throw are deploy's own signals. A surviving intrinsic, a symbol, an unsubstituted `Fn::Sub` placeholder, or a container holding any of those (a nested `undefined` included) is not the deploy's failure signal, so it says nothing about what the deploy does with that key: the deploy may write it, lose it when the bag is serialized (a symbol), or throw where the diff's best-effort pass kept a placeholder (a structural `Fn::Sub` failure). Previewing such a key as carried could hide its change with no warning, so those arms clear `failuresMirrorDeploy`.

With a resource change pending the section stays suppressed as before: there a failure usually means an output waiting on a resource this deploy creates.

## Changes

- `src/analyzer/outputs-diff.ts`: `failedOutputKeys`, `failuresMirrorDeploy`, and the per-arm decisions above; `templateLetsConditionsReachOutputs`; a `forceLegacyRecord` option on `computeOutputsDiff`; its four debug lines that print an output key, error text or condition name now strip control characters.
- `src/cli/commands/diff-recursive.ts`: the merge preview in `computeStackDiff`, `withFailedOutputsUndefined`, and one `diffOutputsAgainst` closure that replaces the two `computeOutputsDiff` calls and serves all three comparison paths. The suppression warning names a pending resource as the likely cause only when a resource change is pending, and the stack name is stripped of control characters in both Outputs warnings and in the rendered stack headers.
- `src/deployment/no-change-outputs-merge.ts`: module doc and the `keptWholeReasonText` doc name the diff as a second caller.
- `docs/cli-diff.md`, `docs/architecture.md`, `.claude/rules/layout-analyzer.md`: describe the preview and when the section is still omitted; `.claude/rules/no-change-outputs-merge.md` names `computeStackDiff` as a caller.
- `tests/integration/output-never-resolved-diff/verify.sh`, `lib/output-never-resolved-diff-stack.ts` and `bin/app.ts`: phase 5c, its env-agnostic second diff (`CDKD_TEST_ENV_AGNOSTIC=true` drops `env`), and their three synth premise variants; `docs/_generated/integ-last-run.tsv` records the passing run.
- `changelog.d/entries/2026-09-15-3101-diff-no-change-outputs-merge-preview.md`.

## Verification

- Unit: `tests/unit/cli/diff-recursive.test.ts` — sibling ADD with no row for a failed output that keeps its stored value, which a warning names; a failed mapping lookup keyed by a secret the diff never fetches is carried with that warning; `treeHasChanges` true; suppression kept with a pending resource CREATE, UPDATE or DELETE; MODIFY plus a REMOVE the merge drops; a failed output with no stored value is never an ADD; the literal export alias is carried; a resolved sibling's export alias row is tagged `isExport`, including a sibling exported as `__proto__`; the alias is carried on a pre-v9 record with no export set; suppression kept when a condition reads a secret-valued parameter or a template parameter cannot be bound, and when a condition verdict can reach an output through an `Fn::If` in its value, its own `Condition` or an `Fn::If` in a mapping, with the warning naming neither a pending resource nor an `Export.Name` reason there; the merge previewed for the env-agnostic CDK shape, whose `CDKMetadataAvailable` gates only its metadata resource; the keep-whole warning names the merge's reason, for both the intrinsic-`Export.Name` and the mixed-generation refusal; the merged-path warning names every failed output with its count and control characters stripped, splitting a stored one (compared at its stored value) from one with no stored value under its own name, and states the withheld-values reason, pinned in full, only when the forced verdict actually withholds a row (not for an ADD-only section) and truthfully when only a failed output's export alias was carried, while a row withheld by another legacy signal, with no carried plaintext, gets no such sentence; a legacy record's values and a deleted key stay withheld on the merge path, including a deleted key whose only secret reference sits in an unchanged resource, and a carried output never exonerates a legacy record whose evidence lived only in that output's resolved value, including when a secret expression is stored under another key, the template holds no secret reference, a plaintext is carried second after a carried secret expression, or the carried plaintext is an export alias whose output key stores a secret expression, while nothing carried or a carried secret expression leaves stored values printed; control characters stripped from the stack name in both Outputs warnings and both render headers, and from the stack and parameter names in the parameter-type warning; a record with no outputs bag at all is read as empty; suppression kept for the merge's keep-whole refusal, for a diff-only failure (`Fn::Sub` placeholder), and for an `Export.Name` that throws or keeps an unsubstituted `Fn::Sub` placeholder (the surviving intrinsic `Export.Name` object is covered in the analyzer tests). `tests/unit/analyzer/outputs-diff.test.ts` — per-arm values of the two new fields, including a symbol and a flag that stays false once a later output fails with a top-level `undefined` or a throw, plus control characters stripped from each debug line, the condition-false skip line included; `templateLetsConditionsReachOutputs` true for each of its four routes and false for a resource's own `Condition`, an `Fn::If` in a resource property, a `{Condition}` inside `Conditions`, a multi-key object whose first key is `Condition`, and no route at all.
- Mutation probes, one change per probe, each restored byte-exact: removing the `hasChanges` gate, the `failuresMirrorDeploy` gate, the `undefined` injection, or the previous export set; dropping the throw arm's key; keeping the flag true on the wider arm, the `Export.Name` throw arm, the undecidable literal arm, or the intrinsic-name arm; and naming every unresolved value. A second set after the maintainer-checklist pass: routing a fully resolved bag through the merge path, emptying the merge's input export set or the comparison's, narrowing the resource gate to CREATE only, keeping the flag true for a symbol, resetting it on a later top-level `undefined` or on a later throw, and building the merge's input bag as a plain object instead of a null-prototype one. A third set after the four-reviewer round: dropping the conditions gate, the keep-whole reason in the warning, the pre-v9 export-set fallback, the intrinsic `Export.Name` arm, and each debug line's control-character strip; then dropping the declared-`Conditions` gate, and, after a Codex round showed a secret-keyed mapping lookup failing only in the diff, turning the carried-output warning into a debug line or dropping the output names from it; and, after a maintainer-checklist pass, truncating the warning's names to the first, hard-coding its count, dropping its control-character stripping, and hard-coding the keep-whole reason. Each turned at least one test red. A fourth set after the maintainer's round-2 review, 20 probes: dropping the conditions-route gate or the resource-change conjunct; in `templateLetsConditionsReachOutputs`, dropping the output-`Condition` loop, the `Resources` or `Conditions` exclusion, the `Fn::If` arm, the `{Condition}` arm, its single-key requirement, or array descent; in the forced legacy verdict, never setting it, `some` to `every`, and `computeOutputsDiff` ignoring it; inverting the resource-change branch of the cause sentence; and dropping each of the six new control-character strips (two on the skip line, two Outputs warnings, two render headers). Nineteen turned a test red. The twentieth, replacing the forced verdict's `typeof stored === 'string'` check with `String(stored)`, is equivalent: a non-string stored value whose string form reads as a secret expression carries that expression into `desired`, where pass 1 of `computeOutputsDiff` already marks the record legacy. A fifth set after the maintainer's round-3 review, 12 probes: in the merged-path warning, inverting the compared or the left-out filter, dropping either part, making the compared part unconditional, and dropping the names' control-character strip; making the withheld-values sentence unconditional or never shown, and reading only the first carried key; and dropping each of the parameter-type warning's three strips. All twelve turned a test red. The redundant `typeof stored === 'string'` check behind the fourth set's equivalent mutant has since been removed. A sixth set after the maintainer's round-4 review, 8 probes: dropping either half of the withheld-values sentence's gate (the forced verdict, or a row actually withheld) and never showing the sentence; inverting the no-stored-value filter, dropping that group, or making its guard unconditional; and re-running the compared guard and the any-carried-key force. All eight turned a test red.
- Integ: `tests/integration/output-never-resolved-diff/verify.sh` phase 5c stores `AttOut` as the marker parameter's `Ref`, then breaks it with an `Fn::GetAtt` both deploy and diff refuse while `Plain` changes. `diff --fail` must exit 1 with exactly the `[~] Plain` row, no suppression warning, and a warning naming `AttOut` as compared at its stored value, with no failed output lacking a stored value, the full withheld-values reason (AttOut's stored value is not a secret reference), the withheld stand-in somewhere in the diff output, and no quoted previous Plain value (`"cdkd-neverres-plain-value"`) in it; the no-change deploy writes `Plain` and keeps `AttOut`'s stored value; the next `diff --fail` exits 0. Real-AWS runs on 2026-09-15: PASS in 291 s; PASS in 365 s after the phase-5c synth premise was tightened to compare against the saved att-output template; PASS in 277 s after rebasing onto current `main` and adding phase 5c's warning assertion; PASS in 392 s after a wording-only change; and PASS in 421 s on the round-2 tree, where phase 5c also runs its diff env-agnostic (`CDKD_TEST_ENV_AGNOSTIC=true`, so CDK declares `CDKMetadataAvailable`) and asserts the same rc, row and warning, behind a synth premise that the env-pinned template declares no conditions, the env-agnostic one declares exactly `CDKMetadataAvailable` gating only `AWS::CDK::Metadata` and gives no condition verdict a route to an output, and that, with the whole `CDKMetadata` resource excluded from the comparison, nothing else differs; and PASS in 286 s on the round-3 tree, where both phase-5c diffs also assert `AttOut` compared at its stored value and the withheld-values reason, the env-pinned diff also asserts nothing was left out of the comparison, and the env-agnostic diff asserts the changed `Plain` value; and PASS in 290 s on the round-4 tree, where both phase-5c diffs assert the withheld-values sentence in full, the withheld stand-in somewhere in the diff output, and that `Plain`'s previous value does not appear quoted; the row-level placement is covered by the unit cases; PASS in 397 s on the round-5 tree; and PASS in 409 s on the round-6 tree, whose change is doc wording, a rules clause and comment re-wraps. Each covered every phase including 5c, destroyed with 0 errors, and left 0 surviving state versions and no leftover resources. The first run of the new phase failed on the fixture itself: `--verbose` prefixes every rendered line with a timestamp, so the row count read 0 while the `[~] Plain` row was printed. The phase now takes its premise from a separate `--verbose` run and its row assertions from a plain one.

Closes #3101


go-to-k:
Reviewed at head `e2ce101d` — 3-axis (spec + code + test) plus the security add-on. Tier resolved to 3-axis by file count (15 files, 811 LOC after excluding `docs/_generated/**`), with two independent up-biases that clamp there: two distinct `fix:` rounds on the branch, and a recent-defect history on the edited code (`src/analyzer/outputs-diff.ts`'s last three commits are all `fix:`, the nearer one being go-to-k/cdkd#3107 — the deploy-side merge this PR mirrors). The security lens ran because the change widens what `cdkd diff` RENDERS from a previously suppressed section.

Both decisions go-to-k/cdkd#3101 demanded hold up under grep rather than under the PR body's claim. All five `resolutionFailed = true` arms were enumerated from source: only the throw (`outputs-diff.ts:521`) and a top-level `undefined` (`:532`, gated at `:537`) name a key; the three arms that cannot name one — `Export.Name` threw (`:561`), undecidable literal alias (`:657`), `Export.Name` stayed intrinsic (`:668`) — all clear `failuresMirrorDeploy`. The flag is monotonic (initialised `true` at `:411`, every other assignment `= false`), so the loop is order-independent in both directions, and `isUnresolvedValue` genuinely recurses (`:242`, `:245-247`) rather than stopping at depth 1. The merge inputs are argument-identical to the deploy's (`deploy-engine.ts:3168-3174`), `hasChanges` covers all four change types including replacement, and the preview's bag reaches no persistence surface — no `saveState`, no events store, no exports index. `previousExportNames` / `resolvedExportNames` are copies, so the v9 `Fn::ImportValue` binding set cannot be mutated by a preview.

No blockers. Four items, none of which change behaviour:

**M1 — required before merge.** `src/cli/commands/diff-recursive.ts:773-786` does not reproduce the deploy's SECOND mixed-generation check (`src/deployment/deploy-engine.ts:3186-3197`), which re-runs `bagHoldsSecretExpression` on the bag as `redactOutputs` will write it and can downgrade `merged` to `kept`. It cannot diverge here — that check only fires on a needle recorded during the observed-capture drain, and the diff never drains (`skipDynamicReferences: true`, `diff-recursive.ts:658`), so the preview's bag can never gain a late expression; the first-order check is inside the shared merge (`no-change-outputs-merge.ts:210-216`) and the preview inherits it. Please write that down at the call site. Two reviewers reached this asymmetry independently and had to derive the answer from three files; the next reader will too.

**M2 — required before merge.** The `isUnresolvedValue`/#1948 residual documented at `src/analyzer/outputs-diff.ts:771-775` — a secret-handling stack that deletes its ONLY secret-bearing output leaves `templateHasSecretReference` false, so `unaccountable()` does not fire and a pre-GHSA stored plaintext prints as a `REMOVE` row's `old:` side — now reaches a RENDERED section on the merge path, where the suppression previously hid it whenever a sibling failed. The residual is already accepted and no new value becomes printable that the fully-resolved path would not also print, so this is not a behaviour change request: one sentence in the merged branch's comment saying the residual's REACH widened here, because this PR is where it changed.

**m3 — optional (take-or-leave).** `tests/unit/cli/diff-recursive.test.ts:2555` (`keeps the suppression when an Export.Name cannot be resolved`) asserts rows only, while its two siblings at `:2524` and `:2542` also assert `expect(suppressed()).toHaveLength(1)`. The row assertion does catch a latch regression here (a merge would emit a `Named` ADD), so this is consistency rather than a hole — but it is the only name-then-clear ordering case, so it is the one worth making unambiguous.

**m4 — optional (take-or-leave).** `diff-recursive.ts:785` duplicates the return at `:814`; an `else` would leave one exit, so a future field on the result shape need not be added twice.

Two findings from the code axis that I am deliberately NOT asking you to act on, so they are not chased later: the merged bag can flip `legacyRecord` on via a carried CONTAINER value holding a secret-expression leaf (`outputs-diff.ts:790-801`), and a condition-false output survives into the merged bag when `conditions` is `undefined` (`:477`). Both are inherited unchanged from the fully-resolved path, which already feeds `desired` the same shapes — the merged path matching it is the correct design, not a defect. The second is further blocked in practice: an unevaluated `conditions` leaves `effectiveTemplate` unfiltered, which produces phantom resource CREATEs, which fails the `hasChanges` gate and keeps the suppression.

Verification reads as sound on its own terms — two real-AWS runs of the new phase 5c with 0 destroy errors and 0 surviving state versions, the `--verbose` timestamp defect genuinely fixed by splitting premise and row assertions across two runs, row assertions carrying a floor as well as a cap, and the fixture's values derived rather than pinned. I re-ran the phase's new `jq` read expressions against a realistic sample and the `^ {4}\[~\] Plain( |$)` anchor against the real `renderOutputChangeLines` output; both match.

Push M1 and M2 (plus m3/m4 if you want them) and I will re-review and merge.

nix-tkobayashi:
Thanks for the review of `e2ce101d`. Pushed as `96619870`. All four items are taken; M1 is taken with the opposite conclusion, with the evidence below.

**M1 — taken, but the second check CAN diverge, so the call-site comment documents the divergence instead.** Your derivation holds for the preview's own bag: this diff drains no observed captures and resolves with `skipDynamicReferences`, so nothing late ever reaches it. The re-check at `deploy-engine.ts:3191-3197` runs on the DEPLOY's bag, though, and that bag can gain a late expression. The existing case `decides the bag AFTER the observed-capture drain and re-checks it as the save redacts it, so a needle recorded during the drain cannot slip past` in `tests/unit/deployment/deploy-engine-outputs-only-change.test.ts` pins exactly that: a needle `LATE -> {{resolve:secretsmanager:late:...}}` is recorded during the drain, the resolved LITERAL output `Plain2: LATE` is value-scan-redacted into that expression, and the save-time re-check turns `merged` into `kept`. On the same template the preview's bag keeps `Plain2` as the literal, the merge's first-order check sees no expression, and the diff renders `[+] Plain2` while the deploy keeps the previous bag. The comment now says the check is not reproduced and gives that mechanism and the test. It deliberately asserts no direction: a first draft said the divergence was always visible, and a Codex round found the other one, where a late needle matching a STORED value makes the save redact that value in whichever bag it keeps, a rewrite no diff path previews, merge or not. That second half predates this PR. `--strict-getatt` is named as the other departure.

**M2 — taken.** The merged branch's comment now says the go-to-k/cdkd#1948 residual reaches a rendered section on this path, where the suppression used to hide it, and that no value becomes printable that the fully-resolved path would not print.

**m3 — taken.** `keeps the suppression when an Export.Name cannot be resolved` now asserts the suppression warning like its siblings.

**m4 — taken.** The merged branch no longer returns early; the result has one return.

**What else changed since `e2ce101d`.** Four read-only reviewers and seven Codex delta rounds ran against this push (three before your review landed, four after), and their findings changed behaviour, so this push is larger than the four items:

- **The merge preview is refused in two more shapes where the preview knows its resolution differs:** a template that declares any `Conditions` (they are evaluated best-effort here, and a verdict reaches an output through routes no scan of it can see — an `Fn::If` in its value, or one inside a mapping), and parameters that did not bind (`conditions` undefined).
- **Where the merge does run, it is no longer silent about what it could not see.** A `logger.warn` names every output in `failedOutputKeys`, compared at its stored value, and says the next deploy may write a different value for any it can resolve. Codex found a route no gate can enumerate: a secret the diff never fetches, used as an `Fn::FindInMap` key, throws here and resolves at deploy. `--fail` is therefore never weaker than the suppression it replaces. A key a binding go-to-k/cdkd#2740 record skipped is not in that list; it is previewed absent on every diff path, as that record documents.
- **A keep-whole verdict's suppression warning** names the merge's reason (`keptWholeReasonText`) instead of "a resource this deploy has yet to create", which cannot be the cause on a stack with no resource change.
- **The three `outputs-diff.ts` debug lines** that print an output key and error text strip control characters (the security reviewer's nit).
- **Tests added:**
  - pre-v9 export-set fallback;
  - intrinsic `Export.Name` arm through `computeStackDiff`;
  - secret-valued condition parameter, and unbound parameters;
  - declared `Conditions`, with and without an `Fn::If`;
  - keep-whole reason in the warning;
  - legacy-plaintext and deleted-key withholding on the merge path;
  - secret-keyed `Fn::FindInMap` carried with the warning;
  - debug-line stripping.
- **Integ:** phase 5c also asserts the warning names `AttOut`, with the fragment pinned against `dist`.

Verification on the pushed head: the branch is one commit rebased onto current `main`. `vp test run` passed 1021 files / 23130 tests with no type errors; `vp run check`, `vp run typecheck:test` and `vp run build` are clean. The real-AWS `output-never-resolved-diff` integ, whose phase 5c now also asserts the warning naming `AttOut`, passed in 277 s on this tree (the final amend only records that run): destroy 2 deleted, 0 errors, 0 surviving state versions, no leftover resources.

go-to-k:
Round 2, reviewed at head `96619870` — 3-axis plus the security add-on again (1248 LOC after excluding `docs/_generated/**`, 15 files).

**M1: you are right and I was wrong.** My derivation only covered the PREVIEW's bag, and the divergence lives in the DEPLOY's. `deploy-engine-outputs-only-change.test.ts`'s late-needle case is exactly the mechanism, and the second half a Codex round found — a late needle matching a STORED value making the save redact it in whichever bag it keeps — is a route I had not considered at all. Documenting the divergence without asserting a direction is the correct call. M2, m3 and m4 are taken as asked. The ledger question the test axis raised is answered by your note that the final amend only records the run; `docs/_generated/integ-last-run.tsv:223` naming the carried-output warning confirms the run exercised this round's assertion.

Two blockers this round. Both are in what the extra Codex rounds added, not in the four items.

**B1 — required before merge. The `Conditions` clause makes the fix inert for the default CDK app shape.** `src/cli/commands/diff-recursive.ts:818` refuses the preview for any template that declares a condition. `aws-cdk-lib/core/lib/private/metadata-resource.js` attaches a `CfnCondition` with the overridden logical id `CDKMetadataAvailable` whenever `Token.isUnresolved(scope.region)`, and `src/synthesis/synthesizer.ts:183` makes cdkd's own synth arm it by injecting `aws:cdk:version-reporting: true`. I synthesised both shapes against this repo's `aws-cdk-lib` rather than reasoning about it:

```
env-agnostic + version-reporting -> Conditions: ["CDKMetadataAvailable"]
env-pinned   + version-reporting -> Conditions: []
```

So every env-agnostic stack — what `cdk init` produces, and the shape a user has unless they pin `env` — carries a non-empty `Conditions` block and keeps go-to-k/cdkd#3101 with no signal. The live fixture cannot catch it: `tests/integration/output-never-resolved-diff/bin/app.ts` pins `env`, so the only real-AWS proof runs the one template shape the gate admits.

The clause is also wider than the hazard it was added for. With `!hasChanges` already true, a condition verdict that diverges from the deploy's cannot reach the resource set — a wrong prune or keep is a CREATE/DELETE and a wrong `Fn::If` in a property is an UPDATE, and either flips `hasChanges` and blocks the merge before this clause is consulted. The routes that survive are an output's own `Condition:` key and an `Fn::If` or `{Condition: …}` reachable from `Outputs` / `Mappings` / `Parameters` — and `CDKMetadataAvailable` gates only `AWS::CDK::Metadata`, which no output can reference. Please replace the blanket test with a predicate over those routes (a declared output carrying `Condition`, or `Fn::If` / `{Condition: …}` appearing outside `Resources`), and give the new predicate a fixture arm or an env-agnostic unit template so the CDK default shape is covered. Keep the `parametersBound` half — the spec axis confirmed it is not redundant with the Conditions test, since `evaluateConditions` returns a defined `{}` for an absent `Conditions` block, so on a zero-`Conditions` template `conditions !== undefined` carries exactly the parameter-binding weight and nothing else.

**B2 — required before merge. The M2 sentence you added is false, and the arm it describes is weaker on the merge path than on the fully-resolved one.** `diff-recursive.ts:839` now says "No value becomes printable that the fully-resolved path would not also print." `legacyRecord` (`src/analyzer/outputs-diff.ts:800-810`) reads `containsSecretDynamicReference(desired[name]) || secretSourceKeys.has(name)`, and on the merge path a carried key's `desired` side is its STORED value rather than the value it would have resolved to — so evidence that exists only in the resolved value is erased.

Concrete: a pre-GHSA record `{F: "<plaintext>", G: "old-g"}`, template `F: {"Fn::GetAtt": ["MySecret", "X"]}` where state holds `{{resolve:secretsmanager:…}}` for that attribute, `G` changed, no resource change, no `Conditions`. `containsSecretDynamicReference` is false on `F`'s raw template value, so `secretSourceKeys` is empty. Fully resolved: `desired.F` is the expression, `legacyRecord` is true, `G`'s `old:` renders as `REDACTED_LEGACY_PLAINTEXT`. `F` failing instead: the merge carries `persisted.F`, `desired.F` is the plaintext, `legacyRecord` is false, and `G`'s `old: "old-g"` prints unredacted — on a path where, before this PR, the section was suppressed and nothing printed at all. `recordProvesPostGhsa` and `unaccountable` are unaffected (carried keys are already in `declaredKeys`), so this is the only weakened arm, but it is the record-wide one.

The existing coverage does not reach it: `tests/unit/cli/diff-recursive.test.ts:2715` gives the failing output `SECRET_JOIN` as its TEMPLATE value, so `secretSourceKeys` holds the key and the record is legacy by the surviving arm. The uncovered case is the one where the evidence lived only in the resolved value. Please make a carried key unable to exonerate — force the legacy verdict on `merged` with `carriedKeys.length > 0` when `templateHasSecretReference && !bagHoldsSecretExpression(currentState.outputs)` — correct the sentence at `:839`, and add the `Fn::GetAtt`-to-a-stored-expression case as a test.

**m5 — required before merge.** `diff-recursive.ts:874` prints `It is usually an output referencing a resource this deploy has yet to create.` whenever `merge` is `undefined`, which now includes every gate refusal on a stack with NO resource change (Conditions declared, parameters unbound, `failuresMirrorDeploy` false). That is the cause your own comment at `:861-864` says cannot apply there. Branch the sentence on `!diffCalculator.hasChanges(changes)` rather than only on `merge?.kind === 'kept'`.

**m6 — required before merge.** `src/analyzer/outputs-diff.ts:482` is the one `logger` line in that file the control-char sweep missed; it interpolates `outputKey` and `output.Condition` raw, both template-controlled.

**n7 — optional (take-or-leave).** `diff-recursive.ts:843` interpolates `stackName` unstripped into the new `logger.warn`, whose message argument the logger does not sanitise. Nested children are named `${stackName}~${logicalId}` and cdkd never applies CFn's logical-id charset validation, so ANSI/bidi can reach the terminal. The same holds at the pre-existing `:870` and `:1857`, so this is an added instance of a standing class, not a new one — worth fixing all three together if you touch it. The count and the name list themselves are correctly key-names-only and stripped.

Not requests, recorded so they are not chased: the new warning firing when `outputChanges` is empty is correct, not noise — that is precisely the "no Outputs section" disambiguation, and the old warning's `wouldHaveChanged.length > 0` guard existed for a different reason (failed keys reading as phantom REMOVEs, which no longer applies once they are carried). `Object.keys(template.Conditions ?? {}).length === 0` versus `template.Conditions === undefined` is moot once B1 replaces the clause. `no-change-outputs-merge.ts:221`'s "The sentence the deploy engine appends" now has a second caller, which is a comment to fix while you are in the file.

Everything else re-verified clean at this head: `failuresMirrorDeploy` monotonicity and all five `resolutionFailed` arms, output-keys-only in `failedOutputKeys`, merge inputs still argument-identical to `deploy-engine.ts:3168-3174`, `hasChanges` still the deploy's own predicate, null-prototype bags on every hop, `previousExportNames` / `resolvedExportNames` as copies so the v9 `Fn::ImportValue` binding set is untouched, no persistence reader on the preview path, a carried key never rendering as an unredacted `ADD`, and `--fail` propagation unchanged by the early-return removal. Both new gate clauses are genuinely non-vacuously tested, both `keptWholeReasonText` reasons are now exercised on the diff path, and the integ phase asserts the new warning's presence alongside the suppression warning's absence rather than being relaxed.

Push B1, B2, m5 and m6 and I will re-review and merge.

nix-tkobayashi:
Follow-up push: `96619870` -> `adc2c699`, wording only (no behaviour change). A final Codex pass over the PR body and the maintainer-checklist pass after it found three overbroad sentences, now corrected in the body and in the branch:

- **`--strict-getatt`.** The `computeStackDiff` comment said the deploy "aborts on the failed output instead of merging". Only a THROWN resolution reaches `handleOutputResolutionFailure`'s strict arm; an output resolving to top-level `undefined` is stored and still reaches the merge, so the sentence now says exactly that. The same overclaim sits in two pre-existing `deploy-engine.ts` comments, and `.claude/rules/architecture.md` errs the other way (an Output refusal lets the deploy exit 0, with no `--strict-getatt` exception). The underlying gap (the flag does not refuse such an output) is filed as go-to-k/cdkd#3168, since deciding the flag's contract is yours and this PR does not touch `DeployEngine`.
- **The wider `isUnresolvedValue` arms.** The `failuresMirrorDeploy` docstring, the comment beside `failedOutputKeys.add`, the `docs/cli-diff.md` bullet and the symbol case's test name said the deploy WRITES such a value. It does not always: a top-level symbol is lost when the bag is serialized, and a structural `Fn::Sub` failure the best-effort diff keeps as a placeholder throws at deploy. They now say such a value is not the deploy's failure signal and so says nothing about what the deploy does with the key. The gate is unchanged; the suppression it keeps is still the safe direction.
- **PR body.** "replacing three copies" of the `computeOutputsDiff` call is corrected to two.

Verification on `adc2c699`: still one commit, on `c4f7ab5e`; `main` has since gained go-to-k/cdkd#3166, which touches no file this branch changes, so the branch was not rebased. `vp test run` passed 1021 files / 23130 tests with no type errors; `vp run check`, `vp run typecheck:test` and `vp run build` are clean. The real-AWS `output-never-resolved-diff` integ passed again on this tree in 392 s (destroy 2 deleted, 0 errors, 0 surviving state versions, no leftover resources), recorded in the ledger by this amend.

go-to-k:
Read the wording pass at `adc2c699` — the three corrections are right, and the second one is a correction to text I had verified and passed. "Such a value is not the deploy's failure signal, so it says nothing about what the deploy does with the key" is the accurate framing; a top-level symbol lost to serialization and a structural `Fn::Sub` that throws at deploy are both cases my reading of the old wording would have got wrong. Filing the `--strict-getatt` contract as go-to-k/cdkd#3168 rather than deciding it inside this PR is the right split.

That push is comment-and-docs only, so it does not move my round-2 review: https://github.com/go-to-k/cdkd/pull/3162#issuecomment-5678464231. It landed at 10:14Z, before both of these pushes, and I have not seen a response to it. Confirmed against `adc2c699` just now, so this is the current state rather than a stale read:

- **B1** — `src/cli/commands/diff-recursive.ts:819` still gates on `Object.keys(template.Conditions ?? {}).length === 0`. Outstanding.
- **B2** — `diff-recursive.ts`'s merged-branch comment still asserts no value becomes newly printable, and `legacyRecord` is unchanged. Outstanding.
- **m5** — `diff-recursive.ts:873` still branches the cause sentence on `merge?.kind === 'kept'` alone. Outstanding.
- **m6** — `src/analyzer/outputs-diff.ts:486` still interpolates `outputKey` and `output.Condition` unstripped. Outstanding.

Please take B1, B2, m5 and m6, or push back on any of them if you think the reading is wrong — B1 in particular rests on a measurement rather than an argument, so if the synth probe in that comment is not representative of what cdkd sees, say so and I will re-measure. Either way I need one of those two before the next round; right now the PR is blocked with nothing moving.

nix-tkobayashi:
Sorry, I missed your round-2 review: it landed before I armed a watch on this PR, and my follow-up only covered the wording pass. Taking all four, plus n7:

- **B1** — your synth measurement stands; I have no reason to think cdkd sees a different template for an env-agnostic app. Replacing the blanket `Conditions` clause with a predicate over the routes that can reach an output (a declared output carrying `Condition`, or `Fn::If` / `{Condition: …}` outside `Resources`), keeping the `parametersBound` half, with an env-agnostic `CDKMetadataAvailable` unit template and an env-agnostic arm in the live fixture.
- **B2** — forcing the legacy verdict on `merged` with carried keys when the template holds a secret reference and the stored bag holds no secret expression, correcting the sentence, and adding the `Fn::GetAtt`-to-a-stored-expression case.
- **m5** — branching the cause sentence on the resource diff.
- **m6** — stripping `outputKey` and `output.Condition` on that debug line.
- **n7** — stripping `stackName` at all three warn sites, and fixing the `no-change-outputs-merge.ts` caller comment.

I will reply here when the push is up, with verification.

nix-tkobayashi:
Pushed as `7f298fd4`. All four taken, plus n7.

**B1 — taken.** The blanket clause is replaced by `templateLetsConditionsReachOutputs` (`src/analyzer/outputs-diff.ts`), true when a declared output carries its own `Condition`, or when an `Fn::If` or `{Condition: …}` appears in any top-level section other than `Resources`. `parametersBound` is kept. One departure from your wording: the scan also skips `Conditions` itself. A `{Condition: …}` there composes a verdict, and it reaches an output only through a reference the scan already sees, in an output or a mapping; scanning it would refuse every template whose conditions compose each other. Coverage:

- unit cases on the predicate, true for each route (the output's own `Condition`, `Fn::If` in its value, `Fn::If` in a mapping, `{Condition}` in another section) and false for a resource's own `Condition`, an `Fn::If` in a resource property, a `{Condition}` inside `Conditions`, a multi-key object whose first key is `Condition` (so only a real reference counts), and no route;
- through `computeStackDiff`: a representative `CDKMetadataAvailable` shape (two regions instead of CDK's full list, and a placeholder `Analytics` value) (the condition plus the `AWS::CDK::Metadata` resource gated on it) previews the merge, and the output's own `Condition` and a mapping `Fn::If` each keep the suppression;
- live: phase 5c runs its diff a second time env-agnostic (`CDKD_TEST_ENV_AGNOSTIC=true` drops `env` in `bin/app.ts`) and asserts the same rc, row and warning. A synth premise fails closed unless the env-pinned template declares no conditions, the env-agnostic one declares exactly `CDKMetadataAvailable` gating only `AWS::CDK::Metadata`, and nothing else differs between them. It also rejects an output `Condition` or an `Fn::If` / `{Condition}` outside `Resources` and `Conditions` in the env-agnostic template directly, so an identical route in both templates cannot pass it.

The docs, the changelog entry and `.claude/rules/layout-analyzer.md` also name the refusal that was already there when condition evaluation is skipped because a condition depends on a secret-valued parameter.

**B2 — taken, with a narrower exoneration than proposed.** `computeOutputsDiff` gains `forceLegacyRecord`, and the `:839` sentence is replaced by why a carried key cannot exonerate. A Codex round found two records the proposed condition (`carriedKeys.length > 0 && templateHasSecretReference && !bagHoldsSecretExpression(currentState.outputs)`) still leaks, both legacy on the fully-resolved path:

- a secret expression stored under ANOTHER key (`{F: <plaintext>, G: "old-g", H: "{{resolve:secretsmanager:…}}"}`). Pass 1's veto excuses only the key holding the expression, so `F`'s resolved value still marks the record legacy, but the bag-wide `bagHoldsSecretExpression` turned the forced verdict off and `G`'s old value printed;
- no secret reference in the template at all, when the expression `F` resolves to lives only in a stored attribute (or a parameter, or another stack's outputs), so `templateHasSecretReference` is false.

Where such an expression can come from is not enumerable, so the verdict is now forced whenever a carried key's stored value is anything but a secret expression, which is the one value pass 1's veto would have excused for that key. The cost is that rows beside a carried plaintext withhold their old values on stacks that handle no secret; the change itself still renders. The go-to-k/cdkd#1948 residual note says it reaches a rendered section on this path only when every carried key's stored value is a secret expression. `computeOutputsDiff`'s docstring, `docs/architecture.md`, `docs/cli-diff.md` (including its `--json` note on `oldValueRedacted`) and the rules note describe the forced verdict. The stand-in stays `<redacted: may be legacy plaintext in state — run \`cdkd scrub\`>`, per `OutputChange`'s one-boolean decision; on a stack with no secret, `cdkd scrub` then finds nothing to repair. Tests: your `Fn::GetAtt`-to-a-stored-expression case as a premise (it resolves, and the record is legacy through `desired`); its failing twin (the same template with the attribute absent from the record, so `F` is carried and `G`'s old value withheld) together with both leaking shapes above; two carried keys, a secret expression first and a plaintext second, so the verdict reads every carried key and not only the first; and the two cases where the verdict stays off (nothing carried, and a carried key whose stored value is a secret expression).

**m5 — taken.** The cause sentence branches on the resource diff: `kept` names the merge's reason; with a resource change pending it keeps "usually an output referencing a resource this deploy has yet to create"; otherwise it says the stack has no resource change and this diff cannot preview what the deploy does with the outputs that failed. Both refusal cases above assert the pending-resource sentence is absent.

**m6 — taken.** Both values are stripped; a condition-false output with an escape in its key and in its condition name pins it.

**n7 — taken.** `stackName` is stripped in both Outputs warnings and in both render headers (`Stack` and `Nested stack`), each with a test. `keptWholeReasonText`'s doc now names both callers.

Verification on `7f298fd4`: still one commit, on `c4f7ab5e`. `vp test run` passed 1021 files / 23153 tests with no type errors (two earlier runs each lost one generator test in `tests/unit/scripts/` to a 60 s timeout under 16-17 peer vitest processes, and one lost a worker; both files pass alone and this branch touches neither). `vp run check`, `vp run typecheck:test` and `vp run build` are clean. Three Codex delta rounds plus a maintainer-checklist pass converged to NO ISSUES. Twenty mutation probes over the new clauses: nineteen turned a test red, and the twentieth, dropping the forced verdict's `typeof stored === 'string'` check for `String(stored)`, is equivalent, since a non-string stored value whose string form reads as a secret expression carries that expression into `desired`, where pass 1 already marks the record legacy. The real-AWS `output-never-resolved-diff` integ passed in 421 s on this tree, including the env-agnostic premise and diff: destroy 2 deleted, 0 errors, 0 surviving state versions, no leftover resources.

go-to-k:
Round 3, reviewed at `7f298fd4` — 3-axis plus the security add-on. I had to read that commit off your fork's branch, for the reason in M8 below; the PR itself still points at `adc2c699`.

**B1 is closed, measured rather than argued.** The spec axis re-ran my synth probe against this head: an env-agnostic stack with `aws:cdk:version-reporting: true` yields `Conditions: ["CDKMetadataAvailable"]` with `Resources.CDKMetadata.Condition` set, and `templateLetsConditionsReachOutputs` answers `false` — the preview runs for the default app shape. The `Resources` skip is sound for the reason your comment gives, verified in the resolver rather than from the comment: the diff's resolver binds `resources: currentState.resources`, `Ref` / `Fn::GetAtt` resolve from those records, and the one read of `context.template.Resources` is exempted under `bestEffort`.

**Your departure on `Conditions` survives a refutation attempt.** No template could be constructed where a condition's definition changes an output's value without an output-level `Condition` or an `Fn::If` / `{Condition}` outside `Resources`. One fact worth having, since it changes nothing but corrects the record: CDK's `CDKMetadataAvailable` is pure `Fn::Or` / `Fn::Equals` and contains no `{Condition}` or `Fn::If`, so the non-skipping form would NOT have refused the default app either. The `Conditions` skip buys composed-condition stacks; the B1 fix came from the `Resources` skip. Your code does not claim otherwise, so this is not a drift.

**B2 — keep the broadening. I am not asking you to narrow it.** The code axis proposed gating the force on a union (`templateHasSecretReference` ∪ a stored resource attribute holding an expression ∪ `imports` / `outputReads`) and argued your "not enumerable" line is overstated. I am declining that, for two reasons. The union still misses a parameter value carrying an expression, which `computeStackDiff` does not hold a complete view of — so it is not a sound gate, only a narrower one. And this predicate has now had two narrowing attempts, mine and the one before it, each of which a later round proved leaked; a third is the point at which the method, not the instance, is wrong. The security axis independently verified the current force closes both shapes the Codex round found and could construct no third: carried export alias in either polarity, a carried key absent from `storedOutputs` (impossible — carry requires `hasOwn(persisted, key)`), a container stored value holding an expression (forced, correctly stronger than the leaf veto), and both render paths withholding.

Two required items, both in the merged-path warning rather than in the gate.

**M8 — required before merge, and first. The PR has not picked up this push.** `refs/pull/3162/head` still resolves to `adc2c699`, `gh pr view` reports `mergeable: UNKNOWN`, and `7f298fd4` has **zero** check-runs — CI has never run on the commit your comment describes. Your fork's branch does carry it (`7f298fd4`, committed 12:29:25Z), so the commit is fine and only the PR sync is stuck. Please re-push so the PR points at it and CI runs; I cannot bind the review marker or merge against a head the PR does not name, and the green checks currently shown belong to the previous commit.

**M9 — required before merge. The merged-path warning claims something untrue of part of the list.** `src/cli/commands/diff-recursive.ts:870-875` enumerates every member of `resolved.failedOutputKeys` and says they "are compared at their stored values". A failed output with no stored value is not carried — `no-change-outputs-merge.ts:182` carries only under `hasOwn(persisted, outputKey)` — so it is absent from the merged bag, renders no row, and is compared against nothing. When no failed output has a stored value the sentence is wholly false, and `carriedKeys` is empty, so nothing else in the output corrects it. Please split the list: name the carried ones as compared at their stored values, and say the rest were dropped.

**M10 — required before merge.** Fold the reason for the withholding into that same warning. On a stack that handles no secrets, one persistently failing output with a stored value now withholds every MODIFY/REMOVE `old:` behind `<redacted: may be legacy plaintext in state — run \`cdkd scrub\`>`, forever, and `cdkd scrub` exits 0 finding nothing — the repo already ruled that outcome out in `no-change-outputs-merge.ts:80-83`. Keeping the force is right; leaving the placeholder as the user's only signal is not. Say in the warning that previous values are withheld because a failed output was compared at a stored value, so the reader has something to act on that is true on a secret-free stack.

**m11 — optional (take-or-leave).** `no-change-outputs-merge.ts:206` pushes export ALIASES into `carriedKeys`, so an alias whose stored value is plaintext forces the verdict even when its output key's stored value is a secret expression. The behaviour is correct and the security axis verified both polarities; no test covers an alias as the forcing key.

**m12 — optional (take-or-leave).** The env-agnostic arm at `tests/integration/output-never-resolved-diff/verify.sh:1590-1600` asserts rc, row count, the `[~] Plain` anchor, the absent suppression and the carried warning, but not `EXPECTED_PLAIN_CHANGED` in the row, which the env-pinned arm does at `:1570`. Assert it for parity of the two diffs. While there: the m5 third-arm test asserts the positive sentence and `not.toContain('yet to create')` but not the absence of an `Export.Name` reason.

**n13 — optional (take-or-leave).** `diff-recursive.ts:855`'s `typeof stored === 'string' &&` is dead — `isSecretBearingReferenceString` already returns false for a non-string. Your mutation note calling it equivalent is right; it is redundant rather than untestable. Separately, `diff-recursive.ts:540` interpolates a template-controlled `declaredType` unstripped into a `logger.warn`; pre-existing and not this PR's, but it is the same class n7 just swept.

Everything else re-verified clean at this head: m5's three arms (the `kept` arm cannot collide with the pending-resource sentence, since `merge` is computed only when there is no resource change), m6 stripping both segments with the emitted line asserted, n7 at all four sites, `failuresMirrorDeploy` monotone with the merged arm unable to report zero failures, `storedOutputs[key]` always an own-property read, pass 1 pure so the short-circuit changes nothing else, null-prototype bags on every hop, the v9 `Fn::ImportValue` binding set untouched, no persistence reader on the preview path, and `--fail` propagation through nested children. The new predicate's tests cover all four routes and all five non-routes non-vacuously, the `CDKMetadataAvailable` case asserts a non-empty delta so the merge provably ran, all six `forceLegacyRecord` cases discriminate in both directions, and the integ premise was executed against mutants rather than read — it fails closed on a missing or empty template.

Push M8, M9 and M10 and I will re-review and merge.

nix-tkobayashi:
Thanks. Taking M8, M9 and M10, and all three optional items (m11, m12, n13).

**M8** — confirmed from here: the fork branch holds `7f298fd4`, but `gh api repos/go-to-k/cdkd/pulls/3162` still reports `head.sha` `adc2c699` with `mergeable` unknown, and `7f298fd4` has no check-runs. The M9/M10 changes go up as a new commit, and I will check that the PR's head moves to it and that CI starts before replying, rather than assuming the push synced.

I will reply here with the new head and verification.

nix-tkobayashi:
Pushed as `b7185ee5`, and the PR now names it: `gh api repos/go-to-k/cdkd/pulls/3162` reports `head.sha` `b7185ee5` with `mergeable` true, and the commit had 11 check-runs within half a minute of the push.

**M8 — taken.** See the line above; CI ran on this head, not the previous one.

**M9 — taken.** The merged-path warning now splits `resolved.failedOutputKeys` the way the merge carries them, by an own stored value in `currentState.outputs` (the `hasOwn(persisted, outputKey)` read at `no-change-outputs-merge.ts:182`): `Compared at their stored values: …` for the carried ones, and `Not in state, so left out of the comparison: …` for the rest. With no failed output stored, only the second part appears, and the no-stored-value case now asserts exactly that, so the `compared.length > 0` guard is pinned. The two-failure case pins both parts, with `A` stored and the escaped `B` not.

**M10 — taken.** When the legacy verdict is forced, the same warning ends with: "Previous values in this Outputs section are withheld because a failed output was compared at a stored value that is not a secret reference, so this diff cannot rule out legacy plaintext in state. That reason no longer applies once every failed output resolves, though other legacy-plaintext checks still can withhold them." (A Codex round caught a first draft that said the values show again once every failed output resolves, which is false when a resolved output then marks the record legacy through `desired`.) The four legacy-record forcing cases and the alias case assert the sentence's opening clause, through "not a secret reference", and both non-forcing cases assert its absence; the rest of the sentence is not pinned by a unit case. Phase 5c asserts it live on both diffs, since `AttOut`'s stored value is the marker's name, with the fragment pinned against `dist`.

**m11 — taken.** A case where the output key `F` stores a secret expression and its carried alias `S:F` stores a plaintext: the alias alone forces the verdict, `G`'s old value is withheld, and the alias's plaintext is absent from the rendered rows.

**m12 — taken.** The env-agnostic arm now also requires `EXPECTED_PLAIN_CHANGED`, and the m5 third-arm cases assert the warning names no `Export.Name` reason.

**n13 — taken.** The redundant `typeof stored === 'string' &&` is gone, and the list-valued case's comment now says it is a behaviour case rather than a pin of the force alone. The parameter-type warning at `:540` strips `stackName`, the parameter name and `declaredType`. The case declares the parameter `List<X` plus an escape plus `>`: any `List<...>` type still reaches the warning, so all three strips are pinned.

Verification on `b7185ee5`: still one commit on `c4f7ab5e`; `main` has since gained three commits, none touching a file this branch changes. `vp test run` passed 1021 files / 23155 tests with no type errors (the run before it passed every test and lost a worker, so it was re-run). `vp run check`, `vp run typecheck:test` and `vp run build` are clean. Three Codex delta rounds and a maintainer-checklist pass converged to NO ISSUES; the checklist pass asked for the no-stored-value pin, the list-type strip case and the changelog's left-out arm, all in this push. Twelve mutation probes over the round-3 clauses (both warning filters, both parts, the compared guard, the names' strip, the withheld sentence unconditional and never, the force reading only the first carried key, and the three parameter-warning strips) each turned a test red. The real-AWS `output-never-resolved-diff` integ passed in 286 s on this tree, asserting the split warning and the withheld-values reason on both phase-5c diffs: destroy 2 deleted, 0 errors, 0 surviving state versions, no leftover resources.

go-to-k:
Round 4, reviewed at `b7185ee5` — code, test and security. I left the spec axis out this round: the delta is warning text plus tests and touches nothing in the go-to-k/cdkd#3101 contract the round-3 spec pass verified.

**M8 is closed** — the PR names `b7185ee5`, reports `MERGEABLE`, and CI ran on this head. **M9 is closed**, and correctly: `isStored` and the merge's own carry test are the same `Object.prototype.hasOwnProperty.call` over the same object, the split matches the merge's output-key carry exactly, and `failed.length === compared.length + leftOut.length` by construction. **m11, m12 and n13 are closed**, each with a non-vacuous case — the alias-only forcing case genuinely discriminates (computed over output keys only, `F`'s secret expression would exonerate the bag in pass 1 and `G` would render `old-g`), and the `List<…>` parameter type was executed against `parameterTypeMayLoseSecretIdentity` rather than assumed, so all three strips are really reached.

**M10 landed, and the sentence it added is wrong in two shapes.** Both are defects in the wording I asked for, so this is my request coming back rather than new scope — but it has to be right, since the whole point of M10 was to stop the message misdirecting.

**M14 — required before merge. The sentence fires when nothing was withheld.** `withheld` (`src/cli/commands/diff-recursive.ts:852`) says the VERDICT was forced, not that any row carries a withheld value; `forceLegacyRecord` redacts only non-`ADD` changes (`src/analyzer/outputs-diff.ts:897`). On an ADD-only or empty section the warning still prints "Previous values in this Outputs section are withheld…" with nothing withheld. This is not a corner: the case pinned at `tests/unit/cli/diff-recursive.test.ts:2411` is exactly it — a plaintext stored value forces the verdict and the section is a single `ADD` row with no previous value at all. Gate the clause on the rendered result, `withheld && outputChanges.some((c) => c.oldValueRedacted === true)`.

**M15 — required before merge. The sentence's subject and the split's subject are different sets, and they contradict each other in one reachable shape.** The split is over `resolved.failedOutputKeys`, output keys only; `withheld` is over `merge.carriedKeys`, which also holds carried export ALIASES. The alias carry at `src/deployment/no-change-outputs-merge.ts:193-207` is NOT nested under the output key's `hasOwn(persisted, outputKey)` — that block closes at `:188` — so a failed output absent from state whose literal `Export.Name` is present and previously published carries the ALIAS alone. Then `compared` is empty, `leftOut` holds the output key, and the message reads "Not in state, so left out of the comparison: Out. … Previous values … are withheld because **a failed output was compared at a stored value**". Both the code and security axes reached this independently. Word the clause from what actually forced it — "because a value carried from state is not a secret reference", or name the alias explicitly. While you are there, "left out of the comparison" understates that same shape: the carried alias does render a row from a stored value, so the output key's absence is not the whole story.

**M16 — required before merge.** Pin the full sentence. Today `WITHHELD_REASON` (`tests/unit/cli/diff-recursive.test.ts:2808`) stops at "not a secret reference" and the integ fragment stops earlier still, so the tail is pinned by nothing — and that tail has already been wrong once, in the draft a Codex round caught. Since M14 and M15 rewrite the clause anyway, extend the constant to the whole sentence in the same push. Your own report named this gap before I did, which is the right way to hand one over.

**m17 — optional (take-or-leave).** The `leftOut.length > 0` guard has no unit pin — only `verify.sh` catches it. Mirror what you did for `compared.length > 0` by adding `not.toContain('left out of the comparison')` to the compared-only case at `:2427`.

**m18 — optional (take-or-leave).** Phase 5c asserts the withheld SENTENCE but never that `EXPECTED_PLAIN` is absent from `DIFF_5C`; the row-level effect is unit-covered only. Asserting the absence live would make the fixture prove the withholding rather than the announcement of it.

Not requests, recorded so they are not chased. The security axis found that `stripControlChars` does not remove `U+2028` / `U+2029` while `displaySafe` does, so an output name carrying one can still forge a line in the new warning, and that `diff-recursive.ts:584` and `:631` interpolate `stackName` and a raw `error.message` unstripped. Both belong to go-to-k/cdkd#2888, which already owns exactly this population and already names `outputs-diff.ts` and `diff-recursive.ts` as the two sites to look at first — this PR adds call sites to it, so no new issue. Do not widen the class here; that issue re-baselines two pinned suites and wants the per-site reading done together.

Everything else re-verified clean at this head: `withheld` computed once and feeding both the diff and the warning so rendered rows and the explanation cannot disagree, every interpolated segment of both name lists stripped and neither able to print a value, the counts numeric, no key reported as compared that the merge did not carry, the shipped tail not re-introducing the "values show again" claim the Codex round corrected, `:540`'s three strips reached by a real type, no persistence reader on the preview path, the v9 `Fn::ImportValue` binding set untouched, and null-prototype bags on every hop. The ledger row records a 286 s run of this head's fixture asserting the split warning on both phase-5c diffs, and the fixture's new read expressions were executed against mutants rather than read — the old wording, a left-out regression and a dropped withheld sentence are each rejected.

Push M14, M15 and M16 and I will re-review and merge. CI has one check still pending on this head; I will confirm it green before merging rather than chaining the two.

nix-tkobayashi:
Thanks. Taking M14, M15 and M16, and both optional items (m17, m18). The go-to-k/cdkd#2888 notes are left to that issue, as you say.

I will reply here with the new head and verification once it is pushed and CI has started on it.

nix-tkobayashi:
Pushed as `dc529dd2`: `gh api repos/go-to-k/cdkd/pulls/3162` reports `head.sha` `dc529dd2` with `mergeable` true, and the commit had 11 check-runs within half a minute of the push.

**M14 — taken.** The withheld-values sentence is now gated on the rendered result, `withheld && outputChanges.some((change) => change.oldValueRedacted === true)`, so an ADD-only or empty section never claims a withheld value. The case at `:2411` (a stored plaintext forces the verdict, the section is one `ADD`) now asserts the sentence is absent.

**M15 — taken.** The reason now names what forces it: "because a value carried from state for a failed output is not a secret reference". A new case carries only the alias (`Out` has no stored value, `S:Out` does and was published): `G`'s old value is withheld, the warning names `Out` under the second group and not as compared, and it carries the full sentence. The second group is renamed from "Not in state, so left out of the comparison" to "No stored value under their own names", which stays true when that output's alias still renders a row. `docs/cli-diff.md`, `docs/architecture.md` and the changelog entry follow.

**M16 — taken.** `WITHHELD_REASON` and phase 5c's `WITHHELD_WARN_FRAGMENT` now hold the whole sentence, and the fragment is still pinned against `dist`.

A maintainer-checklist pass over this round added two things in the same push: a case where the carried value is a secret expression, so the verdict is not forced, while pass 1 still withholds rows through another key, asserting the sentence stays silent (it pins the `withheld &&` half of the gate, which no other case did); and the changelog entry now names every value carried from state, an alias included, and the rendered-row condition.

**m17 — taken.** The compared-only case asserts "No stored value under their own names" is absent.

**m18 — taken.** Both phase-5c diffs now assert the withheld stand-in (`<redacted: may be legacy plaintext in state`, pinned against `dist`) and the absence of `"cdkd-neverres-plain-value"` quoted, which the changed value `"cdkd-neverres-plain-value-changed"` cannot match.

Verification on `dc529dd2`: still one commit on `c4f7ab5e`. `vp test run` passed 1021 files / 23157 tests with no type errors; `vp run check`, `vp run typecheck:test` and `vp run build` are clean. Codex delta rounds and the checklist pass converged to NO ISSUES. Eight mutation probes over the round-4 clauses each turned a test red: either half of the sentence's gate dropped, the sentence never shown, the no-stored-value filter inverted, that group dropped or made unconditional, and the compared guard and the any-carried-key force re-checked. The real-AWS `output-never-resolved-diff` integ passed in 290 s on this tree, asserting the full sentence, the withheld stand-in and the absent previous `Plain` value on both phase-5c diffs: destroy 2 deleted, 0 errors, 0 surviving state versions, no leftover resources.

go-to-k:
Round 5, reviewed at `dc529dd2` — code, test and security. **All three axes came back clean: no blockers, no minors.**

M14, M15, M16, m17 and m18 are all closed, and verified rather than taken on report. The M14 gate is not merely conservative but exact — `computeOutputsDiff` emits no `NO_CHANGE` rows and the renderer renders every entry, so `some(oldValueRedacted === true)` holds exactly when a rendered row really lost a value. The M15 wording is true in every shape, including the alias-only carry the previous wording contradicted, and `No stored value under their own names` is literally true for every member of that group since it is `!hasOwnProperty(storedOutputs, key)` by construction. M16's constant is byte-identical across all three copies — the source literal, `WITHHELD_REASON` and the integ fragment are each 317 characters and string-equal. Your checklist-pass addition earns its place: it is the only case pinning the `withheld &&` half of the gate, with M14's case pinning the other, so both arms are now covered.

Two things worth recording because they answer questions I had raised rather than ones you did. The `unaccountable()`-only withholding that leaves the sentence silent is **not** a regression and **not** a gap: the previous gate was `withheld` alone, so that shape was already silent, and the added conjunct can only subtract prints. It is also not a hole in M10's purpose, because every withheld row still renders the stand-in and `--json` still emits `oldValueRedacted`, and in that shape the `cdkd scrub` advice is apt unassisted — which is the opposite of the condition M10 exists for. Separately, a misattributed cause is unreachable: `withheld` implies `forceLegacyRecord` implies `legacyRecord`, so whenever the gate's left conjunct holds, the forced verdict is already a sufficient cause for every non-ADD row. The integ's new negative assertion is properly anchored — the quoted pattern cannot be satisfied by the changed value, verified by executing it against both samples rather than reading it.

One required item, and it is M15's own fix reaching only one of its sites.

**M20 — required before merge. The retired framing survives in two more places, one of them public.** M15 corrected "a failed output was compared at a stored value" because an alias-only carry contradicts it. That phrase is still in:

- `src/analyzer/outputs-diff.ts:52` — the `oldValueRedacted` doc comment, explaining the flag with the wording now known to be untrue in that shape;
- `docs/cli-diff.md:454` — a published page, stating the same reason to a reader who cannot check it against the code.

`docs/cli-diff.md:197` already carries the corrected framing ("every value carried from state that way"), so the page contradicts itself between two sections. Please sweep the phrase rather than the file — `grep -rn 'compared at a stored value' src/ docs/ .claude/rules/ changelog.d/` finds both, and the changelog entry is already correct. The test axis found the first; the public one is mine, and it is the reason this is required rather than a nit: a doc page is the one copy a user reads.

**m21 — optional (take-or-leave).** The new test's comment at `tests/unit/cli/diff-recursive.test.ts:2917` cites "review M19", which is not a label from this thread — the round-4 set was M14–M16 plus m17/m18, and the checklist item carried no label. Re-point it at whatever it means so the comment and the thread can be reconciled later. I am numbering this round from M20 to avoid colliding with that stray.

Not a request: `WITHHELD_ROW_FRAGMENT` (`verify.sh:1516`) stops before the `— run \`cdkd scrub\`` tail, unlike M16's whole-sentence constants. That asymmetry is fine — it avoids the em dash and backticks in shell, and the fragment is pinned against `dist`, so it cannot silently stop matching.

CI on this head still has `check-build-test` and `once-leak-detect` pending; I will confirm both green before merging rather than chaining the two.

Push M20 and I will re-review and merge.

nix-tkobayashi:
Pushed as `372681ba`: `gh api repos/go-to-k/cdkd/pulls/3162` reports `head.sha` `372681ba` with `mergeable` true, and the commit had 11 check-runs within half a minute of the push.

**M20 — taken.** Swept the framing rather than the file, and it was in more places than the grep you gave finds, because two spell it "compared a failed output at a/its STORED value" and one wraps across lines. Five sites now say a value carried from state for a failed output, an export alias included, is not a secret expression (or, on the doc page, a secret reference): `OutputChange.oldValueRedacted`'s doc comment and both `forceLegacyRecord` explanations in `src/analyzer/outputs-diff.ts` (the `computeOutputsDiff` docstring paragraph and the option's own comment), `docs/cli-diff.md`'s `--json` paragraph, and the forced-verdict sentence in `docs/architecture.md`. A Codex round caught the two `outputs-diff.ts` spellings after I had fixed only the two sites you named, and a line-wrap-aware sweep then found the `architecture.md` one. What remains quotes the retired wording as the thing not to say: the `computeStackDiff` comment explaining why the sentence changed, and the alias-only test's comment. `verify.sh`'s note that the warning names AttOut as compared at its stored value stays, since AttOut is an output key that really is.

**m21 — taken.** The comment no longer cites a label; it says the case closes a gap a maintainer-checklist pass found on the round-4 fixes.

Verification on `372681ba`: still one commit on `c4f7ab5e`. `vp test run` passed 1021 files / 23157 tests with no type errors, on a run with no peer vitest processes (three runs before it each passed every test and lost one `tests/unit/scripts/` generator test to a 60 s timeout or a worker to peer load, and this branch touches neither file). `vp run check`, `vp run typecheck:test` and `vp run build` are clean. Two Codex delta rounds converged to NO ISSUES; the first is the one that found the two `outputs-diff.ts` spellings. The change is comments, two doc pages and one test comment, so no mutation probes were added. The real-AWS `output-never-resolved-diff` integ passed in 397 s on this tree: destroy 2 deleted, 0 errors, 0 surviving state versions, no leftover resources.

go-to-k:
Round 6, reviewed at `372681ba` — spec, code and test. Security was not re-run: this delta changes no executable line, adds no sink, and that axis cleared the mechanism at the two previous heads.

**M20 is closed, and you swept it better than I asked.** Five sites against the two I named, including two alternate spellings and a line-wrapped one my grep could not have found. I ran my own residual sweep across spellings and with newlines joined, and the six remaining hits are all correct: `diff-recursive.ts:878` and `diff-recursive.test.ts:2943` quote the retired wording as the thing not to say, and `architecture.md:376`, the changelog entry and `verify.sh:1506,1596` describe the warning's GROUP NAMES, which is a different claim — `AttOut` really is an output key compared at its stored value. **m21 is closed.** The code axis confirmed mechanically that every one of the 20 changed `src/` lines is a comment continuation and that `diff-recursive.ts` and `no-change-outputs-merge.ts` are byte-identical to the previous head, so round 5's verified invariants carry over untouched, and that the rewritten comments are not over-corrected in the other direction ("an export alias included" is additive, so it does not claim only an alias can force the verdict).

The closing adequacy sweep came back complete: all ten decision points this change introduced — each of the merge gate's four conjuncts on its negative side, `templateLetsConditionsReachOutputs`'s routes and both exclusions, `failuresMirrorDeploy`'s arms, both polarities of `forceLegacyRecord`, both conjuncts of `explainWithheld`, the warning's two groups, and both keep-whole reasons — carry a non-vacuous case, with nothing resting on the integ alone and nothing resting on a mock where live proof would be stronger. The fixture is healthy: exiting traps with rc seeds, an env-agnostic arm that is synth-and-diff only so it creates no AWS resource, fail-closed synth premises, and no assertion that passes on empty output.

Two required items, both on the public page, and the first is the same failure mode as M20.

**M22 — required before merge. `docs/cli-diff.md:362` declares a closed enumeration that this PR makes wrong.** "Three refusal gates decide this:" heads a three-row table of what withholds a previous value. At this head `forceLegacyRecord` is a FOURTH, independent, record-wide trigger, OR'd ahead of pass 1 at `src/analyzer/outputs-diff.ts:869-871`. The page documents it twice elsewhere (`:197-201`, `:452-455`), but this is the section a reader lands on from "why was my value withheld", and it tells them the list is complete. `docs/architecture.md:391-406` does not have the gap.

Add the fourth row — trigger "the no-change merge preview carried a value from state, an alias included, that is not a secret expression", scope Record-wide — and **drop the numeral rather than updating it**: make the sentence "These refusal gates decide this:". That is the actual ask. This is the second enumeration on this one page to go stale in two rounds (round 5's `:454` was the first), and a bare count above a table is a second copy of the table's length that nothing keeps in step. Note the sentence two lines above, "two safeguards apply to that side of the output", is the same shape — check it still holds and give it the same treatment if it does not.

**M23 — required before merge. `docs/cli-diff.md:197` still binds the forcing condition to the wrong set.** "Unless every value carried from state **that way** is itself a secret reference" — "that way" refers back to the failed-output split in the preceding sentence, so it reads as "compared at its stored value". But an export alias carried with no stored value under the output's own name also forces the verdict, alone (`src/deployment/no-change-outputs-merge.ts:198-206`); that is exactly the case M15 corrected and `:454` and `architecture.md:404` now state. Use the same "an alias included" wording here. Fold in `.claude/rules/layout-analyzer.md:20`, which omits the alias case in one clause for the same reason.

**m24 — optional (take-or-leave).** Two wording residuals, neither worth a round on its own. The user-facing warning at `diff-recursive.ts:889` says "is not a secret reference" for a LEAF predicate, so a list-valued output holding an expression at depth makes the sentence assert something a deep reading contradicts — the withholding is still the conservative direction, so this is prose only. And the sweep edited inside comment blocks without re-wrapping them, leaving a 32-character orphan line at `src/analyzer/outputs-diff.ts:54`, a 107-character line at `:787` and the same artifact at `docs/architecture.md:403`; Oxfmt does not reflow comments, so nothing will catch it later.

Everything else verified accurate at this head: the gate list, the condition-reach wording, the warning's group names spelled identically across code and both pages, the forced verdict and its cost, the keep-whole refusals, the changelog entry at 1985 characters with no retired framing, and both rules files naming the right symbols and the right caller. The ledger row records a 397 s run that sits between the two commits, and the delta touches no fixture file, so that run exercised this head's fixture byte-for-byte.

Push M22 and M23 and I will merge. CI on this head has `check-build-test` and `once-leak-detect` still pending with nothing failing; I will confirm green before merging rather than chaining the two.

nix-tkobayashi:
Pushed as `2c48dae4`: `gh api repos/go-to-k/cdkd/pulls/3162` reports `head.sha` `2c48dae4` with `mergeable` true, and the commit had 11 check-runs within half a minute of the push.

**M22 — taken.** `docs/cli-diff.md`'s withheld-values section now reads "These refusal gates decide this:" with no count. The table gains a fourth row, "Carried value in the merge preview": the no-change merge preview carried a value from state for a failed output, an alias included, that is not a secret reference, record-wide. A sentence after the table says why that gate is record-wide. I checked "two safeguards apply to that side of the output" two lines above: it still holds, since the section's two bold paragraphs are the withholding (now four gates in one table) and the control-character stripping, so it keeps its number.

A maintainer-checklist pass over this round found a third enumeration of the same shape, now fixed in the same push: `docs/architecture.md`'s "Two rows the preview cannot decide from the template alone come from state" listed the `skippedOutputs` record and the literal `Export.Name` verdict as "One" and "The other". It now reads "Rows the preview cannot decide ... come from state", the second is "Another", and it names the third this PR adds: on a stack with no resource change, the merge preview carries a failed output's stored value, and its literal alias's, from state instead of resolving them.

**M23 — taken.** The preview paragraph at `:197` now says "Unless every value carried from state for a failed output, an alias included, is itself a secret reference", matching `:454` and `architecture.md`. `.claude/rules/layout-analyzer.md`'s clause says "a value carried from state, an alias included, that is not a secret expression".

**m24 — the re-wraps are taken, the warning reword is not.** The orphan line in `OutputChange.oldValueRedacted`'s doc comment, the long line in `computeOutputsDiff`'s docstring and the one in `docs/architecture.md` are re-wrapped with no word changed. I left the runtime sentence's "is not a secret reference" as it is. As you note, the withholding is the conservative direction, and changing the shipped string would move the pinned full-sentence constant, its integ fragment and the ledger run for a prose-only gain.

The source change in this push is comment re-wrapping only. The bundle keeps JSDoc text, so a rebuilt `dist` is not byte-identical and I could not show the difference is comments alone; `output-never-resolved-diff` was therefore run again on this tree rather than carried from the previous head (result below).

Verification on `2c48dae4`: still one commit on `c4f7ab5e`. `vp test run` passed 1021 files / 23157 tests with no type errors (the runs before it each passed every test and then lost a vitest worker, with no test file named in the error, so it was re-run). `vp run check`, `vp run typecheck:test`, `vp run build` and the rules payload fence are clean. Two Codex delta rounds converged to NO ISSUES, the second after the checklist pass's `architecture.md` fix. The change adds no clause, so no mutation probes were added. The real-AWS `output-never-resolved-diff` integ passed in 409 s on this tree: destroy 2 deleted, 0 errors, 0 surviving state versions, no leftover resources.

go-to-k:
Merged as `d758c852`. Thank you — seven rounds, and the thing worth naming is that three of them were spent on text rather than on the gate, and they were the right three: a warning that claimed outputs were compared at values they never had, a reason clause whose subject was a different set from the split above it, and a published page telling a reader its list of withholding gates was complete. The behaviour was settled by round 3; what took the rest was making the diff say something true about what it had just done.

Two things I want to record because they changed my mind rather than yours.

**On M1, you were right and I was wrong.** I derived that the deploy's save-time mixed-generation re-check could not diverge from the preview, reasoning about the preview's own bag. The divergence lives in the deploy's bag, which gains a late needle during the observed-capture drain — a mechanism an existing test already pinned. You took the item and reached the opposite conclusion with the evidence attached, which is the most useful possible response to a wrong review note. The second half your Codex round found, where a late needle matching a stored value makes the save redact it in whichever bag it keeps, is a route I had not considered at all.

**On B2, I declined the narrowing two reviewers wanted, and your instinct to go broader was the better one.** My proposed gate leaked twice; the union a later round proposed still missed a parameter carrying an expression. Forcing the verdict on any carried non-expression is coarser and costs real information on secret-free stacks, but it is the side to be wrong on, and after two failed narrowings the method was the problem rather than the instance.

Three more that are worth carrying forward. B1 was caught by measurement, not argument — synthesising an env-agnostic stack against this repo's own `aws-cdk-lib` is what showed the blanket `Conditions` clause disarming the fix for the default `cdk init` shape, and the fixture could not have caught it because it pins `env`. Your sweeps repeatedly beat the grep I handed you: M20 named five sites against my two, including two alternate spellings and a line-wrapped one, and in round 6 you found a third stale enumeration in `docs/architecture.md` before I did. And you deleted the counts instead of correcting them — "Three refusal gates" became "These refusal gates", "Two rows" became "Rows" — which is the fix that does not drift again.

I ran `output-never-resolved-diff` independently on my own account before merging rather than taking the ledger row on report: PASS, phase 5c showing the one `[~] Plain` row beside the failing `AttOut` with the deploy writing it and keeping `AttOut`, destroy 2 deleted / 0 errors, 0 surviving state versions.

The two pre-existing doc claims the round-7 sweep surfaced are filed as go-to-k/cdkd#3184 — the anti-drift fence described as three assertions where it holds five, and a literal `Export.Name` described as omitting the Outputs section unconditionally. Neither is yours; both were in the tree at this PR's base.

Issue #3171, opened 2026-09-15
fix(resolver): AWS SDK error text that quotes a name assembled around a 1-3 character secret prints it past the name's log twin


nix-tkobayashi:
## What is wrong

go-to-k/cdkd#3150 masks the NAMES the resolver prints itself (map keys, stack / output / export names, regions, attribute names, the parts of a dynamic reference) through the log twin the pass registered when an `Fn::Sub` assembled the name around a secret shorter than `MIN_NEEDLE_LENGTH` (4). Text the AWS SDK produces is a different route: an SDK error quotes the name it refused back, and that text reaches the needle mask only, whose substring arm has the four-character floor. So a name the resolver's own line prints as `missing-***` can still print as `missing-q7` a few characters later, in the SDK's words.

Reproduced by Codex review against the go-to-k/cdkd#3150 branch, each with `{"Fn::Sub": ["<name>-${P}", {"P": <a secret reference resolving to "q7">}]}` feeding the name:

| Route | Where | What prints |
| --- | --- | --- |
| a dynamic-reference lookup's rejection, re-thrown as it is | `resolveSSMReference` / `resolveSecretsManagerReference`, the `sendWithThrottleRetry` calls | `ParameterNotFound: missing-q7` (a thrown message) |
| `withRetry`'s per-attempt debug line for those lookups | `sendWithThrottleRetry` passes `this.logger` unmasked, and `src/deployment/retry.ts` interpolates the error message into the line it prints before each retried attempt. (Its give-up warning is not reachable from these lookups: their `isRetryable` override leaves every counter that gates it at zero.) | `Retrying ssm:missing-*** ... - ... missing-q7` |
| the CloudFormation-fallback warning's message suffix | `lookupCfnStackOutputs` masks the caught `DescribeStacks` message with `maskSecretsForLog`, which cannot see a sub-floor secret inside the stack ARN the SDK quotes. (`lookupCfnExport` is not a route: `ListExports` sends no template-derived name.) | `stack-q7` inside an access-denied message |
| the `Fn::GetAZs` describe-failure message's suffix | `resolveGetAZs` | `failed to describe availability zones for region 'us-west-***': ... us-west-q7` |

The label of each line (the name the resolver prints itself) is already masked; only the SDK's text is not.

A literal-equal form is NOT this issue: a template that spells the secret-bearing name literally already holds the text.

## Why it was not fixed in go-to-k/cdkd#3150

[...]

Issue #2827, opened 2026-09-08
fix(resolver): the resolver masks a resolved value when it LOGS it and leaves it bare when it THROWS, so the plaintext travels to every caller


go-to-k:
## Summary

`IntrinsicFunctionResolver` masks a resolved value when it LOGS it and leaves it bare when it THROWS. The throw is the one that travels: it crosses into every caller, and a caller that masks at its own boundary can only mask against ITS bag. The sweep for issue go-to-k/cdkd#2803 found the same shape at fourteen throw sites and four `logger.warn` sites in one file.

Issue go-to-k/cdkd#2803 fixed the consumer end for `cdkd import`, and go-to-k/cdkd#2728 did the same for `DeployEngine.handleOutputResolutionFailure`. Both are boundary masks over an error the resolver built unmasked. This issue is the producer end.

## The clearest instance

`src/deployment/intrinsic-function-resolver.ts`, the SSM lookup:

```ts
throw new Error(`Dynamic reference: SSM parameter '${parameterName}' not found or has no value`);
```

`parameterName` comes from the ASSEMBLED reference string, so it carries a plaintext whenever an `Fn::Sub` / `Fn::Join` built the parameter path out of a value the same pass resolved from a secret. What makes this one decisive rather than arguable: the debug line a few lines above and the retry label just below BOTH run that same value through `maskSecretsForLog`. The throw is the sole unmasked consumer of a value its own neighbours treat as a needle.

## The rest of the class

Throws interpolating a value that came back from `resolveValue` or a live lookup:

- `Fn::ImportValue: export '<name>' not found in any stack` -- the export name is the RESOLVED argument, and the file's own comment nearby states that an export name assembled from a secret IS a resolved secret, then masks it for logging and not here.
- Four `Fn::GetStackOutput` throws (`output '<name>'`, `stack '<name>'`, and siblings) -- both resolved; masked for logs a few lines away, unmasked in the throws.
- `Fn::GetStackOutput: '<region>' is not a valid AWS region name` -- resolved region, truncated to 64 chars, not masked.
- Three `Fn::FindInMap` throws (mapping / top-level key / second-level key) -- all three resolved.
- Two `Fn::GetAZs` throws (`'<region>' is not a valid AWS region name`, `failed to describe availability zones for region '<region>'`) -- region derived from `resolveValue`.
[...]

PR #3176, opened 2026-09-15
fix(resolver): mask a sub-floor secret assembled into an intrinsic's name argument through its log twin

### LINKED ISSUE #3150: fix(resolver): name arguments still print a sub-floor secret on resolver log lines and messages after the log twin

nix-tkobayashi:
Closes #3150

## Summary

go-to-k/cdkd#3100 gave the resolver a masked LOG TWIN of each string a write put a recorded secret into, and go-to-k/cdkd#3114 routed the value lines through it. The NAME arguments of the intrinsics were still masked by the needle mask alone, which matches a value under `MIN_NEEDLE_LENGTH` (4) only as the whole text, so `{"Fn::Sub": ["env-${P}", {"P": <two-character secret>}]}` in a map key, a stack / output / export name, a region, an attribute name or a part of a dynamic reference printed `env-q7`.

Routing each site through the twin one at a time kept surfacing further sites, all of the same shape: a raw name handed to `maskSecretsForLog`. So the twin lookup now lives in `maskSecretsForLog` itself, and only names that are TRANSFORMED or COMPOSED before they reach it need a site-level fix.

## Design

| Part | What it does |
| --- | --- |
| `maskSecretsForLog` | Looks up the pass's registered twin first (own bag, then inherited, as `registeredLogTwin` already did for go-to-k/cdkd#3114). No twin: the needle mask as before. A twin: the twin as it is (the raw text with masked spans, so it holds no recorded needle the raw text lacks), or `***` when the needle mask also changes the raw text (the rule `logTwinText` applies to a Join / Sub line). |
| `maskNeedlesForLog` | The previous body, needle mask only. Used by the two detectors that ask the needle mask apart from the position mask (`logTwinText`, `resolveBase64`), so their verdicts are unchanged. |
| `regionLogText` | A region lowercased by `canonicalizeRegion`: the raw value's twin, lowercased, or `***` when lowercasing forms a recorded needle the raw case hid. Used by the `Fn::GetStackOutput` / `Fn::GetAZs` lines (and `lookupCfnStackOutputs`' warning), the `Fn::GetStackOutput` origin, and the producer-region resolver line, which `reresolveCrossStackValue` now hands the caller's region log text. |
| region-scoped clients line and refusal | `clientsForRegion` prints a region's log text when its caller built the region from a template: `Fn::GetAZs` passes it, and a region-pinned producer resolver keeps the masked text its creator held for its own region, masked, stripped of control characters and masked again; its creation line prints the same text. The resolver is cached per canonical region, so a later call whose text masks differently switches the cached resolver's text to `***`, which it keeps from then on. The same text is what the invalid-region refusal prints, which a secret ARN's region reaches with no `isClientSafeRegion` gate in front of it (`us-west-2_${P}` printed `us-west-2_q7` on `origin/main`). |
| `outputNameLogText` | A nested stack's output name sliced off `Outputs.`: the whole name's twin, sliced, or `***` when a mask covers part of the prefix. Used by the nested-stack refusal's missing name and its list of declared outputs. |
| invalid-region refusals | At the `Fn::GetAZs` / `Fn::GetStackOutput` gates, strip-and-cut: `***` whenever the value's twin differs from it. The region-scoped clients refusal is in its own row above. |
| `reresolveCrossStackValue`'s `origin` | Composed from each name's log text, including the `Fn::ImportValue` producer stack names and regions (the latter also masked on the index / scan success lines and the `No state found` line). |
| `dynamicReferenceNameLogText` | `resolveSub` / `resolveJoin` re-enter the dynamic-reference loop with the assembled string and its twin. Each token is paired with the twin's token by position (the twin's tokens come from `secret-redaction.ts`'s shared `dynamicReferenceTokens`, so the go-to-k/cdkd#1936 fence still sees one token pattern in the resolver), and each name parsed out of it (service, secret id, JSON key, version stage / id, SSM parameter name, ARN region) prints the twin's run over the same `:`-separated pieces through `straddleSafeTwin`, including a name assembled inside the same `Fn::Sub`; the refusals print the token's twin the same way. `straddleSafeTwin` prints the twin unless the needle mask also changes the raw text, and then only when the needle-masked raw text IS the twin (`/probe/***` for a 4+ character secret held whole); when they differ a recorded secret straddles a masked span (`q7ab` across `id-***ab`) and it prints `***`. A name whose runs carry different twins prints `***`, and so does every name parsed out of a token whose pieces cannot be paired with the twin's (a mask covering a `:` or a brace). A refusal's token text is the paired twin token, so a masked `:` still prints `{{resolve:ssm-secure:pub-***}}`; it is `***` when the twin holds no token to pair (a masked brace), or when `straddleSafeTwin` finds a recorded secret straddling a masked span in it. A name that is no run of the token is a default the caller synthesized (`AWSCURRENT`, or an absent version id) and prints as it is, unless the token spells it inside a longer piece, where it prints `***`. A token delegated to a region-pinned sibling resolver is handed over as itself, with the parent's straddle-guarded token log text and name mapping passed separately for display, so the sibling prints the same verdict while the twin it registers for its result is unchanged. |

Behaviour change beyond the issue: the registry is keyed by value, so a string equal to one masked elsewhere in the pass now prints masked on its own line too (the over-masking direction the registry already documents). One go-to-k/cdkd#3100 expectation is updated for it.

Review: a code and a security reviewer each found that a parsed dynamic-reference name printed the twin's run as it was, so a longer recorded secret overlapping the mask showed its unmasked part (`prod-***` where `origin/main` printed `***`); both paths now go through `straddleSafeTwin` (a first fix through `logTwinText` over-masked an SSM name the go-to-k/cdkd#2728 suite pins as `/probe/***`). The security reviewer also found the unmasked region-scoped clients line and the lowercasing case above; the test reviewer asked for the whole-secret attribute name, the `Fn::ImportValue` deferred refusal displays, the state-backed `Resolved Fn::GetStackOutput` line, controls, a region-matching state mock, and a positive `Fn::Base64` pin beside each negative one. All are addressed.

Maintainer review round 1: the invalid-region refusal in `clientsForRegion` printed a producer-region guest's region raw although a secret ARN's region reaches it ungated, and its note claimed every such region was operator-typed; it now prints the guest's masked text and the note names what else reaches it. `dynamicReferenceNameLogText`'s fallback for a name that is no run of the token fails closed where the token spells the name, both lookup helpers take the name mapping as a required parameter, and unrecorded-value controls, a whole-message assertion and a `loggedRegionText` rename were added.

Maintainer review round 2: the `Using a producer-region resolver for` line still took a single `maskSecretsForLog` of the text the guest stored through `maskThenStripThenMask`, so a literal region splitting a recorded secret with U+0001 printed it on that line while the refusal masked it; the line now prints the stored text. A guest reused for a second spelling of its region prints `***` once the two spellings mask differently, `resolveSSMReference`'s unreachable parameter defaults are dropped, and a version stage / version id control was added.

Resolved and persisted values are unchanged: neither detector's verdict moves, and nothing new is registered. That constraint is load-bearing -- the log-twin registry also decides what `Fn::Base64` records, so an earlier revision that registered a canonical region's twin would have persisted a plain `{"Fn::Base64": "us-east-1"}` as `***`. Two cases pin it.

Not covered: a template LITERAL equal to a dynamic-reference token masked earlier in the pass still prints as written (the template itself carries that text, so nothing is disclosed that the template does not already hold).

Two contrived bounds a security re-review named, neither printing more than `origin/main` did (which printed the whole name) and left as stated bounds rather than changed:
- a region handed to the region-pinned sibling as a parsed name is lowercased without `regionLogText`'s lowercase-formed-needle guard, so a 4+ character secret that only lowercasing forms, straddling a mask, can show its unmasked part on the producer-region line;
- tokens are paired with the twin's tokens by count, so a mask removing one `{{resolve:` opener while another mask forms a token in the twin could mispair them.

Also not covered, filed as go-to-k/cdkd#3171: text an AWS SDK produces that quotes a name back (a `ParameterNotFound` rejection re-thrown as it is, `withRetry`'s per-attempt retry line for the dynamic-reference lookups, and the `DescribeStacks` fallback warning and `Fn::GetAZs` message that append the SDK's message). Such text reaches the needle mask only, whose substring arm has a four-character floor. One unit fixture sits on that line: the SSM mock answers a `param-` name with `{ Parameter: {} }`, which real SSM does not return (it rejects with `ParameterNotFound`). That shape is what makes `'param-***' not found or has no value` assertable whole; against real AWS the same template reaches the SDK-text residual in go-to-k/cdkd#3171, so the fixture is kept rather than changed.

## Tests

- `tests/unit/deployment/intrinsic-resolver-name-argument-log-twin.test.ts` (new, 104 cases, real resolver): each site family with an `Fn::Sub`-assembled two-character secret, asserting the whole sentence or line prefix with `***` at the secret's position, with controls where the same shape carries an unrecorded value. Includes the sites behind a literal attribute-name check (the check compares the resolved name, which `Private` + a secret `Ip` spells), the dynamic-reference names assembled in an inner and in the same `Fn::Sub`, the ambiguous-region and `ssm-secure` refusals, a secret carrying `:` or a `{{` (every parsed name `***`; a refusal prints the whole token as `***` only for `{{`, which leaves the twin no token to pair, and the token's twin for `:`), two tokens of which the twin holds only one, the producer-region resolver line for a `Fn::GetStackOutput` region, a secret ARN's region and an uppercase `Fn::ImportValue` producer region, a foreign-region secret ARN whose names cannot be paired (the sibling resolver keeps them masked), the `Fn::ImportValue` producer stack names and regions, the ambiguous-region refusal's producer-region list, and the `Fn::GetStackOutput` origin's default region -- the last two for a region equal to a string the pass assembled. Against `origin/main`'s resolver 88 of the 104 fail; the 16 that pass are the 13 unrecorded-value controls, the compile-time pin that both lookup helpers require the name mapping (checked by `vp run typecheck:test`, not at run time), the sliced-output `Fn::Base64` persistence pin and the literal-token needle case, the last two pinning behaviour `origin/main` already has. Since the maintainer's first review the file also covers the region-scoped clients refusal (masked, two mask-order cases, and two controls), the synthesized-default fallback, a guest reused for the same and for a differently masked spelling of its region, and whole-message assertions where it asserted a fragment. Three further cases pin that `Fn::Base64` records nothing new: for a literal equal to a masked region, to a parsed name, and to the public value a region-pinned sibling resolved.
- `tests/unit/deployment/intrinsic-resolver-sub-floor-log-mask.test.ts`: the one expectation above (the file's only case that fails against `origin/main`), and a case pinning `resolveBase64`'s needle operand (a literal input embedding a 4+ character recorded secret is recorded mask-only), which no suite pinned before.
- `tests/unit/cli/import-resolver-error-masking.test.ts`: the go-to-k/cdkd#2803 case that PINNED the residual go-to-k/cdkd#2827 stopped at (a sub-floor plaintext inside a longer JSON-key segment, `key-${Pw}`, still printed) now fails as intended, so it is flipped to `CLOSED by issue #3150` and asserts `key 'key-***' not found`: the JSON key parsed out of the token prints the twin's run. The needle floor itself is unchanged.
- `scripts/check-resolver-mask-coverage.ts` / `tests/unit/scripts/resolver-mask-coverage.test.ts`: pinned counts `maskedExprs` 161 -> 167 and `markers` 109 -> 103 (six `not-in-class` notes retired by the masks they described as absent), and the masks floor moved 155 -> 161 with the population: the suite's block-comment injection at `resolveFindInMap` left 157 masked expressions, which the old floor no longer caught.
- Mutation probes, one change at a time with the file restored between them, run against both files above: the central lookup and its both-masks arm, `logTwinText`'s detector, `regionLogText` and each of its uses, the producer-region log text override and both of its arguments, both invalid-region arms, `outputNameLogText` and its two uses, each component of the four `origin` builders, each `Fn::ImportValue` producer-region mask, the ambiguous refusal's per-region mask, and in the dynamic-reference loop the token pairing, the unpaired and conflicting-run arms, the run mapping and its fallback, the sibling hand-off (its twin, the display argument, and the loop's use of each half), both refusals' token, and each name's use -- 79 probes, every one reds at least one case. After the maintainer's first review, 11 more over the new clauses (the refusal's printed region and each of its arms, both masks around the strip, the fallback's three guards, over-masking of an unmasked name, and the nested refusal's attribute name): 10 red at least one case, and the `break` after a matched run survives as a cost-only change. Two compile-time probes restoring each helper's old identity default make `vp run typecheck:test` fail. The maintainer's second review added four more, all red: the creation line reverted to the single mask, a reused guest keeping its first spelling's text, a cache hit masking the guest unconditionally, and a cache hit comparing the unmasked region text. One mutant is EQUIVALENT and survives by construction: swapping `resolveBase64`'s needle-only operand to `maskSecretsForLog`, whose extra twin lookup is already the other operand's job. Earlier runs left survivors, each resolved: sites made redundant by an intermediate design (reverted to `origin/main`'s spelling), inert code (an early return, a needle pass over a registered twin, token-boundary checks and three per-site twin lookups the central lookup already covers, all deleted), and untested branches (cases added, including a second producer region so the per-region mask is not hidden by the mask over the joined list).

## Verification

- Codex delta review, 16 rounds on the central design. Each finding is fixed in this diff, except round 5's AWS SDK error text, filed as go-to-k/cdkd#3171 rather than patched site by site (round 6 found that class at three more sites). Round 16: NO ISSUES.
- A maintainer-checklist pass (mutation / comment-vs-code / invariant / name / docs parity), 3 rounds, each round's findings fixed.
- After the maintainer's first review: a Codex delta review of the response, 2 rounds (round 1: the backstop note omitted a resolver's own stack region; round 2 NO ISSUES), and the maintainer-checklist pass again, 3 rounds (round 1: the missing first-mask and fallback probes, premise asserts against the masking behaviour, two leftover `operator-supplied` comments, changelog parity; round 2: `maskThenStripThenMask`'s docstring claiming its second mask was unreached, the fallback's guards in the changelog, a compile-time pin for the required parameters; round 3 NO ISSUES).
- `pr-security-reviewer`, `pr-code-reviewer` and `pr-test-reviewer` on the uncommitted diff, and a second `pr-security-reviewer` pass over the committed post-review changes (`straddleSafeTwin`, the sibling's guarded token text, the region-scoped clients debug line, the lowercasing guard): clean, with the two bounds above. The security and code reviewers each found the straddle regression now closed by `straddleSafeTwin`. The security reviewer also found the region-scoped clients DEBUG line and the lowercasing case, and the test reviewer found the gaps listed under Tests. All addressed. That pass predates the region-scoped clients REFUSAL fix, which came from the maintainer's review and is covered by the rounds listed below.
- The residual issue go-to-k/cdkd#3171's body: 3 Codex review rounds, the last NO ISSUES, before filing.
- Mutation probes: 79 one-at-a-time probes over `src/deployment/intrinsic-function-resolver.ts`, each redding at least one case in the two resolver files and `deploy-engine-output-failure-masking.test.ts`, run against the final formatted tree; none survives, and none passes by failing to collect (see Tests for what they cover and the one equivalent mutant).
- `/check` on the second review-response tree (`97ec55c4`, fetched origin/main with no overlapping file): `vp run check`, `vp run typecheck:test`, `vp run build`, regenerated matrices, and the full unit suite: 1023 test files, 23201 passed, 1 skipped, 0 failed, no type errors (a single run rooted in the worktree).
- `/run-integ lambda` (in the integ-broad set this resolver change requires; the broad fixtures with a `verify.sh` do not reach the name-argument paths, so the four fixtures below cover those): re-run on `97ec55c4`: PASS in 111 s, destroy 9 deleted / 0 errors, no leftover resources. The run's own `deployments/` event logs under the stack prefix were removed after the check.
- `/run-integ importvalue-chain` (`Fn::ImportValue` export and producer names): re-run on `97ec55c4`: PASS in 116 s: the A -> B -> C import chain resolved both exports from the index, each stack destroyed with 0 errors, and the exports index was purged of both chain exports; no leftover resources, and the run's `deployments/` event logs were removed after the check.
- `/run-integ getstackoutput-crossregion` (`Fn::GetStackOutput` names and region; its producer output is a literal value, so `reresolveCrossStackValue` returns before building a producer-region resolver, and that line is covered by the unit cases only): re-run on `97ec55c4`: PASS in 94 s: the consumer in us-east-1 resolved its producer's output from us-west-2 and both stacks destroyed with 0 errors; no leftover resources in either region, and the run's `deployments/` event logs were removed after the check.
- `/run-integ secrets-dynamic-ref` (dynamic-reference names; the fixture already seeds the two-character `pin` secret): re-run on `97ec55c4`: PASS in 533 s: every `{{resolve:...}}` resolved to its value, the stack destroyed with 0 errors, the fixture's own sweep reported 0 surviving state object versions, and neither the known password nor the two-character `q7` pin appears in the run's log.
- `/run-integ intrinsics-torture` (`Fn::FindInMap` keys, `Fn::GetAZs` region): re-run on `97ec55c4`: PASS in 73 s: the `Fn::FindInMap` / `Fn::GetAZs` / `Fn::Cidr` / `Fn::Base64` parameters resolved, the stack destroyed its 12 resources with 0 errors, and the fixture confirmed state gone with no orphan SSM parameters.


go-to-k:
Reviewed head `f9b1f7ad`. Tier `3-axis` + security add-on (2093 LOC after excluding `docs/_generated/`, 9 files; a redaction/masking fix triggers the security lens at any tier): `pr-spec-reviewer`, `pr-code-reviewer`, `pr-test-reviewer`, `pr-security-reviewer` in parallel.

CI is fully green on this head — all 15 checks `pass`/`skipping`, including `check-build-test` (11m57s), `once-leak-detect` (11m12s) and both `runtime-compat` legs, with `ci-ok` closing.

What holds up under an independent read, so it is not re-litigated below: the persistence constraint is intact — `resolveBase64`'s detector operands are unchanged (`logTextOfLeaf` untouched; operand 2's `maskNeedlesForLog` is byte-identical to main's `maskSecretsForLog`), the diff adds zero twin-registration lines, and the sibling hand-off keeps `logTwin` as the raw token, so `rememberLogTwin` registers exactly what main did. `straddleSafeTwin` never prints more than main in either the parsed-name or the token path. The coverage floor move 155 -> 161 is necessary, not cosmetic (the instrumented tree sits at 157). `Closes #3150` is earned: every site the issue enumerates, both addendum paths, and all three transform cases carry a citation. Assertions are whole-message `toBe` at the sites that matter, and both two-occurrence sites are pinned at both ends.

Three items are the same class as each other, and the same class as what this PR fixes: a fail-open default or an exemption note that outlives its premise, in the file whose job is masking.

## Required before merge

**M0 — `src/deployment/intrinsic-function-resolver.ts:2905-2921`: the `not-in-class` note's premise is falsified by this PR's own code 25 lines below.**

The backstop throw prints the region raw, justified by:

> every call site today passes either `this.explicitRegion` — the CLI's `--region`, never template-derived

That is no longer true, and this diff is what makes it untrue. `resolverForProducerRegion` (`:7508-7524`) builds a guest with `new IntrinsicFunctionResolver(target)` where `target` is template-derived — a `Fn::GetStackOutput` region or a secret ARN's region, as the PR's own comment at `:7525` says. The guest then calls `this.clientsForRegion(this.explicitRegion)` at `:10110` and `:10409`. The diff itself acknowledges this at `:2941`, where `targetRegion === this.explicitRegion ? this.explicitRegionLogText : undefined` exists *only* because `explicitRegion` can be template-derived — the debug line two frames up prints `us-west-2_***` while the throw prints the plaintext.

Reachable, with no `isClientSafeRegion` gate one arm up (unlike the `Fn::GetAZs` / `Fn::GetStackOutput` sites this PR did mask): `arnRegion` is `secretId.split(':')[3]` with no shape check (`src/deployment/secret-region-classification.ts:132-136`), and the `named-region` arm at `:9727` delegates without a charset check. So

```json
{"Fn::Sub": ["{{resolve:secretsmanager:arn:aws:secretsmanager:us-west-2_${P}:123456789012:secret:x:SecretString:k}}", {"P": {"Ref": "Pin"}}]}
```

with `Pin` -> `q7` builds a guest, `isClientSafeRegion` (`/^[a-z0-9][a-z0-9-]{0,30}$/`, `:1902`) rejects the `_`, and the throw prints `'us-west-2_q7'` — into CLI output, deployment events and the rollback journal. A 32+ character assembled region reaches it identically.

Not a regression (main printed it too), but it is in-class for #3150, it is absent from the PR's stated-bounds list, and the note is what keeps `check-resolver-mask-coverage.ts` green there — so a reader who trusts it stops looking. Two reviewers reached this independently, and I confirmed the call chain by hand.

Please either thread `context` and mask it — `this.maskSecretsForLog(targetLogText ?? stripControlChars(target).slice(0, 64), context)`, or `***` whenever `logTextOfLeaf(targetRegion, context) !== targetRegion`, the shape already used at `:8204` / `:8854`, which the note itself anticipates ("A future caller passing a template-derived region owes this method a `context` and this throw a mask") — or, if you would rather keep the site as it is, correct the note and add the site to the PR's bounds list. Masking is preferable; correcting the note is the floor.

**m1 — `:9302`: `return text ?? name` fails OPEN against an invariant the doc above it asserts.**

The doc at `:9269` states "Every such name is a run of the token's `:`-separated pieces". `:9302` is the arm where that is false, and it prints the raw name; every other arm of the helper returns `SECRET_MASK`. No leak today — all eight callers (`secretId`, `jsonKey`, `versionStage`, `versionId`, `parameterName`, `service`, `regionVerdict.secretName`, `regionVerdict.region`) are contiguous colon-boundary runs, and the only non-run caller is the `'AWSCURRENT'` default synthesized at `:10080`. But the invariant is asserted in prose and unfenced in code, so a future parse that normalizes a name before display prints it in the clear, silently.

Please make it fail closed with the synthesized default special-cased (e.g. `return text ?? (inner.includes(name) ? SECRET_MASK : name)`), or state the "synthesized defaults only" contract in the doc so the next reader knows the fallback is not a general one.

**m2 — `:10024` and `:10389`: dead fail-open parameter defaults.**

`nameLogText: (name: string) => string = (name) => name` on `resolveSecretsManagerReference` and `resolveSSMReference`. All three call sites (`:9837`, `:9845`, `:9902`) pass the real closure, so the default is dead — and it defaults to printing the raw name. Contrast `lookupCfnStackOutputs` (`:7957`) in this same diff, where `regionLogText` is correctly required. Please make both parameters required, so a future call site cannot omit the masking by forgetting an argument.

## Take-or-leave

- **n1 — no unrecorded-value control for the dynamic-reference families** (`tests/unit/deployment/intrinsic-resolver-name-argument-log-twin.test.ts:1203`, `:1264`, `:1553`; ~40 cases over the most new machinery). Over-masking is caught only incidentally there, by sibling cases whose assertions keep an unmasked prefix (`ssm:param-***` at `:1441`) and by `:1413`. A control in the shape of the six `CONTROL:` cases would make it direct.
- **n2 — `:648-655` asserts a fragment** (`toContain` / `not.toContain`) where #3150's plan asks for the whole refusal; the un-asserted first occurrence passes on the needle mask only because that fixture's secret is 7 characters.
- **n3 — `:8169` and `:8818`: a local `regionLogText` shadows the method of the same name**, and both scopes also call `this.regionLogText(...)` (`:8210`, `:8861`). Renaming the local to `loggedRegionText` saves the next reader resolving it by hand.
- **n4 — redundant outer masks at `:4453` and `:9705`**: both wrap already-fully-masked text, and they inflate the `maskedExprs` population the coverage script pins at 167.
- **n5 — `:9291-9301` has no `break` after a run matches**, though a longer run from the same `start` can never equal `name`; `pieces` is unbounded (`[^}]+`) and the closure runs up to 4x per token.
- **n6 — one line for the PR body**: the `{ Parameter: {} }` fixture shape at `:159` is not what SSM returns (it rejects with `ParameterNotFound`). It is what makes `'param-***' not found or has no value` assertable whole, and the real-AWS outcome for that template is the SDK-text residual already filed as #3171 — worth saying so rather than changing the fixture.

## Out of scope for this PR

- The SDK-authored-text class is correctly split out as #3171, not patched site by site. I re-verified all four routes in that issue body and they are accurate, including the parenthetical about `defaultSchedule` being false so the give-up `warn` gate never opens.
- One reviewer raised a second token-mispairing source, claiming `dynamicReferenceTokens` and the loop's local `/\{\{resolve:([^}]+)\}\}/g` disagree on a token containing `{`. That is wrong and needs nothing from you: `DYNAMIC_REFERENCE_INNER` is `[^}]+` (`src/deployment/secret-redaction.ts:2821`, `:2828`), so `DYNAMIC_REFERENCE_TOKEN_SCAN` (`:4554`) matches identically — the doc above it says as much. Noting it only so it is not raised again.

Once M0, m1 and m2 are pushed I will re-review and merge; the `pr-review` marker is not set, so nothing is blocking on my side other than that round.

go-to-k:
Round 2, reviewed head `002f2ec8`. Same tier (`3-axis` + security; 2320 LOC after excluding `docs/_generated/`, 9 files), all four axes re-run against the delta.

**M0, m1 and m2 are closed**, and I checked the fixes rather than the commit message:

- **M0** — the refusal now prints `loggedTarget`, and the replacement `not-in-class` note is TRUE, which is the part that mattered after round 1. Two reviewers enumerated it independently and agree: all 12 `clientsForRegion(` call sites, of which 11 pass `this.explicitRegion` by identity so a guest always takes arm 2, and the 12th (`resolveGetAZs:8889`) supplies a masked `targetLogText` behind the `isClientSafeRegion` gate at `:8849`; and all constructor sites feeding `explicitRegion` (`deploy-engine.ts:1265`, `diff-recursive.ts:470,847`, `export.ts:6264`, `import.ts:1646`, and the `drift.ts` / `scrub.ts` / `rollback-executor.ts` replay siblings, whose region is `secretId.split(':')[3]` off a state-recorded literal). The bare-`target` arm cannot carry a value this pass resolved. `:1324` asserts the throw itself, and `:1354` pins the bare-`target` arm as a control.
- **m1** — `name !== '' && inner.includes(name) ? SECRET_MASK : name` is right, and the `name !== ''` guard is load-bearing (`inner.includes('')` is always true) *and* pinned: `:1434` covers both arms in one case. The added `break` is behavior-preserving — for a fixed `start`, `pieces.slice(start, end).join(':')` grows strictly, so at most one `end` can match, and the cross-`start` conflict arm at `:9310` is untouched.
- **m2** — both `nameLogText` parameters are required, with `@ts-expect-error` lines as the fence.

Also worth recording, because round 1's version of that doc said the opposite: the `maskThenStripThenMask` doc rewrite is **true, not asserted**. `:1367` (`st-1` split by U+0001, caught only by the second mask) and `:1393` (`x`+U+0001+`y7` recorded whole, caught only by the first) each drive the real refusal and each reds if its half is deleted. Persistence is still untouched — `rememberLogTwin` count is 4 at base, at `f9b1f7ad` and at `002f2ec8`. `maskedExprs` 167 / `markers` 103 re-derived by running the script's own `scan()` against the tip: they still hold.

CI: `once-leak-detect` green at 10m53s, `check-build-test` still running; I will confirm it before merging.

## Required before merge

**m3 — `src/deployment/intrinsic-function-resolver.ts:7549`: this round hardened `:7546` and left its sibling three lines below on the weaker masker, with the same input.**

```ts
scoped.explicitRegionLogText = this.maskThenStripThenMask(regionText, context);   // :7546
this.producerRegionResolvers.set(target, scoped);
this.logger.debug(
  `Using a producer-region resolver for ${this.maskSecretsForLog(regionText, context)}`  // :7549
);
```

I confirmed the reachability chain end to end. For a LITERAL ARN token, `straddleSafeTwin` returns on its first arm (`if (twin === raw) return twin`, `:9264`), so `nameLogText(regionVerdict.region)` at `:9743` hands back the region with its control character intact; `canonicalizeRegion` is `.toLowerCase()` only (`src/utils/aws-partition.ts:105`), so `regionText` still carries it. `maskSecretsForLog` finds no twin for a template literal and falls to the needle mask, which cannot see `st-1` split as `s` + U+0001 + `t-1`. So on exactly the input `tests/unit/deployment/intrinsic-resolver-name-argument-log-twin.test.ts:1367` builds, `:7546` yields `us-west-2_***` while `:7549` prints `us-west-2_s\x01t-1` — which a terminal renders as `us-west-2_st-1`. That test asserts `message` only, never `everyLine()`, so the line is unpinned.

This is #2827's control-char-split class rather than #3150's sub-floor class, and it is not a regression against `origin/main`. It is required anyway because *this round* created the asymmetry: two adjacent lines now give different verdicts on the same value, and the `maskThenStripThenMask` doc you just rewrote cites this very route as what earns the second mask. Please use `this.maskThenStripThenMask(regionText, context)` at `:7549` and add an `everyLine()` assertion to the `:1367` case so the line stays pinned.

**m4 — the PR body still describes `f9b1f7ad`.** The body is the durable record this change is read from later, so please refresh it to `002f2ec8`:

- "90 cases" -> **101** (I counted `it(` at both heads: 90 at `f9b1f7ad`, 101 at the tip). The dependent sentence "Against `origin/main`'s resolver 82 of the 90 fail; the 8 that pass are the 6 unrecorded-value controls…" is stale with it — round 2 added three unrecorded-value controls plus two clients-refusal controls.
- "79 probes" and the probe enumeration name none of this round's arms: the `loggedTarget` hoist, `maskThenStripThenMask` at `:7546`, the fail-closed fallback, the now-required `nameLogText`.
- "`/check` … 23187 passed" predates the +11 cases.
- Design table: the "region-scoped clients line" row describes the debug line only, but the refusal is now masked too; the `dynamicReferenceNameLogText` row omits the synthesized-default / fail-closed rule that the changelog entry does carry.
- The Verification section credits the second `pr-security-reviewer` pass with "the region-scoped clients line", but the refusal fix is this round's own commit, after that pass.

Nothing needs REMOVING from "Two contrived bounds": bound 1 still stands at the tip — `canonicalizeRegion(producerRegionLogText)` at `:7542-7544` still has no lowercase-formed-needle guard, and `maskThenStripThenMask` does not supply one.

## Take-or-leave

- **n7 — `:10396-10398`: `decrypt = true` and `service: 'ssm' | 'ssm-secure' = 'ssm'` now sit BEFORE the required `context` / `nameLogText`**, so neither default is reachable by omission any more (both call sites pass all five positionally). Dropping the `= true` / `= 'ssm'` stops the signature advertising an unusable convenience.
- **n8 — `:7526-7547`: `explicitRegionLogText` is set only at the guest's FIRST creation**, and `producerRegionResolvers` is keyed by canonical region. A pass holding both a literal `arn:...:us-west-2_q7:...` token and an `Fn::Sub` one assembling the same region around recorded `q7` prints the literal's text for both. Needs the plaintext to sit in the template literally (the class the body already bounds), and the reverse order over-masks, so this is strictly ≤ `origin/main`. Re-deriving `loggedTarget` per call rather than caching it on the guest would close it.
- **n9 — no unrecorded-value control for version stage / version id** (positive case at `:1265-1280`). They share `nameLogText` with the covered families, so the risk is nominal.
- **n10 — a heads-up rather than a request**: `changelog.d/entries/2026-09-15-3150-name-argument-log-twin.md` is at 1997 bytes against the 2000-character cap `tests/unit/scripts/changelog-entry-size.test.ts` enforces. If m4's refresh touches it, there is almost no headroom.

Round 1's n4, n6 and the old n1/n2 are either addressed or superseded; I am not re-raising them.

Once m3 and m4 are pushed I will re-review and merge, assuming `check-build-test` is green.

nix-tkobayashi:
Thanks for the review. Pushed as `002f2ec8`, with the integ ledger update `97a9d38f` on top. Item by item:

**M0 -- fixed by masking.** `clientsForRegion` now works out the region's printed text before the gate and the refusal prints it: the caller's masked text (`resolveGetAZs`), else a producer-region guest's `explicitRegionLogText`, else the canonical region. I stored the text on the guest rather than threading `context` into `clientsForRegion`: the guest's own call sites (`resolveSecretsManagerReference` / `resolveSSMReference` and the other `clientsForRegion(this.explicitRegion)` callers) have no masked text of the region to pass, while `resolverForProducerRegion` holds both the text and the context. Because the refusal strips control characters and cuts to 64, the guest's text is now stored through `maskThenStripThenMask` rather than one `maskSecretsForLog`, the same order as the `Fn::GetAZs` gate. The note above the throw is rewritten to say which regions reach it: the guest's masked text, `resolveGetAZs`' already-accepted region, and otherwise a resolver's own region as its command built it (the stack's synthesized or recorded region, `--region`, or a replay / drift / scrub resolver's region read out of a literal token). Your template now prints `Refusing to build AWS clients for the region 'us-west-2_***': ...`. New cases: that refusal, its unrecorded-value control (`us-west-2_k9`), and two cases for the mask order, each asserting its premise against `maskSecretsInText` on the raw and the stripped region: a recorded 4-character secret split by U+0001 inside the literal region, which only the mask after the strip sees, and a recorded secret holding U+0001, which only the mask before the strip sees.

**m1 -- fixed.** The fallback is `text ?? (name !== '' && inner.includes(name) ? SECRET_MASK : name)`, and the docstring now states the contract: a name that is no run of the token is a synthesized default (`AWSCURRENT`, or the empty version id), printed as it is unless the token spells it inside a longer piece. The `name !== ''` guard keeps an absent version id printing empty rather than `***`, since every string includes `''`. New case: `x-AWSCURRENT-${P}` as the secret id prints the defaulted stage as `***`.

**m2 -- fixed.** `nameLogText` is required on both methods; their `context` parameter became `ResolverContext | undefined` so a required parameter can follow it. Since every caller passes the argument, no runtime case can see a default come back, so a case holds two `@ts-expect-error` calls that omit it; restoring either default makes `vp run typecheck:test` fail on the unused directive.

**n1 -- done.** Unrecorded-value controls for the inner-`Fn::Sub` JSON key, the same-`Fn::Sub` secret id and SSM parameter name, and a value carrying `:` (which pairs, since both sides split the same way).

**n2 -- done**, and it was worth it: the whole-message assertion showed the first occurrence prints `[Child, ***]`, not `[Child, ***.env]`. The needle mask changes the raw attribute name, so the central lookup's both-masks arm prints `***`. The test now asserts the whole refusal.

**n3 -- done.** The locals in `resolveGetStackOutput` and `resolveGetAZs` are `loggedRegionText`, and so is `lookupCfnStackOutputs`' parameter of the same name.

**n4 -- not taken.** Dropping either outer mask trips `check-resolver-mask-coverage.ts`: it credits a mask inside `.map(...)` to nothing (a method call is judged on its receiver, and the receiver is the unmasked key list or `foreignProducerRegions`), so each site would need a `not-in-class` note claiming the elements were masked, which is the unchecked-convention shape the script's header argues against. The two calls stay, and `maskedExprs` stays at 167.

**n5 -- done**, with a note that a longer run from the same start is a longer string. A probe removing the `break` survives, as it should: it is a cost change, not a behaviour change.

**n6 -- added to the PR body**: the `{ Parameter: {} }` fixture shape is not what SSM returns, and the real-AWS outcome for that template is the SDK-text residual go-to-k/cdkd#3171.

Verification of this round: 11 mutation probes, one change at a time with the resolver restored byte-exact between them, over this file, `intrinsic-resolver-sub-floor-log-mask.test.ts` and `import-resolver-error-masking.test.ts`. Red: the refusal printing the raw region (3 cases), dropping the guest arm of the printed text (6), storing the guest's text through a single mask (1), masking only after the strip (1), the fallback failing open (1), the fallback without the empty-name guard (6), the fallback masking every non-empty name (5), `straddleSafeTwin` masking an unmasked name (16, the controls among them), the refusal printing `***` when it holds no region text (1, which first survived: the only control went through a guest, so a control on an ordinary resolver with an invalid region of its own was added), and the nested refusal printing the attribute name raw (4, including the whole-message case from n2). The `break` probe survives as above. Codex delta review of this round: 2 rounds (round 1 found the backstop note still omitted a resolver's own stack region; round 2 NO ISSUES), then a maintainer-checklist pass of 3 rounds (round 1: the missing first-mask and fallback probes, premise asserts, two leftover "operator-supplied" comments and changelog parity; round 2: the stale second-mask docstring, the fallback's guards in the changelog, and the compile-time pin for m2; round 3 NO ISSUES). Full unit suite on `002f2ec8`: 1023 test files, 23198 passed, 1 skipped, 0 failed, no type errors. The five integ fixtures from the first round re-ran on `002f2ec8`, all PASS with 0 destroy errors and no leftovers: `lambda` (118 s), `importvalue-chain` (132 s), `getstackoutput-crossregion` (167 s), `secrets-dynamic-ref` (538 s, 0 surviving state versions, no `q7` in the log) and `intrinsics-torture` (84 s). None of them reaches the `clientsForRegion` refusal itself, since that needs an invalid region no fixture deploys; the unit cases above cover it.

go-to-k:
Thanks — that writeup answers round 1 item by item, and I had already verified M0, m1, m2 and n1-n3 against `002f2ec8` before your reply landed. Two notes on your dispositions, then the one thing left.

**n4, not taken — agreed, and the reason is right.** I checked it rather than taking it on trust: `scripts/check-resolver-mask-coverage.ts:280-290` decides a method call on its RECEIVER alone and deliberately does not consult arguments, so `foreignProducerRegions.map((r) => this.maskSecretsForLog(r, context)).join(', ')` bottoms out on an unmasked receiver. Dropping the outer mask would turn each site into a finding, and papering over it with a `not-in-class` note is exactly the unchecked-convention shape that file's header argues against. Keep both calls; `maskedExprs` staying at 167 is correct.

**n5 — agreed that the surviving probe is the right outcome.** A longer run from the same `start` is strictly longer, so at most one `end` can match: the `break` is a cost change, not a behaviour change, and a probe that reds would mean the loop had been doing something else.

**n2 was worth it for the reason you found, not the one I asked for.** `[Child, ***]` rather than `[Child, ***.env]` is the central lookup's both-masks arm firing on an attribute name the needle mask also changes — good that the whole-message assertion surfaced it.

## Still open

Your reply does not mention my second comment, https://github.com/go-to-k/cdkd/pull/3176#issuecomment-5682826815, posted at 15:16Z on head `002f2ec8` — about two hours before your reply, so I do not think it crossed. It found two things in the round-2 diff itself. Both are still open at `97a9d38f` (which is ledger-only: `docs/_generated/integ-last-run.tsv` +5/-5, and I checked the five re-run rows against the file's conventions — the shared `16:55:15Z` stamp has ample precedent there, and the durations differ from the first round's, so nothing to raise):

- **m3 — `src/deployment/intrinsic-function-resolver.ts:7549`.** Round 2 hardened `:7546` to `maskThenStripThenMask` and left the debug line three lines below on `maskSecretsForLog`, with the same `regionText`. For a literal ARN token `straddleSafeTwin` returns on its first arm (`twin === raw`, `:9264`) and `canonicalizeRegion` only lowercases, so the control character survives to `:7549`, where no twin exists and the needle mask cannot see `st-1` split as `s` + U+0001 + `t-1`. On the exact input your new `:1367` case builds, `:7546` gives `us-west-2_***` while `:7549` prints `us-west-2_s\x01t-1`. That case asserts `message` only, so the line is unpinned. It is #2827's class rather than #3150's and not a regression against `origin/main` — but this round created the asymmetry, and the `maskThenStripThenMask` docstring you rewrote cites this route as what earns the second mask.
- **m4 — the PR body still describes `f9b1f7ad`.** Your reply already carries most of the replacement numbers: 101 `it(` cases at the tip against the body's 90 (so "82 of the 90 fail" and the "8 that pass" breakdown move with it), 23198 passed against the body's 23187, this round's 11 probes against the body's 79 and an enumeration naming none of this round's arms, plus the two design-table rows and the security-reviewer credit for a fix that postdates that pass.

Please push m3 and refresh the body for m4; n7 from that comment (`decrypt = true` / `service = 'ssm'` now sitting before the required parameters, so neither default is reachable) is optional and I will not hold the merge for it. `check-build-test` and `once-leak-detect` are still running on `97a9d38f`; I will confirm them green and merge once m3 and m4 are in.

go-to-k:
Round 3, reviewed head `97ec55c4`. Same tier, all four axes re-run.

**m3 is closed, and better than what I asked for.** I asked you to swap the masker on the debug line; you computed the text once before the cache lookup and printed the stored value, so the creation line and the refusal can no longer disagree by construction rather than by two call sites staying in step. **n7, n8 and n9 are closed too** — n8 in particular went from a stated bound to a fail-closed arm.

Verified rather than taken on trust:

- **The bare `guestRegionText` in the debug line keeps its mask credit, and the pins hold.** This was my one worry about the restructure, since `tests/unit/scripts/resolver-mask-coverage.test.ts` pins `maskedExprs` 167 / `markers` 103 and round 3 does not touch it. Running the script's own `scan()` over the tip: `statements 136 / maskedExprs 167 / markers 103 / 0 findings / 0 unconsumed` — unchanged. And the credit is not vacuous: swapping the initializer to `const guestRegionText = regionText;` drops it to 166 and raises `FINDING 7562 log guestRegionText`, so `resolveBinding` really does reach the `const`.
- **The cache-hit mutation is monotone and order-independent.** `***` is absorbing, `!==` is symmetric, so A-then-B and B-then-A both land on `***` and no sequence leaves the guest less masked than the most-masked spelling seen. `undefined !== guestRegionText` cannot fire spuriously — the single `.set` is immediately preceded by the assignment.
- **Nothing on the newly-hot path registers.** `regionLogText` / `logTextOfLeaf` / `maskThenStripThenMask` and friends are pure lookups; `rememberLogTwin` is 4 at base, at `002f2ec8` and at the tip, with byte-identical call sites, so the `Fn::Base64` persistence invariant is intact.
- **`maskThenStripThenMask(x)` is never weaker than the `maskSecretsForLog(x)` it replaced** — `stripControlChars` deletes only C0/DEL/C1/bidi, so it cannot remove a `***` span, and anything it rejoins the second mask catches.
- Creation-line, reuse and version-stage/id cases all reach the real path and red for the right reason: the creation-line assertion sits in the control-char-split fixture, which is the one that would print raw under the old spelling.

CI at this head is fully green: `check-build-test` 12m13s, `once-leak-detect` 10m56s, both `runtime-compat` legs, `ci-ok` closing. Nothing failing or pending.

## Required before merge

**m5 — a comment this PR falsified, and the third one in three rounds.**

`src/deployment/intrinsic-function-resolver.ts:3411-3413`, unchanged from `main`:

> `// ANY error, not only a lookup echo. Residual: a plaintext`
> `// shorter than MIN_NEEDLE_LENGTH (4) is embedded here rather than`
> `// whole, so no needle matches it and it still prints.`

The example it sets up two lines above is `key '<password>' not found in secret '<id>'` — exactly what this PR closes. `tests/unit/cli/import-resolver-error-masking.test.ts:403` is flipped to `CLOSED by issue #3150` and asserts `key 'key-***' not found`, and `intrinsic-resolver-name-argument-log-twin.test.ts:386-400` asserts the whole condition warning with `top-level key 'env-***' not found`. "It still prints" is false at this head; what survives is the SDK-text class, which this note does not name.

I am asking for the sweep rather than the one line, because this is the third round and all three findings are one class — prose asserting a premise the same diff falsified. Round 1's M0 was a `not-in-class` justification your own new branch contradicted 25 lines below; round 2's m3 rode on a `maskThenStripThenMask` docstring that round's fix had just made true while leaving its sibling line behind; now this. Fixing the instance I happened to find leaves the next one for round 4.

The surface is small enough to close exhaustively — I grepped it: **three** prose residual/bound claims in the file, `:1705`, `:3411` and `:9564` ("THE BOUND, since this file's job is to state them"). Please re-read each against the tip and correct what this PR moved, and give the same pass to the `not-in-class` justifications your diff touched. The 103 markers are machine-checked for being *consumed* — the scanner reports 0 unconsumed — but nothing checks that a marker's stated *reason* is still true, which is precisely how M0 survived into round 1.

**m4 — the body counts are stale for the third time, so please stop recounting them.**

At the tip it is **104** `it(` cases, not the body's 101 (90 at `f9b1f7ad`, 101 at `002f2ec8`, 104 now), so "86 of the 101 fail; the 15 that pass are the 12 unrecorded-value controls…" no longer adds up — 86 + 15 = 101 ≠ 104, and the `CONTROL:` count moved too. `002f2ec8` appears six times in the body and neither `97a9d38f` nor `97ec55c4` appears at all, so the `/check` line and the five integ lines are bound to a head two commits back with a resolver source change since. Round 3's own three behaviors — the creation line printing the stored text, the reused guest failing closed, the `resolveSSMReference` signature — are absent from the body entirely.

You refreshed these numbers last round and the same push that refreshed them added the test lines that re-staled them. A fourth recount has the same failure mode, so please do one of:

- **drop the counts** — "each site family with an `Fn::Sub`-assembled two-character secret, with controls" says what a reviewer needs; the exact case count does not; or
- **attribute them** — "as of `97ec55c4`: 104 cases" — so a stale number reads as historical rather than wrong; or
- **fence them** — if a count is worth keeping, assert it in the suite so it cannot drift silently.

The `/check` and integ lines are the one place a sha genuinely belongs, so those just need re-binding to the final head. Nothing needs removing from "Two contrived bounds".

## Take-or-leave

- **n11** — no reversed-order reuse case (assembled-then-literal) and no third delegation after the guest is already `***`. The arm is sticky by construction, but a mutant narrowing the guard to one direction survives today.
- **n12** — `:2948`'s `not-in-class` note omits the guest arm that `:2927`'s names, though both describe the same binding in the same method; and `:2927`'s "at the guest's construction" is now incomplete after the cache-hit overwrite.
- **n13** — the `guestRegionText` credit rides on `resolveBinding`, whose header enumerates its own fail-open shapes; a second binding of that name anywhere in the file would silently withdraw it. The `const` factoring is right, just worth a word in the comment.
- **n14** — ragged manual rewrap at `:2917-2921` leaves `` `targetLogText` is `` dangling.

A heads-up that still applies: the changelog entry is at 1996 characters against the 2000 cap, so any correction there has to remove text first.

Push m5 and m4 and I will re-review and merge.

nix-tkobayashi:
It did not cross: your round-2 comment (issuecomment-5682826815) was up about two hours before my reply, and I missed it. Pushed as `97ec55c4`, with the integ ledger update `53c168bb` on top. Taking it item by item:

**m3 -- fixed.** `resolverForProducerRegion` now computes `guestRegionText = this.maskThenStripThenMask(regionText, context)` once, the guest stores it, and the `Using a producer-region resolver for` line prints that same value, so the two adjacent lines can no longer disagree. Both control-character cases (the `st-1` split by U+0001, and the recorded `x` + U+0001 + `y7`) now also assert the creation line through `everyLine()`, not only the refusal.

**m4 -- refreshed.** The body was updated at 17:12Z, just before your follow-up, with the 101-case count, 86 of 101 against `origin/main`, 23198 passed, this round's probes, the refusal row, and the synthesized-default rule in the `dynamicReferenceNameLogText` row. What was still stale was the second `pr-security-reviewer` pass: it now credits that pass with the region-scoped clients DEBUG line and says the refusal fix came later, from your review. The counts now also include this round's cases (104 cases in the new file, 88 of them failing against `origin/main`, and the four probes this round added). Bound 1 stays, as you said.

**n7 -- taken.** `decrypt` and `service` lose their defaults.

**n8 -- taken, fail-closed rather than per call.** The region text is now computed before the cache lookup, and on a hit whose text differs from the cached guest's, the guest's text becomes `***`. I did not re-derive `loggedTarget` per call, because the guest's call sites (`clientsForRegion(this.explicitRegion)` inside the lookup helpers) hold no text of their own to pass. With a disagreement, both spellings print `***` from then on. Two new cases: a literal `us-west-2_q7` ARN token followed by an `Fn::Sub` assembling the same region around a recorded `q7` on one resolver, where the second refusal prints `***`; and the same control-character region twice, where the second refusal keeps `us-west-2_***`, which pins both the equality check and that the cache hit masks its own text before comparing.

**n9 -- taken.** An unrecorded-value control for the inner-`Fn::Sub` version stage and version id.

**n10 -- noted, and it bit.** This round's changes had to be named in the entry (the creation line's mask, strip and mask, and the reuse rule), so it was rewritten to fit: it stays at 1997 bytes, and `changelog-entry-size.test.ts` passes.

Verification: the 11 probes from the last round plus four for this one, re-run one at a time with the resolver restored byte-exact between them. The creation line reverted to the single mask reds the `st-1` case's new line assertion (1 case); the cache-hit arm left on the first spelling's text reds the differing-spelling reuse case (1); the cache hit masking unconditionally, and the cache hit comparing the unmasked region text, each red the same-spelling reuse case (1 each). Of the other 11, 10 still red at least one case each and the `break` probe still survives as the cost-only change. Codex delta review of this round: NO ISSUES on round 1; the maintainer-checklist pass took 2 rounds (round 1: the equality check and the cache hit's own masking were unpinned, which the same-spelling case and its two probes now pin; three comments lagged the new cache-hit arm and the non-empty guard; the changelog lacked the creation line and the reuse rule; the version control's name claimed inner `Fn::Sub`s it does not build; round 2 NO ISSUES). Full unit suite on the fix commit: 1023 test files, 23201 passed, 1 skipped, 0 failed, no type errors. The five integ fixtures re-ran on `97ec55c4`, all PASS with 0 destroy errors and no leftovers: `lambda` (111 s), `importvalue-chain` (116 s), `getstackoutput-crossregion` (94 s), `secrets-dynamic-ref` (533 s, 0 surviving state versions, no `q7` in the log) and `intrinsics-torture` (73 s). As before, none of them reaches the producer-region creation line or the clients refusal for an invalid region; the unit cases cover both.

go-to-k:
**m4 is closed** — I re-read the body just now: 104 cases, 88 of the 104, 23201 passed, and `97ec55c4` six times with no `002f2ec8` left. That matches the tip (`53c168bb` only adds the ledger rows, so the source-commit binding is the right one to name), and 23198 + 3 new cases = 23201 checks out. n7, n8, n9 and n10 are closed too; I reviewed all of them at `97ec55c4` and reported on them in https://github.com/go-to-k/cdkd/pull/3176#issuecomment-5685239162.

That comment is the one thing your reply has not reached — I posted it at 17:51Z, about 19 minutes before you wrote, and its shape (answering round 2 item by item) says it was already drafted. No need to re-read the whole thing; only **m5** is still open, and it is unchanged at the tip:

`src/deployment/intrinsic-function-resolver.ts:3411-3413` still says the sub-floor plaintext "still prints", two lines under the `key '<password>' not found in secret '<id>'` example — which is exactly what this PR closes (`tests/unit/cli/import-resolver-error-masking.test.ts:403` is flipped to `CLOSED by issue #3150` and asserts `key 'key-***' not found`). What survives there is the SDK-text class, which the note does not name.

The reason I asked for a sweep rather than that one line: this is the third round and all three of my required findings have been the same class — prose asserting a premise the same diff falsified (round 1's M0 was a `not-in-class` justification your own new branch contradicted 25 lines below; round 2's m3 rode on the `maskThenStripThenMask` docstring that round had just made true while leaving its sibling line behind; now this). Fixing only the instance I happened to find is what leaves round 4 a fourth one.

The surface is bounded — I grepped the file at the tip and it is still exactly three prose residual/bound claims: `:1705`, `:3411` and `:9564` ("THE BOUND, since this file's job is to state them"). Please re-read those three against the tip, correct what this PR moved, and give the same pass to the `not-in-class` justifications your diff touched — the scanner checks those markers are *consumed*, not that their stated reason is still true, which is how M0 reached round 1 in the first place.

n11-n14 in that comment stay take-or-leave and I will not hold the merge for them; your same-spelling reuse case covers part of n11 already.

CI on `53c168bb` is green so far with `check-build-test` and `once-leak-detect` still running. Push m5 and I will re-review and merge.

go-to-k:
I ran the m5 sweep myself rather than leaving it with you — it was my ask, and it turns out to be much smaller than I implied. This **supersedes** the sweep request in https://github.com/go-to-k/cdkd/pull/3176#issuecomment-5685239162; treat the list below as the whole of m5.

Swept at the tip (`53c168bb`, source identical to `97ec55c4`):

**The two other prose claims are fine — no change needed.**

- `:1705` "Residual: issue #2805" sits on `allSettledKeepingFirstRejection` and is about which rejection surfaces from a drained `Promise.all` feeding `retryClassificationText`. Nothing to do with this PR's class; untouched and still true.
- `:9564` "THE BOUND ... a needle split by a character `stripControlChars` removes" is still exactly what `maskThenStripThenMask` does at the tip (`M(strip(M(x)))`), and the paragraph above it was already brought up to date this round with the #3150 route. True as written.

**So the required part of m5 is one comment: `:3411-3413`.** Its "no needle matches it and it still prints" is false for the `key '<password>' not found in secret '<id>'` example it sets up, which this PR closes. Narrowing it to the #3171 SDK-text class is all it needs.

**On the `not-in-class` justifications — I diffed them rather than sampling.** The PR retires 8 markers and adds 2, which is the 109 -> 103 the suite pins, and the scanner reports 0 findings / 0 unconsumed, so the retirements are consistent with the masks that replaced them. The 2 ADDED ones are worth a look, because they describe the same binding in the same method and each omits an arm the other names. `loggedTarget` has three arms:

```ts
const loggedTarget =
  targetLogText ??
  (targetRegion === this.explicitRegion ? this.explicitRegionLogText : undefined) ??
  target;
```

- `:2948` says "masked by the caller that built the region from a template, or a resolver's own region as its command built it" — arms 1 and 3, **not** the guest arm, which is the one M0 was about.
- `:2927` says "masked at the guest's construction, or a resolver's own region as its command built it" — arms 2 and 3, not arm 1. Omitting arm 1 there is defensible since `isClientSafeRegion` accepted it one arm up and the refusal cannot see it, but "at the guest's construction" is now incomplete anyway: after this round's cache-hit arm the text can become `***` later. Strictly more masked, so nothing unsafe — just no longer the whole reason.

That is the n12 nit from my last comment, now with the diff behind it. I am not making it required: both justifications are *understated* rather than false, which is the safe direction, and the marker text is prose the scanner never reads. Take it if you are touching the file for `:3411` anyway, since a justification that does not name every value reaching its site is the shape that carried M0 into round 1.

Nothing else from the sweep. Please push the `:3411` correction and I will re-review and merge — CI is green on `53c168bb` (`check-build-test` 10m42s, `once-leak-detect` 8m55s, both `runtime-compat` legs, `ci-ok`), the body matches the tip, and no other item is outstanding.

go-to-k:
I pushed the m5 correction to your branch myself rather than leave the PR waiting on a comment edit — `4d290773`, the only thing outstanding. Everything else was already closed at `53c168bb`. Flagging it because it is your branch: if you have local work on it, rebase onto the new head rather than force-pushing over it.

What the commit changes, and nothing else: the note above `evaluateConditions`' `warn` said a sub-floor plaintext embedded in a longer name has no needle and "still prints". It now records that the embedded case is masked BY POSITION out of the log twin (with the two tests that pin it), and names what actually survives at that sink — text an AWS SDK authored and the sink only forwards, #3171. Comment only, no behaviour delta, so no changelog entry.

I also finished the sweep I had asked you for, in the comment above: `:1705` and `:9564` are both still true at the tip and needed nothing, so `:3411` was the whole of it. The two added `not-in-class` justifications stay as they are — understated rather than false, and prose the scanner never reads.

Local gates on the pushed tree: `vp check --fix` and `vp run check` 0 errors, `vp run typecheck:test` clean, `vp run build` clean, regenerated matrices produced no diff, and the full suite 1023 files / 23201 passed / 1 skipped / 0 failed, rooted in this worktree. `maskedExprs` 167 / `markers` 103 still hold.

I will merge once CI is green on this head.

go-to-k:
Heads-up before you see two notifications: my two pushes to this branch (`4d290773`, `b6faf7b5`) created NO workflow run at all — `actions/runs?head_sha=...` returns 0 for both, `statusCheckRollup` is empty, and there is no `action_required` run waiting either. Your own pushes triggered CI normally, so this is specific to the branch being pushed by someone other than you; GitHub did not raise a `synchronize` event for it. `ci.yml` runs only on `push` to main and `pull_request`, with no `workflow_dispatch`, so there is nothing to re-run directly.

I am going to close and immediately reopen this PR, which fires `pull_request: reopened` and creates the runs. Nothing is lost by it — commits, comments and the branch are untouched — and I will confirm the checks appear.

The second push, `b6faf7b5`, fixes two defects a review found in my first one:

- I wrote that the twin is REGISTERED for the assembled string. That is true for the `Fn::FindInMap` case, where the substitution completes and `rememberLogTwin` runs, and false for the `key '<password>' not found` example the note actually cites — there the throw is raised inside `resolveDynamicReferencesWithLogTwin`, before `rememberLogTwin`, and the twin arrives as a parameter. Exactly the failure mode I had been asking you to sweep for, committed by me one round later; the note now states both halves and says not to shorten it back.
- I named only one surviving residual. There is a second: `resolveGetStackOutput` hands the RAW stack name to its state read at `:8331-8333` with no `try` / `catch`, and `S3StateBackend.getState` quotes it back (`src/state/s3-state-backend.ts:432`) through `displaySafe`, which sanitises ASCII rather than masking. The `Fn::ImportValue` sibling catches and masks the same shape at `:7787`, so this site was simply missed. Pre-existing on `origin/main`, not introduced here, and the fix is a code change rather than a comment, so it is filed as go-to-k/cdkd#3234 and the note now names both residuals.

Local gates on `b6faf7b5`: `vp run check` 0 errors, `typecheck:test` clean, `build` clean, regenerated matrices produced no diff, full suite 1023 files / 23201 passed / 1 skipped / 0 failed.

go-to-k:
Closing and immediately reopening to fire `pull_request: reopened` — my pushes to this branch created no workflow run (see the note above). Nothing is lost; reopening now.

go-to-k:
Correcting my note above: the missing CI had nothing to do with who pushed. The branch was **CONFLICTING** against `main` (25 commits behind, clashing on `docs/_generated/integ-last-run.tsv`), and GitHub cannot build a merge ref for a conflicted PR — so no `pull_request` workflow run is created at all. The one run that *did* fire on reopen, `PR inherit issue labels`, is `pull_request_target`, which uses the base branch and needs no merge ref. That asymmetry was the tell, and I read it as "who pushed" instead of "the branch went stale underneath". Your pushes ran because the branch was still clean then.

Fixed in `448bdfb7`, and `mergeable` went `CONFLICTING` → `MERGEABLE` with the `pull_request` runs appearing immediately.

Only the `lambda` row actually overlapped. The ledger keeps one row per test at its latest run, and `main`'s is the later one (`2026-09-16T04:07:16Z`, the go-to-k/cdkd#3199 broad-set run), so `main`'s row wins; the other four rows this branch added are on tests `main` did not touch, and they stay. Re-normalized with `vp run integ-ledger-normalize`.

I resolved just that one row rather than rebasing or merging, deliberately: `flatten-before-rebase-gate` wants the branch collapsed to one commit before a rebase, which would mean force-pushing a rewritten history over your branch, and `git merge origin/main` trips `integ-local-gate` because `main`'s new commits touch `src/local/**`. Fixing the conflicting row directly clears the conflict without rewriting your history and without routing around either gate.

Also in this round, from a further review pass over my own comment edit — the tally in it was wrong a third time, so I stopped writing one:

- A third residual exists that the count missed: `describeAvailableOutputs` renders a PRODUCER's own output keys into two `Fn::GetStackOutput` throws, and this pass never assembled those keys, so no twin covers them. That also falsifies the "composed OUTSIDE this resolver" qualifier, since they are built right here. Its own docstring already owned that bound — which is the convention I had been ignoring.
- So the note now states what IS closed (names **this pass assembled**), says the ways a sub-floor plaintext can still print are **not** enumerated, and points at each known instance where it is OWNED. `.claude/rules/layout-deployment-secrets.md` records five rounds on go-to-k/cdkd#2803 refuting the same shape of sentence; four rounds here repeated it. The changelog entry carried the same single-residual framing and is reworded the same way, at 1998 of 2000 characters.
- The closure qualifier is narrowed from "names the RESOLVER builds into its own text" to "names THIS PASS ASSEMBLED", and the cited example is spelled in its embedded form (`key 'key-<password>'`) so it matches the `key 'key-***'` output it is said to produce — it was previously the whole-value form, which masks to `key '***'`.

Local gates on `448bdfb7`: `vp run check` 0 errors, `typecheck:test` clean, `build` clean, regenerated matrices no diff, full suite 1023 files / 23201 passed / 1 skipped / 0 failed.

Nothing is outstanding from you. I will merge once CI is green.

go-to-k:
Merged — thank you. This was a long one and the quality held up across every round.

What stood out: the diff came with its own falsification work rather than only its result. The mutation probes were one change at a time with the file restored between them, you named the one equivalent mutant instead of quietly counting it as killed, and when a probe survived you treated that as a finding and added the case. The `Fn::Base64` persistence constraint is the part I'd have expected a change this size to break — the twin registry decides what gets recorded, and you pinned it from the start and kept it byte-identical through four rounds of edits.

You also took the review rounds the right way: n4 came back with a measured refusal (the coverage script judges a method call on its receiver, so dropping the outer mask would make it a finding) rather than a change, and n2 turned into a real discovery — `[Child, ***]` instead of `[Child, ***.env]`, which only the whole-message assertion could see.

My last three rounds found one class repeatedly, and then I shipped three instances of it myself in a comment I wrote to fix it: prose asserting a premise the same diff had falsified. That is why the note now states what is closed and declines to tally what is not. Worth knowing for the next change in this file — the convention is that a residual is stated where it is OWNED, which `describeAvailableOutputs`' docstring was already doing and I was not reading.

I pushed four things to your branch: the `:3411` note (twice, the second fixing the first), the ledger conflict that was blocking CI, and two `origin/main` merges. Also filed go-to-k/cdkd#3234 for the one real gap the review turned up — `resolveGetStackOutput` hands a raw stack name to an uncaught state read, and `S3StateBackend` quotes it back through a sanitiser rather than a masker. Pre-existing, not from this PR, and the `Fn::ImportValue` sibling already does it right.

Issue #2934, opened 2026-09-10
feat(deploy): adopt a rollback-orphaned resource instead of colliding with its name


go-to-k:
## Problem

A resource carrying `DeletionPolicy: Retain` that a rollback orphans stays in AWS and leaves cdkd's state. cdkd's generated physical names are deterministic (`generateResourceName` — `{stackName}-{logicalId}`, sha256-truncated over the type's length limit, **no random component**), so the next `cdkd deploy` asks AWS for a name the orphan still holds, fails already-exists, rolls back, and repeats. **Re-running never resolves it.**

Reported externally in go-to-k/cdkd#2902. That reporter recovered only by detaching policies from and deleting two IAM roles through the AWS API directly, plus scheduling deletion of orphaned KMS keys and deleting an orphaned S3 bucket.

The blast radius is wider than "users who chose `Retain`". CDK's stateful L2 constructs — `kms.Key`, `s3.Bucket`, `dynamodb.Table`, `logs.LogGroup`, `cognito.UserPool` and others — apply `RemovalPolicy.RETAIN` **by default**, and those same L2s leave the physical name unset so cdkd generates it. Retain-by-default and generated-name overlap almost exactly.

go-to-k/cdkd#2916 shipped a diagnosis: the deploy fails with a message naming the cause and printing a `cdkd import` command. That leaves recovery as manual work, using `import` for something it was not designed for. This issue replaces it with cdkd recovering by itself.

## Spec change

```
$ cdkd deploy      # fails partway, rolls back
$ cdkd deploy      # (template fixed) — SUCCEEDS
```

- the adopted resource keeps its contents — objects a `BucketDeployment` uploaded before the failure are preserved
- orphan hand-deleted → ordinary CREATE, succeeds
- the template now derives a DIFFERENT name → ordinary CREATE, succeeds; the old orphan stays behind, listed but not adopted
- ownership cannot be confirmed, or another stack already owns it → deploy STOPS, with go-to-k/cdkd#2916's existing diagnosis
- `cdkd diff` shows "adopt", not "create"; a record that is kept but not adoptable gets one info line
- `cdkd destroy` lists the records at its existing confirmation prompt before dropping them
- the state file survives a first-deploy rollback while a record exists

[...]

Issue #2798, opened 2026-09-08
test(unit): the full suite intermittently exits 1 with "Worker exited unexpectedly" while every test passes


go-to-k:
## Symptom

`vp test run` prints a fully green summary and still exits 1:

```
 Test Files  943 passed (943)
      Tests  20488 passed | 1 skipped (20489)
Type Errors  no errors
     Errors  1 error
```

The error is a pool-level crash, not a test failure:

```
Vitest caught 1 unhandled error during the test run.
Error: [vitest-pool]: Worker forks emitted error.
 ❯ EventEmitter.onTaskError .../vitest/dist/chunks/cli-api.BK8pd4xc.js:3459:21
Caused by: Error: Worker exited unexpectedly
```

This is the `Errors` vs `Type Errors` shape the repo already warns about — the
summary a reader scans says PASS, and only the exit code disagrees.

## Measured rate

2026-09-08, on `go-to-k/issue-2757-docs-audit` rebased onto `origin/main`
(`5ea84ad5`), macOS, `vp test run` with no filters:

- 8 full runs total, **2 exited 1** with this error (~25%).
- Not reproducible in isolation: the heaviest new test file on that branch ran
  **6/6 clean** on its own.

## It is not that branch's new tests

The branch adds `tests/unit/scripts/docs-error-strings.test.ts`, whose
`fails against real code` block copies a subtree and spawns the checker several
times — the obvious suspect. Disabling exactly that block (`describe.skip`) and
re-running the full suite **still reproduced the crash on run 4 of 4**, at the
same rate. The pool crash is independent of it.

An earlier attempt to isolate by DELETING the file is not a valid experiment and
should not be repeated: `tests/unit/scripts/rule-file-module-citations.test.ts`
then fails deterministically, because a `.claude/rules/` entry cites the file.

## Why it matters

The CI-green gate blocks a merge unless every check reports pass, so a ~25%
pool-crash rate means roughly one merge attempt in four is blocked by a run in
which nothing actually failed — and the natural response ("just re-run") is also
the response that would mask a real regression.

## Suggested starting points

- Whether the fork pool is hitting a memory ceiling: several suites spawn `node`
  on `.ts` entry points, each paying type-stripping.
- Whether a specific file correlates: bisect with `--shard` and compare rates.
- Whether `poolOptions.forks.maxForks` or `isolate` changes the rate.

## Classification

Dup-check: searched open and closed issues for "Worker exited", "vitest-pool",
[...]

Issue #805, opened 2026-06-12
ECS CapacityProvider create races same-stack infrastructure-role propagation (ServiceAccessDeniedException not retried)


sakurai-ryo:
## Problem

cdkd's event-driven DAG dispatches the Cloud Control `CreateResource` for an `AWS::ECS::CapacityProvider` (Managed Instances) as soon as its same-stack infrastructure role finishes creating. cdkd's fast SDK path creates the IAM role without waiting for propagation, so ECS tries to assume the just-created `InfrastructureRoleArn` before IAM has propagated it and rejects the create:

```
CREATE failed for RunnerCapacityProviderA547DE0A: Invalid request provided: CreateCapacityProvider error: Caught ServiceAccessDeniedException for ECSInfrastructureRole[arn:aws:iam::...:role/...] (Service: Ecs, Status Code: 400, Request ID: ...) (SDK Attempt Count: 1)
```

The CC API resource handler classifies this as a terminal `InvalidRequest` (no internal retry — `SDK Attempt Count: 1`), and cdkd's outer `withRetry` does not recognize the message either: none of the patterns in `src/deployment/retryable-errors.ts` match `ServiceAccessDeniedException` (`cannot be assumed` / `not authorized to perform` / `trust policy` etc. do not appear in the message). The deploy fails fast on a transient error.

This is the ECS manifestation of the general "cdkd fast path beats IAM propagation" class already fixed for RDS Enhanced Monitoring (#794) and the Custom Resource path — CloudFormation tolerates it via deployment latency; cdkd does not.

Surfaced by a real-world stack: an ECS Managed Instances capacity provider whose infrastructure role (trust: `ecs.amazonaws.com`, managed policy `AmazonECSInfrastructureRolePolicyForManagedInstances`) is created in the same deploy. The template is correct; once IAM has settled, the same create succeeds.

## Proposed fix

Add `'Caught ServiceAccessDeniedException'` to `RETRYABLE_ERROR_MESSAGE_PATTERNS` in `src/deployment/retryable-errors.ts` so the deploy engine's existing `withRetry` (8 attempts, ~47s cumulative) absorbs the propagation window — mirroring the `ENHANCED_MONITORING` pattern added for #794. The phrase is anchored on the CC-API/ECS handler wording so a genuine, permanent role misconfiguration only burns the bounded retries before surfacing.

## Verification

[...]

PR #3182, opened 2026-09-15
fix(provisioning,deployment): generate a CapacityProviderName for a Cloud Control create, and keep generated names out of update patches

### LINKED ISSUE #3174: fix(provisioning): a Cloud Control-routed AWS::Lambda::CapacityProvider without CapacityProviderName fails with Resource Handler Internal Failure
## What is wrong

A CDK `lambda.CapacityProvider` created without `capacityProviderName` did not deploy under cdkd. The create failed about one second after submission:

```
Failed to create Provider2281708E: CREATE failed for Provider2281708E: Resource Handler Internal Failure
```

`AWS::Lambda::CapacityProvider` has no SDK provider, so it is routed via Cloud Control (rule 6 in `ProviderRegistry.getProviderFor`). Before the create, `DeployEngine.preparePropertiesForCcApi` calls `applyDefaultNameForFallback`, which only fills a name for types listed in `FALLBACK_NAME_RULES` (`src/provisioning/resource-name.ts`). This type is not listed, so the desired state goes out with no `CapacityProviderName`. The verbose log shows exactly that: `DesiredState for Provider2281708E: keys=["InstanceRequirements","PermissionsConfig","VpcConfig"]`.

The CDK construct allows the name to be omitted, and the schema does not require it (`required` is `PermissionsConfig` and `VpcConfig`; `primaryIdentifier` is `/properties/CapacityProviderName`). In direct Cloud Control calls, omitting it returned `InternalFailure` while the same request with a name succeeded (table below). The generic error does not say what the handler failed on, but the name is the only difference between the two requests.

This blocked the tested Lambda Managed Instances app on cdkd 0.290.6 in ap-northeast-1. Setting `capacityProviderName` explicitly is the workaround.

## Reproduction

cdkd 0.290.6, aws-cdk-lib 2.269.0, ap-northeast-1, 2026-09-15. The app, trimmed to what matters here (the tested handler also logged its remaining time, and `addFunction` also set `perExecutionEnvironmentMaxConcurrency: 4` and `executionEnvironmentMemoryGiBPerVCpu: 2`):

```ts
const vpc = new ec2.Vpc(stack, 'Vpc', {
  maxAzs: 3,
  natGateways: 0,
  subnetConfiguration: [{ name: 'isolated', subnetType: ec2.SubnetType.PRIVATE_ISOLATED }],
});
const sg = new ec2.SecurityGroup(stack, 'LmiSg', { vpc });

const provider = new lambda.CapacityProvider(stack, 'Provider', {
  subnets: vpc.isolatedSubnets,
  securityGroups: [sg],
  architectures: [lambda.Architecture.ARM_64],
});

const fn = new lambda.Function(stack, 'BatchJob', {
  runtime: lambda.Runtime.PYTHON_3_13,
  architecture: lambda.Architecture.ARM_64,
  memorySize: 2048,
  timeout: Duration.minutes(90),
  handler: 'index.handler',
  code: lambda.Code.fromInline('def handler(event, context):\n    return {}\n'),
});
provider.addFunction(fn, { publishToLatestPublished: true });
```

Isolating it from cdkd with direct `aws cloudcontrol create-resource` calls, run back to back with one operator role (`AWSLambdaManagedEC2ResourceOperator` attached, created five minutes before the first call), the same two subnets, the same security group and `InstanceRequirements.Architectures: ["arm64"]`:

| Order | `CapacityProviderName` | Result |
| --- | --- | --- |
| 1 | omitted | `FAILED` / `InternalFailure` / `Resource Handler Internal Failure` |
| 2 | `cdkd-lmi-probe3` | `SUCCESS` |
| 3 | omitted | `FAILED` / `InternalFailure` / `Resource Handler Internal Failure` |

The named create succeeding between two unnamed failures with the same role rules out IAM propagation. (A role only seconds old fails differently, with `InvalidRequest: The operator role is invalid or doesn't have sufficient permissions`.)

A re-run of `cdkd deploy --no-rollback` about a minute after its operator role had been created also failed with `Resource Handler Internal Failure`, about one second after submission.

With `capacityProviderName` set, the same app deploys, `cdkd diff` reports no changes, and `cdkd destroy` removes all 14 resources. The function (routed via Cloud Control for `CapacityProviderConfig`, see go-to-k/cdkd#1616) comes up with `Timeout: 5400` on an `Active` `$LATEST.PUBLISHED`.

## Proposed fix

Add `AWS::Lambda::CapacityProvider` to `FALLBACK_NAME_RULES` with `nameProperty: 'CapacityProviderName'` and options that satisfy the schema: `maxLength: 140`, pattern `[a-zA-Z0-9-_]+` (the default `allowedPattern` already produces a subset of that).

Two things to settle in the same pass:

- An entry in `FALLBACK_NAME_RULES` also admits the type to orphan adoption (`nameProperty` feeds the adoption allow-list; see the comment above `ADOPTION_REFUSED_TYPES` in `src/deployment/orphan-adoption.ts`). `CapacityProviderName` being create-only constrains updates but says nothing about what a duplicate create does, so before admitting the type, check the three tests that set documents: whether a create with an existing name mints a new resource instead of colliding, whether an update is refused, and whether the resulting replacement would delete the retained provider.
- Only this type has been measured. Other Cloud Control-routed types whose name is left out of the template may fail the same way; this issue does not claim they do, and the fix here is scoped to the measured type.

## Verification plan

- Unit: `applyDefaultNameForFallback` fills `CapacityProviderName` when absent, keeps a user-supplied name, and the generated name matches the schema pattern and stays within 140 characters for a long stack name plus logical id.
- Unit: the orphan-adoption decision for this type matches whatever the adoption check above concludes.
- Integ: a new fixture that deploys a `CapacityProvider` with no explicit name and a function attached to it, asserts the provider exists under the generated name, and destroys it with no EC2 or capacity provider left behind. It launches EC2 instances through the capacity provider, so it carries a real charge.

Dup-check: searched open and closed issues for `CapacityProvider`, `CapacityProviderName`, `managed instances`, `Resource Handler Internal Failure`, `physical name Cloud Control` and `name generation cloud control`, and read `FALLBACK_NAME_RULES` and `preparePropertiesForCcApi`. go-to-k/cdkd#1616 (open) wires `CapacityProviderConfig` into the Lambda SDK provider and does not touch the capacity provider's name; go-to-k/cdkd#805 (closed) is the ECS capacity provider's IAM propagation race. None covers a missing default name on a Cloud Control create.
Session-fit: now (do it in this session) -- context test: `src/provisioning/resource-name.ts` read; `src/deployment/deploy-engine.ts` (`preparePropertiesForCcApi` and the create call site) read; `src/provisioning/provider-registry.ts` read; `src/deployment/orphan-adoption.ts` read; `tests/unit/provisioning/resource-name.test.ts` and the orphan-adoption unit tests not read; the integ fixture is new. Loaded files alone make it `now`, the real-AWS reproduction exists only in this session, and the reporter is taking the fix.
Severity: medium -- a capability is missing (a Lambda Managed Instances app with an unnamed capacity provider does not deploy) and a workaround exists (set `capacityProviderName`).
Effort: large (L) -- a new integ fixture must be written, and it launches EC2 through the capacity provider.
Estimate: ~3-4 h -- most of it the integ fixture and its teardown check; the table entry and unit cases are small.

nix-tkobayashi:
Closes #3174

## Summary

A CDK `lambda.CapacityProvider` created without `capacityProviderName` did not deploy under cdkd: the create failed about a second after submission with `Resource Handler Internal Failure`. `AWS::Lambda::CapacityProvider` has no SDK provider, so it is created through Cloud Control, and a Cloud Control create without `CapacityProviderName` fails even though the schema does not require it (go-to-k/cdkd#3174 has the back-to-back measurement: the same request with a name succeeded between two unnamed failures under one operator role).

### The create: a generated name

`src/provisioning/resource-name.ts` adds `AWS::Lambda::CapacityProvider` to `FALLBACK_NAME_RULES` (`nameProperty: 'CapacityProviderName'`, `maxLength: 140`, the default `allowedPattern`, whose output is a subset of the schema's `[a-zA-Z0-9-_]+`). `DeployEngine.preparePropertiesForCcApi` then fills `<stack>-<logicalId>` on the Cloud Control create, hash-truncated past 140 characters. A template-supplied name is kept.

### The update: the generated name is taken back out

Review found that the same fill also ran on the Cloud Control UPDATE path, and that this was wrong for every type in the table. Every site records the template's resolved properties, so the recorded bag (the patch's previous side) never holds a generated name. Filling it on the desired side made each update patch `add` the name: the live value for a resource cdkd created, but for a resource running under another name a real change, refused where the name is create-only and a rename where the type accepts the update. Derived from the committed schema fixtures: 38 of the other 40 rules name a create-only property; `AWS::IAM::Policy.PolicyName` and `AWS::Cognito::UserPool.UserPoolName` are updatable. `AWS::Lambda::CapacityProvider` has no committed fixture; its live `DescribeType` lists `CapacityProviderName` in `createOnlyProperties`.

`withoutGeneratedFallbackName` (new export in `src/provisioning/resource-name.ts`) takes a generated name back out: a name counts as generated exactly when the fill would have produced it (the type has a rule, the resolved template bag leaves the property falsy, and the prepared bag carries it). The property then goes back to what the template holds, deleted when the template has no such key and the template's own falsy value (an empty string, which the fill generates over) when it does; otherwise the prepared bag is returned by identity. `src/deployment/deploy-engine.ts`'s UPDATE call site passes the prepared bag through it. The CREATE site and both replacement-create sites still fill the name.

Three neighbouring comments are corrected to match: the `FALLBACK_NAME_RULES` header (a Cloud Control create, including a type with no SDK provider), the `ADOPTION_ONLY_NAME_PROPERTIES` note (adding a rule changes the name later creates use, not an existing resource's; an adoption-only entry still admits its type to adoption), and `preparePropertiesForCcApi` (it no longer "no-ops" for a Cloud Control-native type).

### The operator role: retried as IAM propagation

The first real-AWS run of the new fixture failed past the name: Cloud Control created both providers under their generated names, then rejected them with `The operator role is invalid or doesn't have sufficient permissions. Verify the role and permissions and try again.` (InvalidRequest, HTTP 400). CDK's `lambda.CapacityProvider` creates that operator role in the same stack, cdkd issues the create seconds after the role, and no `IAM_PROPAGATION_ERROR_MESSAGE_PATTERNS` entry matched the message, so the create was single-shot. go-to-k/cdkd#3174 already recorded this message for a role only seconds old. A probe against a fresh role (us-east-1, 2026-09-15, one sample) got it 8 s after the role's policy was attached and a SUCCESS at 23 s, inside the dense grid's ~47.75 s of backoff. `src/deployment/retryable-errors.ts` now carries that sentence UP TO `sufficient permissions` as an IAM-propagation pattern -- the advisory tail is deliberately left out so a reword of it cannot blunt the match, which does make the `.includes` test broader than the quoted sentence -- so a genuinely missing permission only burns the bounded retries before surfacing.

### Orphan adoption

An entry in `FALLBACK_NAME_RULES` also admits the type to rollback-orphan re-adoption (go-to-k/cdkd#2934). The issue asked for the three `ADOPTION_REFUSED_TYPES` tests to be checked first. Measured against Cloud Control in us-east-1 on 2026-09-15: a create under an existing `CapacityProviderName` failed with `AlreadyExists: Capacity provider already exists` and only one provider was listed, so the first test (the create mints a new resource instead of colliding) does not hold, and the type is not added to the refusal set.

## Tests

- `tests/unit/provisioning/resource-name.test.ts`: the fill when absent (input not mutated) and a user-supplied name kept by identity; a 140-character name kept verbatim, a 141-character one cut to exactly 140 ending in the hash, and a nested child stack name (`<parent>~<logicalId>`, whose `~` the schema rejects) sanitized into the schema pattern at 140; `fallbackNamePropertyFor` for four table members (two with updatable names), a type with no rule, and an adoption-only type; `withoutGeneratedFallbackName` taking the name out without mutating either input, doing so for an updatable member, giving an empty template name back, and returning the prepared bag by identity for a template-supplied name, a bag with no name, and a type with no rule.
- `tests/unit/deployment/orphan-adoption.test.ts`: the type is adopted through the real `explicitNamePropertyFor`.
- `tests/unit/deployment/deploy-engine-cc-update-fallback-name.test.ts` (new): a Cloud Control CREATE receives a generated name; an UPDATE's desired and previous bags carry none while the real change (`Tags`) still reaches the provider; a template-supplied name is still sent on an update; the property-driven replacement's create and the UPDATE-not-supported fallback's create each still receive a generated name; and, for a type with NO rule, what an SDK provider's `preparePropertiesForFallback` hook adds reaches the provider whole.
- `tests/unit/deployment/retryable-errors.test.ts`: the operator-role rejection, verbatim as the fixture's first run surfaced it, is matched by exactly the new entry within the IAM-propagation subset and across every retryable message pattern, is classified as IAM propagation and retryable, and a same-handler Availability Zone rejection stays terminal. A second near-miss, CONSTRUCTED rather than observed, reuses the anchor's opening clause and then diverges, so an anchor shortened short of `sufficient permissions` reds it -- the Availability Zone case shares no token with the anchor and could not constrain its length. Probes: removing the entry reds the three positive cases (the IAM-subset exact match, the classification, and the across-every-pattern exact match); adding a matching pattern outside the IAM subset reds the across-every-pattern case.
- `tests/unit/scripts/integ-aws-commands.test.ts`: unchanged, and deliberately so. An earlier revision of this branch moved the parsed-invocation band to 3600/4400, which LOWERED the floor and ceiling `main` had since re-tracked to 3630/4430; the hunk was dropped on the rebase, the file is byte-identical to `origin/main`, and the new fixture's invocations sit inside main's current band with the suite green.
- Mutation probes, one at a time with the file restored byte-exact between them, each reddening at least one case: the `withoutGeneratedFallbackName` guards (no rule, no name in the prepared bag, truthiness read as definedness), its restore branch, a version that mutates the prepared bag, the 140 cap lowered to 64, `fallbackNamePropertyFor` restricted to one type or without its optional chaining, the engine's UPDATE call site without the wrapper, and the fill removed at the property-driven replacement create and at the UPDATE-not-supported fallback create (each reddens exactly its own case). Three more from the review round: shortening the operator-role anchor to `The operator role is invalid` reds 3 cases including the new near-miss; deleting the `FALLBACK_NAME_RULES` entry reds 9 in `resource-name.test.ts` including `keeps a user-supplied name untouched`, which stayed green before its premise was added; and deleting `withoutGeneratedFallbackName`'s `nameProperty === undefined` arm reds 2 cases across the two files that watch it -- the helper's own `returns prepared by identity for a type with no rule, whatever keys it holds`, which already killed it, AND the engine's no-rule UPDATE case, which did NOT until it carried a key making the otherwise-inert mutant observable. The added key buys the ENGINE-level case, not the first kill.

## Integration

`tests/integration/lambda-capacity-provider-default-name` (new): a VPC with isolated subnets, an unnamed capacity provider with a function attached via `publishToLatestPublished`, and a second unnamed provider with nothing attached.

1. Deploy: both providers are recorded under their generated names on the `cc-api` route and are `Active`; the function's `CapacityProviderConfig` points at the first; the `ProviderArn` output agrees.
2. A clean `cdkd diff`.
3. The spare provider's state record is re-pointed at a provider created out of band under a different name, and the generated spare is deleted. This is the premise the update-path fix is about: a recorded bag that names nothing and a live name that is not the generated one. The record is rewritten directly rather than through `cdkd import`, which records the template's unresolved properties and `provisionedBy: 'sdk'` (`src/cli/commands/import.ts`), two differences this phase is not about.
4. A tags-only UPDATE of both providers: each tag reaches AWS and each provider keeps its ARN and state physical id; a clean diff under that mode.
5. Destroy: every provider, the function, both per-provider operator IAM roles and the VPC are gone, no instance is left in the VPC, and the state file is removed. The role names are read out of the state record rather than derived -- the spare's is exactly 64 characters, the IAM `RoleName` cap, so one more character would take it down the truncate-and-hash path and a derived literal would stop naming the live role -- and a premise asserts both roles are LIVE after the create, so the teardown check cannot pass for the wrong reason.

Not established by the fixture: that step 4 fails on a binary without the update-path fix; the unit case carries that direction.

- `lambda-capacity-provider-default-name`: PASS on the final tree, `verify.sh` rc=0 in 555 s (re-run 2026-09-16T04:06:19Z after the review round edited `verify.sh`); every phase ran (both providers `Active` under their generated names, clean diff, the spare re-pointed at the out-of-band provider, both tags reaching AWS in place, clean diff under the update mode), 2 operator-role propagation retries fired in phase 1 against 447 captured debug lines, destroy 17 deleted 0 errors, both operator roles live before it and gone after, 0 surviving object versions and an empty state prefix. The first run, before the operator-role pattern, FAILED in phase 1 on the propagation rejection described above (78 s) and rolled back with nothing left in AWS. An intermediate run at 563 s was taken on a host with GNU coreutils ahead of `/usr/bin`, so it never exercised the BSD `wc` path the review found; that is what the re-run replaces.
- `lambda` (broad set): PASS on the final tree, `verify.sh` rc=0 in 93 s, destroy 9 deleted 0 errors.
- State bucket: every version under the fixture's prefix swept; under `LambdaStack` only this session's four event objects were removed, leaving other runs' objects and the shared `index.json`.

## Verification

- `/check`: `vp check --fix`, `vp run check`, `vp run typecheck:test`, `vp run build` and the matrix regeneration all pass on the rebased tree. In the runs that completed, the full `vp test run` failed no test (1023 files; 23227 passed, 1 skipped on the base before the review round), but full runs on this host exited 1 on the pool-level `Worker exited unexpectedly` crash tracked in go-to-k/cdkd#2798, with no test failing. Run in isolation, each of the 11 `tests/unit` directories passed with rc=0 and no worker exit. Final clean run after the review round: rc=0, 1025 files, 23489 passed, 1 skipped, no worker exit. The attempt before it, started while another worktree's suite was running, failed two `tests/unit/scripts` critics on `Test timed out in 5000ms` (`gen-handled-property-wiring`, `withretry-interrupt`) and passed on the identical tree next run.
- `/check-docs`: no doc, rule file or the distributed plugin skill describes the name fill, the Cloud Control property preparation or the operator-role rejection; the dense-grid rationale in `.claude/rules/layout-deployment.md` (and its twins in `retryable-errors.ts` / `retry.ts`) is qualified to "usually ... within seconds", and `docs/troubleshooting.md` already said "usually"; `docs/testing.md` names fixtures as examples rather than listing them, so the new one needs no entry.
- Codex delta review: 3 rounds -- r1 the generated name reached the UPDATE patch, a create-only change for a provider under another name (fixed by taking it back out on the update path), r2 the comments and changelog called every table name create-only although two are updatable (reworded from the schema fixtures), r3 NO ISSUES. Maintainer-checklist proxy pass: r1 six findings (the helper's guards, the replacement-create sites and the other table members were unpinned; the fixture's update exercised only the equal-name case; three neighbouring comments were stale; the 140 boundary was unpinned and a comment miscounted it), r2 two findings (the sanitize case stayed green with sanitization removed, because the `_` it used is schema-legal; a fixture comment still described the abandoned `cdkd import` flow), r3 NO ISSUES. The operator-role pattern, added after the fixture's first run, got its own round: Codex delta review 1 round, NO ISSUES; maintainer-checklist proxy pass r1 three findings (the "resolves in single-digit seconds" rationale in `retryable-errors.ts`, `retry.ts` and `.claude/rules/layout-deployment.md` contradicted the 23 s measurement and is now "usually ... within seconds"; the near-miss cases checked fixed strings, replaced by an exact set over every retryable message pattern; a test name credited dense-grid selection it does not exercise), r2 NO ISSUES.

## Review round 2 (maintainer review at `481b51c`)

M0 both `wc -w` captures trim through `| tr -d ' '` (BSD `wc` pads to width 8, so the untrimmed value is `"       0"` on macOS and every `=` against it is false), plus the fixture re-run and fresh ledger row above. M1 the band hunk dropped, see above. M2 `docs/cli-flag-coverage.md` regenerated on the rebased tree. M3 the operator-role assertions, see above. go-to-k/cdkd#3199 (the rollback executor's fourth create site) deliberately left out, as the review asked. All five take-or-leave items taken: `fallbackNamePropertyFor`'s doc names its real readers and `explicitNamePropertyFor` now calls it; the operator-role anchor's comment says it stops at `sufficient permissions` and that the `.includes` test is therefore broader than the quoted sentence; the near-miss and the two premises above; and `isTransientAuthzThrow`'s doc no longer claims EVERY member of the shared union is a front-door rejection -- the go-to-k/cdkd#3174 entry is a resource-handler message, inert there only because no CR `Invoke` failure carries its text, and `delivered` cannot discharge the next such entry because `onDelivered()` runs after `invokeLambda` resolves. `.claude/rules/provider-custom-resources.md` asserted the same two things and is corrected with it.

Round-2 review: Codex delta 3 rounds -- r1 the false `delivered` claim, r2 `delete_provider`'s non-retryable guard missing the AWS CLI's own credential-resolution failures (so an unauthenticated shell burned the full 40 x 15 s per provider, and the post-accept poll had the same spin), r3 NO ISSUES. Maintainer-checklist proxy pass 2 rounds -- r1 four findings (the inert mutant above, the contradicting sentence in `.claude/rules/provider-custom-resources.md`, a stale phase overview and a stale `--verbose` comment in `verify.sh`, and a `wc` site count that was wrong), r2 NO ISSUES.


go-to-k:
Reviewed at head `481b51c`. Tier: 3-axis plus a security pass (1290 LOC after excluding `docs/_generated/**`, 25 files; every touched `src/` file also carries 2 `fix:` commits in its last 3, so the recent-defect bias applies).

**The core change is sound and I want to say so before the requests.** Four reviewers independently re-derived the load-bearing claims rather than taking the body's word for them, and they hold:

- The fill and the removal predicates are exact inverses. Both gate on `props[nameProperty]` truthiness, so an empty string / `0` / `null` template value is generated over and then restored identically, and a template name that happens to equal the generated string returns by identity — the helper never compares values, so hash truncation and `~` sanitization are irrelevant to it.
- "The recorded bag never holds a generated name" was traced, not assumed: `applyDefaultNameForFallback` has exactly one consumer, whose four call sites are `deploy-engine.ts:5452` (create), `:5898` (property-driven replacement create), `:6435` (update — the only wrapped one) and `:6808` (UPDATE-not-supported replacement create); state records take `propertiesToRecord(resolvedProps, …)`, and `CloudControlProvider` returns no `effectiveProperties`. So the patch's previous side is always template-shaped and the update emits no name op at all, rather than a `remove`.
- The 38-of-40 create-only enumeration was re-parsed from the committed schema fixtures: the list, not just the count, is right — `AWS::IAM::Policy.PolicyName` and `AWS::Cognito::UserPool.UserPoolName` are exactly the two exceptions.
- `CapacityProviderName`'s `maxLength: 140` and `[a-zA-Z0-9-_]+` were confirmed against live `DescribeType` rather than the body, and the default `allowedPattern` emits a strict subset.
- Security pass: the generated name is never persisted, never reaches `destroy` (which targets AWS's `Identifier`), and the orphan-adoption admission cannot adopt a provider cdkd did not create — `StackOrphanRecord`s are minted only by cdkd's own rollback, and a name collision on create fails `AlreadyExists` without converting to adoption. The new retry pattern widens a denial to retryable on deploy, rollback and destroy, but every path rethrows after the bounded ~47.75 s; no denial becomes a success or a skip.

Four things to fix before this can merge, then two notes.

---

## Must fix before merge

**M0 — `tests/integration/lambda-capacity-provider-default-name/verify.sh:365` fails unconditionally on a stock macOS host, and the same idiom is at `:249`.**

```sh
INSTANCES_LEFT="$(aws ec2 describe-instances … --output text | wc -w)"
[ "${INSTANCES_LEFT}" = "0" ] || { echo "FAIL: …"; exit 1; }
```

BSD `wc` pads its count to width 8, and `$(…)` strips only the trailing newline, so `INSTANCES_LEFT` is `"       0"` and the comparison is false. Executed verbatim on a macOS host with only `/usr/bin/wc`: `NOT equal, value=[       0]`. 33 other sites in `tests/integration/*/verify.sh` pipe `wc` through `| tr -d ' '` for exactly this reason. Please add it at both lines.

Worth reconciling alongside it: the ledger row (`docs/_generated/integ-last-run.tsv`, `2026-09-15T17:03:38Z`, PASS, 563 s) is not consistent with the committed file on a BSD-`wc` host, and those two lines were not touched by the review-round commit. If the run was on a host with GNU coreutils ahead of `/usr/bin` on `PATH`, that explains it — but the next agent to re-run this fixture burns ~560 s of real AWS and gets a red. The fix needs a re-run and a fresh ledger row either way.

**M1 — the `integ-aws-commands.test.ts` band hunk now LOWERS `main`'s fence; please drop it rather than resolving the conflict either way.**

`main` has since moved to `total > 3630` / `total < 4430` / `verbs > 385` (re-tracked at 3903). This PR sets `> 3600` / `< 4400` and leaves `verbs > 380`. Taking this side on the conflict regresses all three edges — and a ceiling raised while the floor stays put is precisely the silently-widened window the surrounding comment says must not happen.

Re-measured with the test's own `readFixtureScripts` + `extractAwsInvocations`: `main` is **3905**, and the new fixture adds **26**, for **3931** — not the 3915 the new comment cites. 3931 sits inside `main`'s current band, so after the rebase no band change is needed at all and the whole hunk can go.

**M2 — `docs/cli-flag-coverage.md` also conflicts; regenerate it on the rebased tree rather than hand-resolving.**

**M3 — `verify.sh` never asserts the two operator IAM roles are gone after destroy.**

The destroy phase checks the providers, the function, the VPC, the instances and the state file, but CDK's `lambda.CapacityProvider` creates an operator role per provider and a leak of those would pass silently. The fixture is new, so this is the moment to close it; please add an `assert_gone` for each.

---

## Filed separately, not asked of this PR

The rollback executor's `reverse-replacement` replay-CREATE (`src/deployment/rollback-executor.ts:2500` and `:2654`) hands `{ ...resolvedPrevProps }` — the recorded, never-named bag — straight to a provider routed by `prev.provisionedBy`, with no `applyDefaultNameForFallback` anywhere in that file. It is a fourth create site that your three siblings fix and it does not. For this type the replay create then fails with `Resource Handler Internal Failure`, and on the delete-new-first fallback the new resource is already gone and the state entry already deleted, so the provider ends absent from both AWS and state.

I checked whether this belongs in your diff and concluded it does not: that file is byte-identical on `main`, the gap applies to all 41 table types there today, and the one reachable shape for this type is narrow (the physical id IS the name, so the four other create-only properties re-create under the same name and never select the arm — it needs a deploy that ADDS an explicit `capacityProviderName` to a previously unnamed provider and then fails on a later resource). Tracked as go-to-k/cdkd#3199. Please leave it out of this PR.

---

## Take or leave

- `resource-name.ts:515` — "Read by the deploy engine's Cloud Control UPDATE path" is not quite true of `fallbackNamePropertyFor`; `deploy-engine.ts` imports only `withoutGeneratedFallbackName`, and the helper's sole non-test reader is its sibling in the same file. Separately, `explicitNamePropertyFor` at `:588` still does the raw `FALLBACK_NAME_RULES[resourceType]?.nameProperty` lookup instead of calling the new helper.
- `retryable-errors.ts:166` — the comment says "Anchored on the handler's full 'operator role' sentence", but the pattern stops at `sufficient permissions` and drops `. Verify the role and permissions and try again.`; the `.includes` match is a little broader than the comment claims. Relatedly, the near-miss test (the Availability Zone rejection) shares no token with the anchor, so it does not constrain the feared regression — an anchor shortened to `operator role is invalid` would still pass it.
- `custom-resource-provider.ts:1984` — `isTransientAuthzThrow`'s doc asserts every member of the shared union is a front-door rejection where "the handler provably did not run". The new entry is a resource-handler message, so that invariant now has a false member. Inert in practice (the `delivered` flag fences the replay), but the wording should narrow.
- `resource-name.test.ts` "keeps a user-supplied name untouched" (`toBe(props)`) stays green if the `FALLBACK_NAME_RULES` entry is deleted, since it also passes for the no-rule reason.
- The `AlreadyExists` measurement that settled the orphan-adoption question lives only here and in a test comment; go-to-k/cdkd#3174 asked for that determination, so a one-line comment there would keep the answer with the question.

---

Push M0 through M3 and I will re-review and merge. Thanks for the unusually thorough body — the mutation-probe list and the two integ runs made this much faster to check.

nix-tkobayashi:
Pushed as `3beb6374`. All four must-fix items and all five take-or-leave items are in; nothing is deferred except one changelog request I am pushing back on, at the end.

## M0 — the BSD-`wc` trim

Both sites take `| tr -d ' '`. The comment beside the first one says why, but without a site count. My original "33" was wrong, and when I went to correct it I could not land on one number either — how you count depends on whether you include the `wc -l <<< "..."` input-redirection form alongside the piped one, and three of the sites trim with `tr -d '[:space:]'` rather than `tr -d ' '`. A count in a comment that nothing re-derives is exactly the line that goes stale in silence, so it now names both trim spellings and asserts no total.

**You were right about the ledger row, and about the cause.** The run behind it was on a host with GNU coreutils, so the committed file was never exercised on a BSD `wc`. It has been re-run and the row replaced — details under Verification below.

## M1 — the band hunk

Dropped entirely, not resolved. `tests/unit/scripts/integ-aws-commands.test.ts` is now byte-identical to `origin/main` (`git diff origin/main...HEAD` lists it not at all), so the floor is back at 3630, the ceiling at 4430 and `verbs` at 385 — the last of which had already auto-merged to your value, since my commit never touched that line. The suite passes on main's band with the new fixture present, which is the check your re-measurement predicted.

## M2 — `docs/cli-flag-coverage.md`

Regenerated on the rebased tree through `vp run gen:all-matrices`, not hand-merged.

## M3 — the operator roles

Two `assert_gone` calls in the destroy phase, plus a premise that closes the vacuity the fixture would otherwise have: both roles are asserted LIVE right after the phase-1 deploy, so "gone after destroy" cannot pass because the role never existed. The probe is a strict capture of `iam get-role --query Role.Arn`, so a throttle or an auth failure aborts rather than reading as absent.

The names are read out of the state record rather than derived, and that is not defensiveness: cdkd generates `<stack>-<logicalId>` here, and `CdkdLmiCapacityProviderExample-SpareProviderOperatorRole6345D27C` is **exactly 64 characters** — the IAM `RoleName` cap — which I measured rather than estimated. One more character in the stack or construct name takes that role down the truncate-and-hash path, and a derived literal would then quietly stop naming the live role while the assertion still "passed".

The re-run confirms the assertions are live: the destroy log shows `ProviderOperatorRoleDF614C31` and `SpareProviderOperatorRole6345D27C` deleted, and the new premise line printed both names.

## go-to-k/cdkd#3199

Left out, as you asked. Thanks for tracing it and for saying explicitly that it does not belong in this diff — that saved a round.

## Take-or-leave — all five taken

- your `:515` — corrected. The doc now says the reader is its sibling `withoutGeneratedFallbackName`, which the deploy engine calls from the Cloud Control UPDATE path. I verified your claim against the pre-change revision before rewording, and it held: `fallbackNamePropertyFor` had exactly one non-test reader there, `withoutGeneratedFallbackName` in the same file, and `deploy-engine.ts` does not import it at all (it takes `withoutGeneratedFallbackName`, `explicitNamePropertyFor`, `withStackName`, `applyDefaultNameForFallback`, `getCurrentStackName` and `looksLikeCdkdGeneratedName` from that module, but not this one). One consequence worth naming: your other take-or-leave item on the same line adds a SECOND non-test reader, since `explicitNamePropertyFor` now calls the helper instead of repeating the raw table lookup. The doc names both, and says neither is outside this file.
- `retryable-errors.ts:166` — the comment now says the pattern is anchored up to `sufficient permissions` and **no further**, that the advisory tail is deliberately left out so a reword of it cannot blunt the match, and — the part your note was really about — that this does make the `.includes` test broader than the quoted sentence.
- The near-miss — added. You are right that the Availability-Zone case constrains nothing about the anchor's length. The new case reuses the anchor's opening clause and then diverges, so every prefix short of `sufficient permissions` matches it and reds it. It is labelled as CONSTRUCTED and explicitly not a message AWS has been observed to emit, since I did not measure it.
- `custom-resource-provider.ts:1984` — narrowed, and the narrowing got further than your note asked. My first attempt discharged the new member against the `delivered` flag; Codex caught that as false, and it is: `onDelivered()` runs after `invokeLambda` **resolves**, so a throw between acceptance and resolution reaches the classifier with the flag still false. The doc now says the shared union can acquire a non-front-door member, that this one is inert only because no CR `Invoke` failure carries its text, and that `delivered` may not be used to discharge the next one. `.claude/rules/provider-custom-resources.md` asserted both of the same things and is corrected in the same change — it said "every pattern" and "`delivered === false` ... therefore re-invokes nothing".
- `resource-name.test.ts` — premise added, and measured: with `expect(fallbackNamePropertyFor(TYPE)).toBe('CapacityProviderName')` in place, deleting the `FALLBACK_NAME_RULES` entry reds that case (9 of 36 in the file go red; before, that one stayed green for the no-rule reason).
- The `AlreadyExists` determination is now recorded on go-to-k/cdkd#3174 itself, as a comment: https://github.com/go-to-k/cdkd/issues/3174#issuecomment-5691597743

## Two more, found by the passes on this delta

Both are in code this round touched, so I am naming them rather than letting them arrive as a surprise in your re-review.

1. **`delete_provider`'s non-retryable guard was too narrow.** It matched the service's authorization codes only, so a shell with no resolvable credentials still burned the full 40 x 15 s per provider — three providers, half an hour of cleanup — for a condition no wait can fix, while the comment above it claimed an immediate exit. The guard is now a named pattern covering the AWS CLI's own credential-resolution failures too, and it applies to the post-accept poll as well, which had the same 60 x 5 s spin.
2. **The no-rule UPDATE case was asserted rather than pinned.** Deleting `withoutGeneratedFallbackName`'s `nameProperty === undefined` arm left it green, because the mutant is INERT for a type with no rule (`resolvedProps[undefined]` is falsy and `undefined in prepared` is false, so the bag returns unchanged either way). It now pins the invariant (`fallbackNamePropertyFor` is `undefined` for the type, not "EFS happens to have no entry") and carries a key that makes the mutant observable. Measured: the mutant now reds exactly that one case.

## Verification

- **Integ re-run**, `tests/integration/lambda-capacity-provider-default-name`, real AWS `us-east-1`, 2026-09-16T04:06:19Z: **PASS in 555s**, 17 deleted / 0 errors, 2 operator-role propagation retries during phase 1 (447 debug lines captured, so the count is not vacuous), both operator roles live pre-destroy and gone after, 0 surviving object versions, state prefix empty. Ledger row replaced with those numbers.
- **Unit suite**: 1025 files, 23489 passed, 1 skipped, no type errors. `vp run check`, `vp run typecheck:test`, `vp run build` all clean.
- **CI** on `3beb6374`: `check-build-test` and `once-leak-detect` were still running when I posted this; everything else had reported `pass` or `skipping`. I am watching it and will say so here if either goes red.
- **Codex delta review**: 3 rounds — r1 the false `delivered` claim plus two behaviour changes my prompt had failed to describe, r2 the credential-guard gap in item 1 above, r3 NO ISSUES. **Maintainer-checklist proxy pass**: 2 rounds — r1 four findings (the inert mutant in item 2, the contradicting sentence in `.claude/rules/provider-custom-resources.md`, a stale phase overview and a stale `--verbose` comment in `verify.sh`, and the wrong `wc` site count), r2 NO ISSUES.
- **Mutation probes**, one mutation per probe, tree restored byte-exact between them: shortening the anchor to `The operator role is invalid` reds 3 cases in `retryable-errors.test.ts` including the new near-miss; deleting the `FALLBACK_NAME_RULES` entry reds 9 in `resource-name.test.ts` including `keeps a user-supplied name untouched`; deleting the `nameProperty === undefined` arm reds exactly 1 in `deploy-engine-cc-update-fallback-name.test.ts`.

## One thing I did not do

The proxy pass wanted the changelog entry extended to name the two hook replacement arms, the no-rule pass-through and the registration guard. I left it alone: those three are test additions with no behavior delta, and CLAUDE.md scopes an entry to what the shipped binary does. The behavior they cover is already named in the entry ("The same holds for a name an SDK provider's `preparePropertiesForFallback` hook adds"). Happy to add it if you read the scope differently.

nix-tkobayashi:
Correcting one mutation claim in my reply above before you act on it, plus three stale sentences the same pass found in the PR body (the body is now updated; nothing in the tree changed).

**The claim that was wrong.** I wrote that deleting `withoutGeneratedFallbackName`'s `nameProperty === undefined` arm "reds exactly that one case". That was scoped to the file I ran it in. Re-measured across both files that watch the helper, the mutant reds **2** cases:

- `tests/unit/provisioning/resource-name.test.ts` -> `returns prepared by identity for a type with no rule, whatever keys it holds`. This one already killed it, before this round: its prepared bag has carried an `undefined` key since the original commit.
- `tests/unit/deployment/deploy-engine-cc-update-fallback-name.test.ts` -> the no-rule UPDATE case. This one did NOT kill it until this round's added key.

So the finding stands and the fix stands -- the engine-level case really was asserted rather than pinned, which is the thing your checklist asks about -- but what the added key buys is the ENGINE's kill, not the first kill. Calling it "exactly one" overstated what was uncovered.

**Three stale sentences in the PR body**, all predating this round, all now fixed there:

1. It said `retryable-errors.ts` "now carries the handler's full sentence". It does not, and your `:166` note is exactly why -- the pattern stops at `sufficient permissions`. The body now says that, and that the `.includes` test is therefore broader than the quoted sentence.
2. "removing the entry reds the two positive cases" -- there are **three** (the IAM-subset exact match, the classification, and the across-every-pattern exact match).
3. "passed every file and test in every run" contradicted the same paragraph's own report of a run that failed two `tests/unit/scripts` critics on 5000 ms timeouts. Now scoped to the runs that completed.

Found by the Codex pass over the body, which I run because a squash merge makes that body `main`'s commit message. The reply above is otherwise unchanged and its other numbers were verified against `3beb6374`.

nix-tkobayashi:
CI is green on `3beb6374` — `check-build-test` 12m31s, `once-leak-detect` 7m32s, both `runtime-compat` legs, `ci-ok`. `mergeable=MERGEABLE state=CLEAN`. Closing the loop on the line in my reply above; nothing went red, so this is the only follow-up it needs.

go-to-k:
Round 2, reviewed at head `3beb6374`. Code, test and security axes again (the delta touches `src/provisioning/providers/**`, which keeps the security lens on).

**M0 through M3 are closed, and I verified each mechanically rather than taking the report:**

- **M0** — both `wc -w` sites carry `| tr -d ' '`. Thank you for chasing the ledger row rather than just the code; "the run was on a host with GNU coreutils, so the committed file was never exercised on a BSD `wc`" is exactly the right diagnosis, and re-running was the right call. A reviewer independently grepped all 31 `| wc` sites tree-wide and every one trims, so your decision to assert no total was also right — the count I gave you was wrong and the two trim spellings are the reason.
- **M1** — `tests/unit/scripts/integ-aws-commands.test.ts` is byte-identical to `origin/main` (`git diff origin/main pr-head -- <that file>` is empty). Dropping the hunk rather than resolving it was the better of the two, and CI green on this head with the new fixture present is the measurement that closes it.
- **M2** — regenerated, no conflict artifacts.
- **M3** — two `assert_gone` calls plus the phase-1 LIVE premise. The premise is what makes it non-vacuous and I would have accepted the pair without it. Your 64-character argument for reading the names out of the state record holds: `CdkdLmiCapacityProviderExample-SpareProviderOperatorRole6345D27C` is exactly at `maxLength: 64`, so one more character goes down truncate-and-hash and a derived literal would quietly stop naming the live role.

All five take-or-leave items landed. Two things I want to name, because they are the reason this round was cheap: you corrected your own mutation claim before I acted on it, and you fixed three stale sentences in the PR body unprompted. A reviewer re-measured the corrected claim independently and "two, one of which already killed the mutant before this round" is right.

Three things to fix, then it merges.

---

## Must fix before merge

**M4 — `src/provisioning/providers/custom-resource-provider.ts:1049` now contradicts the paragraph this PR added at `:1969`.**

The new text says `delivered` may NOT be used to discharge a non-front-door entry, because `onDelivered()` runs after `invokeLambda` resolves. `:1049` still makes exactly that retired inference:

> **Why it can afford that while its sibling cannot**: it fires only when `delivered === false`, so the handler has been invoked ZERO times and a replay re-runs NOTHING

You corrected `.claude/rules/provider-custom-resources.md` and the `isTransientAuthzThrow` JSDoc in the same change, so the rule file and `:1969` now agree with each other and `:1049` disagrees with both. A softer form sits in the table row at `:955`. Re-attribute the guarantee to the classifier's front-door property, with `delivered` as the second fence — the wording you already wrote at `:1969` is the model.

**M5 — make the front-door property STRUCTURAL again, not a prose obligation.**

This is the one I would most like changed, and it is a consequence of the PR being correct about everything else. Before this round, "every pattern in that union is a front-door rejection" was an invariant a reader could rely on. This PR is what made it false — correctly, by adding a Cloud Control resource-handler message to a SHARED list — and then documented the hazard in prose. The failure mode it guards is invoking a non-idempotent customer handler twice.

The residual: a future lane adds a post-acceptance message to `IAM_PROPAGATION_ERROR_MESSAGE_PATTERNS` for its own type, nothing reds, and nothing walks that author past your paragraph. A unit case in `custom-resource-provider-thrown-retry.test.ts` pinning the CR-visible union as an explicit literal set closes it — adding a shared-list entry then reds and forces the author through the reasoning. `tests/unit/deployment/retryable-errors.test.ts:1172-1180` already shows that shape is cheap.

Worth stating so you can push back if you disagree: I checked the claim that no CR `Invoke` failure can carry the operator-role text, and it holds today with high confidence — every throw site reaching that catch with `delivered === false` was enumerated (`prepareInvocation` S3, `publishToSns` SNS, `invokeLambda`), and `waitForBackingLambdaReady` `markNonRetryable`s unconditionally so it is refused on wording-independent grounds. So M5 is about enforcement, not about a live bug.

**M6 — `verify.sh:118` `could not be found` in `no_retry` is unanchored, and it sits on the leak signal.**

All three reviewers reached this independently. `not_found` is checked first but does not match it — `not ?found` needs "not found", and "could not **be** found" has `be` in between — so a genuine not-found phrased that way breaks the loop with `gone` unset and prints the "may still exist and MUST be checked" WARN for a provider that is in fact gone. Loud and fail-safe, so not a leak; but it makes an operator chase a phantom on the one signal the fixture's leak contract rests on. Either move it into the `not_found` alternation or drop it — `Unable to locate credentials` / `NoCredentialProviders` already cover the CLI credential case you added it for.

---

## Take or leave

- `verify.sh:284-285` — the friendly "the logical ids above no longer match the synthesized template" message is unreachable: `read_state` raises `KeyError` under `set -e` when the id drifts, so the run dies on a traceback before the `[ -n "${role_name}" ]` arm. Use `.get(...)` chaining or drop the claim. The same shape is at `:302`, where the strict capture aborts before the `*)` FAIL text.
- `verify.sh:256-263` — the retry-line guard keys on Cloud Control's async `CREATE <id>: FAILED` poll line, which a synchronously-rejected create never logs, so it passes vacuously for that shape. Fine as a heuristic; the header at `:24-30` overstates it as covering "a provider whose create logged FAILED".
- The new near-miss fences the anchor head-only. A tail-anchored shortening (`or doesn't have sufficient permissions`) survives both it and the AZ case. A second near-miss with a different subject and the identical tail would close it. Measured, for the record: every prefix of length 1-45 reds your new case; the first survivor is length 46, a cut inside `sufficient`, which can only under-match.
- Nothing lints the `| wc` trim, unlike the sibling capture-shape rule. Not this PR's job — noting it because M0 is the second time that class has bitten.

---

## On the changelog

Your reading is right and I would have made the same call. CLAUDE.md scopes an entry to what the shipped binary does; three test additions with no behavior delta write none, and the hook arm is already named. Leave it.

---

Push M4, M5 and M6 and I will merge. This has been a genuinely good PR to review — the measurements, the self-correction, and leaving go-to-k/cdkd#3199 out when I asked all made it faster.

nix-tkobayashi:
Pushed as `20897e09`. M4 and M5 are in as asked. **M6 took your second option, not because I preferred it but because the first is not available** — details below, and it is the one thing worth your eye before merging.

## M4 — the contradicting siblings

Both sites corrected, and I went looking for the rest of the family rather than fixing only what you named. Your two were `:1049` (now `:1053`) and the table row at `:955`; review of this round found a THIRD, at `:995`, where the `transientAuthzMaxRetries` doc reads "a pre-delivery throw invokes the handler ZERO times" with the same missing attribution. All five sites — `:955`, `:995`, `:1053`, `isTransientAuthzThrow`'s own doc, and `.claude/rules/provider-custom-resources.md` — now say the same thing: the guarantee comes from the classifier admitting front-door rejections only, and `delivered` is the second fence, for a delivery that COMPLETED.

## M5 — structural, and it found a second member

Done as you specified: a literal-set pin over `IAM_PROPAGATION_ERROR_MESSAGE_PATTERNS` in `custom-resource-provider-thrown-retry.test.ts`, order included. Measured: adding one entry to the shared list reds exactly that case.

**Your framing turned out to be understated.** The companion case holds a ledger of members examined and found NOT to be front-door, and I wrote it asserting "exactly one" — the go-to-k/cdkd#3174 entry. That was wrong, and Codex delta round 2 caught it: `Caught ServiceAccessDeniedException` (issue #805) is the same shape, and its own comment in `retryable-errors.ts` says so outright — *"any CC-provisioned type that validates a same-stack role at create time can hit this"*. So the shared list already carried a non-front-door member before this PR added one. The ledger now carries both, with the provenance and the unreachability argument per entry.

Three things about that case are deliberate, since a ledger is the kind of test that drifts into asserting nothing:

- it pins cardinality (2) **and distinctness**, so the length cannot be met by the same entry twice — the proxy pass's first pass at this survived exactly that mutant;
- its name says what it pins (`holds a distinct, non-empty ledger of examined members, all still in the list`), not what a reader might hope it pins;
- it states in-line what it does **not** establish: that these are the RIGHT two is a judgement about AWS's behaviour no assertion here can make, and swapping a member for another shared-list entry keeps every assertion green. The per-entry notes are the evidence; review is the check.

The admission rule is now one rule rather than two conflicting ones (the proxy pass flagged that the block header and the ledger comment gave different instructions): front-door → add to `PINNED` and stop; not front-door → admissible only if provably unreachable from a CR `Invoke` / `Publish` failure, then `PINNED` **and** the ledger with the argument; if you cannot establish that, split it out of the shared list or gate the CR consumer.

## M6 — your first option is blocked by a tree-wide lint

I tried it first. Moving `could not be found` into the `not_found` alternation **fails `tests/unit/scripts/integ-verify-probe-not-found.test.ts`**, which pins the canonical not-found signature byte-for-byte across every fixture. It flagged the fixture at the alternation — and then, after I reverted that, flagged my *explanatory comment*, because the classifier's rule is "any line containing the literal `not ?found` that is not the canonical string", comments included.

So I took the drop. The residual is worth naming precisely, because it is now a property of the shared signature rather than of this fixture: a provider that really is gone but is reported as "could not **be** found" still ends the loop with `gone` unset and prints the "may still exist and MUST be checked" WARN. Loud and fail-safe, exactly as you said — and repo-wide, not something this fixture may diverge on locally. That is recorded at the site, in a comment that describes the gap rather than quoting the pattern, since quoting it re-trips the lint.

The credential case the entry was added for keeps `Unable to locate credentials` / `NoCredentialProviders`, as you predicted. A missing PROFILE now pays the full budget; I judged that below the bar for widening a signature the whole tree shares. Say the word if you would rather have it back some other way.

## Take-or-leave — all four handled

- `verify.sh:284-285` / `:302` — you are right that both messages were unreachable. The role reads are `.get(...)`-chained so a drifted logical id reaches the named guard instead of a `KeyError` traceback, and the live-ARN read is now an `if !` branch with stderr captured, so its FAIL sentence prints (with AWS's text) while staying fail-closed in every direction.
- `verify.sh:256-263` — the header overstated it and now says what the guard is: it keys on Cloud Control's ASYNC `CREATE <id>: FAILED` poll line, and a synchronously-rejected create logs none, so it passes vacuously for that shape. Called a heuristic in the text, not a total check.
- The tail-anchored survivor — closed. A second near-miss carries the anchor's identical tail under a different subject (`The scaling role ...`), with a premise asserting that subject matches no other retryable pattern, so a green verdict is about the anchor rather than about a sibling entry. Measured: shortening to `or doesn't have sufficient permissions` reds it, and leaves the head-only case green — which is the gap you identified.

  One correction to your note, and it changes nothing about the fix: *"the first survivor is length 46, a cut inside `sufficient`, which can only under-match"* is backwards. A shorter `.includes` needle matches MORE, so a 46-character cut ending `... doesn't have s` would ADMIT an unrelated `... doesn't have sufficient capacity`. Neither near-miss catches that, because both diverge from the anchor before that point; what catches it is the exact-set pair, which compares the entry against `ANCHOR` itself. I have written that division into the comment — the near-misses fence what the classifier DISCRIMINATES, the exact-set cases fence what the entry IS.
- The missing `| wc` lint — filed as go-to-k/cdkd#3213 rather than done here, per your "not this PR's job". It carries the two shapes present in the tree (the piped capture, and the inline here-string comparison at `local-ecs-service-connect/verify.sh:247` / `:251` which never lands in a variable), the two accepted trim spellings, and the two questions I did not want to decide unilaterally: whether an uncompared result should be flagged, and whether an arithmetic use is in scope at all — `[ "${N}" -gt 0 ]` accepts a padded value, so a lint over every `wc` would be wider than the defect.

## On the changelog

Thanks for settling it. Left as it is.

## Verification

- **Integ re-run** (round 3 edited `verify.sh` again): PASS in 510s, 2026-09-16T06:02:17Z, 17 deleted / 0 errors, 2 operator-role propagation retries against 449 captured debug lines, both operator roles live before the destroy and gone after it. Teardown measured AFTER the version sweep rather than read off the run log: `cdkd/CdkdLmiCapacityProviderExample/` lists 0 objects, `list-object-versions` over that prefix returns 0 across `Versions` + `DeleteMarkers`, and `lambda list-capacity-providers` returns 0 in the region. Ledger row replaced.
- **Unit suite**: 1025 files, 23492 passed, 1 skipped, no type errors — three more cases than round 2 (the M5 pair and the tail near-miss). `vp run check`, `vp run typecheck:test`, `vp run build` clean.
- **Codex delta review**: 3 rounds — r1 the "can only under-match" claim I had copied from your note without checking it, plus a ledger guard an empty array satisfied, r2 the `Caught ServiceAccessDeniedException` counterexample, r3 NO ISSUES. **Maintainer-checklist proxy pass**: 2 rounds — r1 three findings (ledger membership asserted rather than pinned, a `verify.sh` comment claiming a fix the code did not provide, and the two conflicting admission instructions), r2 NO ISSUES.
- **Mutation probes** this round, one at a time, tree restored byte-exact between them: one entry added to `IAM_PROPAGATION_ERROR_MESSAGE_PATTERNS` reds exactly the new literal-set case; the tail-anchored shortening reds the new tail near-miss plus the two exact-set cases and leaves the head case green.

go-to-k:
Merged as `f7b5882`. Thank you — this was an unusually good PR to review.

Three things stood out, and they are the reason it took three rounds rather than six:

- **You corrected your own claims before I acted on them, twice.** The mutation count that was scoped to one file, and the three stale sentences in the PR body. A reviewer cannot catch what a report quietly overstates; volunteering it is what made the numbers in this PR worth reading.
- **You checked my reasoning instead of just applying it.** Two of my notes were wrong and you caught both: the `.includes` polarity (a shorter needle matches MORE — my "can only under-match" was backwards), and M5's framing, where I said this PR made the front-door invariant false. It was already false via `Caught ServiceAccessDeniedException` (issue #805), whose own comment says so outright. The ledger is better for carrying both entries than it would have been carrying the one I asked for.
- **You pushed back on M6 with a measurement rather than an opinion.** I did not know the canonical not-found signature is compared byte-for-byte tree-wide, comments included; I verified it after your reply and your first option is genuinely unavailable. Framing the residual as a property of the shared signature rather than of this fixture is also the more accurate read.

The one habit worth keeping: every number in this PR was measured and attributed, and where you could not measure something you said so — the CONSTRUCTED near-miss is labelled as not-AWS-observed, and the ledger states in-line that "these are the right two" is a judgement no assertion can make. That is the difference between a test suite and a green light.

go-to-k/cdkd#3199 is fixed separately in go-to-k/cdkd#3212, and go-to-k/cdkd#3213 (the `wc` trim lint you filed) is a good catch — the two shapes and the two open questions in its body are exactly the right framing for it.

PR #3210, opened 2026-09-16
fix(state): cap how deep cdkd state list --tree nests, so one long parent chain cannot empty the whole view

### LINKED ISSUE #3155: fix(state): cdkd state list --tree prints nothing for a parent chain thousands of records deep

nix-tkobayashi:
Closes #3155

## What was wrong

Four steps in the `cdkd state list --tree` path recurse once per level of the tree, and none of them sits inside `renderTreeMode`'s per-stack guard, which covers only the state READ:

- `buildStackTree`'s child sort (`src/cli/commands/state-list-tree.ts`);
- `renderChildren`, behind `renderStackTreeAscii` (the text view);
- `stackTreeToJson` (the `--json` view);
- `JSON.stringify(stackTreeToJson(roots), null, 2)` in `renderTreeMode` (`src/cli/commands/state.ts`), which recurses into whatever nesting it is handed.

So a long enough chain of `parentStack` links ended the command with a `RangeError` and printed no rows at all — for every stack in the bucket, not just the chain's own.

## What changed

`buildStackTree` now caps a node's depth at `MAX_STACK_TREE_DEPTH` (100) levels below its root. A node that would sit deeper is placed at the root instead — the treatment an orphan or a parent-loop member already gets — and the chain resumes there as its own tree.

Capping the TREE rather than rewriting the walks is what the issue asked for, and it is the only thing that works: two of the four steps are bounded by the output, not by how it is built. `JSON.stringify` recurses into whatever nesting it is handed, so an iterative `stackTreeToJson` still hands it a structure too deep to serialize; and the text view indents four characters per level, so a deep enough chain needs more indentation than a string can hold.

The depth is computed in ONE non-recursive pass over the parent links, mirroring the loop detection directly above it: walk up to the first node whose depth is known, then assign downward. So the build itself costs no stack frames either, and the answer depends only on the chain above a node — never on the order the records arrive in.

That walk needs the links it follows to be acyclic, and every parent-loop member is seeded at depth 0 to give it that: 0 is the depth a root has, and a cycle consists of loop members only, so a walk up from anywhere meets a seeded node and stops. **It is stated there rather than derived inside the walk deliberately.** The first cut excluded loop members inside the parent lookup instead — which works, but leaves termination depending on a fact settled thirty lines above, and a walk that reaches a loop it does not recognize does not FAIL: it never returns. Not hypothetical — the mutation probe deleting that exclusion came back as seven test TIMEOUTS rather than assertion failures. To be exact about what the change buys: it does NOT remove that mode (deleting the seeding line hangs the same way, and the same seven cases catch it). It states the reason at the point of use and drops a helper.

## Scope of the behavior change

- Every record still appears exactly once.
- A tree within the cap is built and rendered byte-identically, so no ordinary listing changes. CloudFormation stops at five levels of nested stacks, so an app that also has to deploy through CloudFormation cannot exceed that — but **cdkd deploys nested stacks through its own engine and imposes no depth limit**, so a deeper tree is reachable by deploy and not only by a hand-edited bucket. It now stays readable instead of failing. (The first draft of this PR, and of the doc and changelog text, claimed no deploy could reach the cap; that was wrong and is corrected.)
- A re-rooted record keeps its own `parentStack` / `parentLogicalId` / `parentRegion`, so `--tree --json` still reports the parent it named.

## Files

- `src/cli/commands/state-list-tree.ts` — the cap, the non-recursive depth pass, and the `buildStackTree` doc comment
- `src/cli/commands/state.ts` — a note at the `JSON.stringify` site saying what bounds it
- `docs/cli-state.md` — the cap in the `--tree` section and a row in the malformed-value table
- `.claude/rules/layout-cli.md` — `buildStackTree`'s promotion list was an enumeration that had gone stale; the same hunk also corrects a pre-existing sentence, since `renderTreeMode` reads one record per reference THAT CARRIES A REGION (a legacy region-less one is skipped)
- `tests/unit/cli/state-list-tree.test.ts`, `tests/unit/cli/state-list.test.ts`, and a new `tests/unit/cli/state-list-tree-cap-doc-sync.test.ts` (the cap fence)
- `changelog.d/entries/2026-09-16-3155-state-list-tree-depth-cap.md`

## Verification

Run against the PRE-FIX module (`git show origin/main:src/cli/commands/state-list-tree.ts`, with only the new constant appended so the file compiles), `tests/unit/cli/state-list-tree.test.ts` reports `5 failed | 22 passed (27)`. Two of those five — the 30000-record chain and the loop-chain — throw `RangeError: Maximum call stack size exceeded` inside `sortRecursive`, so those inputs really do exceed the runner's recursion limit rather than merely exercising the new arithmetic; the past-cap, resume and branching cases fail on their assertions. The at-cap and order-independence cases pass pre-fix, by design (the describe block says why). Post-fix: `27 passed`.

Cases added:

- a 30000-record chain — `buildStackTree`, `renderStackTreeAscii` and `JSON.stringify(stackTreeToJson(...))` all return, every record appears exactly once, and no node sits deeper than the cap;
- a chain exactly AT the cap (one tree) and one just PAST it, pinning where the boundary sits. The past-cap case asserts the re-rooted node through `stackTreeToJson` with all three parent fields, AND its rendered TEXT row at column 0 with no `└── ` connector — the default view is what `docs/cli-state.md`'s "shown at the root instead" is about;
- a chain that RESUMES under the re-rooted node. Re-rooting EVERY node past the cap is already rejected by the 30000-record case's root count, but only arithmetically (29900 roots instead of 298); this says it as a shape, on an input small enough to read;
- **a deep chain hanging off a parent LOOP**, the shape recorded on go-to-k/cdkd#3155 by the maintainer: go-to-k/cdkd#3157 correctly roots every loop member, which makes such a chain their subtree and so reachable by every recursive step;
- a BRANCHING node exactly at the cap with two over-cap children — every other case is a straight chain, so nothing else says whether the depth is per-node or per-chain; each child re-roots independently and keeps its own subtree;
- an order-independence case, since the depth pass memoizes — forward, fully reversed, and INTERLEAVED (a full reversal is degenerate: every start after the first short-circuits, so only an interleaved order makes a walk stop at a memoized node at non-zero depth straddling a cap boundary);
- a command-level case in `tests/unit/cli/state-list.test.ts` driving `state list --tree` and `--tree --json` over a 5000-record mocked chain, which is the only way to cover the `JSON.stringify` step — it lives in the command, not in the tree helper.

The existing tree cases (ordering, box-drawing prefixes, orphans, self-link, 2- / 5-record and disjoint loops, the JSON shape) stay green, which is what pins that a tree within the cap renders as before.

Mutation probes, one mutation per probe, the file restored byte-exact between them and the restore verified by `git diff HEAD` printing nothing. **The table is produced by a script re-run at the current head, not re-typed** — the first version of it was measured before the branching case and the loop-seeding refactor existed, and five of its eight rows plus the baseline were wrong by the time they were published.

| # | Mutation | Cases red |
| --- | --- | --- |
| M1 | cap boundary `depth + 1 > MAX` -> `>= MAX` | 7 |
| M2 | drop the cap entirely (`depth = depth + 1`) | 6 |
| M3 | attach regardless of depth (`if (parent)`) | 12 |
| M4 | cap value 100 -> 1000 | 3 |
| P1 | delete the `depthOf.has(start)` short-circuit | **0 — not pinned** |
| P2 | depth seed `-1` -> `0` | 4 |
| P3 | reverse the downward assignment loop | 5 |
| P4 | drop `parentLogicalId` / `parentRegion` from a re-rooted node | 1 |

Baseline for those three files at this head: **105 passed**.

Two rows worth reading rather than counting. **P1 is deliberately `0`**: that clause is a short-circuit, not a guard — a start that already has a depth walks nowhere and assigns nothing — and the line says so, because a clause no test can pin otherwise reads like one nobody checked. **M4 reds three** rather than two because the new `docs`-vs-constant fence is one of them: drift the constant and the user-facing page's two bare `100`s are caught.

Two earlier rows are gone rather than restated. They probed the loop-member exclusion in its pre-refactor position (inside the parent lookup) and its post-refactor one (the depth seeding); the first anchor no longer exists at this head, and the second needs two edit sites to express, which breaks one-mutation-per-probe. The seeding is covered by the six pre-existing loop cases, which go red — as timeouts, since the mutant does not terminate.

Sibling sweep, scoped to the walkers that build a recursive structure from STATE: `buildStackTree` is the only consumer of the arbitrary `parentStack` link, and two other walkers recurse on state-derived nested keys. `buildCdkdStateStackTree` (behind `state show --show-nested` and `cdkd export`) and `buildDiffTree`'s state-only arm — whose recursion is `buildDeletedSubtree`, `src/cli/commands/diff-recursive.ts:1311-1373` — both derive each child as `<parent>~<childLogicalId>`, so S3's 1024-byte key limit bounds them at a few hundred levels, and `loadStateOrEmpty` stops the walk earlier still by returning an empty record for a missing child. No cap is owed on either.

`buildDiffTree`'s TEMPLATE arm follows the synth output through `readNestedTemplate` / `indexNestedChildTemplates`, and it has no visited set and no self-reference check, so a hand-modified relative `aws:asset:path` can close a cycle and the walk follows it. It is bounded all the same, and by the SAME key limit as the two above: `buildDiffTree` reads state at every node (`src/cli/commands/diff-recursive.ts:1186`) against a `childStackName` that grows by one `~<childLogicalId>` per level (`:1236`). Measured against the built binary, a self-referencing assembly ends after about 190 levels with `Your key is too long`, naming a stack that does not exist and mentioning neither the cycle nor the asset path. So what that arm lacks is a diagnosis, not a bound. Filed separately as go-to-k/cdkd#3239; out of scope here.

This paragraph audits only the walkers above. cdkd's other recursive nested-stack paths — the child `DeployEngine` per level in `NestedStackProvider`, the per-level recursion in `runDestroyForStack`, `importNestedStackChildrenRecursive`, and `walkCfnStackTree` — are NOT audited here, and the first two are not bounded: that is the same "cdkd imposes no depth limit of its own" this PR states elsewhere. Whether an `await`-driven recursion accumulates synchronous frames at all is not measured here either, and it is the fact that decides most of them at once; an audit of any of those arms must settle it first rather than read a bound off the key-limit argument above.

## Live exercise

Against the real state bucket (`cdkd-state-596912130800`, `us-east-1`), with the BUILT binary, not the source:

- seeded 205 state records forming one parent chain (`Cdkd3155LiveChain0..204`);
- `cdkd state list --tree --json` and `cdkd state list --tree` both returned 0;
- all 205 records present exactly once, deepest node at 100, re-rooted at records 0 / 101 / 202, the re-rooted node keeping `parentStack` / `parentRegion` / `parentLogicalId` and its one child, 205 text rows, deepest text indent exactly 400 characters;
- teardown purged all 205 object versions and delete markers under `cdkd/Cdkd3155LiveChain`, and a re-listing returned 0 — the bucket is versioned, so `aws s3 rm` alone would have left the rows readable.

The live run exercises the CAP, not the crash: reproducing the `RangeError` live needs thousands of records, and seeding thousands of objects into the shared bucket is not worth it when the unit measurement above already pins it.

## Review

Four read-only reviewers (spec, code, test, security) over **two rounds**, plus three Codex rounds and a maintainer-checklist proxy pass. **No blocker in either round.** What the passes changed, since several were errors rather than polish — and the second round's findings were almost all in prose I had published, which this repo squash-merges into the commit message:

**Round 1 (against the pre-review head):**

- **A false claim.** The first draft said no deploy could reach the cap because CloudFormation stops at five nested levels. Wrong: cdkd deploys nested stacks through its own engine and imposes no limit, so the five-level bound binds only an app that also has to deploy through CloudFormation. Corrected in the constant's doc, `buildStackTree`'s doc, `docs/cli-state.md` and the changelog entry.
- **"the build costs no stack frames"** was true of the depth pass and false of the child sort. Scoped.
- **Termination is a proof obligation discharged 40 lines away**, and a walk meeting an unrecognised cycle does not fail, it hangs. The comment now DERIVES the property (parent links are a functional graph, so cycles are disjoint; `settled` is written only after a walk completes, so the first walk to touch a cycle runs it whole and seeds every member) instead of asserting it.
- Test gaps closed: the branching case, the interleaved order, the rendered text row, and the `docs`-vs-constant fence — `docs/cli-state.md` spells `100` twice and nothing stopped it drifting. That fence was probed in both directions: changing the constant reds it, and deleting one of the two sentences reds it too.

**Round 2 (against the head those fixes produced):**

- **The mutation table was stale** — measured before the branching case and the loop-seeding refactor, never re-run. Five of eight rows and the baseline were wrong. It is now emitted by a script run at the current head.
- **A measured number that did not reproduce.** The constant's doc claimed the pre-cap `--tree --json` ceiling was "4k records"; two reviewers independently measured ~2k. The number is gone, and so is the framing that produced it: pre-fix and post-fix are DIFFERENT failures — a call-stack limit that moves with V8's budget, versus a deterministic maximum-string-length refusal — and the doc now says that instead of a count nothing re-checks.
- **A comment that was simply false.** The interleaved-order arm claimed it makes a walk stop at a memoized node of non-zero depth. Instrumented: the FORWARD order does that ~300 times; this one does it zero times. Reworded to what it measures.
- **A real flake risk.** The 30000-record case ran 0.8-1.2 s alone but 4.5 s beside nineteen peer vitest processes, against vitest's 5 s default. It and the 12000-record loop case now declare `30_000`.
- **A dead filter.** The doc fence excluded the literal `5` to avoid CloudFormation's limit — but the page spells that in words, so the filter never fired, and it would have hidden a legitimate future cap of 5. Removed, with the reasoning at the line.
- The fence's JSDoc claimed the same shape as `custom-resource-response-prefix-sync.test.ts`; that one DERIVES its population from git precisely so a copy elsewhere cannot escape, and this one names its page. Stated as the known limitation it is.

Two findings recorded as **won't-do**, both re-confirmed in round 2:

- *A defensive `if (path.length > byKey.size) break;` in the depth walk.* Declined: the invariant is proven independently by two reviewers, and the clause could be pinned by no test — the same objection that made the `depthOf` short-circuit get a comment marking it inert. The mechanism settles it: a `break` there leaves `current` unseeded, so `depthOf.get(current)!` is `undefined`, `undefined + 1 > 100` is false, and every node on the path silently becomes a root with `NaN` depth — a wrong tree, quietly, in a view whose whole contract is that every record appears exactly once. A `throw` instead would empty the view, which is the bug being fixed.
- *The text view gives no signal that the cap re-rooted a chain.* Declined: `--tree --json` keeps `parentStack` so a tooling consumer can tell, and a parent-loop member is already rooted silently today. Marking one and not the other would be the inconsistency.


go-to-k:
Reviewed at head 5e67a9a. Three read-only reviewers (spec, code, test) in parallel; **no blocker and no major on any axis**, and `Closes #3155` is earned — every acceptance item in the issue's "Proposed fix" and "Verification plan" carries a file:line, and the pre-fix `RangeError` plus the in-cap byte-identical claim were both re-derived independently rather than read off the body.

The depth pass's termination argument was re-derived from the code rather than accepted from the comment, and then confirmed empirically: 400 randomized parent graphs (self-links, back-edges, orphans, region-less parents, disjoint cycles, branches) x 6 orderings each = 2400 builds, with zero hangs, zero throws, zero duplicates, zero drops, `maxDepth <= 100` throughout, and a byte-identical `stackTreeToJson` for every permutation.

**Three items, all required before merge.** None is a defect in shipped behavior; all three sit in files this PR adds, and each is a few minutes' work. n1 is the one that would otherwise cost a later round on its own, and n2 / n3 are bundled into it rather than left to drift, so this is one round rather than three.

**M0 — `tests/unit/cli/state-list-tree-cap-doc-sync.test.ts:40`: the assertion is stricter than its own doc comment.** The comment says the regex is "deliberately not anchored to the two sentences that carry it today, so a THIRD one added later with a stale number is caught too" — but `toEqual([MAX_STACK_TREE_DEPTH, MAX_STACK_TREE_DEPTH])` also reds on a third sentence carrying the CORRECT value, and on any unrelated future numeral-plus-"levels" on the page. As written, the fence produces a false positive on a doc edit that is right. Make the assertion say what the comment says:

```ts
expect(stated.length).toBeGreaterThanOrEqual(2);
expect(stated.every((v) => v === MAX_STACK_TREE_DEPTH)).toBe(true);
```

That still reds when one of the two sentences is deleted (the count floor), which is the case the current `toEqual` was chosen for.

**M1 — `tests/unit/cli/state-list.test.ts:1166`: declare the `30_000` timeout.** It is the only one of the three large-input cases without one, while `state-list-tree.test.ts:365` and `:514` both carry it and explain why. The 5000-record case measures ~14 ms, so the flake risk is genuinely low — the reason to add it is that a reader cannot tell the omission from an oversight, and the two siblings establish the convention.

**M2 — `changelog.d/entries/2026-09-16-3155-state-list-tree-depth-cap.md`: trim it.** 1983 of the 2000-character budget, so the next edit to that entry reds `tests/unit/scripts/changelog-entry-size.test.ts`. Around 150 characters is enough headroom; the paragraph explaining why an iterative rewrite of any single step could not work is the part that is already stated in full in `MAX_STACK_TREE_DEPTH`'s doc comment.

Nothing else is owed. I am holding the merge until these land; push them and I will re-review from the new head.

nix-tkobayashi:
All three landed. Pushed as `9b41c0fa`.

**M0 — the fence's assertion now says what its comment says.** `toEqual([MAX, MAX])` is replaced by a floor plus a per-element check:

```ts
expect(stated.length).toBeGreaterThanOrEqual(2);
expect(stated.every((v) => v === MAX_STACK_TREE_DEPTH)).toBe(true);
```

You were right that the old form reds on a correct doc edit, so I probed the replacement in four directions rather than taking the shape on trust — each mutation applied alone, `docs/cli-state.md` and the source restored byte-exact between them:

| Probe | Result |
| --- | --- |
| constant `100` -> `101` | RED |
| delete one of the two cap sentences | RED (the floor) |
| **add a THIRD sentence stating the cap correctly** | **PASS** — the false positive you named; the old `toEqual` redded here |
| rewrite CloudFormation's limit as `5 levels` instead of "five" | RED |

The last one is deliberate rather than incidental, and the comment at the line says so: the page spells CloudFormation's limit in words today, so it does not match, and if someone rewrites it as a numeral this fence firing is the outcome to want over a `!== 5` filter that would also hide a genuine cap of 5.

**M1 — declared.** `tests/unit/cli/state-list.test.ts` now carries `}, 30_000)`, with a comment saying the reason is the convention rather than the risk, since ~14 ms is nowhere near the default.

**M2 — trimmed to 1872 of 2000, 128 characters of headroom.** I dropped the paragraph you named; the entry now points at `MAX_STACK_TREE_DEPTH`'s doc comment for why an iterative rewrite of any single step could not work. I re-read the whole entry after the cut rather than only the edited sentence — a cap-driven trim is a new sentence, and the surrounding clauses had to still parse without the removed one. While there, the entry's file list was missing `tests/unit/cli/state-list-tree-cap-doc-sync.test.ts`; it is there now, inside the headroom.

Nothing else changed: no source behaviour, no other test. Full suite and the changelog fences are green, and the `verify-pr` / `pr-review` markers are re-bound to the new head.

One note on your termination result, since it is stronger evidence than what the PR body claims: 2400 builds across 400 graphs x 6 orderings covers the permutation axis my own order-independence case only samples at three points. If you would take it, I am happy to add a seeded randomized case in that shape as a follow-up rather than in this PR — it is a coverage addition, not a fix, and this PR is otherwise ready.

go-to-k:
Round 2, reviewed at head 9b41c0fa. **M0 / M1 / M2 all land.** Three reviewers again (spec, code, test), and again **no blocker and no major on any axis**; `Closes #3155` re-derived at the new head and still earned.

What I verified rather than took from your reply:

- **The shipped bytes did not move.** `git rev-parse 5e67a9a:src` and `9b41c0fa:src` are the same tree hash, and so are `docs` and `.claude` — a recursive content hash, so round 1's verdict on the depth algorithm carries without re-deriving it. The branch was never rebased either: both heads share the parent `b07b8b6d2`, and the `baseRefOid` move you may see in the API is just main's tip advancing. Nothing in the 8 commits main gained since touches `state-list-tree.ts`, `state.ts`, or adds a caller of anything this PR changes.
- **Your M0 probe table reproduces, all four rows**, re-measured independently against scratch copies rather than read off the table. Two rows you did not claim also hold: a third sentence with a STALE number reds, and removing every `<n> levels` match reds on the floor — so the floor is genuinely what stops `stated.every(...)` passing vacuously on an empty array.
- **M2 verified by execution**, not by byte count alone: `changelog-entry-size`, `-uniqueness` and `-policy-sync` run green against the entry materialized at 9b41c0fa, and the delegation target really does carry the dropped paragraph (`state-list-tree.ts:41-48`).

**Six items, all required before merge.** None is a behavior defect; all are one-line-to-two-line edits, and one is a PR-body edit that does not move HEAD. Bundled into one round for the same reason as last time.

**M3 — `tests/unit/cli/state-list-tree-cap-doc-sync.test.ts:44`: the shape I asked for in M0 was wrong, and this is my error, not yours.** `expect(stated.every((v) => v === MAX_STACK_TREE_DEPTH)).toBe(true)` fails as `expected false to be true` — it prints neither the offending number nor what the page says. Your own sibling case states the opposite convention 1200 lines away, at `tests/unit/cli/state-list.test.ts:1214-1216`: *"not a `startsWith` predicate collapsed to a boolean — a failure should print what WAS rendered."* The flexibility M0 was after does not cost the diagnostic:

```ts
expect(stated.length).toBeGreaterThanOrEqual(2);
expect(stated.filter((v) => v !== MAX_STACK_TREE_DEPTH)).toEqual([]);
```

**M4 — close the arrival-order gap on the fixtures you already have, and no, not as the randomized suite.** Taking your offer first, decisively: **declined, and not as a follow-up issue either.** A seeded 400x6 suite buys a real cell but pays for it with a failure mode whose repro is a seed, and the cell is narrower than the offer. What your cases actually sample is one topology across three orders — `state-list-tree.test.ts:516` is a linear chain — while the branching case (`:423`) and the loop-hang case (`:465`) each call `buildStackTree` exactly once, in one order. So the unsampled cell is *branching / loop topologies x arrival order*, not "more chains". Close it deterministically on the fixtures those two cases already build: assert that `JSON.stringify(stackTreeToJson(buildStackTree([...entries].reverse())))` equals the forward result, in each of those two cases. Two lines, no randomness, and it samples the actual gap. With that in, the randomized suite buys nothing worth a seed.

**M5 — `tests/unit/cli/state-list.test.ts:1218`: delete the `~14 ms`, do not re-measure it.** It does not reproduce at this head: three runs measured 38 ms, 66 ms, 65 ms. The sentence's conclusion — that the convention, not the flake risk, is the reason for the bound — survives at any of those, so the number is load-bearing for nothing. A hardware-specific millisecond figure in a committed comment drifts by construction, and this PR already retired one measured number that did not reproduce.

**M6 — `tests/unit/cli/state-list-tree-cap-doc-sync.test.ts:12-15`: the docstring still describes the old assertion.** It says *"It asserts a COUNT as well as the value. Without that, deleting one of the two sentences leaves the other matching…"*, which was true of `toEqual`. It is now a FLOOR, and the guarantee it states holds only while the page states the cap exactly twice — add a third correct sentence and a later deletion passes. Say that it is a floor sized to today's corpus, which is the same caveat the `:17-22` paragraph already makes about the named population.

**M7 — the changelog entry's file list is missing `.claude/rules/layout-cli.md`.** The entry lists six paths; the PR changes seven besides the fragment itself. `.claude/rules/**` appears in the file list of seven existing entries, so this is a gap against the header's "changed files", not a convention against listing it. You have 128 characters of headroom, which is enough.

**M8 — correct the PR body's sibling sweep; the reason is wrong for one of `buildDiffTree`'s two arms.** The body says the other recursive nested-stack walkers are safe because they derive each child as `<parent>~<childLogicalId>`, so their depth is bounded by S3's 1024-byte key limit. That is right for `buildCdkdStateStackTree` (`src/cli/commands/export.ts:3407-3433`) and for `buildDiffTree`'s state-only arm (`src/cli/commands/diff-recursive.ts:1252-1268`). It is **not** right for its template arm (`:1194-1248`), which recurses on the synth output via `readNestedTemplate` / `indexNestedChildTemplates` and reads no child state — nothing there terminates at the key limit, and its depth is bounded by the user's own nested-template nesting instead. The conclusion is unchanged, so this is a body edit only. It matters because this repo squash-merges the body into the permanent commit message, and this paragraph is the PR's only recorded audit of those walkers — a later lane reading the current wording would skip re-auditing that arm. A body edit does not move HEAD or invalidate a marker.

M3, M5 and M6 all sit inside the M0 / M1 fix, which is the second consecutive round finding items inside the previous round's own fix. That is on me for prescribing M0's assertion shape without its diagnostic, so every item above carries the exact final form rather than a direction — there should be nothing left to interpret. Push them and I will re-review from the new head.

nix-tkobayashi:
All six landed, in the final forms you gave. Pushed as `ad0b5cdb`; this reply is later than it should have been.

**M3 — the fence prints the offending values.** `every(...).toBe(true)` is now `expect(stated.filter((v) => v !== MAX_STACK_TREE_DEPTH)).toEqual([])`, with the floor unchanged. Checked the diagnostic rather than assuming it: rewriting one cap sentence to `150` now fails as `expected [ 150 ] to deeply equal []`. That was the convention the sibling case already stated, and I missed it when I took M0's shape.

**M4 — declined suite accepted as declined, and the gap closed on the existing fixtures.** Both cases now assert that `JSON.stringify(stackTreeToJson(buildStackTree([...entries].reverse())))` equals the forward result: the branching case (a start at a leaf walks up through the capped parent before its sibling has a depth) and the loop case (the walks reach the 2-cycle from the far end of the hanging chain, so the seeded members are met from below first). No randomness, no follow-up issue.

**M5 — the `~14 ms` is gone, not re-measured.** The sentence keeps its conclusion: the convention set by the two siblings is the reason for the bound, not the flake risk.

**M6 — the docstring describes the floor.** It now says the count check is a FLOOR sized to today's corpus, that deleting either of the two sentences reds, and that the guarantee holds only while the page states the cap exactly twice — a third correct sentence followed by a later deletion passes. Framed as the same kind of caveat the next paragraph makes about the named population.

**M7 — `.claude/rules/layout-cli.md` added to the entry's file list.** 1903 of 2000; the three changelog fences are green.

**M8 — the sibling sweep is corrected per arm in the PR body.** It now records three walkers and three reasons: `buildCdkdStateStackTree` and `buildDiffTree`'s state-only arm are bounded by the 1024-byte key limit on `<parent>~<childLogicalId>`, and `buildDiffTree`'s template arm is NOT — it recurses on the synth output through `readNestedTemplate` / `indexNestedChildTemplates`, so its depth is bounded by the user's own nested-template nesting. I read `diff-recursive.ts:1194-1248` before writing that rather than taking the line numbers on report, and I added a sentence telling a later audit to re-derive the template arm rather than read it off the key-limit argument. Body-only; HEAD did not move for it.

One thing I cut while writing M8 and want on the record: a draft sentence claimed the template arm "recurses through `await`, so it does not accumulate synchronous stack frames." I had not verified it — a call whose synchronous prefix reaches the recursive call before its first real yield would stack up anyway — so it is not in the body. If the question matters for that walker, it is its own audit.

No source behaviour changed: `src`, `docs` and `.claude` are untouched by this round; the edits are the fence, two test files, the changelog fragment and the PR body. Full suite green; `verify-pr` / `pr-review` re-bound to the new head.

go-to-k:
Round 3, reviewed at head ad0b5cdb. **M3 through M8 all land**, and the code and test axes are clean: your reverse-order assertions were instrumented rather than taken on trust, and they sample cells nothing else reaches — the branching case reversed stops at a node memoized at depth 100 and re-roots across a cap boundary mid-walk, and the loop case reversed drives `path.slice(onPath.get(current)!)` at index 12000 instead of 0, which is the "cycle reached from outside behind a long non-cycle prefix" branch your production doc comment argues and nothing pinned before. The fence's red set is identical to the round-2 form and now prints the offending value. 105 passed, 0 failed. The 97 bytes of changelog headroom are fine; I am not asking for more.

**One item is a real problem, and I caused it.** M8 asked you to correct the sibling sweep per arm. What that request did was promote a scoped parenthetical — the original body named two walkers and said their depth was key-limit bounded — into a **universal**: `body:74` now reads *"The other recursive nested-stack walkers are bounded, but not all for the same reason, and the reason is recorded per arm because this paragraph is the only audit of them."* That sentence is false, and this body is squash-merged into the permanent commit message.

At least four more recursive nested-stack walkers exist, and the two most exposed are ones this PR elsewhere says are **unbounded**:

- `NestedStackProvider` spawns a child `DeployEngine` per nesting level (`src/provisioning/providers/nested-stack-provider.ts:53-60`, `:81`). This is the writer of the very depth the cap exists for — `state-list-tree.ts:37-38` and `docs/cli-state.md:162-163` both say cdkd's engine "imposes no limit of its own."
- `runDestroyForStack` recurses per nested level, and the repo already records that it is not bounded: `src/cli/commands/destroy-runner.ts:974` says *"Deep nesting + high `--concurrency` can legitimately exceed Node's default"*.
- `importNestedStackChildrenRecursive` (`src/cli/commands/import.ts:2943`, self-call `:3147`) — a second producer of the arbitrary `parentStack` link.
- `walkCfnStackTree` / `injectRetainPoliciesRecursive` (`src/cli/commands/retire-cfn-stack.ts:912`, `:930`) — bounded, but by CloudFormation's own nesting, a third reason the paragraph never records.

A later lane reading that paragraph would be **misinformed, not merely under-informed**: it would conclude the deploy and destroy recursions had been audited and found bounded.

**The fix is to shrink the claim, not to audit four more walkers.** Auditing the deploy / destroy / import / retire recursions does not belong in a `state list --tree` PR, and I am not asking for it. Three items, all required, all prose, none touching `src`:

**M9 — replace the sibling-sweep paragraph (`body:74-80`) with the scoped version.** Exact text:

> Sibling sweep, scoped to the walkers that build a recursive structure from STATE: `buildStackTree` is the only consumer of the arbitrary `parentStack` link, and two other walkers recurse on state-derived nested keys. `buildCdkdStateStackTree` (behind `state show --show-nested` and `cdkd export`) and `buildDiffTree`'s state-only arm — whose recursion is `buildDeletedSubtree`, `src/cli/commands/diff-recursive.ts:1280-1342` — both derive each child as `<parent>~<childLogicalId>`, so S3's 1024-byte key limit bounds them at a few hundred levels, and `loadStateOrEmpty` stops the walk earlier still by returning an empty record for a missing child. No cap is owed on either.
>
> `buildDiffTree`'s TEMPLATE arm is a different case and is NOT covered by that argument: it recurses on the synth output through `readNestedTemplate` / `indexNestedChildTemplates`, reads no child state, and has no visited set or depth guard, so a hand-modified relative `aws:asset:path` can close a cycle. Filed separately as go-to-k/cdkd#3239; out of scope here.
>
> This paragraph audits only the walkers above. cdkd's other recursive nested-stack paths — the child `DeployEngine` per level in `NestedStackProvider`, the per-level recursion in `runDestroyForStack`, `importNestedStackChildrenRecursive`, and `walkCfnStackTree` — are NOT audited here, and the first two are not bounded: that is the same "cdkd imposes no depth limit of its own" this PR states elsewhere. Whether an `await`-driven recursion accumulates synchronous frames at all is not measured here either, and it is the fact that decides most of them at once; an audit of any of those arms must settle it first rather than read a bound off the key-limit argument above.

That last sentence is the one you cut as unverified. Cutting an unverified claim was right; what was not right — my fault for asking for a paragraph that invited it — is that the conclusion stayed while its only premise left. Recording it as an open question is true, is cheap, and is what makes the instruction to a later lane actionable.

**M10 — `docs/cli-state.md`: say that a re-rooted record keeps its parent link.** Your own won't-do rests on it — *"`--tree --json` keeps `parentStack` so a tooling consumer can tell"* — but no user-facing text says so. The cap paragraph at `:158-164` and the table row at `:199` say only "shown at the root" and "no record is dropped"; a consumer told nothing cannot tell. One sentence in the cap paragraph is enough, and it keeps the changelog budget out of it entirely. Assert it in the docs and the won't-do becomes honest.

**M11 — `tests/unit/cli/state-list-tree.test.ts:522`: "members" should be singular.** The depth walk halts at the first node carrying a depth, so only `LoopA` is met; `LoopB` is never walked to in that pass. "the seeded loop is met from below" is what you measured.

Everything else is closed. The `diff --recursive` template-arm cycle is go-to-k/cdkd#3239 and is not yours to fix here. **This is the last round I intend to open on this PR**: M9 through M11 are prose with exact text supplied, so push them and I will confirm rather than re-review.

nix-tkobayashi:
M9 through M11 landed, using the text you supplied. Pushed as `e2cb243a`.

**M9 — sibling sweep replaced with the scoped version, verbatim, in the PR body.** Before pasting it I checked each code reference it makes, because this paragraph becomes the permanent commit message: `buildDeletedSubtree` is defined at `diff-recursive.ts:1280` and calls itself at `:1332`; `loadStateOrEmpty` is at `:258`; `NestedStackProvider` builds a child `DeployEngine` per level; `destroy-runner.ts:974` carries the "Deep nesting + high `--concurrency` can legitimately exceed Node's default" note; `importNestedStackChildrenRecursive` is at `import.ts:2943` and calls itself at `:3147`; `walkCfnStackTree` is at `retire-cfn-stack.ts:912` and calls itself at `:930`; go-to-k/cdkd#3239 is open under the title the paragraph gives it. The claim that the other walkers are all bounded is gone (the read-back shows no match for it), and the question about `await`-driven recursion is back in the body, stated as unmeasured.

You're right about what went wrong with the cut. Dropping the unverified sentence was fine. Leaving the conclusion standing after its only premise was gone was the mistake. Writing it up as an open question is the version a later lane can act on.

**M10 — `docs/cli-state.md` now says a re-rooted record keeps its parent link.** The cap paragraph gains one sentence: *"A re-rooted record keeps its own parent link: `--tree --json` still reports the `parentStack`, `parentLogicalId` and `parentRegion` it names, which is how a consumer tells it apart from a genuine top-level stack."* The won't-do about the text view now rests on something a user can read. I checked that the new sentence adds no `<n> levels` match: the page still has exactly two, so the cap fence's floor is unaffected. The changelog entry is untouched.

**M11 — singular, with what the walk does spelled out.** The comment now says the seeded loop is met from below, and that only `LoopA` is reached because the walk stops at the first node that already has a depth, so `LoopB` is never visited in that pass.

This round touches `docs/cli-state.md`, one comment in `tests/unit/cli/state-list-tree.test.ts`, and the PR body. `src` and the changelog fragment are unchanged. The full suite passes, and `verify-pr` / `pr-review` are re-bound to the new head.

go-to-k:
**Correction: the M9 text I supplied contains a false claim, and you pasted it verbatim because I told you to. That is my error, not yours.** You checked every code reference in it — which is exactly right, and it would not have caught this, because what is wrong is a claim about BEHAVIOUR that no line number attests to.

The TEMPLATE-arm paragraph says that arm "reads no child state". It does. `buildDiffTree` calls `loadStateOrEmpty(stackName, region, stateBackend)` at the top of EVERY node (`src/cli/commands/diff-recursive.ts:1186` on this PR's base), and `childStackName` is `` `${stackName}~${logicalId}` `` (`:1236`) — so the nested state key grows by one segment per level on the template arm exactly as it does on the state-only one, and S3's 1024-byte key limit bounds both.

I measured it against the built binary rather than reasoning about it, on a two-file `cdk.out` whose child template declares a nested row resolving back to itself:

```text
StateError: Failed to get state for stack
'Cdkd3239Live~Child~Loop~Loop~ ... ~Loop' (us-east-1): Your key is too long
```

About 190 levels, then it stops, exit 1. So the arm is NOT unbounded; what it lacks is a diagnosis — it burns ~190 pointless state reads and then reports a stack name that does not exist, mentioning neither the cycle nor the asset path.

Worth naming the trap, because it is the one this PR has been fighting all along: a unit test with a fake state backend DOES run away, since the fake enforces no key length. The mock endorses an assumption the wire does not. That is why the number above comes from the binary.

It is also worth saying that **the pre-M8 body was closer to right than what M8 and M9 produced**. It said these walkers were bounded by the key limit, and for this arm that was correct. My M8 asked you to split the claim per arm on a premise I had not measured, and M9 then hardened the wrong half into the paragraph that becomes the commit message. Two rounds of my requests made that sentence worse.

**M12 — replace the TEMPLATE-arm paragraph (`body`, the one beginning "`buildDiffTree`'s TEMPLATE arm is a different case").** Required before merge, for the same reason M8 was: the body is squash-merged into the permanent commit message. Exact text:

> `buildDiffTree`'s TEMPLATE arm follows the synth output through `readNestedTemplate` / `indexNestedChildTemplates`, and it has no visited set and no self-reference check, so a hand-modified relative `aws:asset:path` can close a cycle and the walk follows it. It is bounded all the same, and by the SAME key limit as the two above: `buildDiffTree` reads state at every node (`src/cli/commands/diff-recursive.ts:1186`) against a `childStackName` that grows by one `~<childLogicalId>` per level (`:1236`). Measured against the built binary, a self-referencing assembly ends after about 190 levels with `Your key is too long`, naming a stack that does not exist and mentioning neither the cycle nor the asset path. So what that arm lacks is a diagnosis, not a bound. Filed separately as go-to-k/cdkd#3239; out of scope here.

Nothing else in the paragraph changes: the two state-derived walkers, the not-audited list, and the open `await` question all stand. go-to-k/cdkd#3239 has been corrected the same way — its first body claimed call-stack exhaustion and `Severity: medium`; it now records the measurement above and `Severity: low`.

M10 and M11 land as written; I have nothing further on either. **M12 is the last item**, and it is a body edit, so it does not move HEAD or invalidate your markers. Push it and I will confirm rather than re-review.

nix-tkobayashi:
M12 is in: the TEMPLATE-arm paragraph is replaced with your text, word for word. This was a body edit only, so HEAD is still `e2cb243a` and the markers did not move. I read the body back after the PATCH: "reads no child state" no longer appears, and "lacks is a diagnosis, not a bound" does.

Before pasting I checked your line numbers against `origin/main`, not against my branch. In this paragraph that distinction matters. `:1186` is the `loadStateOrEmpty` call near the top of `buildDiffTree` (the function starts at `:1103`), and `:1236` is where `childStackName` is built. Both are right on main. Because this PR does not touch `diff-recursive.ts`, main's numbers are also the numbers after the squash.

That check turned up one more thing, and I fixed it in the same edit. The first half of the paragraph — the M9 text, which I had confirmed against my branch — gave `buildDeletedSubtree` as `diff-recursive.ts:1280-1342`. That range is from this branch's older base. On main the function runs `1311-1373`, 31 lines lower. Left alone, the commit message would have quoted two different bases for the same file in one paragraph, and one of them would be wrong for the commit. Both references now point to main.

Agreed on the trap. This PR had already hit it once: a checker whose fake allows an input the real system rejects. A fake state backend with no key-length limit lets the walk run forever where S3 stops it at 1024 bytes. The ~190-level figure from the built binary is the number I would trust, and I'm glad the paragraph now carries it rather than my account of the mechanism.

go-to-k:
M12 lands, and your catch on the `buildDeletedSubtree` range is the right one — quoting two different bases for one file in a single paragraph would have been worse than either number alone. **No findings outstanding. This PR is done.**

One coordination item, and it is caused by my work rather than yours, so it is mine to raise: **this PR should merge BEFORE go-to-k/cdkd#3243.**

Your reasoning that "main's numbers are also the numbers after the squash" holds only while nothing else changes `diff-recursive.ts` first. go-to-k/cdkd#3243 does — it adds the ancestor-chain guard — and it moves every line this paragraph cites:

| citation | on `main` today | after go-to-k/cdkd#3243 |
| --- | --- | --- |
| `loadStateOrEmpty` in `buildDiffTree` | `:1186` | `:1204` |
| `childStackName` | `:1236` | `:1254` |
| `buildDeletedSubtree` | `1311-1373` | `1377-1439` |

Merged in this order, your three citations are correct at your squash and go-to-k/cdkd#3243's are correct at its own. Merged the other way round, this PR's permanent commit message ships three line numbers that are wrong the moment it lands — the exact failure the M12 round was about, arriving through a different door.

go-to-k/cdkd#3243 is rebased onto current `main` and green, so it can follow immediately; nothing about it needs this PR's content.

Thank you for the four rounds on this one. Two things in particular were better than what I asked for: re-measuring the mutation table at the current head rather than re-typing it, and cutting the unverified `await` sentence instead of shipping it — that instinct is what let the M8 error be caught at all, because you checked the claim's references even when the claim came from me.

go-to-k:
**M13 — rebase onto current `main` before this merges.** Required, and it is one command; everything else on this PR is done.

I took your branch into a local worktree to run the gates before merging, and its full suite fails on one case:

```text
rule-file-payload.test.ts > every budgeted path still fits its cap once this
branch is merged into origin/main

`src/deployment/deploy-engine.ts`: this branch adds 0 B to what that path
loads. Against origin/main's current 65644 B that projects to 65644 B, over
the 61000 B cap.
```

**It is not a defect in your change, and the message is misleading about why.** `3f1ed63e` is not an ancestor of `e2cb243a`, so this branch still carries the pre-go-to-k/cdkd#3237 budget row (`['src/deployment/deploy-engine.ts', 43_000, 61_000]` at `:380`). go-to-k/cdkd#3237 raised it to `66_500` in the same commit that grew the payload. The fence projects `main`'s CURRENT payload onto the BRANCH's copy of the table, so "the overrun arrived on origin/main" is true of the bytes and wrong about the cause.

I know because I filed that as a repo-wide breakage yesterday (go-to-k/cdkd#3248) and had to close it: `main` was never red. The remedy both times is `git rebase origin/main`.

Your PR does not touch `rule-file-payload.test.ts`, so the squash itself is safe either way — main's copy of the table survives. What the stale base blocks is VERIFICATION: I cannot get a green local suite on this head, so I cannot set the `check` marker, and the merge gate reads that marker rather than the PR page. Your CI is green for the same reason mine was — its base was frozen before go-to-k/cdkd#3237 landed, so it is attesting to a `main` that no longer exists.

After the rebase, please confirm the full suite is green at the new head and re-bind your markers. I will then merge this one FIRST, ahead of go-to-k/cdkd#3243, for the line-number reason in my previous comment — and since go-to-k/cdkd#3243 is already rebased onto current `main`, your rebase also puts the two PRs on the same base, so the citations settle consistently.

nix-tkobayashi:
M13 done. Rebased onto `origin/main` at `97206c84`; the new head is `0b1887bf`. The rebase applied without conflicts.

- The budget row the branch was carrying is now main's: `['src/deployment/deploy-engine.ts', 55_000, 66_500]`.
- Full suite at `0b1887bf`: rc=0, 1034 files / 24104 tests passed, no errors. The first run had every test pass but exited 1 on a `Worker exited unexpectedly` pool error while six peer vitest processes were running. I re-ran once the machine was quiet, and that run is clean.
- `check` / `docs` / `verify-pr` / `pr-review` are re-bound to `0b1887bf`.
- The body's citations still hold at this base: `diff-recursive.ts:1186` is the `loadStateOrEmpty` call in `buildDiffTree`, `:1236` builds `childStackName`, and `buildDeletedSubtree` runs `1311-1373`. That means the merge-before-go-to-k/cdkd#3243 ordering still applies.

go-to-k:
Merged as `c70dcecc`. Thank you — four rounds, and the PR got better at each one.

Three things I want on the record, because they are what made the rounds converge rather than repeat:

**You re-measured instead of re-typing.** The mutation table was regenerated by a script at the current head after the branching case and the loop-seeding refactor existed, and five of its eight rows turned out to have been wrong. Nothing downstream would ever have re-checked those numbers.

**You cut the unverified sentence.** The `await`-driven-recursion claim went out of the body because you had not measured it, and you said so unprompted. That instinct is what made the M8 error findable at all: you check a claim's references even when the claim comes from the reviewer, which is exactly how the wrong per-arm split got caught.

**You checked my line numbers against the right base.** `buildDeletedSubtree` at `:1280-1342` was from your older base and would have shipped two different bases for one file inside one paragraph. I supplied that range; you caught it.

For symmetry, the errors in those rounds that were mine: M8 asked you to split a claim I had not measured, and M9 then hardened the wrong half of it into the paragraph that becomes the permanent commit message. The corrected version is in the merge commit.

go-to-k/cdkd#3243 follows now, on this base, so its own line citations settle against a `main` that already contains this.

Issue #2715, opened 2026-09-06
chore(hooks): twelve suites run under bash 3.2 while running their hook under 5.x, so their 3.2 coverage is a claim about the test


go-to-k:
`run-tests.sh` drives every suite under both bashes AND exports `HOOK_BASH`
alongside each shell, precisely so a suite can run the HOOK under that
interpreter too. Twelve suites never read it. They spawn the gate as a
subprocess with a bare `bash "$HOOK"`, and the hooks are `#!/usr/bin/env bash`,
so the SUITE runs under 3.2 while the SUBJECT takes whatever comes first on
PATH -- 5.x. For those twelve, "passes under bash 3.2" is a claim about the
test, not about the thing under test.

That is not theoretical. go-to-k/cdkd#2650 hit it: `gate_strip_prefix` used the
inline bracket expressions `[\(\{]` / `[\)\}]`, and a backslash inside a POSIX
bracket expression is an ordinary MEMBER rather than an escape, so bash 3.2's
engine strips a leading backslash where 5.x does not. `\\cd /tmp ; echo hi >
<tracked>` came out as `cd /tmp` under 3.2 and the gate followed a `cd` real
bash does not run -- a fail-open in a blocking gate, invisible to every local
run, caught only by CI (the sole runner carrying 3.2 as both `bash` and
`/bin/bash`). Probed after the fix: reverting it leaves the gate suite 76/0
unshimmed and 75/1 under `HOOK_BASH=/bin/bash`.

The two suites in that PR now honour `HOOK_BASH`. **This issue deliberately
asserts no COUNT of the ones that do not**, because three defensible predicates
gave three answers on the same tree -- 12 (`bash "$HOOK"` spelled exactly), 13
(a reviewer's), and 28 (any literal-interpreter spawn of a hook path). A wrong
number in the prose would outlive whichever list someone acts on, so the
predicate and the command are the deliverable instead:

  suites that spawn a hook under an interpreter they name literally, and never
  read `HOOK_BASH`

```sh
for f in .claude/hooks/*.test.sh; do
  grep -q 'HOOK_BASH' "$f" && continue
  grep -qE '(^|[^A-Za-z_-])(bash|sh)[[:space:]]+("?\$\{?[A-Z_]*(HOOK|GATE)[A-Z_]*\}?"?|[^|&;]*\.claude/hooks/[^ ]*\.sh)' "$f" || continue
  printf '%s\n' "${f##*/}"
done
```

[...]

Issue #2686, opened 2026-09-06
chore(hooks): verify-pr has no sha binding, so a second lane in the same worktree inherits a green it never earned


go-to-k:
`.markgate.yml` gives the `verify-pr` gate `requires: [check, docs]` and, deliberately, **no own `include:`**:

```yaml
  verify-pr:
    # ... "No own `include:` — freshness is purely the AND of
    # children, avoiding the git-tree default."
    requires: [check, docs]
```

So the parent marker has no scope of its own to digest. Once `/verify-pr` has set it ONCE in a worktree, it never stales by itself — it is only ever MASKED by a stale child. Running `/check` and `/check-docs` un-masks it, and the gate that is supposed to physically block `gh pr create` / `gh pr merge` on a PR whose live behavior was never exercised goes green for a PR `/verify-pr` has never seen.

That is normally invisible, because a single-PR session runs `/verify-pr` anyway. It is NOT invisible in an IN-PLACE `/work-issues` run, where several lanes share ONE worktree and markgate state resolves under `git rev-parse --git-dir` — so lane N inherits lane N-1's parent marker.

## Measured, 2026-09-06, in this worktree

A `/work-issues` lane opened PR go-to-k/cdkd#2685 on branch `go-to-k/fix-2613-updatecontext-fence`, having run `/check` and `/check-docs` but never `/verify-pr`.

BEFORE the children were refreshed:

```
verify-pr               mismatch   21m13s ago  child check is stale
```

AFTER `markgate set check` + `markgate set docs` (the only variable moved):

```
check                   match      2m40s ago   -
docs                    match      2m37s ago   -
verify-pr               match      33m17s ago  -
```

The AGE is the tell: 33 minutes, against children 2 minutes old. That marker was set by the PREVIOUS lane in this worktree, for a different branch, before this branch existed. `gh pr create` for go-to-k/cdkd#2685 then went through with no block.

The sibling `pr-review` gate is NOT affected in the same way, and the contrast is the useful part: it also read `match` (42 minutes old), but its hook additionally compares the gitignored `.markgate-pr-review-sha` sentinel against the PR's HEAD, and that sentinel held `c139d9d67e6f93cc5464d53c275ea55be88664ad` while HEAD was `47955214271a373c21750502b37dbd8b2966e3bf` — so it would still have refused the merge. `verify-pr` has no such second binding.

## Why this is the dangerous direction

[...]

Issue #2354, opened 2026-08-27
chore(hooks): a QUOTED command LEADER evades the gate trigger, so `"sudo" git commit` reaches git ungated (go-to-k/cdkd#2333 residue)


go-to-k:
Dup-check: searched open issues for `gate_strip_prefix`, `quoted wrapper sudo gate`, and ran a body-window scan for `gate_strip_prefix` / `quoted wrapper` / `sudo. git`. No hits. The three neighbours are all closed or covered: go-to-k/cdkd#2333 (quoted/escaped STRUCTURAL token -- the verb and the leading `-`) and go-to-k/cdkd#2339 (backtick substitution inside a double-quoted span) are fixed by the lane that found this, and go-to-k/cdkd#2156 converged the trigger onto command position. None covers a quoted LEADER.

## The shape

go-to-k/cdkd#2333 made the gate trigger compare the verb and the leading `-`
against a token with its quoting REMOVED, via a length- and offset-preserving
dequoted view consulted by `gate_matches` after the raw text. That closed
`git "commit"`, `git \-C /tmp commit`, `gh "pr" merge 1` and `"git" commit`.

One structural position is still matched literally, and it is the one BEFORE
the command word. `gate_strip_prefix` runs ahead of the view and matches its
leader set (`sudo`, `env`, `command`, `xargs`, and the rest) as literal text,
so quoting the LEADER evades every gate:

```
"sudo" git commit -m x
'sudo' git commit -m x
s"u"do git commit -m x
\sudo git commit -m x
```

Each is an executable command that really does run the gated verb, and each
reaches git with no gate armed -- the same silent, exit-0 miss go-to-k/cdkd#2333
describes, one token to the left.

## Why it was NOT fixed in the lane that found it

Deliberate, and recorded in-code rather than left to be discovered. The
go-to-k/cdkd#2333 / go-to-k/cdkd#2339 lane had already made two structural
changes to `command-match.sh` (a backtick arm on `flush_line`'s in-quote
branch, and the dequoted view consulted at five sites). A third structural
change to the same file in the same lane is the cascade shape this repo has
paid for repeatedly -- most recently five rounds on go-to-k/cdkd#2027, where
each round's fix created the next round's blocker in the same two functions.

It is also not a one-line extension of the existing fix. The view is consulted
AFTER the raw text specifically so the change is additive-only; that ordering
was forced by the lane's own differential fence, which measured that testing
[...]

Issue #559, opened 2026-05-23
make markgate gate hooks cwd-aware so parallel agents in separate worktrees stop colliding on main tree's state


go-to-k:
## Summary

Every markgate-backed gate hook in `.claude/hooks/` intentionally `cd`s into the main worktree before calling `markgate verify`. This defeats markgate's per-worktree marker isolation and forces all parallel agents to converge on the main tree's working files, which is the actual source of the cross-agent edit-race documented in memory rule `feedback_cross_agent_main_tree_contention.md`.

## Background

markgate 0.3.3 stores state at `<git rev-parse --absolute-git-dir>/markgate/`. That path is per-worktree by design:

- Main worktree: `main/.git/markgate/`
- Side worktree: `main/.git/worktrees/<name>/markgate/`

A `markgate set check` from a side worktree lands in the side worktree's state dir without touching main tree's. That's the intended isolation, and it works correctly out of the box (verified against markgate 0.3.3 on 2026-05-24, see https://github.com/go-to-k/markgate/issues/64 for the empirical repro).

## The cdkd-side override

cdkd's gate hooks (`pr-review-gate.sh`, `check-gate.sh`, `verify-pr-gate.sh`, `integ-destroy-gate.sh`, `integ-broad-gate.sh`, `integ-local-gate.sh`, `integ-schema-migration-gate.sh`) all start with this resolution:

```bash
# .claude/hooks/pr-review-gate.sh:30-48
SCRIPT_REPO="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"

if git_common=$(git -C "$SCRIPT_REPO" rev-parse --git-common-dir 2>/dev/null); then
  case "$git_common" in
    /*) abs_common="$git_common" ;;
    *)  abs_common="$SCRIPT_REPO/$git_common" ;;
  esac
  REPO="$(cd "$(dirname "$abs_common")" 2>/dev/null && pwd)" || REPO="$SCRIPT_REPO"
else
  REPO="$SCRIPT_REPO"
fi
...
cd "$REPO" 2>/dev/null || exit 0
```

The comment makes the intent explicit:

> Prefer the main working tree (shared across `git worktree` instances) so markgate marker state is consistent regardless of which worktree triggered the hook.

The trade-off: markgate state is "consistent" (= always main tree's view) at the cost of per-worktree isolation.

## Concrete impact

Two agents working in parallel can't both use markgate-gated commands cleanly:

[...]

PR #3214, opened 2026-09-16
chore(hooks): require verify-pr-gate's sha binding only in the repo that defines it

### LINKED ISSUE #3209: chore(hooks): verify-pr-gate demands a cdkd-only sentinel from sibling repos, so the mirror PR that retro.md section 10-c mandates cannot be opened
## What happens

`verify-pr-gate.sh` refuses `gh pr create` / `gh pr merge` for a **sibling repo** even after that repo's own checklist is complete and its own markers are set, and the refusal is one the sibling cannot clear:

```
Blocked by verify-pr-gate: ... marker bound to: <unset>
```

Measured today on both mirror PRs of the go-to-k/cdkd#3152 retro (go-to-k/cdk-local#736, go-to-k/cdk-real-drift#1914).

## Mechanism

The gate resolves the directory the command will run in, then decides in three steps (`.claude/hooks/verify-pr-gate.sh`):

1. target is not a git repo -> pass through;
2. target top has no `.markgate.yml` -> pass through (the deliberate opt-in scope: "a session can still run git / gh against OTHER repos that have no markers at all, and blocking there is pure friction");
3. otherwise require BOTH `markgate verify verify-pr` rc=0 in that tree AND `<target top>/.markgate-verify-pr-sha` equal to the target's HEAD (the sha binding added by go-to-k/cdkd#2686).

Step 2 is what makes step 3 unreachable for a sibling. cdk-local and cdk-real-drift BOTH carry a `.markgate.yml` with a `verify-pr` gate, so they are in scope — but neither repo has ever written the sentinel:

- `git -C <sibling> grep -l markgate-verify-pr-sha origin/main` -> nothing, in either repo;
- their `.gitignore`s list `.markgate-pr-review-sha` (and cdk-real-drift's bughunt sentinels), never `.markgate-verify-pr-sha`;
- the sentinel is a cdkd-only device: go-to-k/cdkd#2686 added it so lane N cannot inherit lane N-1's `verify-pr` parent marker, a mechanism neither sibling chose.

So `markgate verify verify-pr` reports `match` in the sibling's tree, `recorded_sha` stays empty, and the comparison fails.

## Why the prescribed remedy does not clear it

`.claude/rules/hooks.md` -> "Working on a sibling repo from a cdkd session" says: **complete the TARGET repo's own checklist and set its markers legitimately, then retry. Never route around the block, and do not 'fix' the target repo to match cdkd.** Both mirror lanes did exactly that (each set its repo's `check` / `docs` / `verify-pr` at the pushed HEAD, tree clean) and the retry still refuses, because what is missing is not a marker but a FILE FORMAT the sibling does not define. Writing it into the sibling is the second half of the same sentence -- "do not fix the target repo to match cdkd".

The asymmetry is sharper than that: cdk-real-drift's OWN `verify-pr-gate` exempts a PR touching no `src/**`, which is exactly the shape of both mirror PRs. cdkd's gate has no such exemption, deliberately (hooks.md records why: a `.claude/hooks/**`-only PR touches no `src/**`, so that exemption would have waived `/verify-pr` for the change that introduced a remote-code-execution path). That part is right and must not change. What is wrong is requiring cdkd's *mechanism* in a repo that never adopted it -- itself an instance of "do not port cdkd's stricter gates down to a sibling".

## Why it matters: the gate blocks a step the skill mandates

`.claude/skills/work-issues/references/retro.md` §10-c: **"The session that FINDS the lesson lands all three"** repos, one `chore:` PR per repo, and session budget is not a deferral reason. So the flow REQUIRES a cdkd-rooted session to open PRs in both siblings, and this gate is what stops it. Today that cost an `AskUserQuestion` round-trip and a hand-written attestation (below); left as is, every future mirror run pays the same.

## Scope: this is the only gate of the class (measured)

- `pr-review-gate.sh` — both siblings DO define `.markgate-pr-review-sha` (gitignored in both), and a docs-only mirror PR resolves to the `inline` tier, which passes through with no marker at all. Not affected.
- `integ-broad-gate.sh` — its `CROSS_CUTTING_REGEX` names cdkd `src/**` paths, so a sibling's diff never matches. Not affected.
- `check-gate.sh` / the `docs` gate — plain markgate gates both siblings define themselves. Not affected.

## Proposal A (the fix)

Require the sha binding only when the target repo IS the repo that defines it; keep requiring `markgate verify verify-pr` for a foreign target, so cdkd's POLICY still applies while cdkd's MECHANISM stops being demanded of a repo that has none.

Identity test: compare the target's `git rev-parse --git-common-dir` (canonicalised) against the same value for the repo the hook itself belongs to. The common dir is what makes every cdkd worktree one repo, so per-worktree marker isolation (go-to-k/cdkd#559) is untouched and only a genuinely different repository takes the relaxed path.

KNOWN BOUND to state in the code: for a foreign target, `verify-pr` is `requires: [...]` with no `include:` of its own there too, so a marker stale by inheritance can pass. That is the sibling's own design, and hooks.md forbids porting cdkd's stricter gate down to fix it.

## Proposal C (interim, and worth keeping either way)

Until A lands, `hooks.md`'s sibling-repo section should say what a session is expected to DO when the refusal is unclearable — today it prescribes a remedy that does not work and forbids the two obvious alternatives, which is how a hand-written attestation becomes an unexamined precedent.

## Full disclosure: what was done today

With the maintainer's explicit go-ahead, for each sibling PR I wrote `<worktree>/.markgate-verify-pr-sha` = HEAD after confirming the tree was clean and that repo's own `check` / `docs` / `verify-pr` were fresh at that HEAD, opened the PR, re-bound it after a fix-round force-push, merged, and removed the file with the lane worktree. Untracked, in a throwaway worktree, no tracked file and no target-repo policy changed. It is a true attestation, and it was still cdkd's `/verify-pr` skill's "settable ONLY by this skill -- setting it by hand to bypass the gate defeats the point" satisfied by hand. That is the precedent this issue exists to replace.

Dup-check: searched open and closed cdkd issues for `verify-pr-gate` / `sibling repo` / `markgate-verify-pr-sha` / `cross-repo gate`; go-to-k/cdk-real-drift#1863 is the mirror-image complaint from the other side (a session working that repo from another repo's workspace runs THAT repo's gates) and names no remedy for this direction. Nothing covers the sentinel requirement.

Session-fit: now (do it in this session) -- the hook, its test harness and `hooks.md`'s sibling section were all read in this session while diagnosing the refusal, and the evidence (the exact refusal, both siblings' gate shapes, the §10-c conflict) exists only here
Severity: medium -- no wrong result ships, but a flow step the skill MANDATES cannot be completed from the session that owns it, and the workaround is a hand-written attestation that erodes the gate it satisfies
Effort: small (S) -- one identity comparison in one hook plus cases in its existing `.test.sh` harness, riding the ordinary `/check` cycle; no `src/**`, no integ
Estimate: ~1 h -- most of it the hook's test harness (both the same-repo and foreign-repo directions) and the `hooks.md` wording

go-to-k:
Closes #3209

## What was broken

`verify-pr-gate.sh` refused `gh pr create` / `gh pr merge` in cdk-local and
cdk-real-drift with a block no legitimate action could clear:

```
Blocked by verify-pr-gate: the `verify-pr` marker is fresh but bound to a different commit.

  HEAD is:          8da4f674e13f2fe97c4150ade434eeaeea7086ce
  marker bound to:  <unset>
```

Both siblings carry a `.markgate.yml`, so they are inside the gate's opt-in
scope, and both declare a `verify-pr` gate — but the gate ALSO compared
`<target top>/.markgate-verify-pr-sha`, a cdkd-only sentinel added by
go-to-k/cdkd#2686 that neither sibling has ever written. Re-measured on this
branch rather than relayed from the issue:

```
$ git -C <sibling> grep -l markgate-verify-pr-sha origin/main   # both repos
(no tracked reference on origin/main)
$ grep -n markgate <sibling>/.gitignore
45:# markgate pr-review sentinel
46:.markgate-pr-review-sha            # ...and no verify-pr sentinel, in either repo
$ grep -n '^  verify-pr:' <sibling>/.markgate.yml
203:  verify-pr:                       # cdk-local; cdk-real-drift at 300
```

What was missing was a FILE FORMAT rather than a marker, which is why
`.claude/rules/hooks.md`'s prescribed remedy — complete the target's own
checklist and set its markers — could not reach it, and why writing cdkd's
sentinel into a sibling is the second half of the same sentence ("do not fix the
target repo to match cdkd"). That blocked a step
`work-issues/references/retro.md` 10-c MANDATES: the session that finds a lesson
lands the mirror PR in all three repos.

## The fix

Require the sha binding only when the target repo IS the repo the hook file
lives in. Keep `markgate verify verify-pr` for a foreign target, so cdkd's
POLICY still applies while cdkd's MECHANISM is no longer demanded of a repo that
has none.

Identity is the canonicalised `git rev-parse --path-format=absolute
--git-common-dir` of the target compared against the same value for
`${BASH_SOURCE[0]}`'s directory. The header spells out why `BASH_SOURCE` is
right for REPO IDENTITY and was wrong for the MARKER LOOKUP pre-go-to-k/cdkd#559:
"which marker store" is per-worktree (markgate keys it on `--absolute-git-dir`)
and only the payload `cwd` answers it; "which repository" is deliberately
worktree-INVARIANT, which is exactly what the common dir is. go-to-k/cdkd#559's
isolation is untouched — the marker lookup still comes from the payload `cwd`,
every cdkd worktree resolves to the SAME common dir, and only a genuinely
different repository takes the relaxed path. **Fail closed**: if either common
dir is unresolvable, the sentinel stays required.

### ...and a second half the issue did not anticipate

A command can name a repo OTHER than the one the target directory is in, and
the relaxation is sound only while those are the same thing. Measured from a
cdk-local checkout, `gh pr merge <N> --squash --repo go-to-k/cdkd` went **2 → 0**:
cdk-local's fresh marker clearing the merge of a CDKD pull request, with neither
cdkd's marker nor its sentinel consulted, and nothing downstream catching it
(`pr-review-gate` and `ci-green-gate` resolve the PR in the target cwd with no
`-R` of their own).

A denylist for `-R` / `--repo` closed that spelling, and THREE review rounds
produced eight more that gh honours and it did not see — `GH_REPO=` in three
carriers, a PR-URL selector bare AND quoted, the flag NAME quoted or split, and
`-R` inside a combined short-flag cluster. Enumerating spellings has no
termination proof; `implement.md` 5-f' names the instrument change and its
stricter option — REFUSE the construct rather than model it. So the guard is an
**allowlist**. Relax only when:

- the hook's own environment carries no `GH_REPO` (a hook inherits the
  session's, and gh honours it over the local repo);
- the RAW command text contains no `GH_REPO` anywhere — raw, because
  `env "GH_REPO=x" gh …` hides it from the stripper, and whole-command, because
  `export GH_REPO=x; gh …` puts it in a different segment;
- every TOKEN of every matched gh segment is literally readable
  (`gate_word_is_literal`) — a `$VAR`, a substitution, a brace expansion, a
  backslash or an unbalanced quote stops the relaxation instead of passing
  through it;
- no token, with its quote characters removed, matches `--repo*`, `-R*` **or a
  combined short-flag cluster carrying `R`** (`-[!-]*R*`). gh's `-R` is an
  ordinary cobra short flag, so it clusters: measured, `gh pr view -cR
  go-to-k/cdkd 3214` and the glued `-cRgo-to-k/cdkd` both resolve the cdkd PR
  from a non-repo directory, and `-sdR <slug>` / `-sR<slug>` are the same shape
  on `pr merge`. A prefix-only `-R*` test let every one of those through;
- and no token is a URL / PR selector: the token's **first
  whitespace-delimited word** carries `://` or `/pull/<digits>`.

**That test took three tries, and the two it replaced are worth reading.**
"The token carries no QUOTE" failed because agents quote URLs routinely —
measured rc=0 for `gh pr merge "https://github.com/go-to-k/cdkd/pull/5"` while
the bare form returned 2. "The whole token carries no WHITESPACE" then failed on
a **false premise**: a selector CAN carry trailing whitespace and gh still
resolves it, because gh `url.Parse`s the argument and PREFIX-matches
`^/OWNER/REPO/pull/(\d+)`, so anything after the number is ignored. The
first-word test is what the whitespace version was reaching for — a URL inside
PROSE stays exempt, because prose starts with a word that is not a URL.

- and the segment did not reach the verb through **`xargs`**, which INJECTS
  argv from stdin — the words this predicate reads are then not the words gh
  receives.

**That last one is a STRUCTURAL distinction, not another spelling**, which is
why it closes in one arm instead of another round. Of everything
`gate_strip_prefix` strips ahead of the verb — `env` `command` `nohup` `sudo`
`exec` `time` `timeout <arg>` `VAR=value` `-FLAG`, the shell keywords
`then do else elif if while until !` `{` `(`, a `case … in` head and its arm
labels — **`xargs` is the only member that reads stdin and appends it as
arguments**; the rest exec the command with the argv written in the text (`env`
changes the ENVIRONMENT, which is what the `GH_REPO` clauses above are for). It
reads `gate_segments_raw`, because `gate_segments` emits each segment with the
wrapper ALREADY stripped and a first attempt at the check saw nothing.

**Never stricter than `origin/main`**: refusing to relax falls back to the
binding requirement, which was unconditional before this change. Verified across
nine shapes × {origin/main, pre-guard, branch}: no shape where the branch
refuses and `origin/main` passes. Two deliberate over-refusals, both fail-closed:
a `-R` naming the target's OWN repo, and `-tRelease`-style short flags whose
VALUE begins with `R`.

### The refusal now says what to do

When the allowlist RETRACTS the relaxation, the block names the target, names
what in the command resolved another repo, and prescribes re-running from the
target repository's own checkout with literal arguments. Previously that path
printed the second-lane text — sending the reader to `/verify-pr` and to a
sentinel a sibling cannot write, the one situation where the message is actively
misleading, and with `GH_REPO` exported in a shell profile EVERY sibling PR
lands there.

### The residual class, as one thing

The predicate answers "does this TOKEN name another repo" from the token's
TEXT. Two questions that is structurally unable to answer:

- **(a) WHICH token gh will read as the selector.** It does not know, so it
  scans every token and models prose with the first-word heuristic — and that
  heuristic is exactly where the `--body "https://x is the link"` over-refusal
  comes from. This one IS closable, by deciding on the SELECTOR POSITION
  instead of on every token: **go-to-k/cdkd#3256**.
- **(b) A repo reaching gh through a channel that is not the segment's literal
  argv at all** — an `upstream` remote in the target checkout
  (**go-to-k/cdkd#3235**), a `GH_REPO` / `GH_HOST` assembled at run time, a
  `gh alias` expanding to a `-R`. **No** reading of the command text closes
  this class, positional or otherwise; only asking what gh itself would resolve
  does.

## Known bounds, stated in the code rather than chased

- A foreign `verify-pr` has no `include:` of its own there either, so a marker
  stale by inheritance can pass — the sibling's own design, which hooks.md
  forbids porting cdkd's stricter gate down to fix.
- A second CLONE of cdkd is foreign by the same test, for the same reason.
- A symlinked `.claude` whose physical location sits in another repo is a
  RESOLVABLE-but-WRONG identity, which "fail closed" does not cover. Needs write
  access to the checkout.
- **gh resolves a repo from the target CHECKOUT too, and that is wider than
  `gh repo set-default`.** Measured in a fresh repo with `origin` = cdk-local,
  `upstream` = cdkd, NO `gh-resolved` key and no interaction:
  `gh repo view --json nameWithOwner` answers **go-to-k/cdkd** — gh prefers an
  `upstream` remote over `origin`, so a plain `git remote add upstream <other
  repo>` re-points every later bare `gh pr merge <N>` there. Removing the
  upstream returns the answer to cdk-local. **Filed as go-to-k/cdkd#3235**, not
  guarded here: the honest fix is a remote-slug COMPARISON, a structural fix
  `references/verify.md` 8-a says to file rather than take at the end of a
  cascade, and a presence test would refuse a sibling which ran
  `gh repo set-default` on ITSELF. Landing it would also retire the
  second-clone bound.
- The two environment tests cannot see a name assembled at RUN TIME
  (`export "GH_RE"PO=…`, a sourced `.env`), and `GH_HOST` has the same shape.
  Stated rather than enumerated — enumerating spellings is the instrument this
  guard was rewritten to get away from.
- `env "GH_REPO=x" gh …` — the assignment QUOTED — matches no verb in the
  SHARED matcher, so no gate fires at all. **Not this gate's and not new**:
  measured with `gate_matches`, `env FOO=1 git commit -m x` matches while
  `env "FOO=1" git commit`, `env 'FOO=1' git commit`, `env "FOO=1" git push`,
  `env "FOO=1" gh pr merge 42` and `nohup env "FOO=1" git commit` do not. Folded
  as a checklist row onto go-to-k/cdkd#2354, which already carries the
  quoted-LEADER half of the same `gate_strip_prefix` root cause.

## Live before/after

Both hooks run from installed copies against one fixture shaped like the
siblings (`.markgate.yml` present, `verify-pr` declared, no sentinel), markgate
shimmed FRESH so marker state cannot explain the difference:

```
== BEFORE (origin/main hook), sibling target, fresh marker, no sentinel
   exit: 2
   Blocked by verify-pr-gate: the `verify-pr` marker is fresh but bound to a different commit.
     HEAD is:          8da4f674e13f2fe97c4150ade434eeaeea7086ce
     marker bound to:  <unset>

== AFTER (branch hook), same fixture
   exit: 0

== AFTER, same fixture but marker STALE (policy must still apply)
   exit: 2
   Blocked by verify-pr-gate: the `verify-pr` marker is stale (digest differs).

== CONTROL go-to-k/cdkd#2686: SAME-repo target, fresh marker, no sentinel
   BEFORE exit: 2
   AFTER  exit: 2   (still "bound to a different commit")
```

One fixture defect found and fixed while writing that: the first draft put the
"after" hook in a bare temp dir, its own common dir was unresolvable, and the
arm refused — correctly, for the wrong reason.

## Tests

`.claude/hooks/verify-pr-gate.test.sh`: **104 cases (43 on `origin/main`)**,
green under bash 5.3.9 and 3.2.57, and green with the HOOK itself running under
3.2 (a `bash` shim first on PATH, since this suite does not yet honour
`HOOK_BASH` — go-to-k/cdkd#2715's list). The whole harness
(`bash .claude/hooks/run-tests.sh`) is green locally, and `hooks.yml` runs it
again on this exact tree.

The suite now installs the SUBJECT into each fixture repo rather than running
`.claude/hooks/verify-pr-gate.sh` directly. Without that, the subject's repo is
always cdkd, every throwaway fixture is FOREIGN to it, and **every pre-existing
sha-binding case would silently switch to exercising the relaxed path** — the
go-to-k/cdkd#2686 cases would have passed for the wrong reason. A case now
states its repo identity by choosing which copy it runs. A vacuity guard asserts
the fixtures' premises out loud before the cases run.

It is also **hermetic** now: the suite `unset`s `GH_REPO` and `GH_HOST` at the
top. It read the ambient environment before, and a reviewer running it with
`GH_REPO=x` exported got 69/10 — every relaxation case refusing, for a reason no
case named.

### The four gh-resolvable shapes the whitespace test let through

Measured on gh 2.92.0 from a NON-REPO directory, `gh pr view <sel> --json url`
resolved cdkd's PR for every one of these, and each took the relaxed path:

| selector | gh resolves it | before | after | mutant R9 |
| --- | --- | --- | --- | --- |
| `…/pull/42` (bare) | yes | 2 | 2 | — |
| `…/pull/42 ` (one trailing space) | yes | **0** | **2** | red |
| `…/pull/42?x=a b` | yes | **0** | **2** | red |
| `…/pull/42/files x` | yes | **0** | **2** | red |
| `…/pull/42` + NBSP | yes | **0** | **2** | red |
| `--body "see https://… "` (control) | — | 0 | **0** | — |
| `--body="see https://…"` (control) | — | 0 | **0** | — |
| `--body "https://… is the link"` | — | 0 | **2** | over-refusal, pinned |

A LEADING space, tab or newline needs nothing here: gh itself refuses those.
A scheme-less `github.com/<o>/<r>/pull/<n>` is ALSO refused, and its case says
so — measured, gh does NOT accept that form (it falls back to a branch), so the
case pins an over-refusal rather than a live selector shape.

### `xargs`, the one wrapper that injects argv

| command | gh resolves it | before | after | mutants |
| --- | --- | --- | --- | --- |
| `printf <pr-url> \| xargs gh pr merge --squash` | yes | **0** | **2** | R12, R13 red |
| `printf <pr-url> \| xargs -n1 gh pr merge --squash` | yes | **0** | **2** | R12, R13 red |
| `xargs gh pr merge --squash < url.txt` | yes | **0** | **2** | R12, R13 red |
| `gh pr merge 42 --squash` (control) | — | 0 | **0** | — |
| `gh pr create --title xargs …` (the word as an ARGUMENT) | — | 0 | **0** | — |

Standing hole, not a regression: the pre-allowlist commit gives the same
verdict. **R12** drops the arm; **R13** points it at the STRIPPED segments
instead of the raw ones, which is the defect the first attempt shipped — the
wrapper is already gone by then, so the check saw nothing.

Nineteen mutation probes on the final tree, from a harness that prints byte
deltas and anchor counts (so an inert probe is reported, not assumed absent),
restores byte-exact from a `cp` backup, and exits non-zero if any mutant
discriminates nothing:

```
BASELINE pass=104 fail=0
P1  sentinel compare `=` -> `!=`                        46 / 58 red
P2  `target_is_foreign=0` -> `=1` at init               81 / 23 red
P3  foreign pass `-eq 1` -> `-eq 2`                     88 / 16 red
P4  drop `status -eq 0` from the foreign pass          100 /  4 red
P5  drop both `-n` guards -> `if true`                 103 /  1 red
P6  `--git-common-dir` -> `--git-dir`                  103 /  1 red
P8  identity reads the cwd, not `$target_dir`           99 /  5 red
P9  allowlist never consulted                           73 / 31 red
R1  make the `://` / `/pull/` test unmatchable          95 /  9 red
R9  RESTORE the whole-token + whitespace-exemption test 99 /  5 red
R10 make the first-word extraction a no-op             101 /  3 red
R11 drop `status -eq 0` from the retraction early exit 103 /  1 red
R4  drop the short-flag CLUSTER pattern                101 /  3 red
R5  drop the `GH_REPO` ENV clause                      103 /  1 red
R6  make the whole-command `GH_REPO` test unmatchable  101 /  3 red
R7  drop the tailored retraction MESSAGE branch        102 /  2 red
R12 drop the xargs arm entirely                        101 /  3 red
R13 point the xargs arm at the STRIPPED segments        101 /  3 red
P7  CONTROL, comment only                              104 /  0
ALL PROBES OK (0 problems) — final restore verified byte-exact
```

**R9 is the round-4 receipt**: restoring the pre-fix test (whole token PLUS the
old `*[[:space:]]*` exemption) reds exactly the four shapes above and nothing
else. R10 — making the first-word extraction a no-op — reds only the `--body`
controls, which is what proves the first-word half is the part that keeps prose
exempt. P2 is still the one to read: forcing `target_is_foreign=1` reds
TWENTY-THREE cases, every one a same-repo binding case including all of
go-to-k/cdkd#2686's.

One probe was RETIRED rather than kept green: swapping the selector test's input
form discriminated ZERO cases, so the third derived variable is gone and the
code now says the form is load-bearing for the FLAG test and not for the
SELECTOR test.

The identity block's PLACEMENT (above the `cd -P`) is fenced by moving it: the
suite stays green except the relative-invocation case, which flips **passing →
refusing** because from inside a foreign target the hook's own repo becomes
unresolvable and it fails closed.

## Docs

- `.claude/rules/hooks.md` → "Working on a sibling repo from a cdkd session":
  the prescribed remedy is now SUFFICIENT for `verify-pr`, and writing cdkd's
  sentinel into a sibling is explicitly not the remedy.
- `.claude/rules/hooks.md` → "Two gates bind their marker to a COMMIT": the
  `verify-pr` row said the binding is required, unqualified; now scoped.
- `.claude/rules/gate-sibling-repos.md`: "the other four markgate gates are NOT
  affected" is measured against the gate NAME, the only question that file
  answers — scoped, with a pointer to this fix.
- `.claude/rules/hooks-class-fences.md`: its exemption for this suite is
  RETRACTED, not reworded. Re-measured — the gate-name mutant
  (`verify verify-pr` → `verify check`) leaves the suite GREEN, because the
  refusal path calls `markgate status verify-pr` and the suite's argv-trace grep
  is satisfied by that; only `markgate-gate-name-class.test.sh` reds, whose own
  header has listed `verify-pr-gate Pass: 22 Fail: 0` since it was written.

No `changelog.d/` entry: no `src/**` change, so the shipped binary's behaviour
is unchanged (CLAUDE.md's rule for hooks / agent instructions).

## Review

Tier from `/review-pr`: base **inline** by `fc < 5`, NOT down-biased
(agent-instruction files are deliberately outside the docs bucket), raised to
**3-axis + security** — the recent-defect trigger fires on `verify-pr-gate.sh`
(the two preceding commits touching it, go-to-k/cdkd#2775 and go-to-k/cdkd#2766,
are both fail-open fixes, and this PR relaxes the mechanism the second added);
the diff RELAXES a gate, so "what does it now let through" is a security
question; and the test harness was restructured, which is a test-adequacy
question a code reviewer would not ask.

**Round 1** — 1 security blocker, 3 minor, 4 nits.

| finding | disposition |
| --- | --- |
| **security** — `gh -R <slug> pr merge` / `--repo` from a foreign cwd took the relaxed path (measured 2 → 0) | FIXED |
| **test gap** — no case had cwd != resolved target with a FRESH verdict, so a mutant reading `${hook_cwd:-$PWD}` survived all 54 | FIXED — 4 cases, both directions × both spellings (probe P8) |
| **spec** — a sentence I ADDED to `hooks-class-fences.md` was false | FIXED — re-measured and RETRACTED |
| **code** — the `/var` → `/private/var` rationale for `cd`+`pwd -P` is refuted | FIXED — the line stays, its rationale now says defensive and UNFENCED |
| **code** — a symlinked `.claude` in another repo makes every cdkd target foreign | STATED in KNOWN BOUNDS. Needs write access to the checkout, and the `.gitignore`-based discriminator offered instead would replace the mechanism go-to-k/cdkd#3209 decided on |
| **code** — placement rationale named only the `.` fallback | FIXED — widened; the placement now has a case |
| **code** — `gate_common_dir` sits in the shared library's namespace | FIXED — `__vpg_common_dir` |
| **spec** — PR body said "was 37"; `origin/main` is 43 | FIXED |
| **security** — `--path-format` needs git >= 2.31; older git fails closed | ACCEPTED, safe direction |

**Round 2** (delta-scoped, all four again) — 2 blockers, both against round 1's
fix rather than the original change.

| finding | disposition |
| --- | --- |
| **security + code** — the `-R` denylist is defeated by `GH_REPO=` (three carriers), a PR-URL selector, and a quoted or split flag NAME | FIXED — denylist replaced by the allowlist above; 19 cases including 6 controls |
| **test + spec** — the relative-invocation case fenced nothing (its cwd WAS the target), and the commit message named the flip in the wrong direction | FIXED — case re-pointed at a foreign target, verified by MOVING the block; direction corrected |
| **spec** — `hooks-class-fences.md` recorded "54/54 on that branch" in the commit that made it 64 | FIXED — figure deleted; `origin/main`'s 43/43 kept as a dated measurement |
| **code** — KNOWN BOUNDS omits `gh repo set-default` | STATED, with why the honest guard is a slug comparison |
| **spec** — "only the class fence reds" is consistent with, not independently confirmed to the exclusion of, every other suite (three 3.2-round latency FAILs under concurrent load, in suites that do not source this hook) | ACCEPTED as stated; the three are not re-run in isolation |

**Round 3** (independent, coordinator-side) — 1 blocker, 7 more. Every item
re-measured here before acting.

| finding | disposition |
| --- | --- |
| **BLOCKER** — a QUOTED URL selector took the relaxed path; the quote exemption was reached before the `://` test | FIXED — whitespace-free test on the unquoted value; 3 new cases (double-quoted, single-quoted, scheme-less) + probes R1/R2 |
| **security** — `-R` inside a combined short-flag cluster (`-sdR`, `-sR<slug>`) got through a prefix-only test | FIXED — `-[!-]*R*`; 3 cases + 2 controls + probe R4 |
| **code** — the retracted path printed the second-lane text, pointing at a sentinel a sibling cannot write | FIXED — retraction REASON kept and a tailored block printed; the generic `/verify-pr` remediation is skipped there; 2 cases (new text present, ordinary refusal unchanged) + probe R7 |
| **test** — the `GH_REPO` env clause had no case, and the suite inherited the ambient env (69/10 with one exported) | FIXED — case + probe R5, and the suite `unset`s `GH_REPO` / `GH_HOST` |
| **test** — the three-form loop was dead (`$noq` subsumes the others) | FIXED — one form, and the comment now states which test the choice is load-bearing for and which it is not |
| **security** — the `gh repo set-default` bound understates it: an `upstream` remote wins with no `gh-resolved` and no interaction | CORRECTED to what was measured, and the structural fix FILED as go-to-k/cdkd#3235 rather than taken at the end of a cascade |
| **nit** — the env tests cannot see a run-time-constructed name; `GH_HOST` same shape | STATED as KNOWN BOUNDS |
| **nit** — two stale "line 95" references | FIXED |

Also corrected in the same commit, from round 2's own report: the HOOK's
placement paragraph still carried the false claim (the test file's had been
fixed but not the hook's).

**Round 4** (independent, coordinator-side) — the blocker RECURRED inside round
3's own fix.

| finding | disposition |
| --- | --- |
| **BLOCKER** — a selector carrying trailing whitespace still resolves: gh PREFIX-matches `^/OWNER/REPO/pull/(\d+)`, so `…/pull/42 `, `…/pull/42?x=a b`, `…/pull/42/files x` and a trailing NBSP all took the relaxed path. The premise "a selector never has whitespace" was false | FIXED — first-word test; 4 blocking cases + 3 controls + probes R9/R10 |
| **nit** — the scheme-less-selector comment read as if gh ACCEPTS that form | FIXED — reworded; measured, gh falls back to a branch, so the case pins an over-refusal |
| **nit** — a STALE marker plus a retraction printed only the staleness, so clearing the marker re-blocked with a different message | FIXED — the early exit now requires `status -eq 0`; both reasons print, and the repo-override guidance is shared by both paths (2 cases + probe R11) |

**Round 5** (independent, coordinator-side) — the cascade ended on a
completeness PROOF rather than on "no more spellings found": gh's `ParseURL`
requires `http`/`https` + `:` at index 0 and a scheme cannot contain whitespace,
so every selector gh ACCEPTS carries `://` in its first whitespace-delimited
word, which the test always sees. One item remained.

| finding | disposition |
| --- | --- |
| `xargs` INJECTS argv from stdin, so the command text is not what gh receives — three forms relaxed (standing, not a regression) | FIXED — structural arm over `gate_segments_raw`; 3 blocking cases + 2 controls + mutants R12/R13 |
| the residual class needed stating as one thing, split into the positionally-closable half and the not-closable half | LANDED as a KNOWN BOUND naming go-to-k/cdkd#3256 (a) and go-to-k/cdkd#3235 (b) |

Out of scope by the coordinator's instruction and filed separately as
go-to-k/cdkd#3242: `GATE_RE_GH_PR_CREATE_OR_MERGE` does not tolerate flags
between `pr` and `merge`, so `gh pr -R <slug> merge 42` fires no gate at all.
Pre-existing, not touched here.

## Not verified here

- No real-AWS integ: no `src/**` change, so the deploy/destroy tiers do not
  apply. This lane ran no AWS work.
- `verify-pr-gate.test.sh` does not honour `HOOK_BASH`, so `run-tests.sh`'s 3.2
  round runs the SUITE under 3.2 and the HOOK under whatever PATH gives. Covered
  here by running the suite with a 3.2 shim first on PATH; the general fix is
  go-to-k/cdkd#2715's and is not in scope.
- The second-clone bound and the load-guard rationale for the three added
  library functions are reasoned, not measured.

https://claude.ai/code/session_01NKuq3XKEPEytBDbWD3xG4W


go-to-k:
Merged — thanks. Five rounds, and the two that mattered were the ones that stopped hunting spellings: the completeness proof off gh's own `pullURLRE` (every selector gh accepts carries `://` in its first whitespace-delimited word), and reading `xargs` off the RAW segments after the first attempt silently saw nothing. The KNOWN BOUND naming what a text-based predicate structurally cannot answer is what lets go-to-k/cdkd#3256 and go-to-k/cdkd#3235 be scoped instead of re-derived.

Issue #3211, opened 2026-09-16
fix(state): a per-field malformed state record reaches a provider verbatim in destroy-runner and deploy-engine (the per-FIELD twin of #3018's bag class)


go-to-k:
## Summary

`parseStateBody` deliberately validates only the ROOT shape and the schema version (the placement decision recorded above its refusal, go-to-k/cdkd#2947): it is the read path for deploy, destroy, rollback AND the recovery commands, so refusing a broken record there would make it unreadable by the two commands whose job is to remove it. The consequence is that each consumer must tolerate a malformed shape WHERE IT DEREFERENCES ONE.

go-to-k/cdkd#3018 owns that class at the BAG / null-ENTRY level (`Object.entries(state.resources)` / `Object.keys(..)` with no `?? {}` and no null-entry guard, 21 sites / 9 files). go-to-k/cdkd#3191 owns a non-object `properties` in `diff-calculator.ts`. What neither owns is the **per-FIELD** twin: a record that IS an object, whose `properties` / `physicalId` is not the type its declared type says, handed straight to a provider. Found by the security review of go-to-k/cdkd#3149 (which added exactly these checks on the journal side, where refusing is safe because the journal is discardable) and extended by its round-2 review.

Sites, measured on `40485f35`:

- `src/cli/commands/destroy-runner.ts:1439` — `resource.properties` as `provider.delete`'s 4th argument (the auto-delete tag, `EmptyOnDelete`, final-snapshot gating all read it); `:1204` `Properties: resource.properties || {}`; `:401` / `:415` read named keys off it.
- `src/deployment/deploy-engine.ts:5194`, `:5963`, `:6291`, `:6388`, `:7253` — `currentResource.properties` as a provider argument or as `previousProperties` (the round-1 review named only `:6388`; the round-2 review found the rest).

A planted or hand-edited `state.json` with `properties: "abc"` therefore reaches a provider verbatim. This is NOT a privilege gain — anyone who can write that object can equally plant a well-typed wrong value — so it is a robustness / diagnosability issue, not an exposure.

## Proposed fix

Guard where each consumer dereferences, the same shape go-to-k/cdkd#3018 uses, NOT at the parse boundary (that would re-open the decision go-to-k/cdkd#2947 recorded). Per site: treat a non-object `properties` as `{}` with a warning that names the logical id and the type, or refuse the single resource and continue — matching whatever go-to-k/cdkd#3018 settles for the bag-level case, so the two do not diverge.

[...]

Issue #3203, opened 2026-09-16
fix(rollback): a replay arm hands a provider an EMPTY desired bag when previousState.properties is absent, stripping every property off a live resource


go-to-k:
## Summary

Both the code and the security review of go-to-k/cdkd#3149 measured the same thing while checking one of my claims, and it is a defect in its own right: **a rollback arm can hand a provider an EMPTY desired property bag over a LIVE resource**, which a patch provider applies by removing every property.

Every replay arm coalesces the desired bag:

- `src/deployment/rollback-executor.ts:3031` — `desiredProps ?? {}` (the `revert` arm's `provider.update`)
- `src/deployment/rollback-executor.ts:3432` — `desiredProps ?? {}` (the `--revert-failed` arm)
- `src/deployment/rollback-executor.ts:2330` — `(await resolveReplayProps(...)) ?? {}` (the reverse-replacement re-create)

`desiredProps` comes from `previousState.properties`. go-to-k/cdkd#3149 refuses a non-OBJECT `properties` at the journal parser, but an ABSENT one is deliberately tolerated there (`previousState` is forwarded verbatim from a state record `parseStateBody` does not validate, so requiring presence would refuse a journal cdkd itself wrote). So `previousState: {}` or `{ "physicalId": "p" }` — reachable from a hand-edited state record, — flows to `provider.update(id, physicalId, type, {}, currentProps)`.

`{}` is not a no-op: it is a complete desired state that says "this resource has no properties". `JsonPatchGenerator.generatePatch` emits `op: 'remove'` for every key present in previous and absent from desired (`src/provisioning/json-patch-generator.ts:69-77`), and `cloud-control-provider.ts:632` / `:662` call it with no empty-desired guard. The user asked for a rollback and gets a stripped resource.

**The third site fails differently and the first draft of this issue got it wrong**: `:2330` is the reverse-replacement RE-CREATE, not an update. There `{}` creates a DEFAULT-configured resource and the arm then DELETES the live new one — a delete-class outcome, not a strip.

**This is not introduced by go-to-k/cdkd#3149** — the `??` sites predate it, and the refused non-object case produced the identical `{}` through the same coalesce. What go-to-k/cdkd#3149 changes is that the remaining route to it is now exactly the tolerated-absent case, which makes the gap nameable.

## Proposed fix

[...]

Issue #3191, opened 2026-09-15
fix(analyzer): cdkd deploy's change calculation reads a non-object properties bag unguarded


go-to-k:
## What is wrong

`cdkd deploy`'s change calculation reads a resource's stored `properties` bag
with no readability check. `parseStateBody` validates nothing inside the state
record's root, so `properties` may be any JSON value — a string, a number, a
list — from a hand-edited record, a torn write, or anyone holding
`s3:PutObject` on the state bucket.

`src/analyzer/diff-calculator.ts` reads it at `:223`, `:301-302`, `:380`, `:394`
and `:409`, and the file imports nothing from
`src/state/malformed-resources-bag.ts` (grep: no match). Measured on this tree:

```
Object.keys("abcdef")        -> ["0","1","2","3","4","5"]
JSON.stringify("abcdef")     -> "abcdef"
compare vs any desired object -> different
```

So the deep compare treats the planted value as a current-state object with
per-character keys, and every property the template actually declares reads as
**absent in current, present in desired** — a change.

## Why it matters, and the part that is not yet pinned

**Measured:** the read is unguarded and a non-object bag produces a spurious
property-change set.

**Not yet executed, and the reason this is filed at high severity:** replacement
is decided per property change (`diff-calculator.ts:554`, `:688`, via
`this.replacementRules.requiresReplacement(...)`, with
`createOnlyChangeRequiresReplacement` imported at `:15`). A desired property
that is create-only would then read as changed, which is the input
`requiresReplacement` turns into a REPLACE. If that path is reachable, a planted
or torn `properties` bag makes `cdkd deploy` **replace a live resource** — data
loss on a path an operator did not ask for.

Determining that is the first task on this issue: build the case
(current `properties` a string, desired declaring a create-only property such as
an S3 `BucketName`) and record whether the outcome is UPDATE or REPLACE. If it
is UPDATE only, this drops to medium and becomes a spurious-update defect. Do
not assume either way — the whole point of this issue is that nobody has run it.

## Relationship to the merged work

go-to-k/cdkd#3157, go-to-k/cdkd#3185 and go-to-k/cdkd#3190 closed this class for
the READ-ONLY views: `cdkd state list --long`, `cdkd state show`,
`cdkd state resources`. go-to-k/cdkd#3159 wired `refuseMalformedState` into
`scrub`, `import`, `orphan` and `rollback`, and `repairMalformedResourcesForReadOnly`
[...]

PR #3221, opened 2026-09-16
fix(rollback): refuse the nested journal records the executor dereferences, and record why state.json may not be refused the same way

### LINKED ISSUE #3149: fix(state): validate the nested record shapes both parsers hand to the executor (previousState, attemptedProperties, provisionedBy, oldResourceRetained, and state.json's per-record fields)

go-to-k:
Closes #3149

## Summary

go-to-k/cdkd#3140 refused an operation's own scalar fields; the NESTED records inside one were still unchecked casts. `rollback-journal.json` is writable by anyone with `s3:PutObject`, so a planted operation could hand a provider a value no handling makes correct:

- `previousState: "x"` loses the whole desired bag — and every arm coalesces (`desiredProps ?? {}` at `rollback-executor.ts:3031` / `:3432`, `(await resolveReplayProps(..)) ?? {}` at `:2330`) — so the provider receives an EMPTY desired bag over a LIVE resource; `JsonPatchGenerator.generatePatch` emits `op: 'remove'` for every key present in previous and absent from desired (`json-patch-generator.ts:69-77`), called with no empty-desired guard at `cloud-control-provider.ts:632` / `:662`.
- `previousState.properties: "abc"` — and the operation's own `properties`, the delete-path twin (`provider.delete(id, physicalId, type, op.properties, ..)` reads it for the auto-delete tag, the `EmptyOnDelete` gate and final-snapshot gating) — reach a provider verbatim, in the same argument position `attemptedProperties` occupies for a failed op.
- `previousState.physicalId` / `.resourceType` are rendered, compared AND written back into `state.json`.
- `attemptedProperties` is the failed-op twin of that provider argument, and `?? current.properties` passes a truthy `[]` straight through.
- `oldResourceRetained: "no"` selects the readopt arm through `??` (`rollback-executor.ts:890`), skipping the re-create and re-pointing state at the OLD physical id.

Each is refused by index and TYPE, the value never echoed, naming the same way out as go-to-k/cdkd#3140.

### The strictness rule, because it decided two non-obvious calls

**Refuse what no handling can make correct; tolerate what the state boundary tolerates.** Every check is TYPE-when-present rather than required-presence, and `provisionedBy` is not refused at all. `previousState` is forwarded VERBATIM from `currentState.resources[logicalId]` (`deploy-engine.ts`), a record `parseStateBody` deliberately does not validate — so requiring a nested field to be present, or refusing the `provisionedBy` enum, would refuse a journal **cdkd itself wrote** from a record the read boundary accepts. An unrecognised `provisionedBy` routes to the SDK provider (`=== 'cc-api'` is the only test any consumer makes) and is written into `state.json` by the rollback. Three mutation rows pin both tolerances (refusal, and FORWARDING), so a later "tightening" cannot take them silently.

### The issue's `state.json` half is REFUSED, not deferred

`parseStateBody` already records the decision (the comment above its root-shape refusal, go-to-k/cdkd#2947): inner shapes are deliberately not validated there because it is the read path for deploy, destroy, rollback **and** the recovery commands — so a refusal makes a broken record unreadable by the two commands whose job is to remove it. My own issue body proposed overturning that without citing it; it was filed from a reviewer trace rather than from the file. The remainder is owned by go-to-k/cdkd#3018 (the 21-site bag / null-entry class) and go-to-k/cdkd#3191, with the per-FIELD consumers the round-2 review found now filed as go-to-k/cdkd#3211.

**The journal is a different question because it is DISCARDABLE**: no recovery command parses it (`deleteState` sweeps the key in its `finally`), so refusing locks nothing out. Recorded in the issue, in the parser's comments, and in `.claude/rules/layout-state-types.md`.

### Known-open, named rather than implied

An ABSENT `previousState.properties` is tolerated here by the rule above, and still reaches the same `?? {}` — go-to-k/cdkd#3203, filed from this branch's own review and corrected by the round-3 one (the re-create arm's failure is delete-class, not a strip). It predates this change: the refused non-object produced the identical `{}` through the same coalesce.

## Test plan

- [x] **Unit tests pass** — 1024 files / 23,461 passed / 1 skipped, root asserted to this worktree, exit 0.
- [x] `tests/unit/types/rollback-journal.test.ts` +3 cases (an `it.each` over a non-object `previousState`; one bundled case over the six nested fields; a TOLERANCE case pinning that an absent nested field, any `provisionedBy`, and the sibling fields no check names all parse — and that the values are FORWARDED, not merely accepted).
- [x] **Three review rounds** (code + test + security each round), every item applied:
  - r1 → the `provider.update(.., undefined)` claim was FALSE in shipped source (every arm coalesces; the harm is an empty bag, which is worse) — corrected in the parser comment, the changelog, a test comment and the commit; `op.properties` was the unchecked twin of `attemptedProperties` and is now refused; three test gaps the review measured GREEN are fenced (`previousState.properties: null`, `attemptedProperties: null | []`, and the tolerance case, which pinned only "not refused" while its rationale rests on forwarding).
  - r2 → "an absent field is a shape every arm already handles" is false by go-to-k/cdkd#3203's own finding; the tolerance rationale named the wrong reason (tolerance is about LOCKOUT — normalising locks nobody out), so the assertion now pins the decision *this parser is a validator, not a normaliser* and records that a journal value currently WINS over the state record at `op.provisionedBy ?? current.provisionedBy`; two assertions had dropped their `segments[0].failedOperations[0].` prefix; the tolerance fixture gained the sibling fields (`updateReplacePolicy`, `attributes`) the review measured as unfenced.
  - r3 → the two unowned per-FIELD state consumers moved off a comment on the issue this PR CLOSES into go-to-k/cdkd#3211 (with the fuller site list the review found), and go-to-k/cdkd#3203's body corrected.
- [x] **Mutation table** — measured on the final tree, one mutation per probe over the whole `tests/unit/`, hash-verified restore, every row's count equal to its named-case count (checked mechanically), and re-measured unchanged after round 3:

  | mutation | cases red |
  | --- | --- |
  | previousState object-check dropped | 4 |
  | previousState: array arm dropped (an array passes as an object) | 1 |
  | previousState.physicalId check dropped | 1 |
  | previousState.resourceType check dropped | 1 |
  | previousState.properties check dropped (a STRING bag reaches the provider) | 1 |
  | previousState.properties: array arm dropped | 1 |
  | attemptedProperties check dropped | 1 |
  | oldResourceRetained check dropped (a truthy "no" takes the readopt arm) | 1 |
  | TYPE-when-present -> REQUIRED (refuses a journal cdkd wrote from a tolerated record) | 1 |
  | op.properties check dropped (delete-path twin of attemptedProperties) | 1 |
  | attemptedProperties: null / array arms dropped | 1 |
  | previousState.properties: null arm dropped | 1 |
  | provisionedBy FORWARDING dropped (parser deletes an unrecognised value) | 1 |
  | provisionedBy refused too (would lock rollback out of a journal cdkd wrote) | 1 |

  - **previousState object-check dropped** (4): `refuses a previousState that is a number (issue go-to-k/cdkd#3149)`; `refuses a previousState that is a string (issue go-to-k/cdkd#3149)`; `refuses a previousState that is an array (issue go-to-k/cdkd#3149)`; `refuses a previousState that is null (issue go-to-k/cdkd#3149)`
  - **previousState: array arm dropped (an array passes as an object)** (1): `refuses a previousState that is an array (issue go-to-k/cdkd#3149)`
  - **previousState.physicalId check dropped** (1): `refuses the nested fields the executor dereferences, each TYPE-when-present`
  - **previousState.resourceType check dropped** (1): `refuses the nested fields the executor dereferences, each TYPE-when-present`
  - **previousState.properties check dropped (a STRING bag reaches the provider)** (1): `refuses the nested fields the executor dereferences, each TYPE-when-present`
  - **previousState.properties: array arm dropped** (1): `refuses the nested fields the executor dereferences, each TYPE-when-present`
  - **attemptedProperties check dropped** (1): `refuses the nested fields the executor dereferences, each TYPE-when-present`
  - **oldResourceRetained check dropped (a truthy "no" takes the readopt arm)** (1): `refuses the nested fields the executor dereferences, each TYPE-when-present`
  - **TYPE-when-present -> REQUIRED (refuses a journal cdkd wrote from a tolerated record)** (1): `TOLERATES what the state boundary tolerates: an absent nested field, and any provisionedBy`
  - **op.properties check dropped (delete-path twin of attemptedProperties)** (1): `refuses the nested fields the executor dereferences, each TYPE-when-present`
  - **attemptedProperties: null / array arms dropped** (1): `refuses the nested fields the executor dereferences, each TYPE-when-present`
  - **previousState.properties: null arm dropped** (1): `refuses the nested fields the executor dereferences, each TYPE-when-present`
  - **provisionedBy FORWARDING dropped (parser deletes an unrecognised value)** (1): `TOLERATES what the state boundary tolerates: an absent nested field, and any provisionedBy`
  - **provisionedBy refused too (would lock rollback out of a journal cdkd wrote)** (1): `TOLERATES what the state boundary tolerates: an absent nested field, and any provisionedBy`
  </details>

- [x] **Live-tested against the BUILT bundle and the real state bucket**: journal with `previousState: "x"` planted at `cdkd/Cdkd3149Probe/us-east-1/` (empty `resources`, so no AWS resource is touched). main's binary replayed it and exited 2 with `Cannot restore Victim — resource no longer in state, skipping`; this branch refuses with `Rollback journal for 'Cdkd3149Probe' is malformed: segments[0].operations[0].previousState must be an object when present (got string). Remove the stack's rollback-journal.json (next to its state.json) to discard it.` No lock left behind (`lock.json` 404). The post-fix arm was RE-PLANTED first — the pre-fix run popped the segment, which would otherwise have made the second arm read "Nothing to roll back" and pass vacuously. Probe swept with `s3_purge_prefix_versions ... all`; `s3_assert_versions_swept` = 0.
- [x] **Integration**: n/a — the diff is `src/types/**` plus its test, a changelog entry and a rules note; outside every integ gate's scope (`markgate verify integ-destroy` / `integ-broad` both fresh, unchanged by this branch).
- [x] **Documentation** — `changelog.d/` entry (1868 chars); the `.claude/rules/layout-state-types.md` clause now states the journal-vs-state asymmetry and the `provisionedBy` exemption.

https://claude.ai/code/session_01RfVFucBDnVV62aVkC1LycW


go-to-k:
Merged as dece0b76 — thanks. Three review rounds, and the one that mattered most was the round-1 catch that my own `provider.update(.., undefined)` claim was false: every arm coalesces, so the harm was an empty desired bag, which is worse. That correction produced go-to-k/cdkd#3203, and the round-2/3 passes moved two more findings (go-to-k/cdkd#3211) off a comment that this merge would have buried.

PR #3231, opened 2026-09-16
fix(deployment): retry a Lambda CapacityProvider create on the security-group and ec2:DescribeSecurityGroups propagation wordings

### LINKED ISSUE #3227: fix(deployment): a Lambda CapacityProvider create fails on a transient "One or more security group IDs are invalid" that no propagation pattern matches, so it is never retried
## What is wrong

Creating an `AWS::Lambda::CapacityProvider` alongside the security group it references fails the deploy:

```
Failed to create Provider2281708E: CREATE failed for Provider2281708E: One or more security group IDs are invalid.
Check that the IDs are correct and try again. (Service: Lambda, Status Code: 400)
```

The security group is valid. A second `cdkd deploy` over the same resources, minutes later, creates the provider with the same `VpcConfig` and succeeds.

No entry in `IAM_PROPAGATION_ERROR_MESSAGE_PATTERNS` (`src/deployment/retryable-errors.ts`) matches this text, so `isIamPropagationError` and `isRetryableTransientError` both answer false for it and the create is not retried. Checked by calling both functions on the logged message, not by reading the list. Two shapes were observed:

- the message is the FIRST error: no retry at all, the deploy fails on the single attempt;
- the message is the SECOND error, after `The operator role is invalid or doesn't have sufficient permissions`, which go-to-k/cdkd#3182 added to the list while fixing go-to-k/cdkd#3174 and which the same check confirms matches: cdkd retries once, hits this message, and stops with `Provider2281708E: gave up after 1 IAM-propagation retry over 0.25s of propagation backoff`.

Reproduced on three fresh deploys of the same stack (cdkd 0.290.11, aws-cdk-lib 2.269.0, ap-northeast-1, 2026-09-16), each from an empty state.

## What the probes show, and what they do not

The operator-role entry go-to-k/cdkd#3182 added covers the first window of a fresh role. Probing `aws cloudcontrol create-resource` directly with an operator role created seconds earlier, the error changes before the create succeeds (two roles, same order, success 13 s and 14 s after the role was created):

| Attempt | Result |
| --- | --- |
| 1 | `InvalidRequest` -- `The operator role is invalid or doesn't have sufficient permissions` |
| 2 | `InvalidRequest` -- same |
| 3 | `InvalidRequest` -- `The operator role <arn> doesn't have permission to perform ec2:DescribeSecurityGroups` |
| 4 | `SUCCESS` |

The third message names `ec2:DescribeSecurityGroups`, the call that reads the security group, and it matches no pattern either.

What this does NOT establish is why the cdkd runs reported invalid security group IDs. Two possible explanations are that the operator role's permissions were still propagating, or that the just-created security group was not yet visible to Lambda; the measurements do not distinguish between them. The exact wording was not reproduced outside cdkd in six probe shapes -- a fresh security group in the default VPC with an aged role; a fresh VPC, subnets and security group with an aged role; a fresh role with an older security group (two roles); a fresh role and security group created together; VPC, subnets, security group and role all created together; and a create of a capacity provider name deleted three seconds earlier. Every one either succeeded or failed with a role wording above. In the failing deploys the capacity provider create started 0.19 s after the security group was created (run A) and 1.84 s after it (run B). That says the create followed the security group closely; it does not measure how long the failing condition lasted. So the probes leave a propagation-shaped explanation open rather than ruling it out.

## Timeline from one failing run (`--verbose`)

```
07:18:17.486 EC2Provider    Successfully created SecurityGroup LmiSgFC1639E4: sg-0c8bb9bca18902823
07:18:17.676 CloudControl   Creating resource Provider2281708E (AWS::Lambda::CapacityProvider)
07:18:23.485 CloudControl   CREATE Provider2281708E: FAILED            <- operator role is invalid ...
07:18:24.592 DeployEngine   Retrying Provider2281708E in 0.25s (attempt 1/26)
07:18:28.569 DeployEngine   gave up after 1 IAM-propagation retry over 0.25s of propagation backoff
                            - CREATE failed ...: One or more security group IDs are invalid.
```

The stack is the one from go-to-k/cdkd#3174: a VPC with two isolated subnets, a security group, an operator role, a capacity provider and a function attached to it, all created in the same deploy.

## Proposed fix

Add the two unmatched wordings to the propagation patterns in `src/deployment/retryable-errors.ts`, beside the operator-role entry go-to-k/cdkd#3182 added:

- `One or more security group IDs are invalid`;
- `doesn't have permission to perform ec2:DescribeSecurityGroups` -- the third-stage wording (`not authorized to perform` does not match `doesn't have permission to perform`). The narrower spelling is deliberate: `doesn't have permission to perform ec2:` would also cover every other EC2 action, and only this one was observed.

`isIamPropagationError` is a substring test over the whole message, with no scoping to Lambda, to this resource type or to the HTTP status, so both entries apply to any message carrying the clause. That is the trade-off the file already accepts for `Invalid IAM Instance Profile` and the Glue entries, and it cuts both ways here: a genuinely wrong security group id, or a role genuinely missing `ec2:DescribeSecurityGroups`, would spend the bounded retries before surfacing. Both cases belong in the comment beside the entries, the way the existing entries state theirs.

Worth settling in the same pass: whether the dense grid's budget covers the window. The probes reached a successful create 13 s and 14 s after the role was created, well inside the ~47.75 s the comments quote, but the window for the security group wording was never measured. Run A is not evidence either way -- it stopped after one retry and 0.25 s of propagation backoff because the next error was unmatched, not because a budget ran out.

## Verification plan

- Unit: `isIamPropagationError` returns true for both messages (`tests/unit/deployment/retryable-errors.test.ts`), with a control that an unrelated message does not match.
- Unit: a create that fails with the security group wording on the first attempt is retried; a role wording followed by the security group wording keeps retrying instead of giving up at one; and a persisting match still terminates on the budget.
- Integ: `tests/integration/lambda-capacity-provider-default-name`, added by go-to-k/cdkd#3182, builds the same shape -- a VPC with `natGateways: 0`, a security group, capacity providers with no `capacityProviderName` and a function attached -- so it needs a run rather than a new fixture. Since the failure is intermittent, a single green run does not prove the retry engaged; the run should be read for the retry line, or the classifier cases relied on for the deterministic part. It launches EC2 through the capacity provider, so the run carries a real charge.

Dup-check: searched open and closed issues for `security group IDs are invalid`, `security group propagation`, `CapacityProvider`, `IAM_PROPAGATION_ERROR_MESSAGE_PATTERNS` and `propagation retry`, and read `IAM_PROPAGATION_ERROR_MESSAGE_PATTERNS` in `src/deployment/retryable-errors.ts` for an existing entry. By title, go-to-k/cdkd#805 (closed), go-to-k/cdkd#2901 (closed) and go-to-k/cdkd#2018 (open) are the neighbouring propagation-retry cases, all about IAM-role wordings; none names a security group wording, and the classifier check above confirms no pattern in the file matches this text.
Session-fit: context test -- `src/deployment/retryable-errors.ts`: read, including the pattern list, `isIamPropagationError`, `retryClassificationText` and the rationale comments, and both classifiers were executed against the logged messages; `src/deployment/retry.ts`: read around the classification branch and the give-up summary; `src/deployment/deploy-engine.ts`: read around `withRetry` and the Cloud Control create path; `tests/integration/lambda-capacity-provider-default-name/lib/*.ts`: read; `tests/unit/deployment/retryable-errors.test.ts` and `tests/unit/deployment/retry.test.ts`: not read. Decision: now (do it in this session) -- the implementation paths the fix touches are read, the unit tests still need inspection, the real-AWS reproduction and the probe sequence exist only in this session, and the reporter is taking the fix.
Severity: high -- the first deploy of a Lambda Managed Instances stack failed on all three attempts from an empty state, which is normal operation for anyone adopting the feature; it self-heals only because the user re-runs the deploy.
Effort: medium (M) -- pattern entries plus unit cases, and a run of the existing capacity-provider integ fixture.
Estimate: ~2 h -- most of it the fixture run and confirming the retry budget covers the observed window.

nix-tkobayashi:
Closes #3227

## What was wrong

A Lambda Managed Instances stack whose capacity provider is created alongside its security group and operator role failed its FIRST deploy:

```
CREATE failed for Provider2281708E: One or more security group IDs are invalid.
Check that the IDs are correct and try again. (Service: Lambda, Status Code: 400)
```

A re-deploy minutes later succeeded with the same `VpcConfig`. The create is issued within two seconds of both resources, and two more wordings of that race matched no `IAM_PROPAGATION_ERROR_MESSAGE_PATTERNS` entry -- verified by calling `isIamPropagationError` and `isRetryableTransientError` on both, not by reading the list -- so neither was retried.

The shape that actually broke deploys was a SEQUENCE. The operator-role wording go-to-k/cdkd#3182 made retryable matched, cdkd retried, the next attempt returned the security-group wording, and the loop reported `gave up after 1 IAM-propagation retry over 0.25s of propagation backoff` -- which reads like an exhausted budget and was a budget that never opened.

## The fix

`src/deployment/retryable-errors.ts` -- two IAM-propagation entries beside the go-to-k/cdkd#3182 one:

- `One or more security group IDs are invalid`
- `doesn't have permission to perform ec2:DescribeSecurityGroups` -- which the list's `not authorized to perform` does not cover (the handler writes "doesn't have permission to perform"). Narrow to the ONE action observed; `doesn't have permission to perform ec2:` would cover every other EC2 action, none of them measured.

Both ride the dense 26-retry / 47.75s grid. A probe against a seconds-old role reached a successful create 13s and 14s after the role was created, inside that budget. **The cause is not established and the entries do not claim one**: the operator role's permissions still propagating and the security group not yet visible to Lambda both fit, and the wording did not reproduce outside cdkd in six probe shapes.

Neither entry is scoped to Lambda or a status code, so a genuinely wrong security group id, or a role genuinely missing the permission, spends the bounded retries before surfacing -- the trade this table already makes for `Invalid IAM Instance Profile`. That trade is pinned by a unit case rather than left implicit.

## The fence from go-to-k/cdkd#3182 fired, and its rule was too narrow

The shared list is also consumed by the Custom Resource provider's `isTransientAuthzThrow`, which replays a Lambda `Invoke` PRE-DELIVERY. go-to-k/cdkd#3182 pinned that list as a literal set precisely so a new entry cannot arrive unexamined -- and adding these two redded it, as designed.

Walking its admission rule surfaced a real gap. The rule admitted a non-front-door entry only if it provably could not reach a CR `Invoke` failure. The security-group wording cannot be shown unreachable: `@aws-sdk/client-lambda` carries an `InvalidSecurityGroupIDException` for `Invoke`, and its service-supplied message is not in the bundle, so a collision could not be ruled out by measurement.

**The rule is generalised rather than bypassed.** Front-door-ness is now judged ON THIS PATH -- a wording that is post-acceptance where it originates can only ever reach a CR `Invoke` failure as something else -- and an entry is admissible if on this path it is:

- **(a) unreachable** -- no CR `Invoke` / `Publish` failure can carry the text; or
- **(b) front-door when reached** -- the only rejection on this path that could carry it is itself decided before the handler can run.

Both replayed calls are in scope -- `Invoke` AND `Publish` -- so an argument about `Invoke` alone does not admit an entry, and an entry may rest on a different arm for each.

The argument must also cover the WHOLE matched substring, not the one message that prompted the entry. That clause is what corrected the `ec2:` entry's first argument, which leaned on the observed message naming an operator role -- a word the pattern itself does not contain.

Both new entries rest on (b) for `Invoke` -- `EC2AccessDeniedException` and `InvalidSecurityGroupIDException` are raised while Lambda sets up the function's VPC networking, during setup and before the handler can run, so a replay re-invokes nothing -- and on (a) for `Publish`, which carries no Lambda VPC-networking text. **Stated precisely**: that rests on the documented meaning of those two exception classes, not on a message observed here -- the same standing the existing go-to-k/cdkd#3174 and go-to-k/cdkd#805 arguments have, and the ledger case says in-line that it cannot pin it.

The rule is carried at the four sites it lives: the fence's header, its ledger (each entry now tagged (a) or (b)), `isTransientAuthzThrow`'s doc, and `.claude/rules/provider-custom-resources.md`.

The same analysis moved one more rationale. The pre-delivery arm skips the exec-env recycle, and three sites justified that with "the denial is on cdkd's own principal, not the function's role". That is false for the new EC2 case, which is about the function's OWN role. The general reason holds for both -- the rejection is decided before the handler can run, so no environment was ever created and there is nothing to recycle -- and every site now says that, keeping the principal explanation as the common case rather than the reason.

## Tests

- `tests/unit/deployment/retryable-errors.test.ts` -- both wordings classify as IAM propagation and retryable; each is matched by exactly its own entry; `not authorized to perform` does not cover the `ec2:` wording; the same phrasing for a different EC2 action does not match; a genuinely wrong id DOES match (the stated trade); an unrelated rejection from the same handler stays terminal. **Each negative case carries a positive witness in the same case**, so it cannot pass on an empty table -- measured: deleting the `ec2:` entry reds all three negatives, deleting both reds the control.
- `tests/unit/deployment/retry.test.ts` -- the reported SEQUENCE driven through the real `withRetry`: the issue's four-stage probe (role, role, `ec2:DescribeSecurityGroups`, SUCCESS) with the verbatim live wording; the security-group wording on the first attempt is retried; the role wording followed by the security-group wording keeps retrying instead of giving up at one; a persisting match still terminates on the dense budget. That last case asserts the op call count and the TOTAL slept ms derived from the grid constants, not `sleeps.length` -- the interruptible backoff slices one delay into several `sleep` calls (measured: 49 calls for 26 retries, 47,750 ms), so a call count would fence the slicing rather than the budget.
- `tests/unit/provisioning/custom-resource-provider-thrown-retry.test.ts` -- `PINNED` gains both entries; the ledger grows to four, each tagged with its argument.

Mutation probes, one at a time, tree restored byte-exact between them:

- deleting the security-group entry reds 9 of the 296 cases in the three files: three retry-engine cases, both fence cases, and four classifier cases (one of them the control, through its positive witness);
- deleting the `ec2:` entry reds 8 of 296: the four-stage retry-engine case, both fence cases, and five classifier cases (three of them the negatives, through their positive witnesses).

Both counts were re-measured on the final, rebased tree (the three files hold 296 cases there; go-to-k/cdkd#3237 added cases to `retryable-errors.test.ts`).

## Verification

- `vp run check`, `vp run typecheck:test`, `vp run build`: clean. Unit suite on the final, rebased tree: 1031 files, 23870 passed, 1 skipped.
- Codex delta review: 2 rounds -- r1 NO ISSUES, r2 a full re-review of the final tree after the proxy rounds changed it, NO ISSUES. Maintainer-checklist proxy pass: 3 rounds -- r1 three findings (the `ec2:` argument covering less than its substring, the (b) argument contradicting the fence's documented rule plus a stale "RIGHT two", and two negative cases surviving inert mutants), r2 one finding (the no-recycle rationale's "whose principal" reason), r3 NO ISSUES.
- **Integ, read for the retry line rather than the verdict** -- the issue warned that a green run alone does not prove the retry engaged, and the first run bore that out:
  - `lambda-capacity-provider-default-name`, **us-east-1**: PASS 443s, 17 deleted / 0 errors. Retried ONLY the operator-role wording; neither new entry fired. A green run that proves nothing about this PR.
  - `lambda-capacity-provider-default-name`, **ap-northeast-1** (where the issue reproduced 3/3): PASS 510s, 17 deleted / 0 errors. **The `ec2:` entry fired live**: `SpareProvider8B33A338` got the operator-role wording on attempt 1 (retried by the go-to-k/cdkd#3182 entry), then `... doesn't have permission to perform ec2:DescribeSecurityGroups on the capacity provider ...` on attempt 2, **retried it** (`Retrying SpareProvider8B33A338 in 0.5s (attempt 2/26 ...)`), and created. Attempt 2 is exactly where a build without this entry gives up, which is the reported failure shape.
  - The security-group wording did not appear in these two runs; it did in the sixth run below.
  - Broad set, since `retryable-errors.ts` is in the cross-cutting scope: `lambda` PASS 109s (9 deleted / 0 errors) and `drift-revert-vpc` PASS 284s (21 deleted / 0 errors) -- the propagation-retry path and a security-group-bearing update path, the two a pattern over-matching elsewhere would disturb.
  - A fifth run, after the review round: its comment edits to `retryable-errors.ts` staled the `integ-destroy` digest, so the fixture was re-run in ap-northeast-1 rather than the marker reset by hand -- PASS 447s, 17 deleted / 0 errors, retried only the operator-role wording. The ledger row is that run; its note keeps the 510s run as the one where the `ec2:` entry fired.
  - **A sixth run, on the tree rebased onto `224f3d57`** (ap-northeast-1): PASS 430s, 17 deleted / 0 errors. **The security-group entry fired live, in the issue's fatal shape**: `SpareProvider8B33A338`'s create FAILED on its FIRST attempt with `One or more security group IDs are invalid` (11:14:24.469Z), was retried (`attempt 1/26`, 11:14:25.565Z) and reached SUCCESS at 11:14:28.430Z -- the single-attempt failure a build without the entry reports. The ledger row is this run.
  - All six runs: 0 surviving object versions after the sweep, 0 capacity providers or fixture roles left in either region.
- **One correction the live run forced.** The `ec2:` unit fixture had been reconstructed from the issue's abbreviated probe table while its comment claimed it was verbatim. The live wording carries a tail the table dropped (`on the capacity provider. Grant the operator role ec2:DescribeSecurityGroups permission and try again.`). The fixture is now verbatim from that run, and the narrowing case derives its other-action string from it. The pattern is a prefix of the sentence, so it matched either way -- the fix is to the claim, not the classifier.
- One observation, not changed here: the fixture's `operator-role propagation retries during phase 1` count greps only the operator-role wording, so it reported `1` for a run that retried twice. It is reported rather than asserted, and the retry line above is the stronger reading; widening that grep would mean another fixture re-run for a number nothing depends on.

## Review round 1 (maintainer review at `12b295ed`)

M0 (the real-AWS runs) and M1 (the changelog's live-coverage claim) were already carried by `573fecd6`, pushed while the review was being written; after the sixth run the changelog names both live occurrences precisely. M2: the admission rule reads the same at every site -- `Invoke` AND `Publish` in `isTransientAuthzThrow`'s doc, the fence header and `.claude/rules/provider-custom-resources.md`, which also gains the whole-matched-substring clause it had omitted. M3: the `ec2:` entry states its own cost -- a role genuinely lacking the permission, or an SCP / permissions-boundary deny, spends the full dense budget. M4: "decided before an execution environment is engaged" became "decided before the handler can run" at every no-recycle site and at the (b) arm, since the VPC rejections are raised DURING environment setup; the pre-existing front-door sentence about authorization at the API door keeps its wording, which is accurate for that class. m5: beside the entries, the list's five call sites, derived by grep -- `withRetry`'s retryability and cadence reads, the Custom Resource replay, `destroy-runner.ts`'s delete loop (~35s per resource), and `isTerminalDeleteFailure`, which reads the verdict inverted. m6: the four-stage retry-engine case above.

**The budget question, answered.** The dense grid is measured to cover the role and `ec2:` stages (the probe created 13s and 14s after the role; a live run went from the `ec2:` wording on attempt 2 straight to a create), and one live security-group occurrence, which created about 4s after its first rejection. That last one is a single sample: it shows the grid CAN cover the security-group window, not that it always does. The issue's "re-deploy minutes later succeeded" is an upper bound on the window -- the condition had cleared by then -- not evidence that the window is long. The changelog headline describes the retry rather than promising a first-try deploy.

Round-1 review: Codex delta NO ISSUES; maintainer-checklist proxy pass 2 rounds -- r1 a remaining "before any execution environment is engaged" in the rules file and a changelog headline promising a first-try deploy against its own body, r2 NO ISSUES.

After review round 1 the branch was rebased onto `224f3d57`: go-to-k/cdkd#3237 also edits `retryable-errors.ts` (the name-collision classifier, a disjoint part of the file) and the integ ledger conflicted on the `lambda` row, now owned by `main`'s later run. 

## Review round 2 (maintainer review at `81641dbd`)

Every round-1 item was verified as discharged. Two sentences remained, both comment-only. M5: `isTransientAuthzThrow`'s bold sentence said the union being mostly authorization / request-validation rejections is what makes replaying an `Invoke` safe; it now states the true rule -- every admitted rejection is decided before the handler can run -- with the front-door characterisation as the commonest way to meet it. M6: the claim that both delete-side readers are inert is split to what was checked -- the DynamoDB `isTerminalDeleteFailure` reader is genuinely inert, while for the type-generic `destroy-runner.ts` loop it is narrowed to no delete observed here. Rebuilding `dist` at `81641dbd` and at the final head and normalising comments and content-hashed chunk names, the bundled code is identical across all 8 output files, so the ledger's 11:21Z run covers the final runtime; the same normaliser detects a one-literal mutation of the bundle. The optional structured-ledger suggestion was not taken, and go-to-k/cdkd#3274 (the fixture's retry counter) is out of scope here.


go-to-k:
Reviewed at head `12b295ed`. 3-axis + security (344 LOC / 7 files → 1-reviewer by size, up-biased by `src/provisioning/providers/**` and by both touched `src/` files carrying 2–3 `fix:` in their last 3 commits).

The classifier change itself is sound and I have no objection to the two entries. Four independent checks came back clean on the parts that carry the most risk:

- Mutation counts re-measured from scratch, one mutant at a time: deleting the security-group entry reds 9/270, deleting the `ec2:` entry reds 7/270 — both match the body exactly, and no negative case survives its mutant. The positive-witness claim holds per case.
- The fence is genuinely two-way: appending an unexamined entry to the source list reds `custom-resource-provider-thrown-retry.test.ts:1018`.
- Argument (b) verified from `@aws-sdk/client-lambda`'s own error models rather than from the body — and it is stronger than claimed: every `Invoke` exception that could relay EC2 text (`EC2ThrottledException`, `EC2UnexpectedException`, `ENINotReady`, …) is documented init-time, so (b) survives even if the substring arrives via a sibling exception. No double-invoke, no double-create.
- No new persistence or display class for the `sg-` id / role ARN: the give-up warn goes through the per-resource masking logger, and the same message already reached `deployments/{runId}.jsonl` via the rethrow before this change.

## Required before merge

**M0 — two real-AWS runs, not one.**

`lambda-capacity-provider-default-name` is an acceptance item of #3227 itself (its Verification plan), not only a gate. The ledger's row for it is `2026-09-16T06:02:17Z`, which predates the issue (`07:56:14Z`) — that is #3182's run, against a tree without these two entries. So `Closes #3227` is not earned at this head. Please run it and read it for the retry line, as the body already commits to.

Separately, `src/deployment/retryable-errors.ts` matches `CROSS_CUTTING_REGEX` in `.claude/hooks/integ-broad-gate.sh`, so **`integ-broad` also blocks the merge and a narrow feature fixture does not satisfy it**. One of: `bench-cdk-sample` / `lambda` / `microservices` / `drift-revert` / `drift-revert-vpc` / `multi-stack-deps` / `multi-resource` / `remove-protection` / `export`.

**M1 — the changelog entry asserts coverage that does not exist yet.** `changelog.d/entries/2026-09-16-3227-…md` ends `Live coverage: tests/integration/lambda-capacity-provider-default-name/` while the body says `Integ: pending`. M0 resolves this; if the run does not happen, the sentence has to go. (Size is fine — 1954/2000.)

**M2 — the generalised admission rule does not read the same at its four sites.** This is an agent-instruction surface, so a weaker spelling propagates.

- `src/provisioning/providers/custom-resource-provider.ts:1992-1998` states both arms as "**`Invoke`** failure"; the fence header at `custom-resource-provider-thrown-retry.test.ts:1002-1005` states them as "`Invoke` **/ `Publish`**". `isTransientAuthzThrow` gates the SNS `Publish` replay too, so the code doc is the weaker bar — it would admit an entry on an Invoke-only argument. For these two entries `Publish` is trivially (a), but neither site says so. Please add the `Publish` clause to both.
- `.claude/rules/provider-custom-resources.md:23` carries (a)/(b) but **omits the whole-matched-substring clause** entirely — by the PR's own account that clause is the load-bearing new half, and it is what corrected the `ec2:` argument.

**M3 — the `ec2:` entry's comment is missing its half of the trade #3227 asked for.** The issue says both cases belong "in the comment beside the entries". `retryable-errors.ts:203-205` states the wrong-`sg-`id trade; `:177-186` never states that a role genuinely missing the permission — or an SCP / permission-boundary deny — burns the budget first. The body and the changelog say it; the code does not. One sentence at `:186`.

**M4 — reword the no-recycle universal at the three sites M2 already touches.** "decided BEFORE an execution environment is engaged" (`custom-resource-provider.ts:1064`, `:1783`, `:1993`) is weaker than the code it justifies: `InvalidSecurityGroupIDException` and `EC2AccessDeniedException` are raised *during* environment setup (ENI attach), not before it is engaged. The load-bearing clause — no environment was ever *created*, nothing warm to recycle, the handler never ran — is correct. "before the handler can run" says it literally. While there, `custom-resource-provider-thrown-retry.test.ts:327-329` is the fourth no-recycle site and still phrases it the old "the denial is on cdkd's principal" way; it is true of that fixture's message, but it is the last site stating the rule as a "whose permission" one.

## Optional (take or leave)

**m5 — the shared list has five consumers, not three.** The two the PR does not mention: `src/cli/commands/destroy-runner.ts:1500`, where a match on a DELETE costs ~35s per resource on its own 3-attempt grid; and `src/provisioning/providers/dynamodb-delete-budget.ts:706`, where `isTerminalDeleteFailure` uses the predicate **inverted** — a match makes a failure non-terminal and skips `compensateRemovedDeletionProtection`. Inert for these two wordings (a DynamoDB delete carries no security-group text), but a reader with inverted polarity is worth the one line the other three got.

**m6 — no retry-engine case for the `ec2:` entry.** Deleting it reds none of the three `retry.test.ts` cases; the issue's 4-stage probe (role, role, `ec2:`, SUCCESS) is never driven through `withRetry`, only the SG wording is.

## Not blocking this PR — please just record the answer

#3227 asked to settle "whether the dense grid's budget covers the window". The body's answer is the role-wording probe the issue already contained and already disqualified, and the one SG-side timing datapoint in evidence — "a re-deploy **minutes** later succeeded" — sits *outside* 47.75s, so the question has a counter-signal rather than just an absence. The M0 run read for the retry line is the instrument that closes it; if it comes back green without the retry engaging, say so rather than leaving it implied.

nix-tkobayashi:
Pushed as `330d9da1`. Every required item and both optional ones are in, and the budget question is answered below as you asked -- including the part the runs did not close.

One thing first, since you track process: this reply comes about 105 minutes after your review, and the first 45 of them were a miss on my side rather than work. My PR watcher armed AFTER your comment landed and took the latest comment id as its baseline, so it only ever looked for comments newer than yours, and I found the review by reading the thread directly. The rest went into the fixes below, their review rounds, one real-AWS re-run, and three full-suite attempts. The fix on my side is to read what already exists before arming a watcher.

## M0 -- done, and it overlapped with your review rather than following it

`573fecd6`, pushed while this review was being written, already carried the runs. They postdate the issue, and they are against the tree with both entries:

- **`lambda-capacity-provider-default-name`, ap-northeast-1** (where the issue reproduced 3/3): PASS 510s, 17 deleted / 0 errors. **The `ec2:` entry fired live**: `SpareProvider8B33A338` got the operator-role wording on attempt 1, then `... doesn't have permission to perform ec2:DescribeSecurityGroups on the capacity provider ...` on attempt 2, logged `Retrying SpareProvider8B33A338 in 0.5s (attempt 2/26 ...)`, and created. Attempt 2 is where a build without the entry gives up.
- **The same fixture, us-east-1**: PASS 443s, but it retried ONLY the operator-role wording. Neither new entry fired. I am listing it because a green run that proves nothing is exactly what your M0 warns about.
- **Broad set**, since `retryable-errors.ts` matches `CROSS_CUTTING_REGEX`: `lambda` PASS 109s (9 deleted / 0 errors) and `drift-revert-vpc` PASS 284s (21 deleted / 0 errors) -- the propagation-retry path and a security-group-bearing update path, the two a pattern over-matching elsewhere would disturb.

The ledger rows are those runs. One more, on this head: the comment edits in this round touched `retryable-errors.ts`, which `.markgate.yml` lists in the `integ-destroy` digest, so the marker went stale with no behaviour change. I re-ran the fixture in ap-northeast-1 rather than resetting it by hand -- PASS 447s, 17 deleted / 0 errors, 0 surviving versions. **That run retried only the operator-role wording.** So the row now reads 447s, and its note keeps the earlier 08:50Z run as the one where the `ec2:` entry fired.

## M1 -- resolved by M0, and made precise

The live run exercised the `ec2:` entry, not the security-group one, so "Live coverage" alone would have overstated it. The entry now reads that the `ec2:` wording was retried on attempt 2 in ap-northeast-1 and that **the security-group wording did not occur in any run and is unit-covered only**. Its headline also stopped promising a first-try deploy -- see the budget answer at the end for why.

## M2 -- the rule reads the same everywhere

- `isTransientAuthzThrow`'s doc now states both arms for `Invoke` **and** `Publish`, and says outright that an argument about `Invoke` alone does not admit an entry. You were right that it was the weaker bar.
- For these two entries it names the arm per call: (b) for `Invoke`, (a) for `Publish`, since SNS carries no Lambda VPC-networking text. The ledger says the same per entry.
- `.claude/rules/provider-custom-resources.md` gains `Publish` and the whole-matched-substring clause. That clause was the half that corrected the `ec2:` argument, and it had been missing from the one surface other sessions read first.

## M3 -- done

Beside the `ec2:` entry: an operator role genuinely lacking `ec2:DescribeSecurityGroups`, or a deny from an SCP or a permissions boundary, spends the full dense budget (~47.75s) before the create surfaces its real error.

## M4 -- done at every site, with one deliberate exception

"Before the handler can run" is now the wording at both `custom-resource-provider.ts` no-recycle sites, the (b) arm in its `isTransientAuthzThrow` doc, the fence header and ledger, two sentences in the rules file, and the fourth no-recycle site you found -- the test comment still phrased as "the denial is on cdkd's principal". That comment now says the principal is true of *that fixture's* message but is not the rule.

**The exception:** the pre-existing front-door sentence about authorization at the API door keeps "before any execution environment is engaged". For that class it is accurate, and the ENI distinction you drew does not apply to it.

## m5 -- taken, and derived rather than copied

Before writing the count I grepped for it. There are **five** call sites, not four modules. `withRetry` reads the list twice, for two different jobs: retryability at `retry.ts:380`, and cadence at `:381`, where `isIamPropagationError` picks the dense grid. The comment beside the entries enumerates all five.

I checked both of your figures in the code:

- `destroy-runner.ts` retries up to 3 times at `5000 * 2^attempt`, so 5s + 10s + 20s = 35s.
- `isTerminalDeleteFailure` returns `!(isRetryableTransientError(...) ...)`, and the negation feeds the `'not-applicable'` early return that skips `compensateRemovedDeletionProtection`.

## m6 -- taken

`retry.test.ts` now drives role, role, `ec2:DescribeSecurityGroups`, SUCCESS through the real `withRetry`, with the verbatim live wording. Measured: deleting the `ec2:` entry reds exactly that case, where before it redded none of the retry-engine cases.

## On argument (b)

Thank you for checking it against the SDK's error models rather than the body. I had argued it only for the two exception classes I named. Your finding, that every `Invoke` exception able to relay EC2 text is documented as init-time, makes it hold against a sibling exception too. That is a stronger argument than the one I wrote, and it does not depend on which exception carries the substring.

## The budget question -- the answer, including what stays open

**Covered, and measured:**
- The role stage and the `ec2:` stage. The issue's probe created 13s and 14s after the role.
- The live ap-northeast-1 run moved from the `ec2:` wording on attempt 2 straight to a successful create.
- Both sit well inside 47.75s.

**Not covered, and not closed by the M0 runs:** the security-group window. That wording did not occur in any of the five runs, so the instrument you named never observed it. The only timing evidence for it remains the one you pointed at -- a re-deploy **minutes** later succeeding -- and that sits outside the budget. So this PR retries that wording, but it does not establish that the dense grid outlasts the condition behind it.

The changelog now says so, rather than letting "deploys on the first try" imply it. If a future run shows the security-group rejection outlasting the grid, the remedy is a longer schedule for that one wording, not a wider pattern.

## Verification

- Codex delta review of this round: NO ISSUES. Maintainer-checklist proxy pass: 2 rounds.
  - r1 found two items, both fixed: a remaining "before any execution environment is engaged" in the rules file, and the changelog headline promising a first-try deploy against its own body.
  - r2: NO ISSUES.
- Unit suite on the final tree: 1028 files, 23637 passed, 1 skipped. Two earlier attempts failed only on `Test timed out in 5000ms` in `tests/unit/scripts` critics while other sessions' suites shared the machine; those cases pass in isolation, and the run above was taken with no peer suites running.

nix-tkobayashi:
Follow-up to my reply above. Two things changed since I posted it, and one of them answers the question I said the runs had left open.

## The branch was rebased

`#3237` landed on `main` and edits `retryable-errors.ts` (the name-collision classifier), and GitHub reported the PR `CONFLICTING`. The one real conflict was the integ ledger, where both sides had written the `lambda` row. The branch is flattened, rebased onto `224f3d57`, and normalized; `main`'s later `lambda` run now owns that row, which is correct for a last-run ledger. The two changes touch disjoint parts of that file, and the unit suite passes on the rebased tree: 1031 files, 23870 passed, 1 skipped. Head is now `81641dbd`.

## The security-group wording fired live

Because the rebase pulled a `retryable-errors.ts` change into the `integ-destroy` digest, I re-ran `lambda-capacity-provider-default-name` in ap-northeast-1 on the rebased tree. That run produced **the security-group wording for the first time in six runs**, in exactly the shape the issue reported as fatal:

```
11:14:24.469Z  CREATE SpareProvider8B33A338: FAILED        <- "One or more security group IDs are invalid"
11:14:25.565Z  Retrying SpareProvider8B33A338 in 0.25s (attempt 1/26) - ... One or more security group IDs are invalid ...
11:14:28.430Z  CREATE SpareProvider8B33A338: SUCCESS
```

It was the create's FIRST attempt, with no operator-role wording before it -- the "message is the first error: no retry at all" case from the issue. A build without the entry fails the deploy right there. PASS 430s, 17 deleted / 0 errors, 0 surviving object versions.

## What that does to the budget answer

My reply said the security-group window was unmeasured. That is no longer true, but only as far as one sample goes:

- **Measured:** this occurrence cleared about **4s** after the rejection, well inside 47.75s. So the dense grid CAN cover the security-group window.
- **Not established:** that it always does. One sample is not a bound, and the entry's comment and the changelog both say so.
- **On the "minutes later" re-deploy:** I called it a counter-signal, following your note, and I should correct my own reading. A re-deploy succeeding minutes later only shows the condition had cleared by then. It is an upper bound on the window, not evidence the window is long, so it does not contradict a 4s occurrence.

Two changelog sentences this run made false are corrected: "did not occur in any run", and "the security-group window is NOT measured". The entry's comment in `retryable-errors.ts` records the timing with its one-sample caveat, beside the matching line on the `ec2:` entry.

go-to-k:
Round 2, reviewed at head `81641dbd` (rebased onto `224f3d57`). Same tier as round 1 — 3-axis + security, 435 LOC / 8 files.

**Every round-1 item is discharged, and `Closes #3227` is earned at this head.** M0 (a) and (b), M1, M2, M3, M4, m5 and m6 all verified independently rather than taken from your reply:

- **M0** — `integ-last-run.tsv:158` at `2026-09-16T11:21:11Z` postdates the issue and runs on the rebased tree, and `drift-revert-vpc` at `09:00:51Z` is a broad-set row. I checked the timing objection myself: at 09:00:51Z the head was `12b295ed`/`573fecd6`, which round 1 read and confirmed carried both entries, and `integ-broad` is sentinel-bound rather than code-bound, so the rebase does not stale it. The `lambda` row correctly went to main's later run.
- **M2 / M4** — the rule now reads the same at every site, and your "deliberate exception" is right: the surviving `before any execution environment is engaged` sits on a pre-existing sentence scoped by "**Nearly** every pattern", describing the API-door class, where it is the *stronger* statement. Every site that states the rule says "before the handler can run". I had this one wrong as a blanket ask in round 1.
- **m6** — re-measured, not read: deleting the `ec2:` entry now reds exactly one retry-engine case (`retry.test.ts:42`), where round 1 measured zero. The wording in both files is the full live sentence, not a synthetic.
- **The rebase** — I verified your disjointness claim against #3237 directly. `isNameCollisionError` returns false for both new entries and for both full live wordings; `IAM_PROPAGATION_ERROR_MESSAGE_PATTERNS.filter(isNameCollisionError)` is empty across all 35 entries; ELBv2's collision text does not match `isIamPropagationError`; and `isRecreateRetryableError` is false for both, so neither destructive site (`--replace` delete-first, rollback delete-new-first) sees any change. Ledger resolution is clean — 307 rows both sides, sort invariant intact.
- **Argument (a) for `Publish`** — enumerated all 34 SNS exception classes rather than accepting the claim. `Publish` calls no EC2 API and has no security-group concept, and `publishToSns` wraps nothing, so there is no `.cause` laundering route. (a) holds.

Thank you for the process note about the watcher baseline, and for listing the us-east-1 run that proved nothing — that is exactly the right thing to surface rather than bank.

## Required before merge

Two sentences. Both are the same class as M4, which is why I am asking rather than leaving them.

**M5 — `src/provisioning/providers/custom-resource-provider.ts:1981-1984` still carries the rationale M4 removed everywhere else.** The bold sentence reads "Nearly every pattern in that union is an AUTHORIZATION or REQUEST-VALIDATION rejection, **and that is what makes replaying an `Invoke` safe here**". The *characterisation* is fine and I withdrew my round-1 objection to it above — but the causal clause attached to it is now false for this PR's own two entries, which are not front-door and are safe for the other reason. The paragraph two below states the real rule correctly, so nothing is wrong downstream; the problem is that the bold sentence is the one a hurried reader stops at, and it is the last surviving copy of the justification this PR spent a round removing. Reword the clause to the (b) form ("the handler provably did not run") so the emphasised sentence is the true one.

**M6 — `src/deployment/retryable-errors.ts:232`: "Both delete-side readers are inert for these wordings, which no delete carries" asserts more than the PR measured.** `destroy-runner.ts`'s loop is type-generic, and a Cloud Control DELETE of an `AWS::Lambda::CapacityProvider` plausibly re-validates the operator role. The consequence if it is wrong is only ~35s, and I am not asking you to measure it — just to say "no delete observed here carries" so the sentence claims what you checked. The `dynamodb-delete-budget.ts` half is genuinely inert and I verified it: `isTerminalDeleteFailure` classifies only the two DynamoDB providers' throws, and `DeleteTable` has no `ec2:` or security-group surface.

## Optional (take or leave)

**m7 — the per-call arm tags you added to the ledger are comments, so nothing reads them.** `custom-resource-provider-thrown-retry.test.ts:1088-1112` writes `Invoke`: (b) / `Publish`: (a) per entry, but the case's assertions are length, distinctness, membership and strict-subset — flipping (b)→(a) on either entry reds nothing. The test discloses this at `:1125-1130`, so it is not hidden. A structured ledger (`{ pattern, invoke, publish }`) would at least make tag *presence* mechanical. Not worth a round on its own; worth doing if you touch the file again.

## Out of scope for this PR — filed as #3274

The fixture-counter defect you recorded in the body is real and I confirmed it at `tests/integration/lambda-capacity-provider-default-name/verify.sh:282`: the count greps the operator-role wording alone, so a run that retried on either #3227 wording under-reports — including your 11:21Z run, whose security-group retry does not appear in it at all. It is an informational `echo`, so no gate is weakened, but it is wrong in the direction that hides work, on the fixture whose purpose is to exercise that engine. I filed it as #3274 rather than ask you to widen this PR, since the file is not in your diff. Please do not pick it up in this branch.

## One note on evidence, no action needed

The 08:50Z run where the `ec2:` entry fired was carried by `573fecd6`, which the flatten removed, and the current ledger row overwrote it. Nothing is hidden — the note and this thread preserve it — but that datapoint and the `~4s` timing now rest on your report rather than on anything reconstructible from the repo. The row itself carries the stronger datapoint, so this is a remark about the evidence class, not a request.

nix-tkobayashi:
Pushed as `afee5469`. Both required sentences are fixed. I'm not taking m7, for the reason you gave, and I'm leaving go-to-k/cdkd#3274 alone.

## M5 -- the bold sentence now states the true rule

`isTransientAuthzThrow`'s emphasised sentence now reads: replaying an `Invoke` is safe because every rejection this admits is decided before the handler can run, so the handler provably did not run. The front-door characterisation stays, but as the commonest way to meet that bar rather than the reason. So a reader who stops at the bold sentence now stops at the correct rule. Nothing below it needed to change.

## M6 -- the comment now claims only what was checked

I split the one sentence into two claims:

- **`isTerminalDeleteFailure`** is stated as genuinely inert, with the reason you verified: it classifies only the two DynamoDB providers' throws, and `DeleteTable` has no EC2 or security-group surface. It stays flagged as the reader a future entry is likeliest to get wrong.
- **The `destroy-runner.ts` loop** is type-generic, so there the claim is narrowed to no delete *observed here*. The comment names the case that could match -- a Cloud Control DELETE of a capacity provider that re-validates its operator role -- and bounds the cost at up to ~35s of backoff, only if the rejection persisted through every attempt. An earlier draft said such a delete "would" match and "would cost ~35s", and review caught that overstatement.

## m7 -- not taken

As you said, this isn't worth a round on its own, and this change doesn't touch that test file. The in-line disclosure at `:1125-1130` still says what the ledger does not pin.

## go-to-k/cdkd#3274

Understood. I haven't picked it up on this branch, and won't.

## On the `integ-destroy` gate at this head

Both edits are comment-only, but `retryable-errors.ts` is in the `integ-destroy` digest. So the marker is stale for a change that cannot alter behaviour. I didn't reset it by hand, and I didn't re-run the fixture. A re-run would add little: the failure is intermittent, so a run may not exercise either entry at all. Instead I measured something stronger:

- I rebuilt `dist` at `81641dbd` and at this head.
- I normalised comments and content-hashed chunk names in both.
- **The bundled code is identical across all 8 output files.**
- To check the comparison can fail, I mutated one pattern literal in the bundle, and the same normaliser detected it.

So the runtime this head ships is exactly the one the 11:21Z row ran. If your merge flow wants a fresh row regardless, say so and I'll run it.

## Verification

- Codex review of this round: 2 rounds.
  - r1 found the M6 overstatement described above.
  - r2: NO ISSUES.
- Unit suite: 1031 files, 23870 passed, 1 skipped.

go-to-k:
Round 3, reviewed at `afee5469`. **M5 and M6 are both discharged, and I have nothing further to ask for on this PR.**

- **M5** — `custom-resource-provider.ts:1980-1986`. The emphasised sentence now leads with the general rule ("decided before the handler can run, so the handler provably did not run") and demotes front-door-ness to "the commonest way to satisfy that, but not the only one (see below)". That is the right shape: a reader who stops at the bold text now stops at the rule that actually covers this PR's entries.
- **M6** — `retryable-errors.ts:232-239`. Split as asked, and the `destroy-runner.ts` half is better than what I asked for: "no delete OBSERVED here carries these wordings", the case that could match named, and the cost bounded conditionally ("if the rejection persisted through every attempt") rather than asserted. Your Codex round catching the earlier "would match / would cost" overstatement is the same correction I would have made.
- **m7** — agreed, not taken. It was offered take-or-leave and this change does not touch that file; the in-line disclosure at `:1125-1130` is enough.

## Your `integ-destroy` question — I'll run it, nothing needed from you

Your bundle-identity measurement answers the question it set out to: comments normalised, chunk hashes normalised, all 8 outputs identical, and a mutated pattern literal proving the comparison can fail. I accept that the runtime at this head is the one the 11:21Z row ran, and I agree a re-run is low-information here — the race is intermittent, so a fresh run may exercise neither entry, which is exactly what your us-east-1 run showed.

The re-run is still needed, but for a mechanical reason rather than an evidential one: markgate markers are per-worktree, so a run recorded in YOUR worktree does not satisfy the gate in the one I merge from. Whoever merges has to hold the marker. So I will take the run rather than hand it back to you — please do not spend a second one on it.

Sequencing, so you know what to expect: I have a small lane of my own in flight (the go-to-k/cdkd#3274 fixture fix) that owns my worktree right now. Once it merges I will put this head in that tree, run `lambda-capacity-provider-default-name`, bind the review marker to `afee5469`, and merge. No further pushes needed from you — and if you do push, the review marker unbinds and this round re-opens, so please hold unless something is actually wrong.

nix-tkobayashi:
Understood -- holding. No further pushes on this branch, and I will not run the fixture again; the head stays at `afee5469` for your run and merge.

go-to-k:
Merged as `7c9fa40f` — thank you, this was a genuinely good PR to review.

Three things stood out across the rounds. You surfaced the us-east-1 run that retried only the operator-role wording and said in as many words that it proved nothing — volunteering the evidence that weakens your own case is the habit that makes a review cheap, and it is rare. You corrected your own reading of the "minutes later" re-deploy as a counter-signal before I had to. And when the rebase pulled go-to-k/cdkd#3237 into the same file, you stated the disjointness as a claim to be checked rather than an assurance, which is exactly how I was able to check it quickly.

The bundle-identity measurement at the end deserves its own mention: normalising comments and content-hashed chunk names across all 8 outputs, then mutating a pattern literal to prove the comparison could fail, is a stronger answer than the re-run it replaced. I still took the run, but only because markgate markers are per-worktree, not because the argument was short of anything.

Two corrections you made were better than what I asked for. On M6 I asked you to soften one sentence; you split it into two claims and bounded the `destroy-runner` cost conditionally, and your own review round caught the "would match / would cost" overstatement in between. On M5 the demotion of front-door-ness to "the commonest way to satisfy that, but not the only one" is the right shape — I had asked for a reword and you found the structure.

For the record, my round-1 M4 was wrong as a blanket ask: your retained "before any execution environment is engaged" on the pre-existing API-door sentence was correct, and you were right to push back with a reason rather than just applying it.

go-to-k/cdkd#3274 stays mine, as agreed.

PR #3243, opened 2026-09-16
fix(diff): name the cycle when a nested template's asset path closes one, instead of failing 190 levels later on an S3 key

### LINKED ISSUE #3239: fix(diff): cdkd diff --recursive reports a key-length error instead of the cycle when a nested template's relative aws:asset:path closes one
> **Corrected after a live measurement.** The first version of this body said the walk "recurses until the call stack is exhausted" and rated it one level more severe than the classification line below now does. That is wrong: the template arm IS bounded, by the same S3 key limit as the state arm, and the real cost is a misleading error rather than a crash. Measured against the pre-fix binary and rewritten below.
>
> (The prior rating is deliberately not quoted here. `scripts/check-issue-classification-labels.ts` takes the FIRST `Severity:`-plus-token match anywhere in the body, so quoting a superseded value outranks the real classification line and the CI check reports the stale one as though it were current.)

## What happens

`cdkd diff --recursive` walks nested stacks through a TEMPLATE arm that follows each child's `Metadata['aws:asset:path']` with no visited set and no self-reference check. A relative path that points back at its own template — directly, or around a longer cycle — makes `buildDiffTree` follow the cycle.

`indexNestedChildTemplates` (`src/cli/commands/diff-recursive.ts:240-254`) resolves each child as `path.join(dir, assetPath)` and refuses only ABSOLUTE paths:

```ts
if (isAbsoluteCrossPlatform(assetPath)) {
  throw new Error(
    `Nested stack '${logicalId}' has Metadata['aws:asset:path']='${assetPath}' which is ` +
      `absolute. CDK emits relative asset paths for nested templates; an absolute path ` +
      `indicates the synth output was hand-modified or generated by a non-CDK toolchain. ` +
      `Refusing to load.`
  );
}
result[logicalId] = path.join(dir, assetPath);
```

A relative path takes the second branch unconditionally. `grep -cn 'visited\|seenTemplates\|maxDepth\|MAX_.*DEPTH' src/cli/commands/diff-recursive.ts` returns **0**.

The refusal above already names the threat model: "the synth output was hand-modified or generated by a non-CDK toolchain". A hand-modified assembly is exactly the input that guard exists for, and the relative-path case walks straight past it.

## What it actually costs — measured, not inferred

Built binary, a two-file `cdk.out` whose child template declares a nested row resolving back to itself:

```text
StateError: Failed to get state for stack
'Cdkd3239Live~Child~Loop~Loop~ ... ~Loop' (us-east-1): Your key is too long
```

Around 190 levels, then it stops. **The walk does not run away**, because `buildDiffTree` calls `loadStateOrEmpty(stackName, ...)` at every node (`diff-recursive.ts:1203`) and `childStackName` grows by one `~<logicalId>` per level (`:1253`), so S3's 1024-byte key limit bounds the TEMPLATE arm exactly as it bounds the state-only one.

So the defect is a DIAGNOSIS defect, and the cost is:

- no diff printed at all;
- ~190 pointless S3 state reads before it gives up;
- an error naming a 190-segment stack that does not exist, which mentions neither the cycle nor the asset path that caused it.

A unit test using a fake backend DOES run away, because the fake enforces no key length. That divergence is a mock endorsing an assumption the wire does not — which is why the measurement above is a live run against the built binary.

## Relationship to go-to-k/cdkd#3155

Same family, different severity. go-to-k/cdkd#3155 capped `cdkd state list --tree` because a deep `parentStack` chain in a hand-edited state bucket genuinely exhausted the call stack and emptied the whole view. Here nothing is unbounded; the walk just cannot say why it stopped.

Surfaced during the review of go-to-k/cdkd#3210, while auditing that PR's claim about the sibling nested-stack walkers. That PR is not the place to fix it: it touches `src/cli/commands/state-list-tree.ts` and `src/cli/commands/state.ts` only.

## Proposed fix

Track the resolved absolute template paths of the ancestor chain — root-to-node, NOT a global visited set, so two sibling rows legitimately sharing one child template still diff — and refuse on a repeat, naming the logical id, the owning stack and the path that closed the cycle. A refusal rather than a cap: a cyclic assembly has no correct diff to render, and truncating one would under-report changes the next deploy would still make.

## Verification plan

- A unit case whose child `aws:asset:path` resolves back to the parent, asserting the refusal names the row.
- A case for a longer cycle (A -> B -> A), since a self-reference check alone would miss it.
- A case for a legitimate deep-but-acyclic chain, confirming it still diffs.
- A diamond case (two sibling rows naming one template), which a global visited set would wrongly refuse.
- A live run of the built binary on a cyclic assembly and on the same assembly with the cycle removed.

Dup-check: searched open and closed issues for `diff-recursive cycle`, `aws:asset:path recursion`, `indexNestedChildTemplates`, `nested template infinite recursion` and `diff --recursive depth`; no hit. go-to-k/cdkd#555 and go-to-k/cdkd#464 introduced the recursive nested-stack walks and do not cover cycles.
Session-fit: now (do it in this session) — `src/cli/commands/diff-recursive.ts` was read in this session while auditing go-to-k/cdkd#3210's sibling-walker paragraph, so the file is warm; nothing external blocks it.
Severity: low — `cdkd diff --recursive` prints no diff and reports a confusing key-length error on a hand-modified or non-CDK-generated assembly; nothing is unbounded, no wrong state is written, and the non-recursive `cdkd diff` is unaffected.
Effort: small (S) — an ancestor set threaded through one recursive walk plus unit cases; no integ fixture.
Estimate: ~1-2 h — choosing refusal-vs-cap, ancestor-chain-vs-global, and the cycle fixtures.

go-to-k:
Closes go-to-k/cdkd#3239

## What was wrong

`cdkd diff --recursive` locates each nested child through its `Metadata['aws:asset:path']`. `indexNestedChildTemplates` refuses an ABSOLUTE path — and its message already names the threat model, "the synth output was hand-modified or generated by a non-CDK toolchain" — while a RELATIVE one is joined unconditionally. There was no visited set and no self-reference check, so a relative path resolving back onto the same nesting chain was followed.

## What that actually cost — measured, not inferred

**Not a crash.** The walk was already bounded: `buildDiffTree` reads state at every node (`src/cli/commands/diff-recursive.ts:1186` on this PR's base) against a `childStackName` that grows by one `~<childLogicalId>` per level (`:1236`), so S3's 1024-byte key limit stops the template arm exactly as it stops the state-only one.

Against the BUILT binary, on a two-file `cdk.out` whose child template declares a nested row resolving back to itself:

```text
StateError: Failed to get state for stack
'Cdkd3239Live~Child~Loop~Loop~ ... ~Loop' (us-east-1): Your key is too long
```

About 190 levels, exit 1, no diff printed. So the defect is a DIAGNOSIS defect: ~190 pointless state reads, then an error naming a stack that does not exist and mentioning neither the cycle nor the asset path.

This corrects the issue as originally filed, which claimed call-stack exhaustion and rated it one level more severe than it is. go-to-k/cdkd#3239's body now carries the measurement above, and its label was corrected to match. (The superseded rating is deliberately not quoted: the classification checker takes the FIRST such value anywhere in the text, so quoting a stale one outranks the real line.)

## What changed

`buildDiffTree` threads the resolved absolute template paths of the ancestor chain down the walk and REFUSES the first repeat, naming the logical id, the owning stack and the path that closed the cycle.

Three decisions, each stated because a reviewer would otherwise have to reconstruct it:

- **A refusal, not a depth cap**, for a reason local to this command: a cyclic assembly has no correct diff to render, so truncating the walk would under-report changes the next deploy would still make. A view whose contract is that every record appears would answer differently.
- **The ancestor chain, not a global visited set.** Two sibling rows may legitimately name the same child template; a global set would refuse that diamond as a cycle. Only a repeat along one root-to-node path is one, so nesting depth itself is never limited.
- **`path.resolve` is DEFENSIVE, not load-bearing.** Synth builds the root's entries with `join(assemblyDir, assetPath)` and every deeper path comes from this module's own `path.join`, so on production inputs a raw string comparison behaves identically. It resolves anyway to cover a future caller that hands `buildDiffTree` an unnormalized map, which would otherwise miss the first repeat and refuse one level late.

The STATE arm is untouched and owed nothing.

## Files

- `src/cli/commands/diff-recursive.ts` — the ancestor set, the refusal, and the comment recording the real bound
- `docs/cli-diff.md` — a `--recursive` subsection covering the refusal and why it is a refusal rather than a truncation, plus the exit-code list
- `tests/unit/cli/diff-recursive.test.ts` — five cases
- `changelog.d/entries/2026-09-16-3239-diff-recursive-template-cycle.md`

## Verification

**Live, with the BUILT binary.** The cyclic assembly above exits 1 with the refusal naming `Loop` under `Cdkd3239Live~Child`. The same assembly with the cycle removed diffs normally and walks the nested child (`2 to create` at the root, `1 to create` in `Cdkd3239Live~Child`) — the control that shows the refusal does not fire on well-formed input.

**Cases added:** a self-referencing child; a longer `a -> b -> a` cycle a self-reference check alone would miss; a non-canonical ROOT spelling, asserting WHICH node refuses; a 40-level acyclic chain that must still diff; and a diamond (two sibling rows naming one template) that must still diff.

**Mutation probes**, one mutation per probe, the tree restored byte-exact between them and each restore verified with `git diff HEAD`:

| Mutation | Result |
| --- | --- |
| refusal condition forced `false` | vitest worker runs away |
| `path.resolve(...)` -> the raw string | 1 red |
| ancestor chain -> a shared global set | 1 red (the diamond case) |
| drop the threading at the recursive call site | vitest worker runs away |
| blank the PATH interpolation in the refusal message | 1 red |

The two runaway rows are a UNIT-only symptom, and the test file's header says so: `fakeBackend` enforces no key length, so the mock endorses an assumption the wire does not. That is exactly why the failure mode quoted above comes from the binary rather than from a test.

**Two earlier versions of the path-normalization case were VACUOUS** and are not what shipped. The first spelled the cycle with `..`, which `path.join` already collapses. The second asserted only that a refusal happened — which a raw comparison still produces, one level later. Both measured green with `path.resolve` removed. The shipped case asserts which node refuses, which is the observable difference.

## Not done, deliberately

`.claude/rules/layout-cli.md` is not updated. Adding the note this change would warrant reds the rule-payload fence: a 669 B addition put two budgeted paths over their caps and took the `>4000 B` line count from 7 to 8. The measurement and the per-path headroom are recorded in go-to-k/cdkd#3245 — taken at `ea3070466`, before this branch was rebased, so treat the numbers there as of that base.

The existing text does not become false, only less complete: it describes `readNestedTemplate` / `indexNestedChildTemplates` as template-file loaders, which is still true. The behaviour is carried by `docs/cli-diff.md` and the changelog entry.

## Test plan

- [x] Unit tests pass (1033 files, 24100 tests) at `f4ed3a2e`, rebased onto `97206c84`
- [x] Live-tested against the built binary, both cyclic and acyclic
- [x] Documentation updated (`docs/cli-diff.md`, incl. the exit-code list)
- [ ] Integration test: none — `cdkd diff` performs no AWS mutation, and no integ gate's scope covers `src/cli/commands/diff-recursive.ts`


go-to-k:
Review round 1 at head `d9ebdf83` — three read-only reviewers (spec, code, test). **No blocker. One major, five minor/nit, all addressed in `a8854b3f`.**

The major is worth naming because it is the failure this PR's own test file warns about, reproduced inside the PR: **the refusal message's PATH was asserted by nothing.** A probe replacing that interpolation with a literal left all 145 cases green, because every case stopped at `resolves to nested template` and read no further — and the message IS the deliverable here, since the whole change converts a `Your key is too long` into a diagnosis. Case 1 now asserts the row, the resolved path and the diagnosis separately, and the same probe reds it.

Four of the others were FALSE CLAIMS in prose I wrote, each caught by a reviewer re-deriving rather than reading:

- The source comment and changelog said `cdkd state list --tree` "caps its depth". It does not, on this branch's base — that cap lives in the unmerged go-to-k/cdkd#3210. Reworded to stand on this command's own reasoning.
- The `path.resolve` justification said the root's paths "come from synth and are not normalized". Synth builds them with `join(assemblyDir, assetPath)`, so they are. `path.resolve` is defensive, the comment now says so, and the test case that pins it is labelled as covering a shape production cannot emit.
- `docs/cli-diff.md` documented the PRE-FIX failure mode, which the page template forbids ("Describe current behaviour. Not how it got that way."). Cut to what the command does today.
- Issue go-to-k/cdkd#3239's title still said "recurses forever" and its `severity:medium` label survived the body correction. Both fixed.

One finding is deliberately NOT in this PR: `NestedStackProvider` walks the same `aws:asset:path` chain on `create` / `update` with no ancestor set, and that walk **provisions real AWS resources and writes a child state record at every level** before the same key limit stops it. Materially worse than the read-only arm fixed here, and in the `integ-destroy` gate scope. Filed as go-to-k/cdkd#3247.

Also added: both interpolations in the refusal now go through `displaySafe`. The message exists FOR a hand-modified assembly, so `logicalId` and the asset path are attacker-controlled, and `formatError` sanitizes only `cause`.

**CI caveat, and it is not about this PR.** `check-build-test` shows green here, but that run's base was frozen before go-to-k/cdkd#3237 merged. Against current `main` the identical tree fails `rule-file-payload.test.ts`: `deploy-engine.ts` (65442/61000) and `rollback-executor.ts` (70371/66000) are over their caps with this branch's delta at **0 B**. The fence's own message says it is "not yours to fund" and that trimming another lane's entry "is not yours to spend", so this PR does not fund it. Filed as go-to-k/cdkd#3248 — it reds every branch, and it is invisible from any PR page whose CI started earlier.

**Do not merge this PR until go-to-k/cdkd#3248 is fixed and this branch is rebased onto the result** — the green above attests to a base that no longer exists.

go-to-k:
**Retracting the merge block in my previous comment — it rested on a wrong diagnosis, and the fix was a rebase.**

I said this PR must not merge until go-to-k/cdkd#3248 was fixed, because `rule-file-payload.test.ts` reported `deploy-engine.ts` and `rollback-executor.ts` over their caps with a 0 B delta from this branch. That issue is now closed as not-a-defect: go-to-k/cdkd#3237 raised both budget rows in the SAME commit that added the rule file (`61_000 -> 66_500`, `66_000 -> 71_500`), and this branch had forked before it, so the fence was projecting `main`'s current payload onto this branch's STALE copy of the budget table.

What misled me is worth stating, since the message is designed to be trusted:

> the overrun arrived on origin/main, so it is not yours to fund

That is true about the payload and wrong about the cause. The fence cannot tell "main grew" from "your table is stale", and it asserts the first. I stopped at the message instead of checking whether the row under it had moved — taking a claim on trust, which is the exact failure this PR's review round caught four times in my own prose.

Rebased onto `97206c84` and re-verified at head `f4ed3a2e`:

- `rule-file-payload.test.ts`: **362 passed** (was 1 failed)
- full suite: **1033 files, 24100 passed, 1 skipped, 0 failed**, no type errors
- `vp run check` 0 errors, `typecheck:test` clean, build clean
- live re-run against the REBUILT binary: the cyclic assembly exits 1 with the refusal naming `Loop` under `Cdkd3239Live~Child`; the same assembly with the cycle removed diffs normally (`2 to create` at the root, `1 to create` in the nested child)

One thing the rebase itself surfaced, since it is the kind of stale-artifact trap worth naming: `tests/unit/cli/version.test.ts` failed straight after the rebase (`expected '0.290.12' to be '0.290.15'`) because `dist/` was still built from the pre-rebase tree. A rebuild fixed it — but a lane that read that as a real failure would have gone looking in the wrong place.

No blocker outstanding on this PR. The review findings from the previous comment are all addressed at this head; the deploy-path twin remains out of scope as go-to-k/cdkd#3247.

go-to-k:
Review round 2 at `f4ed3a2e` — code and test axes. **Two majors and three minors, all addressed in `831dd0ad`.** No blocker.

**The test axis's major is the one worth reading: `displaySafe` was unpinned.** Dropping it from all three interpolations left the file 145/145 green, while the changelog asserted the sanitization and the code comment called it security-load-bearing — a claim with nothing watching it. This file already carried the right convention 1700 lines down (the `renderOutputChangeLines` control-character pin), so the fix was to follow it.

Getting that pin honest took three iterations, and each one is a lesson I have been handing out this week:

- A single hostile logical id pins `logicalId` and **leaves `stackName` free** — measured: dropping `displaySafe` from `stackName` alone stayed green, because at the refusal point the stack name is `Parent~<outer id>` and the outer id was clean. Poisoning one value per interpolated value is what the case needs, not one hostile string.
- The PATH argument could not be poisoned through the template at all, because the refusal only fires on a path that was already READ at a higher level — so the file has to exist. It is pinned through a control-character **file name**, which macOS and Linux both accept and round-trip (probed before relying on it).
- All three now red independently.

**The code axis's major was mine again: the PR body still carried both round-1 false claims** after I had corrected the source, the commit message and the docs. Fixed, along with the stale test counts and a mutation-table row the body was missing. The inherited `severity:medium` label was also still on this PR after go-to-k/cdkd#3239's body was corrected to `low`; relabelled.

The minors: the comment said "both interpolations" for three values and omitted `stackName` from its justification, though `stackName` is template-derived below the root; the sibling absolute-path refusal in the same file was still raw, and since the cycle refusal cites that exact threat model as its reason to sanitize, leaving its twin unsanitized was the inconsistency — both now go through `displaySafe`; and the docs and changelog claimed the refusal catches the FIRST repeat without the root-template caveat the source comment already carried.

That caveat is now measured rather than reasoned. Built binary, an assembly whose child template points back at the ROOT template file:

```text
Error: Nested stack 'ToB' under stack 'Cdkd3239Root~ToB~ToRoot' resolves to
nested template '.../b.json', which is already being diffed higher up the
same nesting chain. ... Refusing to diff.
```

One level deeper than the first repeat, exactly as documented, because the root's own template path is never seeded into the ancestor set.

**One thing deliberately left unpinned, stated so nobody re-hunts it.** Swapping `resolvedChildPath` for the raw `childTemplatePath` in the message measures green everywhere. That is an identity, not a gap: at any refusal site the child's path came from `indexNestedChildTemplates`' own `path.join`, which normalizes, and only the ROOT entry can be non-canonical — and the root never refuses, since its ancestor set is empty. The comment at the assertion says this.

Filed rather than widened: go-to-k/cdkd#3277 — `AssemblyReader`'s absolute-asset-path refusal (`src/synthesis/assembly-reader.ts:310-316`) is a third copy of the same message with the same three values raw, and it is the one a hostile assembly reaches FIRST, during synth. Different layer, so not this PR.

Suite at this head: **1033 files, 24101 tests, all pass.** Live re-run of both the cyclic and acyclic assemblies still behaves as before.

Issue #1931, opened 2026-08-18
fix(dynamodb): AWS::DynamoDB::Table destroy still fails hard on the transient index-busy refusal that #1830 fixed for GlobalTable


go-to-k:
Surfaced by the spec review of PR #1930 (the #1830 / #1857 lane).

## The gap

#1830 asked that the sibling `AWS::DynamoDB::Table` provider be audited in the
same pass. PR #1930 fixed the retry for `AWS::DynamoDB::GlobalTable` only:
`DynamoDBTableProvider`'s `DeleteTable` (`src/provisioning/providers/dynamodb-table-provider.ts:2152`)
still has no index-busy retry, so a `Table` whose GSI is mid-update when destroy
reaches it surfaces the same hard `PartialFailureError` that #1830 removed for
the GlobalTable type:

```
Attempt to change a resource which is still in use: Cannot delete table while
indexes are being created, updated, or deleted.
```

The condition is transient and self-resolving, and application auto-scaling can
start an index capacity change at any moment — so any `AWS::DynamoDB::Table`
with an autoscaled GSI can hit it, exactly as the GlobalTable type could.

## Why it was not done in #1930

The retry landed provider-locally (deliberately — `src/deployment/retryable-errors.ts`
was owned by another lane), so extending it to the Table type is not a
one-line reuse: it needs the same bounded re-arm poll, its own test seam, and
a real-AWS destroy arm to prove the retry actually fires. That arm is the
expensive half — the GlobalTable one took three designs before it discriminated
(the first two passed with the fix REVERTED, or missed the race window), and it
is recorded in #1930's body.

## Suggested direction

Port the shape #1930 landed rather than inventing a second one:

- a message-keyed classifier (AWS returns this as a plain `ResourceInUseException`,
  a name it shares with terminal conflicts, so a name-keyed predicate would
  retry things that can never succeed);
- a re-arm between attempts that waits on the CONDITION, bounded well below the
  per-resource deadline `destroy-runner.ts` enforces (the GlobalTable version
  caps the worst case at ~8.8 min);
- one `logger.warn` on the first retry, since `withRetry` announces only at
  debug level;
- a live arm whose driver CREATEs an index rather than re-provisioning one — a
  capacity change on an empty index settles in under a second and is ACTIVE
  again before `DeleteTable` arrives, which is why #1930's second design failed
  twice.

Consider whether the classifier + retry should move into a shared module read
[...]

PR #3246, opened 2026-09-16
fix(providers,rollback): read DynamoDB Integers the way CloudFormation does, and let a rollback replay a falsy log-group retention

### LINKED ISSUE #3141: fix(providers): a rollback revert of a log group whose record spells RetentionInDays falsy is refused where it used to take the delete arm
## The behaviour

Since go-to-k/cdkd#3137 (issue go-to-k/cdkd#2699), `LogsLogGroupProvider.applyUpdate` refuses a `RetentionInDays` of `0` / `'0'` / `false` / `null` on a template bag, admitting `0` only under `UpdateContext.desiredFromAwsReadback` (the `drift --revert` producer). The rollback executor's revert arms hand `update()` the journaled `previousState.properties` with a context carrying NO `desiredFromAwsReadback`, and `UpdateContext` has no `replayingState` counterpart to `CreateContext`'s. So a record whose `properties` spell the retention as one of those four — a pre-go-to-k/cdkd#2699 template cdkd accepted and recorded (the old code routed the whole falsy family to `DeleteRetentionPolicy`), or `drift --accept` writing the readback `0` into `properties` on a record with no `observedProperties` — is REFUSED on a rollback where it used to take the delete arm. The rollback op fails and the resource stays at the failed deploy's retention.

Recorded in-code as an accepted residual at `src/provisioning/providers/logs-loggroup-provider.ts` (the comment block above `desiredFromAwsReadback` in `applyUpdate`); surfaced by the go-to-k/cdkd#3137 security review as a rollback that cannot complete for that population.

## The fix

Thread `replayingState` into `UpdateContext` (as `CreateContext` already carries it), set it at the rollback executor's revert arms, and under it interpret the falsy family the way the binary that WROTE the record did — the delete arm — with a warning naming the spelling (the same "replay reproduces what the record certifies" rule go-to-k/cdkd#3137 applied to the CREATE replay for a `'30.0'` / `'0x1e'` spelling). Template bags keep the refusal.

## Why it is not in go-to-k/cdkd#3137

`src/deployment/rollback-executor.ts` and `src/types/resource.ts` are held by the open PR go-to-k/cdkd#3104 (`fix/3092-rollback-executor-log-injection`); a second lane editing the executor's revert arms collides with it. Resume query: `gh pr view 3104 --json state` — once MERGED, this is a small change in the provider + executor + `UpdateContext` with a rollback-executor unit case driving `previousState.properties.RetentionInDays: 0` through `update()`.

Dup-check: searched open issues for `rollback RetentionInDays replay`, `UpdateContext replayingState`, `falsy retention rollback` — no hit; go-to-k/cdkd#2699 is closed by go-to-k/cdkd#3137 and names this residual only in its PR body.

Session-fit: next (not this session) — external input: the executor file is held by open PR go-to-k/cdkd#3104; `gh pr view 3104 --json state` clears it.
Severity: medium — a rollback of a log-group replacement fails for a record spelling the retention falsy, leaving the resource at the failed deploy's retention until the record is hand-corrected.
Effort: small (S) — one flag through `UpdateContext`, one branch in `applyUpdate`, one executor unit case; the provider file is already read.
Estimate: ~1 h — the executor unit fixture is most of it; no real-AWS run needed beyond the existing `loggroup-class-guard` fixture.

### LINKED ISSUE #3147: fix(providers): the AWS::DynamoDB::Table capacity forwarders read ProvisionedThroughput through a bare Number(), the sibling of the toFiniteNumber population outside its grep

go-to-k:
Two issues in one lane. Both are about a provider reading a value it was handed
under the wrong rule: one reads a template Integer more loosely than
CloudFormation does, the other cannot tell a state record from a template.

Closes #3147
Closes #3141

## #3147 — `AWS::DynamoDB::Table` Integer forwarders

Issue #3135 moved every `toFiniteNumber`-reading DynamoDB forwarder onto
`coerceCfnInteger`, CloudFormation's MEASURED DynamoDB Integer grammar
(`/^[+-]?\d+$/`, NO trim; the live A/B table is on `toCfnInteger`'s doc). Its
sweep was `grep -n toFiniteNumber src/provisioning/providers/dynamodb-*.ts`, and
this provider's capacity reads never appeared on it — they go through a bare
`Number()`, which is wider still.

### Per-site classification

Sweep command (code lines only; the full output is in the session log):

```
grep -n "Number(\|toFiniteNumber(" \
  src/provisioning/providers/dynamodb-table-provider.ts \
  src/provisioning/providers/dynamodb-globaltable-provider.ts
```

Counts: 20 code occurrences in `dynamodb-table-provider.ts` (25 including
comment mentions), 5 in `dynamodb-globaltable-provider.ts` (11 including
comments).

| site | class | verdict |
| --- | --- | --- |
| `readCapacityNumber` (L150) — per-GSI `GlobalSecondaryIndexUpdates[].Update.ProvisionedThroughput` | FORWARD | moved to `coerceCfnInteger` |
| `hasUsableDeclaredCapacity` (L172-173) | MIRROR of the above | moves with it (it CALLS it) |
| `hasUsableTableCapacity` (L206) | MIRROR of the table-level forwarder | moved to `coerceCfnInteger` + the `< 1` rule AWS adds |
| `create()` table-level `ProvisionedThroughput` (L1025-1026) | FORWARD | `tableCapacityForSend` |
| `update()` billing-flip table-level (L1786-1787) | FORWARD | `tableCapacityForSend` |
| `recoveryPeriodInDays` (L2803) — `UpdateContinuousBackups` | FORWARD | `coerceCfnInteger`, announced drop |
| `liveCapacityAlreadyMatches` (L247-250) | GUARD (live-vs-desired compare) | unchanged — `toFiniteNumber` |
| `warmThroughputAlreadyMatches` (L530-531) | GUARD | unchanged |
| `skipZeroCapacityIndexUpdate` (L3897-3898) | GUARD | unchanged |
| GlobalTable `targetValue = Number(tttCfg['TargetValue'])` (L3890) | FORWARD, but a CFn **Double** | unchanged and out of scope — the file's own comment records that grammar as unmeasured |
| GlobalTable L8193-8194 / L8206-8207 | GUARD (live readback compares) | unchanged |

### Per-site decision for a rejected spelling

Never silent, never a throw — the arm each site already had for an unusable
value:

- **table-level (both forwarders)**: ANNOUNCE and OMIT the member. DynamoDB
  then rejects the request naming it, which is the arm this type already takes
  for an absent required capacity (the flip block's own comment says a
  `ProvisionedThroughput`-less flip to PROVISIONED is deliberately left to fail
  at AWS as CFn parity). Deliberately NOT the GlobalTable sibling's announced
  5/5 fallback: state records the DESIRED bag, so a table created at 5 against a
  recorded `" 7 "` compares EQUAL on every later `cdkd diff` and the divergence
  is invisible after the one warning scrolls past. `tableCapacityForSend`'s doc
  carries both reasons.
- **ABSENT member**: still the silent `?? 5` default, byte-identical to before.
  `null` takes that arm too, so the forwarder and `hasUsableTableCapacity`
  agree about it.
- **per-GSI**: unchanged arm — the index is omitted and named in the existing
  `unspecified` warning.
- **`RecoveryPeriodInDays`**: no `?? default` exists, so the announced arm is
  the DROP an absent member already takes; PITR is enabled at DynamoDB's own
  default period instead of the call failing on `NaN`. Warn rather than throw
  because `applyPointInTimeRecovery` is reached from `update()`, which is
  replay-reachable.

### A pre-existing divergence closed in the other direction

`Number(null)` is `0`, so `hasUsableTableCapacity` read a DECLARED `null` member
as unusable and refused the destructive pre-flip index removal — while the
forwarder's `?? 5` sent 5 and AWS accepted the flip. The warning said "AWS
rejects the flip either way" over a flip that succeeds. Both now treat `null` as
absent.

## #3141 — a rollback cannot replay a falsy log-group retention

Since #2699, `LogsLogGroupProvider.applyUpdate` refuses a `RetentionInDays` of
`0` / `'0'` / `false` / `null` on a template bag, admitting `0` only under
`UpdateContext.desiredFromAwsReadback`. The rollback executor's revert arms hand
`update()` the journaled `previousState.properties` and set no such flag —
correctly, since that bag is a template recorded earlier, not an AWS readback,
and setting it there would delete a live configuration elsewhere. `UpdateContext`
had no `replayingState` counterpart to `CreateContext`'s, so a record spelling
the retention falsy was REFUSED where it used to take the delete arm: the
rollback op failed and the resource stayed at the FAILED deploy's retention.

`UpdateContext` now carries `replayingState`, and under it the provider
reproduces what the record CERTIFIES, mirroring the create-path downgrade
#3137 shipped:

- a spelling the writing binary read as a positive integer through `Number()`
  (`'0x1e'`, `'30.0'`) is RESTORED as that integer — refusing it would drop a
  compliance retention the failed deploy never touched;
- everything else takes the DELETE arm, which is where it used to land;
- both warn, naming the spelling.

Template bags and `drift --revert` keep the refusal unchanged. The flag licenses
the downgrade and nothing else — `UpdateContext`'s doc says so explicitly, and
the `{Rules: []}` arms in `s3-bucket-provider.ts` read `desiredFromAwsReadback`
and are untouched.

### `UpdateContext` producer sweep

```
grep -rn 'replayingState\|desiredFromAwsReadback' src/
```

Four producers, and only two set the new flag:

| producer | `replayingState` | `desiredFromAwsReadback` |
| --- | --- | --- |
| `deploy-engine.ts` in-place UPDATE | no | no |
| `rollback-executor.ts` `revert` arm | **yes** | no |
| `rollback-executor.ts` `revert-failed-update` arm | **yes** | no |
| `drift.ts` `--revert` | no | yes |

The five providers that re-create inside their own `update()` still pass no
`CreateContext`; the information now exists on that path, but none uses it — a
route that opened, not a constraint that lifted. The four rule / doc files that
asserted `UpdateContext` "carries no `replayingState`" are corrected.

## Tests

56 new cases in `tests/unit/provisioning/dynamodb-table-provider-cfn-integer.test.ts`
(a 19-row spelling matrix run through BOTH table-level forwarders independently,
a 7-row `RecoveryPeriodInDays` matrix, and six cases for the per-index forwarder
and the two mirror predicates), 12 new rows on `UPDATE_MATRIX` in the retention
suite, plus a negative control in `deploy-engine-provider-secret-masker.test.ts`
pinning that the template path declares no `replayingState`.

### Mutation probes

Each mutated the PRODUCTION file after the fixes were committed, one mutation
per probe, restored from a byte-exact copy with `git diff` empty afterwards.

| probe | reds |
| --- | --- |
| `readCapacityNumber` back to `Number()` | 2 — the per-index forwarder case and its `hasUsableDeclaredCapacity` mirror |
| `tableCapacityForSend` reject arm returns the 5 default instead of omitting | 26 — every `sent: undefined` row on both forwarders |
| `hasUsableTableCapacity` back to `Number()` | 1 — "refuses the pre-flip REMOVAL when the TABLE capacity reads unusable" |
| its `null` skip narrowed to `=== undefined` | 1 — "treats a DECLARED null table capacity as absent" |
| `recoveryPeriodInDays` back to `Number()` | 6 — 5 create rows plus the update-path drop |
| `create()` wiring back to `Number(pt?.[x] ?? 5)` | 10 — only the `create:` rows |
| flip wiring back to `Number(pt[x] ?? 5)` | 10 — only the `flip:` rows |
| drop `replayingState` on the `revert` arm | 3 |
| drop `replayingState` on the `revert-failed-update` arm | 4 |
| disable the whole update-path replay downgrade | 11 — every `[replay]` row except the usable-value control |
| disable only the `legacyReading` restore arm | 3 — the `'0x1e'` / `'30.0'` rows alone |
| SET `replayingState` at the deploy engine's UPDATE site | 1 — the negative control |

The create-path wiring and the flip-path wiring red disjoint row sets, so
neither vouches for the other.

## Live verification — the parent ran FOUR fixtures, all PASS

Neither behaviour delta is observable without real AWS, so the lane did not live-test them. The parent did, on `64ab0588` (src identical to this HEAD, which adds only the ledger rows):

| fixture | result | why this one |
| --- | --- | --- |
| `lambda` (broad set) | PASS, 97 s, 9 deleted / 0 errors | the cross-cutting deploy/destroy paths this diff reaches through `rollback-executor.ts`, with DynamoDB Table + GlobalTable present |
| `loggroup-class-guard` | PASS, 91 s, all 7 phases | the log-group fixture for the #3141 retention arm |
| `dynamodb-ondemand` | PASS, 223 s, 4 deleted / 0 errors | **the changed path itself** — its own summary names `BillingMode/ProvisionedThroughput in-place UPDATE`, which is the flip arm #3147 moved onto `coerceCfnInteger` |
| `dynamodb-gsi-update` | PASS, 1099 s, 1 deleted / 0 errors | the per-GSI capacity reader, plus #1931's index-busy `DeleteTable` absorption on the destroy |

Orphan scan clean after each (state prefix holds only `deployments/`; no table, function, role or log group left by name). Ledger rows committed; `integ-destroy` + `integ-broad` set from this worktree.

`lambda` alone would NOT have covered the changed path — it exercises no billing flip — which is why the two DynamoDB fixtures were both run rather than one chosen.

## Not in scope

- `dynamodb-globaltable-provider.ts`'s `TargetValue` read stays on `Number()`:
  it is a CloudFormation **Double**, whose grammar is unmeasured, and the file
  already records that.
- No CLI flag, dependency, or state-schema change.

## Parent round (3-axis + security; the lane's own round died with its session, so nothing here was pre-reviewed)

**Spec / code / security: Clean.** The replay flag is an executor-side literal no journal value can set (two producers, both revert arms); the MIRROR predicates agree with their forwarders across the input cross-product; a rejected spelling reaches neither the wire as `NaN` nor the `?? 5` default (that default's population is exactly `undefined`/`null`).

**Test round found one real gap, applied here:** the UPDATE-path replay downgrade announced nothing any test read — silencing BOTH `logger.warn` calls left the file green at 118/118, so a rollback could DELETE a live retention policy in silence. One case per arm (REMOVED / RESTORED) plus a silence control; the same probe now reds exactly that case. Also applied: accepted-spelling controls (`'+8'`, `'010'`) for `RecoveryPeriodInDays` and the per-index flip reader, which were fenced in the rejecting direction only — narrowing `readCapacityNumber` to `/^\d+$/` previously reddened nothing and now reds the new control.

**One false claim corrected:** `.claude/rules/layout-provisioning.md` said every `ProvisionedThroughput` read in this provider goes through `coerceCfnInteger`; `create()`'s per-GSI block forwards each entry VERBATIM (a cast, structurally invisible to the `Number(` sweep that produced #3147's list). The sentence now names the sites that really coerce, and that site plus the `RecoveryPeriodInDays` phantom-drift shape are filed as go-to-k/cdkd#3255. The `update()` context comment ("no masker is threaded; the retention refusal names no value") went stale once the restore warn interpolates its reading — corrected to say what that value is, a day count.

**Rebased onto `origin/main`:** the rules-payload fence fired on a STALE-BASE artifact, not an overrun — `name-collision-classification.md` does not exist at the old merge base, so main's move of that text read as this branch re-inlining it. After the rebase the fence passes.


go-to-k:
Merged — thanks. Your lane was killed by a rate limit mid-review-round, so the whole tier ran fresh here: spec, code and security came back Clean, and the test round found the one gap that mattered — the UPDATE-path replay downgrade announced nothing any test read, so a rollback could have deleted a live retention policy in silence. Four fixtures exercised it live, including `dynamodb-ondemand`, whose own summary names the `BillingMode`/`ProvisionedThroughput` in-place UPDATE this PR changed.

Issue #3265, opened 2026-09-16
fix(providers): AWS::DynamoDB::Table forwards OnDemandThroughput verbatim at all three send sites


go-to-k:
## `OnDemandThroughput` is forwarded verbatim at every send site

The sweep behind go-to-k/cdkd#3255 item 1 (per-GSI `ProvisionedThroughput` on the create path) enumerated every capacity-shaped forward in `src/provisioning/providers/dynamodb-table-provider.ts` and found one more class it did not close: `OnDemandThroughput`, whose two members (`MaxReadRequestUnits` / `MaxWriteRequestUnits`) are `Long` fields in the SDK model and are handed over exactly as the template spelled them.

```
$ grep -n "OnDemandThroughput" src/provisioning/providers/dynamodb-table-provider.ts | grep -v "^ *[0-9]*: *[/*]"
1136:        createParams.OnDemandThroughput = properties['OnDemandThroughput'] as OnDemandThroughput;
2045:              OnDemandThroughput: properties['OnDemandThroughput'] as OnDemandThroughput,
3763:            ...(gsi.OnDemandThroughput ? { OnDemandThroughput: gsi.OnDemandThroughput } : {}),
```

- `:1136` — `create()`'s table-level `CreateTable` forward.
- `:2045` — `update()`'s table-level `UpdateTable` forward.
- `:3763` — `applyGsiUpdates`' per-index `Create` action.

So a CFn-legal `MaxReadRequestUnits: '100'` reaches AWS as the STRING `"100"` in a `Long` field, and a `' 100 '` / `'0x64'` / `'1e2'` CloudFormation refuses the template for is forwarded unchanged — the same shape issues go-to-k/cdkd#3135 / go-to-k/cdkd#3147 / go-to-k/cdkd#3255 closed for `WarmThroughput` and `ProvisionedThroughput`.

## Why it is a SEPARATE issue rather than part of go-to-k/cdkd#3255

[...]

Issue #3264, opened 2026-09-16
fix(providers): applyGsiUpdates still forwards per-GSI ProvisionedThroughput verbatim at three send sites


go-to-k:
## The class issue go-to-k/cdkd#3255 item 1 closed on the CREATE path only

go-to-k/cdkd#3255 item 1 moved `create()`'s per-`GlobalSecondaryIndexes[]` `ProvisionedThroughput` onto `readCapacityNumber` (CloudFormation's measured DynamoDB Integer grammar, `/^[+-]?\d+$/`, no trim), so create and the BillingMode flip finally agree about one template value. The sweep that produced that fix found the SAME verbatim forward at three more per-index send sites, all inside `applyGsiUpdates` in `src/provisioning/providers/dynamodb-table-provider.ts`:

```
$ grep -n "ProvisionedThroughput = gsi.ProvisionedThroughput\|ProvisionedThroughput: gsi.ProvisionedThroughput" src/provisioning/providers/dynamodb-table-provider.ts
3701:            adopted.ProvisionedThroughput = gsi.ProvisionedThroughput;
3761:              ? { ProvisionedThroughput: gsi.ProvisionedThroughput }
3842:            update.ProvisionedThroughput = gsi.ProvisionedThroughput;
```

- `:3701` — the adopted-index repair arm (the Create was skipped because AWS already holds the index).
- `:3761` — the GSI `Create` action.
- `:3842` — the GSI `Update` action.

Each hands `UpdateTable` the block exactly as the template spelled it, so a CFn-legal `ReadCapacityUnits: '7'` goes out as a STRING in a `number` field and a `' 7 '` CloudFormation refuses the template for goes out too — the identical defect go-to-k/cdkd#3255 item 1 fixed one method away.

## Why this was NOT folded into that PR

These three take a different PREMISE from the create-path site, and the difference is a design decision rather than a mechanical repeat. On the update path the raw block is ALSO the operand of three guards, and all three read it through `toFiniteNumber`, a LOOSER grammar than `coerceCfnInteger`:

- `liveCapacityAlreadyMatches(live, gsi.ProvisionedThroughput)` — the issue go-to-k/cdkd#1630 idempotency skip;
- `skipZeroCapacityIndexUpdate(name, physicalId, gsi.ProvisionedThroughput, ...)` — the issue go-to-k/cdkd#1767 `{0, 0}` placeholder skip;
[...]

Issue #3255, opened 2026-09-16
fix(providers): the DynamoDB Table provider forwards per-GSI ProvisionedThroughput verbatim on create, and an announced RecoveryPeriodInDays drop records a value AWS does not hold


go-to-k:
### 1. `create()`'s per-GSI `ProvisionedThroughput` is forwarded VERBATIM

go-to-k/cdkd#3147 moved the `AWS::DynamoDB::Table` provider's capacity forwarders onto `coerceCfnInteger` (CloudFormation's measured DynamoDB Integer grammar, `/^[+-]?\d+$/`, no trim). The per-GSI block on the CREATE path was not among them, and neither the issue's enumeration nor the PR's sweep could see it, because it is a CAST rather than a call:

```ts
createParams.GlobalSecondaryIndexes = Array.isArray(declaredGsis)
  ? (declaredGsis as GlobalSecondaryIndex[]).map((entry) =>
      this.coerceIndexWarmThroughputForCreate(logicalId, entry, maskSecrets)
    )
  : (declaredGsis as GlobalSecondaryIndex[]);
```

Only `WarmThroughput` is coerced per entry; each entry's `ProvisionedThroughput` goes to `CreateTable` exactly as the template spelled it, so a CFn-legal `ReadCapacityUnits: '7'` reaches the SDK as a STRING in a `number` field, and a `' 7 '` CloudFormation rejects is forwarded too. The BILLING-FLIP path's per-index reader (`readCapacityNumber`) is already on the grammar, so create and flip currently disagree about the same property.

Found independently by the spec and code reviewers on go-to-k/cdkd#3246; the rules sentence in `.claude/rules/layout-provisioning.md` was narrowed there to stop claiming this site.

### 2. An announced `RecoveryPeriodInDays` DROP records a value AWS does not hold

go-to-k/cdkd#3246 made a rejected `RecoveryPeriodInDays` spelling drop the member and announce it, and `UpdateContinuousBackups` then SUCCEEDS with AWS's default period while the state record keeps the declared `' 7 '`. `readCurrentState` reads back AWS's default, so the resource reports drift on every run and `cdkd drift --revert` re-issues the same call forever — the permanent-phantom-drift shape `.claude/rules/provider-replay-and-refusals.md` prescribes `effectiveProperties` for. This provider returns none. (The table-capacity omit has no such problem: that call FAILS, so nothing is recorded.)

[...]

PR #3268, opened 2026-09-16
fix(providers): coerce per-GSI create capacity, and stop hand-quoting pasteable aws commands

go-to-k:
Two HALF-issues, both in the provider layer. **Neither issue is closed** — each keeps an open half that is deliberately out of scope, named below. `Refs`, not `Closes`.

## 1. `Refs go-to-k/cdkd#3255` (item 1 only) — per-GSI `ProvisionedThroughput` on the CREATE path

`create()` forwarded the whole `GlobalSecondaryIndexes` array to `CreateTable` verbatim except for each entry's `WarmThroughput`, so a CFn-legal `ReadCapacityUnits: '7'` reached the SDK as a STRING in a `number` field, and a `' 7 '` CloudFormation refuses the template for was forwarded too. The BillingMode-flip path's per-index reader (`readCapacityNumber`) has read CloudFormation's measured DynamoDB Integer grammar (`/^[+-]?\d+$/`, no trim) since go-to-k/cdkd#3147, so create and flip disagreed about the same template value.

Both now read through `readCapacityNumber`. New `coerceIndexCapacityBlock`; `coerceIndexWarmThroughputForCreate` became `coerceIndexThroughputForCreate` (it now handles both blocks in one pass over the entry); `warnUnusableTableCapacity` became `warnUnusableProvisionedCapacity` and is called from all THREE forwarders that can drop a member; `TABLE_CAPACITY_MEMBERS` became `PROVISIONED_CAPACITY_MEMBERS`, since the membership is the same at both levels and a second list is exactly the drift that constant exists to prevent.

### The rejected-spelling decision, stated

**A rejected spelling drops the MEMBER, not the whole `ProvisionedThroughput` block.** Three reasons:

1. It is what `tableCapacityForSend` already does one level up for the identical grammar, so create's two capacity forwarders answer the same way and print the same `NO capacity was substituted` sentence.
2. It keeps a sibling member the template spelled LEGALLY. Dropping the block would discard information a later `cdkd diff` cannot recover.
3. It is the more informative failure at AWS: `CreateTable` names the missing member at its exact index position, where a missing block yields only `ProvisionedThroughput must be specified for index: <name>`.

Both directions FAIL the deploy — DynamoDB requires both members on a PROVISIONED index — so this is a choice of error TEXT, never of outcome. That is also why the drop owes **no `effectiveProperties`**: the phantom-drift rule in `.claude/rules/provider-property-fidelity.md` exists for a SUCCEEDING call plus a record describing the declared value, and here the call fails and nothing is recorded (the same reasoning `tableCapacityForSend` states for the table-level omit).

Two further decisions, both in `coerceIndexCapacityBlock`'s doc comment: an ABSENT member stays absent (the per-index reader deliberately has no `5` default — go-to-k/cdkd#1588), and a `ProvisionedThroughput` that is not a plain object is forwarded VERBATIM (the fail-open direction this file takes everywhere).

**OUT OF SCOPE, and why:** item 2 of go-to-k/cdkd#3255 (an announced `RecoveryPeriodInDays` drop records a value AWS does not hold, needing `effectiveProperties` plumbed into a provider that has none). It is a different mechanism on a different property with its own `canonicalizeDesiredProperties` twin question, and nothing here touches it. go-to-k/cdkd#3255 stays open for it.

## 2. `Refs go-to-k/cdkd#3136` (the hand-quoted half only)

Four sites wrapped a TEMPLATE- or state-borne value in `'...'` by hand inside a command the user is told to paste, so a `'` in the value broke out of the quoting and a control byte forged lines on the operator's terminal and inside the durable `deployments/*.jsonl` store:

- `ssm-parameter-provider.ts` — the partial-create cleanup warning (`aws ssm delete-parameter --name`) and `refuseUnwritableParameterId`'s thrown remedy (`aws ssm get-parameter --name`);
- `s3-bucket-provider.ts` — BOTH partial-create cleanup arms (`aws s3api delete-bucket --bucket`).

All four now render through the exported `renderDisableCommand` — `displaySafe(asciiOnly)`, then `shellQuote`, then SUPPRESS the whole command when sanitizing CHANGED the value, because a command naming the sanitized id would act on a DIFFERENT resource. The two S3 arms share one `manualBucketDeletionClause`, so they cannot come to render the same bucket name differently. A suppressed site says so and points at the console rather than printing a dead end.

**A clean value now renders BARE** (`--bucket my-bucket`), which is what `shellQuote` emits for `[A-Za-z0-9._/@:+-]` and what go-to-k/cdkd#2669 established for the log-group site. Three existing assertions on the hand-quoted form moved with it.

`replacement-protection-advice.ts` is a doc-only change: its header claimed the export exists "for exactly one caller", which this PR makes false.

**OUT OF SCOPE, and why:** the ~38 remaining sites in go-to-k/cdkd#3136 interpolate a value into PROSE rather than into a pasteable command. Each needs its own provenance call and `renderDisableCommand` does not apply to prose, so they are a separate pass. go-to-k/cdkd#3136 stays open for them.

## Sweep commands and counts

**Class 2 — hand-quoted pasteable commands (re-derived, not taken from the issue):**

```
$ grep -rnE "aws [a-z0-9-]+ [a-z0-9-]+[^\`]*'\\\$\{" src/provisioning/providers/
efs-provider.ts:1206
s3-bucket-provider.ts:6520
s3-bucket-provider.ts:6547
ssm-parameter-provider.ts:357
ssm-parameter-provider.ts:783
    5
```

FIVE, not the four the task named and not the three the issue lists. The fifth, `efs-provider.ts:1206`, is deliberately NOT converted and the reason is recorded on `renderDisableCommand`'s doc comment: its value comes from `acquireIdempotencyToken`, which mints `cdkd-<sha256 hex>` and nothing else (provably `[a-z0-9-]`), and its `'...'` are JMESPath string delimiters INSIDE a double-quoted shell argument — `shellQuote`'s `'\''` escaping would produce invalid JMESPath rather than safer shell. Post-fix the same sweep returns that one line and nothing else.

**Class 1 — remaining verbatim per-index forwards on the create path:**

```
$ grep -n "createParams\.[A-Za-z]* =" src/provisioning/providers/dynamodb-table-provider.ts | wc -l
11
$ grep -n "ProvisionedThroughput = gsi.ProvisionedThroughput\|ProvisionedThroughput: gsi.ProvisionedThroughput" src/provisioning/providers/dynamodb-table-provider.ts
3701 / 3761 / 3842   (all three inside applyGsiUpdates, i.e. the UPDATE path)
```

- `LocalSecondaryIndexes`: CFn's `LocalSecondaryIndex` (like the SDK's, checked in `@aws-sdk/client-dynamodb` `models_0.d.ts`) declares only `IndexName` / `KeySchema` / `Projection` — neither throughput block — so an LSI has nothing to coerce.
- Replica blocks: `AWS::DynamoDB::Table` has none (`grep -n "Replicas" dynamodb-table-provider.ts` returns nothing); they belong to `AWS::DynamoDB::GlobalTable`, with its own forwarders.
- The three `applyGsiUpdates` sites were deferred at first for a real reason (their raw block is also the operand of three `toFiniteNumber`-reading guards, so coercing only the wire value risked making guard and send disagree) — then round 2 TOOK them, because leaving them INVERTED the divergence this PR exists to close: the same entry would succeed on a fresh create and be rejected by AWS when added by a later update. All three now read through the same `coerceIndexCapacityBlock`, while the guards stay on `toFiniteNumber` by design. go-to-k/cdkd#3264 is CLOSED by this PR.
- `OnDemandThroughput` is forwarded verbatim at three sites too, consistently, so it needs its own decision rather than being folded in here. Filed as go-to-k/cdkd#3265.

## Mutation-probe receipts

Every probe: commit first, mutate PRODUCTION, run the file, read WHICH assertion fired, `cp` back byte-exact, `git diff` empty.

| # | Mutation (production) | Receipt | Assertions that fired |
|---|---|---|---|
| 1 | `coerceIndexCapacityBlock` returns `undefined` unconditionally | 262258 -> 262231 B; anchor 1 -> 0 | 18 cases, incl. all four ACCEPTED-spelling rows and every dropped row |
| 2 | drop the whole block instead of the member (`delete rewritten[member]` -> `rewritten = {}`) | 262258 -> 262248 B; anchor 1 -> 0 | 11 cases, incl. `drops the MEMBER and keeps its legal sibling` |
| 3 | default an ABSENT per-index member to 5 | 262258 -> 262313 B; anchor 1 -> 0 | exactly 1: the `Symbol(absent) -> Symbol(absent)` row |
| 4 | hardcode the warn scope's index name | 262258 -> 262189 B; anchor 1 -> 0 | exactly 1: `names the INDEX in the warning, and leaves a sibling index untouched` |
| 5 | revert the SSM cleanup site to hand-quoting | 41869 -> 41953 B; anchor 1 -> 0 | 5: that site's whole shape table; the other three sites stayed green |
| 6 | revert the SSM import-refusal site to hand-quoting | 41869 -> 41967 B; anchor 1 -> 0 | 5: that site only |
| 7 | revert ONLY the S3 indeterminate arm to hand-quoting | 390342 -> 390353 B; anchor 1 -> 0 | 6: that arm's 5 plus `renders the SAME clause from both arms`; the cleanup-failed arm stayed green |
| 8 | drop the SUPPRESS rule (`safeId !== identifier`) in `renderDisableCommand` | 16549 -> 16515 B; anchor 1 -> 0 | 14: all 12 of this PR's suppressed cases (4 sites x 3 shapes) plus the 2 pre-existing log-group ones |
| 9 | drop `shellQuote` from `renderDisableCommand` | 16549 -> 16537 B; anchor 1 -> 0 | 5: the single-quote case at each of the 4 sites plus the both-arms case |

One probe was VOID and is recorded rather than dropped: the first attempt at probe 5 used `perl -0pi -e 's/\Q${manualStep}`\E/.../'`, and `\Q...\E` does NOT stop Perl interpolating `${manualStep}` — the pattern collapsed to a bare backtick and hit the wrong occurrence. The receipt caught it (anchor still 1, suite still green); the probe was redone with a literal-safe replacer.

## Tests

- `tests/unit/provisioning/dynamodb-table-provider-cfn-integer.test.ts` — a new per-GSI create matrix of 16 rows with a per-OUTCOME literal floor, plus member-drop / index-naming / non-object / no-block / both-blocks / no-mutation cases. Every row asserts with `toBe`, which discriminates `7` from `'7'` — a matrix read through `Number(...)` would have passed against the unfixed provider.
- New `tests/unit/provisioning/pasteable-command-hand-quoted-ids.test.ts` — all four sites driven through their real `create()` / `import()`, each over five value shapes. The two OUTCOMES are both pinned: a `'` is printable ASCII, so sanitizing does not change it and the command is still shown SHELL-QUOTED; a control byte, a newline and a non-ASCII character each change the value and SUPPRESS the whole command. Every case also pins the negative — that the RAW hand-quoted form is absent — which is what discriminates against the pre-fix providers.
- Full suite: 1033 files, 23977 passed, 1 skipped, 0 type errors.

## Live verification — the parent ran BOTH, PASS

Run on `fed837f7` (src identical to this HEAD, which adds only the ledger rows):

| fixture | result | why this one |
| --- | --- | --- |
| `dynamodb-gsi-update` | PASS, 1649 s, 1 deleted / 0 errors | the per-GSI CREATE path AND the three `applyGsiUpdates` send sites round 2 converted; also re-proves go-to-k/cdkd#1767 (no phantom drift) and go-to-k/cdkd#1931 (destroy absorbs an index-busy `DeleteTable`) |
| `dynamodb-ondemand` | PASS, 309 s, 4 deleted / 0 errors | the BillingMode flip — the per-index reader this PR made create agree with, unchanged by the coercion |

Orphan scan clean after each (no table left by name; the state prefix holds only `deployments/`). Ledger rows committed; `integ-destroy` set from this worktree.

## Orchestrator round (3-axis, run fresh — the lane's own reviewers saw an earlier head)

Code and security: **Clean** on their focus traces. The four converted sites are the complete set of per-index capacity forwards that were verbatim (a fifth exists and has been on the grammar since go-to-k/cdkd#3147); the guards above each still read the DECLARED block; and no hostile value (`'`, backtick, `$( )`, newline, NUL, ESC, U+202E, non-ASCII) escapes the quoting at any of the four pasteable sites — each is POSIX-quoted or the command is suppressed whole. Test: **Clean**, every probe reddening, including each of the four send sites reverted independently, and the suppression fenced in BOTH directions.

One REGRESSION and three prose defects, applied in `5d40d322`:

- **Regression.** `IndexName: 2024` (unquoted YAML) is truthy, becomes a name-map key and reaches the scope builders as a NUMBER, on which the masker's `String.prototype.replace` THROWS. `create()` was guarded when `indexScope` was added; the EAGER scopes this round put on the update sites were not, so an ordinary capacity update with a non-empty secret bag newly died on a diagnostic path. The five scope literals now route through a guarded `indexScopeAt`; a case drives the real `update()`, and removing the guard reds exactly it.
- `coerceIndexThroughputForCreate`'s JSDoc had been orphaned by a second doc block landing under it — moved onto its function.
- Three comments claimed these WARN lines forge `deployments/*.jsonl` lines. They do not (the recorder persists a thrown `error.message`, not a warn), so the clause was dropped where false and kept at the two sites where a thrown `PutParameter` error genuinely reaches it.
- The SSM create suppression fallback stopped short of the wrong-target reason its own comment claimed the siblings share — appended.

Follow-ups filed by this lane and still open: go-to-k/cdkd#3265 (`OnDemandThroughput` verbatim at three sites) and go-to-k/cdkd#3269 (the PROSE half of these messages, i.e. the rest of go-to-k/cdkd#3136).

Refs #3255
Refs #3136


go-to-k:
Merged — thanks, and the round-2 catch was the important one: coercing only `create()` would have INVERTED the very divergence this PR exists to close. My round found one regression on top — the eager scopes the round added made a numeric `IndexName` throw on an ordinary capacity update — now guarded and fenced. Both DynamoDB fixtures exercised it live.

Issue #3266, opened 2026-09-16
chore(hooks): two PostToolUse informers miss a gh flag between the group word and the verb, so their warnings vanish on the spelling #3242 closed


go-to-k:
## What happens

Two PostToolUse informers hand-roll a `pr[[:space:]]+merge` ERE with no flag
absorber between the group word and the verb, so both go silent on the spelling
go-to-k/cdkd#3242 closed everywhere else:

- `.claude/hooks/main-tree-git-cwd-detector.sh` — `GH_PR_MERGE_VERB`
- `.claude/hooks/post-merge-sync-reminder.sh` — an inline ERE fed to `cmd_matches_verb`

Measured against this repo's library (`cmd_matches_verb`, both hooks' own EREs):

| command | cwd-detector | sync-reminder |
| --- | --- | --- |
| `gh pr merge 42 --squash` | MATCH | MATCH |
| `gh -R o/r pr merge 42` | nomatch (pre-existing, declared) | MATCH |
| `gh pr -R o/r merge 42` | **nomatch** | **nomatch** |
| `gh pr --repo=o/r merge 42` | **nomatch** | **nomatch** |
| `gh pr -Ro/r merge 42` | **nomatch** | **nomatch** |

## Why it is not a bypass, and why it is still worth closing

Both refuse nothing. `main-tree-git-cwd-detector` warns that a verdict-bearing
command ran in the MAIN tree while feature worktrees are live;
`post-merge-sync-reminder` reminds you to sync after a merge. The cost is a
missed warning, not an ungated merge — and the cwd-detector's own header already
declares the LEFT-slot gap as accepted for exactly that reason.

What makes it worth a ticket anyway is that the silence is now ASYMMETRIC with
every blocking gate. After go-to-k/cdkd#3242 the eleven blocking hooks fire on
`gh pr -R <slug> merge <n>`; these two do not, so the one spelling most likely
to be typed by an agent that has just been refused is also the spelling whose
reminders disappear.

## Why it was not folded into go-to-k/cdkd#3242

That issue says to fix defeated constants in the PR "if they share the pattern,
or file them if they are a different mechanism". These are a different
mechanism on three counts, and the PR records the same reasoning at the survey
comment in `lib/command-match.sh`:

1. They are hand-rolled EREs, not `GATE_RE_GH_*` constants, so the family fence
   added by that PR — whose population is read out of the library's own
   assignments — cannot see them by construction.
[...]

Issue #2717, opened 2026-09-06
chore(hooks): tier the guard layer by blast radius, and delete what only enforces a convention


go-to-k:
## The measurement

Taken 2026-09-07 on this tree:

```
hooks              47 files   21,686 lines
hook suites        50 files   21,069 lines
.claude/rules/hooks*           2,312 lines
-----------------------------------------------
guard layer                   ~45,000 lines
src/**            327 files  235,948 lines
```

**The guard layer is 19% the size of the product it guards**, and 42 of the 47
hooks BLOCK rather than warn. That number is the reason to open this: not any
single hook, each of which is defensible on its own -- 37 of the 47 cite a
concrete incident number in their own header, so this is an evidence-based
layer, not a speculative one.

The problem is that "each one is individually justified" has no stopping rule,
and the marginal hook is now being paid for in a currency the early ones were
not: review rounds on the guard layer itself, and lockouts.

## The one change worth making first, independent of everything else

**`main-tree-edit-gate` is the only hook that can take away the ability to
repair it.** Derived, not recalled:

```
PreToolUse hooks matching Edit or Write:
  Edit|Write|Bash          main-tree-edit-gate.sh      sources the shared matcher, 8 exit-2 sites
  Edit|Write|NotebookEdit  worktree-owner-gate.sh      does not source it, 1 exit-2 site
```

So a syntax error in `lib/command-match.sh` -- which every gate sources and
which is under active development -- makes the ONE hook that matches
`Edit|Write|Bash` fail closed on all three. Edit, Write and Bash are refused
together, and the file that must be edited to recover is the file that cannot
be edited. **This happened four times in the go-to-k/cdkd#2650 session**, three
of them from a single apostrophe inside an awk comment, and each time the
maintainer had to run the repair by hand from outside the session.

A safety mechanism must not be able to remove the operator's means of repair.
Options, cheapest first:

1. Make the library-load failure fall back to the Edit/Write arm only for the
   `file_path` case, which needs no shell parsing at all -- the gate's Edit and
   Write arms read `tool_input.file_path` directly and do not use the matcher.
   Only the Bash arm needs it, so only the Bash arm should fail closed on it.
2. Or split the hook: `main-tree-edit-gate` (Edit|Write, no matcher dependency)
   and `main-tree-edit-bash-gate` (Bash, fails closed as today).

Either way the class disappears. This is `severity: high` in practice even
though nothing is "wrong" today.

## The tiering

[...]

Issue #2614, opened 2026-09-05
chore(hooks): main-tree-edit-gate parses its own `cd` and misses a quoted command word


go-to-k:
## The shape

`.claude/hooks/main-tree-edit-gate.sh` blocks mutating a git-tracked file in a
worktree currently on `main` / `master`. On the `Bash` arm it resolves a
leading `cd` itself rather than through the shared matcher:

```sh
# .claude/hooks/main-tree-edit-gate.sh:61
if [[ "$cmd" =~ ^[[:space:]]*cd[[:space:]]+([^[:space:]\&\;\|]+) ]]; then
```

The verb `cd` is matched as LITERAL text, so a QUOTED or ESCAPED command word
does not match and `base_dir` is never updated. A relative write target then
resolves against the payload cwd instead of the directory the command actually
runs in:

```
cd /main-tree && echo x > tracked.txt        -> base_dir = /main-tree   (blocked)
"cd" /main-tree && echo x > tracked.txt      -> base_dir unchanged      (miss)
\cd /main-tree && echo x > tracked.txt       -> base_dir unchanged      (miss)
```

Note the token is already UNQUOTED one line later (`cdt="${cdt%\"}"` …), so the
hook expects quoting on the VALUE and not on the verb — the same asymmetry
go-to-k/cdkd#2333 found in the shared matcher.

## Why it was not fixed in go-to-k/cdkd#2333

That issue's acceptance criterion 3 says "audit hook-local parsers too". The
audit was done and is recorded on `gate_dequote_structural` in
`.claude/hooks/lib/command-match.sh`; the derived population is:

```sh
for f in .claude/hooks/*.sh; do
  grep -q 'tool_input.command' "$f" || continue
  grep -qE 'gate_matches|gate_segments|cmd_matches_verb|gate_verb_rest|gate_pr_selector|gate_target_dir|gate_tokens|gate_argv|strip_noncommand_spans' "$f" \
    || echo "NO SHARED MATCHER: $f"
done
```

It returns two hooks. `main-tree-dirty-detector.sh` is a non-blocking
PostToolUse observer keyed on write-ish tokens with no verb grammar — genuinely
out of scope. `main-tree-edit-gate.sh` is this issue. The criterion's own named
target, `main-tree-branch-gate.sh`'s quoted-span collapsing sed, no longer
exists: that hook reads the shared stream now.

go-to-k/cdkd#2333 closed the gap at `gate_segments`, which covers every reader
[...]

PR #3271, opened 2026-09-16
chore(hooks): absorb a gh flag between the group word and the verb

### LINKED ISSUE #3242: chore(hooks): a flag between `gh pr` and the verb fires no gate at all — `gh pr -R <slug> merge <n>` merges past verify-pr, ci-green and bughunt-clean
## What happens

`gh pr -R go-to-k/cdkd merge 42` fires **no gate at all** — not `verify-pr-gate`, not `bughunt-clean-gate`, not `ci-green-gate` — because the shared verb pattern does not tolerate a flag between `pr` and the verb:

```
GATE_RE_GH_PR_CREATE_OR_MERGE   # lib/command-match.sh
^gh( -C)? +pr +(create|merge)
```

`gh` itself accepts the flag in that position and resolves the repo from it (measured against gh 2.92.0: `gh pr -R go-to-k/cdkd view 3214` answers the cdkd PR from a directory that is not a repo). So the command merges a PR while every gate that guards `gh pr merge` sits out, **including in the same repo with a stale marker** — the case those gates exist for. Measured rc=0 where the equivalent `gh pr merge 42 -R go-to-k/cdkd` returns rc=2.

Found while reviewing go-to-k/cdkd#3214 (the sibling-repo relaxation); this hole is PRE-EXISTING on `main` and unrelated to that PR's change, which is why it is filed rather than folded (`references/verify.md` section 8-a: the structural fix does not go in late in a cascade).

## Spellings measured

| command | gate fires | gh resolves |
| --- | --- | --- |
| `gh pr merge 42 -R go-to-k/cdkd` | yes (rc=2) | cdkd |
| `gh pr -R go-to-k/cdkd merge 42` | **NO (rc=0)** | cdkd |
| `gh pr --repo go-to-k/cdkd merge 42` | **NO** | cdkd |
| `gh pr -R=go-to-k/cdkd merge 42` | **NO** | cdkd |
| `gh pr create -R … ` | yes | — |

The `-C` tolerance in the pattern shows the shape was anticipated for one flag and one position only.

## Why it matters

`verify-pr-gate` is the structural enforcement of "a PR whose live behaviour was never exercised cannot be opened or merged", and `ci-green-gate` of "CI must be green before a merge". Both are defeated by moving a flag three words to the left. No credential or third-party artifact is involved — the risk is that an agent (or a copied command line) merges past the gates believing they ran, which is exactly the silent-pass class `.claude/rules/hooks.md` says a gate must never have.

## Proposal

Widen the shared pattern to tolerate flags between `pr` and the verb, rather than patching each gate: the matcher already strips quoted spans and heredoc bodies, so the change is to the verb pattern plus its fixtures. Consider the same audit for every other `GATE_RE_*` that pins a subcommand position (`gh issue ...`, `git ...`), since the defect is positional rather than specific to this verb — the fix should come with a fence over the flag-between-words shape for the whole family, not one regex.

Both readers of this constant must be re-verified after the change: `verify-pr-gate.sh` and `bughunt-clean-gate.sh`. `ci-green-gate.sh` matches separately — check whether it has the same shape.

Dup-check: searched open and closed issues for `GATE_RE_GH_PR`, `command-match`, `gate bypass`, `gh pr merge flag` and the body window for `pr +(create|merge)`. go-to-k/cdkd#2354 is the neighbouring class (a QUOTED command LEADER — `env "GH_REPO=x" gh …` — evading the same matcher) and names the leader, never the flag position; this is a different span of the same pattern. Nothing covers it.

Session-fit: now (do it in this session) -- `lib/command-match.sh`, `verify-pr-gate.sh` and its 90-case harness were all read and edited in this session (go-to-k/cdkd#3214), gh's real flag-position behaviour was measured here, and the fix is a pattern widening plus fixtures in a suite already open; verify with `bash .claude/hooks/run-tests.sh`
Severity: high -- a complete bypass of every gate keyed on `gh pr merge`, reachable by an ordinary flag placement gh itself accepts, with no error and no trace that the gates did not run
Effort: small (S) -- one shared pattern plus fixtures per reader, riding the hook harness that is already green in this session; no `src/**`, no integ
Estimate: ~1 h -- most of it enumerating the positional family (which other GATE_RE_* pin a subcommand position) and the fence that stops the shape returning

go-to-k:
Closes #3242

## What was wrong

`gh` accepts `-R` / `--repo` in **two** slots and resolves the repo from either
identically. Measured on gh 2.92.0 from a directory that is not a repo:

```
$ cd /tmp && gh pr -R go-to-k/cdkd view 3214 --json number -q .number
3214
```

The shared matcher absorbed only the slot LEFT of the group word (`GATE_GH_C`),
so every `gh <group> <verb>` constant matched **nothing** when the flag moved
three words right — and a gate that recognises no command does not fire.

## rc table — real hooks, this worktree, markers stale

`verify-pr-gate` reads `GATE_RE_GH_PR_CREATE_OR_MERGE`; `pr-body-item-number-gate`
reads `GATE_RE_GH_BODY_CARRIER` (a body file carrying a bare `#N`).

| command | verify-pr | pr-body-item-number |
| --- | --- | --- |
| `gh pr merge 3242 --squash` | 2 → 2 | – |
| `gh -R go-to-k/cdkd pr merge 3242 --squash` | 2 → 2 | – |
| `gh pr -R go-to-k/cdkd merge 3242 --squash` | **0 → 2** | – |
| `gh pr --repo go-to-k/cdkd merge 3242 --squash` | **0 → 2** | – |
| `gh pr -R=go-to-k/cdkd merge 3242 --squash` | **0 → 2** | – |
| `gh pr -Rgo-to-k/cdkd merge 3242 --squash` | **0 → 2** | – |
| `gh pr --repo=go-to-k/cdkd merge 3242 --squash` | **0 → 2** | – |
| `gh pr -R go-to-k/cdkd -q .x merge 3242` | **0 → 2** | – |
| `gh pr -R go-to-k/cdkd "merge" 42 --squash` (quoted verb) | **0 → 2** | – |
| `gh pr -R go-to-k/cdkd create --title t` | **0 → 2** | – |
| `gh issue create --title t --body-file <bare #N>` | – | 2 → 2 |
| `gh issue -R go-to-k/cdkd create … <bare #N>` | – | **0 → 2** |
| `gh issue --repo=go-to-k/cdkd comment 42 … <bare #N>` | – | **0 → 2** |
| `gh pr -Rgo-to-k/cdkd edit 42 … <bare #N>` | – | **0 → 2** |
| `gh issue -R go-to-k/cdkd "create" … <bare #N>` | – | **0 → 2** |
| `gh pr -R go-to-k/cdkd list` / `view 42` / `diff 42` | 0 → 0 | 0 → 0 |
| `gh issue -R go-to-k/cdkd list … <bare #N>` | – | 0 → 0 |

`pr-body-item-number-gate` is the one gate `.claude/rules/hooks.md`'s blocking
criterion says must never sit out — its residue is a permanent `referenced`
event on a THIRD PARTY's issue — and it was defeated on **both** group words.

## Pattern, before → after

```
before  GATE_RE_GH_PR_CREATE_OR_MERGE="^gh${GATE_GH_C}[[:space:]]+pr[[:space:]]+(create|merge)([[:space:]]|$)"
after   GATE_RE_GH_PR_CREATE_OR_MERGE="^gh${GATE_GH_C}[[:space:]]+pr${GATE_GH_V}[[:space:]]+(create|merge)([[:space:]]|$)"

GATE_GH_V="${GATE_FLAGS:-}"      # the SAME absorber, one slot to the right
```

The issue quoted `^gh( -C)? +pr +(create|merge)`. That text is from an **old
revision** and is not what `main` carries; the behaviour it described is real
and reproduced above.

It is the same constant, not a variant, and that is the point: the stopping rule
`GATE_FLAGS` relies on — *a bare token in FIRST position IS the subcommand* — is
gh's grammar here too, so `gh pr list` / `gh pr view 42` still match nothing.
`GATE_FLAGS` is `(...)?`, so inserting it can only ADD matches: a strict
superset by construction, and the differential fence reports **0 lost cells**.

`gate_dequote_structural` takes the symmetric half. It consumed the token
straight after the group word, so `gh pr -R <slug> "merge" 42` rewrote `-R`,
stopped, and left the QUOTED verb undequoted while the left slot handled the
same shape. It now walks flag tokens to the first BARE one.

**Two later rounds found defects that walk INTRODUCED**, both fixed here:

- it rewrote every token it stepped over, so an unenumerated flag's QUOTED VALUE
  was dequoted and taken as the verb. `gh pr --search "merge" list` and
  `gh pr -R o/r --json "merge" view 42` went nomatch → MATCH against
  `origin/main` — the second a READ command arming ci-green, pr-review and the
  four integ gates with the selector answering 42. A quoted token after an
  unenumerated flag is emitted VERBATIM now; only a BARE token becomes the verb,
  so the declared `list --label merge` over-approximation is unchanged. A flag
  that already carries its value (`--name=v`, `-Rvalue`) consumes nothing, or
  the glued spelling lost its verb.
- it took a PRIVATE `GATE_STRUCT_MAXTOK`, so one segment could buy 24
  `_gate_struct_next` full-string regexes **per slot** instead of 24 in total.
  That walk is quadratic in the remaining string, so the cost lands on command
  LENGTH — see Latency below. It shares the outer `n` now.

## Positional-family survey, per constant

| constant family | pins a word position? | defeated? | disposition |
| --- | --- | --- | --- |
| `GATE_RE_GH_PR_{CREATE,EDIT,MERGE,CREATE_OR_MERGE,MERGE_OR_EDIT,WRITE}` | `gh … pr <verb>` | **yes** | fixed |
| `GATE_RE_GH_{ISSUE_CREATE,ISSUE_EDIT}` | `gh … issue <verb>` | **yes** | fixed |
| `GATE_RE_GH_{LABEL,BODY,PROSE}_CARRIER` | `(pr\|issue\|release) <verb>` | **yes** | fixed |
| `GATE_RE_GH_API` | `gh <flags> api` — one word | no | unchanged |
| all 14 `GATE_RE_GIT_*` | `git <flags> <verb>` — one word | no; there is no second word | unchanged |
| `GATE_RE_GIT_REBASE` | strict `GATE_GIT_GLOBAL` allowlist | no; narrowness is deliberate | unchanged |
| `GATE_RE_CDK_{DEPLOY,DESTROY}` | `(npx )?cdk <flags> <verb>` | same shape (`npx -y cdk deploy` misses) | **no live reader** — noted at the constant, NOT widened |
| `GATE_RE_VP_RUN_TEST` | `vp run test` — two literal words | same shape (`vp run --silent test` misses) | **no live reader** (`vp-run-test-path-gate` deleted by go-to-k/cdkd#2717) — noted, NOT widened |
| `GATE_RE_DELSTACK` | bare command word | no | unchanged |

`ci-green-gate.sh` does **not** match separately (the issue's guess): it reads
the shared `GATE_RE_GH_PR_MERGE` and is fixed by the widening.

**Two gate-local EREs DO share the defect**, which the first draft of this
survey wrongly denied: `main-tree-git-cwd-detector`'s `GH_PR_MERGE_VERB` and
`post-merge-sync-reminder`'s inline pattern are nomatch on every between-slot
spelling. Filed as go-to-k/cdkd#3266, **not fixed here**: both are non-blocking
PostToolUse informers (a missed reminder, never an ungated command), they carry
a deliberately different flag-VALUE contract, and they sit outside this fence's
population by construction. The claim that holds is the one go-to-k/cdkd#2614
earned — no *blocking* hook parses a command outside the shared matcher.

**go-to-k/cdkd#3273** is the other deferral, from the round-2 security review:
`ci-green` and `pr-review` read the PR NUMBER from the command but never the
`-R` slug, so they judge the cwd repo's PR of that number —
`gh pr -R go-to-k/cdk-local merge 42` from a cdkd worktree asks about cdkd's 42.
Pre-existing for the LEFT slot; this change makes it reachable by one more
spelling, which is still better than the no-gate state it replaces.
`verify-pr-gate` walks the whole argv and is unaffected.

**go-to-k/cdkd#3284** is the third deferral: a verb quoted AFTER an unenumerated
flag's value (`gh pr --json url "merge" 42`, `-q .x "merge"`, `-t x "merge"`,
`gh issue -R o/r --json url "create"`) is still not dequoted. Measured identical
on `origin/main`, so not a regression — it bounds how much of the class this PR
closed. The issue carries the rule that closes it without enumeration (measured
on gh 2.92.0: `gh pr --json number "view" N` resolves while `gh pr --web "view"`
errors `unknown flag`, so an unknown flag there consumes exactly one token
verbatim) and the one-condition fix shape. Not taken here: it is a third change
to the shared dequote walk, which already produced two of this PR's own
blockers, and verify.md 8-a says the structural fix does not go in late in a
cascade.

## Readers re-verified

Eleven hooks read one of the fixed constants: `bughunt-clean`, `ci-green`,
`gated-command-preamble`, `gh-label-validity`, `integ-broad`, `integ-destroy`,
`integ-local`, `integ-schema-migration`, `pr-body-item-number`, `pr-review`,
`verify-pr`.

| suite | tally |
| --- | --- |
| `lib/command-match.test.sh` | 916 / 0 (was 655) |
| `lib/command-match-differential.test.sh` | 26 / 0, **0 lost cells** |
| `verify-pr-gate.test.sh` | 113 / 0 (was 104) |
| `bughunt-clean-gate.test.sh` | 29 / 0 (was 23) |
| `ci-green-gate.test.sh` | 35 / 0 (was 22) |
| `pr-body-item-number-gate.test.sh` | 56 / 0 (was 50) |
| all six suites under bash **3.2** | identical tallies |
| `gh-label-validity-gate.test.sh` | 13 / 0 (was 9) |
| `main-tree-edit-oracle.test.sh` | 4 / 4, **0 fail-opens**, 0 latency breaches |
| unit suite, post-rebase | 1035 files / 24188 tests, 0 type errors |

## The fence

A **family fence**, not a case per gate per spelling:

1. **Population from the library SOURCE** — every `GATE_RE_GH_*` assignment is
   read out of `command-match.sh`; any whose group word is not followed by
   `${GATE_GH_V` fails, naming the constant. That catches the constant written
   next month by someone copying a neighbour, which no dynamic case can.
2. **Cross product** — 26 (group, verb, constant) triples × 7 flag spellings,
   each asserted to MATCH, plus a left-slot control per row.
3. **The other direction** — gh READ verbs × flag spellings asserted NOT to
   match, `merged-at` / `created-by` for the trailing terminator, quoted-mention
   guards, and the declared accepted false refusals.

**Corrections reviewers forced, all worth reading.** The axis-size checks sat
AFTER the loop they guard, so under `set -u` an emptied array aborted the file
at the `for` line — red under bash 5.x for a stated reason, red under **3.2 by
accident** saying nothing about an axis. They run BEFORE the loops now. The normalisation
first rewrote only the two literal alternations in the tree and was defeated in
one line by `(pr|issue|release)`; it now collapses any parenthesised alternation
of group words. And the block cardinality guard derived its expected count from
the very arrays whose collapse it detects — emptying one shrank both sides and
it reported `ok … ran all 41 cases`. **The axis sizes and the product are
LITERALS now.** Both are the "population derived from the thing being checked"
defect arriving inside the fence written to avoid it.

The differential's generated corpus gained a flag-POSITION axis (136 → 272
inputs) because every gh row in it put the token left of the group word — the
grid was one POSITION behind, not one case behind — and its `gained` count is
now floored rather than merely printed.

**A fourth round hardened the fence itself**, which is the artifact whose job is
to stop this class returning:

- **The lint was a blacklist of two remembered shapes.** Four synthetic
  un-widened constants all passed it — `(pr|label)…` (an alternation carrying a
  word the normaliser did not know), `gist…` (a group word outside the hardcoded
  three), `pr +create` (a literal space), and `pr${GATE_NOPE:-}…` (a FAKE
  absorber) — while its failure message claimed to check for `${GATE_GH_V`. It
  asks **positively** now (the literal absorber must follow the group word), over
  gh's **real** command groups, so an unknown spelling fails closed. All four
  are probed and caught. *(My first probe run was itself vacuous — the names
  carried digits and `^GATE_RE_GH_[A-Z_]+=` never matched them, so the
  population stayed 12. The population count is the receipt.)*
- **The family block pinned the verb ALTERNATION, not the absorber's SHAPE.**
  Relaxing `GATE_GH_V` to any token sequence reddened quoted-mention and
  selector cases elsewhere and **nothing in the block** — every case there
  carries a real `-R`, so the shape was never exercised. Three BARE-token shapes
  (`gh pr list merge 42`, and the issue and release twins) now red under exactly
  that mutant. The comment that conflated the two is corrected.
- **`gh-label-validity-gate`** is the live reader of `GATE_RE_GH_LABEL_CARRIER`
  and had **no** between-slot case, so stripping the absorber from that constant
  reddened the family fence and left its own suite green. Four cases added.

**A known bound, stated rather than detected**: this fence's subject is the
absorber, not each constant's verb list. Widening the verb list of one of the
four constants with **no live reader** (`PR_EDIT`, `MERGE_OR_EDIT`,
`ISSUE_EDIT`, `ISSUE_CREATE`) is invisible to every suite here. Wiring a reader
to one of them owes its own case.

## The accepted false-refusal cost, corrected TWICE

The first declaration said "exactly one shape". **That was false.** The right
slot is structurally more permissive: the absorber sits between the group word
and the VERB, so an arbitrary run of tokens may intervene, and only the VERB
WORD need appear later — where the left slot needed the literal `pr merge`. So a
SINGLE-quoted body carrying the word reaches the gates:

```
gh pr -R <slug> comment 3242 --body 'Ready to merge once CI is green'   -> refused
gh pr -R <slug> comment 3242 --body "Ready to merge once CI is green"   -> inert
```

A second review round showed even that was too narrow: the class is any BARE or
SINGLE-QUOTED **strictly interior** verb word after a between-slot flag, not
`--body` alone. Surveyed against `origin/main` over ~70 realistic read/comment
shapes, **9 newly refuse** (`list --search merge`, `--label merge`,
`--template create`, `comment -b 'LGTM, merge it'`, `issue list --label create`,
…). The EDGES are tighter than any prose about them claimed, and are now three
cases rather than a sentence: `--search 'create'` (first word) and
`--search 'x create'` (**last** word) are both inert — the verb alternation is
followed by `([[:space:]]|$)` and the character after a span's final word is the
closing quote — while `--search 'x create y'` matches.

The trade is **kept**, on the precedent `_GATE_WORD_LOOSE_FLAG` already sets in
this file (a false refusal is loud and one rephrase away; a bypass is silent),
and the double-quoted twin is pinned inert so the rephrase is real. Priced
rather than argued: the between-slot spelling appears **0 times** in this repo's
own instructions outside the rules describing this bug (left-slot 6, post-verb
18). `false-refusal-survey.sh` gained between-slot write AND read shapes — its
five gh shapes were all left-slot and so structurally blind to this widening.

## Mutation receipts

Every probe restored byte-exactly (`git diff` empty afterwards).

| mutation | must red | observed |
| --- | --- | --- |
| `GATE_GH_V=""` | everything | command-match FAIL; verify-pr 52/61, bughunt 17/12, ci-green 15/18, pr-body 31/25 |
| the 11 insertions removed, definition kept | the widening alone | 121 cells; verify-pr 107/**6**, bughunt 25/**4**, ci-green 26/**7**, pr-body 52/**4** — exactly the new cases |
| absorber removed from ONE constant | part 1 names it | `absorber missing from: GATE_RE_GH_PR_EDIT (group word 'pr' not followed by ${GATE_GH_V)` |
| population grep repointed | the floor, not a clean run | `only 0 GATE_RE_GH_* constants found … the scan collapsed` |
| verb alternation → `[a-z-]+` | part 3 polarity | 4 family cells + verify-pr 108/5, every `passes through` case 0 → 2 |
| trailing `([[:space:]]\|$)` dropped | the prefix case | `merged-at is not merge (want 1, got 0)` |
| `gen_gh_groups` emptied | the gained floor | `only 136 gained, floor 250` |
| dequote walk reverted | the quoted-verb cases | 5 cases red; real gate `gh pr -R … "merge" 42` 2 → 0 |
| ci-green selector → first digits in the command | the go-to-k/cdkd#2129 class | reds **only** the two new DECOY rows — the original five all stayed green, which is why the decoys were added |
| ci-green argv expectation 9999 | vacuity | `gate asked gh about PR '2195'` — the trace is real |
| dequote walk's round-2 branch disabled | the quoted-flag-value cases | 3 cases red |
| private `__dq_hops` budget reinstated | the shared-budget cases | both-slots k=12 goes nomatch → MATCH |
| one flag spelling dropped from `__ghv_flags` | the axis guard, BY REASON, on both bashes | `holds 6 spellings, expected exactly 7`, 881/2 under 5.x AND 3.2 |
| a constant appended as `pr([[:space:]]+create)` | part 1's new paren arm | `GATE_RE_GH_PROBE (group word 'pr' followed by '(' with no ${GATE_GH_V)` |

## Latency — the first measurement was taken on the wrong axis

Two `GATE_FLAGS` copies in one regex is the shape that backtracks, and a hook
killed by the 10 s PreToolUse timeout is a SILENT PASS. The regex itself is
linear: 3,813 B regex-only 0.0030 → 0.0085 s; 9,409 B through `gate_matches`
0.073 → 0.093 s.

**That measurement missed the axis that mattered.** It was taken at ~9 KB, and
`_gate_struct_next` ERE-matches the ENTIRE remaining string per token, so the
real cost is token count × command LENGTH. With a private budget the walk could
run 24 tokens per slot, doubling the product. Measured through the real
`pr-body-item-number-gate`, both slots filled:

| tail | `origin/main` | private budget | shared `n` (shipped) |
| --- | --- | --- | --- |
| 150 KB | 4.27 s | 7.02 s | **5.09 s** |
| 200 KB | 7.47 s | 8.98 s | **6.77 s** |

The cheapest input that KILLS the hook fell from ~260 KB to ~155 KB — a silent
pass disarming every gate at once, the class `GATE_STRUCT_MAXTOK`'s own header
records as re-opened once and reverted. Sharing the outer `n` restores it.

**The fence for this is deterministic, not a clock.** A timing case was written
first and REJECTED: a reviewer reinstated the private budget and the suite
stayed green at 2–3 s against an 8 s budget, because the doubling is worth only
+1–2 s and machine noise exceeds it. The reviewer's proposed replacement — `k`
flags in the RIGHT slot alone — was measured here and does not discriminate
either (shared and private give byte-identical verdicts for k=1..30, because one
slot never exceeds one budget). What does discriminate is filling BOTH slots:
shared `n` matches to k=11 and stops at k=12, while a private budget still
matches at k=20. Three cases straddle that boundary, with `k` derived from
`GATE_STRUCT_MAXTOK` so retuning the cap moves the fence instead of retiring it.

## This PR gates its own creation

The pattern changed here is the one `verify-pr-gate` keys on, so `gh pr create`
for this PR goes through the very constant being widened. After the change
`GATE_RE_GH_PR_CREATE_OR_MERGE` matches the plain `gh pr create …` used to open
it **and** `gh pr -R <slug> create …`, `gh pr --repo=<slug> create …`,
`gh pr -R<slug> create …` and the quoted-verb `gh pr -R <slug> "create" …` —
where before only the first two spellings armed it. The gate did fire for this
PR: `verify-pr` was bound to the final HEAD before `gh pr create`, which is
what let it through.

## Not done

No `src/**` change, so no integ and no changelog entry (`.claude/hooks/**` and
`.claude/rules/**` write none per CLAUDE.md). `pr-review` is deliberately NOT
set by this lane.

https://claude.ai/code/session_01NKuq3XKEPEytBDbWD3xG4W


go-to-k:
Merged — thanks. Three things this lane did that are worth carrying: it caught its own first lint-probe run as vacuous (digit-bearing synthetic names never matched the population regex, and the population COUNT was the receipt); it rejected two reviewer proposals on measurement rather than implementing them; and it closed the lint by asking POSITIVELY over gh's real command groups, so an unknown group word fails closed instead of passing silently.

Issue #3218, opened 2026-09-16
fix(resolver): resolveValue's object walk has no per-key try, so a failing key abandons a nested secret reference one level down


go-to-k:
## Summary

`IntrinsicFunctionResolver.resolveValue`'s object walk is a sequential
`for await` over `Object.entries` with **no per-key `try`**:

```ts
for (const [key, val] of Object.entries(obj)) {
  const resolvedVal = await this.resolveValue(val, context);
```

So one key's failure abandons every later key in the same object. Arrays are
NOT affected — `allSettledKeepingFirstRejection` settles every element before
rethrowing — but objects are.

go-to-k/cdkd#3196 scoped `cdkd scrub`'s resolve to each TOP-LEVEL property,
which stops a failure in property A from abandoning property B. It does not
reach one level down:

```json
"Environment": {"Variables": {"A": {"Ref": "NoSuchThing"},
                              "B": "{{resolve:secretsmanager:prod/db:SecretString:password}}"}}
```

That is ONE resolve unit (`Environment`). `A` throws, `B` is never fetched,
records no needle, and its stored plaintext is never rewritten — the identical
defeat recipe with the keys moved inside `Variables`. It is the dominant
secret-bearing Lambda shape.

## What it is NOT

- **Not go-to-k/cdkd#3196**, which is the top-level `Properties` bag.
- **Not go-to-k/cdkd#3181**, which is the per-TOKEN loop INSIDE one string leaf
  (`resolveDynamicReferences`).

This sits between them: the per-KEY walk of a nested object. Fixing either
sibling leaves this standing.

## Severity, and why it is not higher

Visibility SURVIVES. `carriesDynamicReference` walks deep, so the abandoned
scan is still detected and the record is still named in a warning at default
verbosity (`resource 'Fn' property 'Environment'`) — go-to-k/cdkd#3178's
machinery reports it. What is lost is the RESOLUTION: no re-run repairs the
leaf, because the walk stops in the same place every time.

## Suggested direction

A per-key `try` in the object walk, recording the failure and continuing, with
the failures surfaced to the caller so the existing refusals still refuse. Same
shape as go-to-k/cdkd#3181's remedy and subject to the same two costs, which
are why neither is a drive-by:

1. `src/deployment/intrinsic-function-resolver.ts` is in `integ-broad`'s
   `CROSS_CUTTING_REGEX` with no hunk filter, so merging needs a broad-set
   real-AWS run.
[...]

Issue #3181, opened 2026-09-15
fix(resolver): resolveDynamicReferences has no per-token try, so one failing reference abandons every later one in the same leaf


go-to-k:
## Summary

`IntrinsicFunctionResolver.resolveDynamicReferencesWithLogTwin` collects every
`{{resolve:...}}` token in a leaf and then walks them in ONE loop with **no
per-token `try`**. Any throw inside it abandons every remaining token in that
leaf, so a reference AFTER a failing one is never fetched and records no needle.

go-to-k/cdkd#3160 called this "the better fix" and shipped the other half —
`cdkd scrub` now COUNTS such a leaf and refuses to report the stack clean
(go-to-k/cdkd#3178). That closes the false-clean VERDICT. It does not restore
the lost resolution, and nothing else tracks the loop itself, so this issue
exists to keep it tracked.

## Why the counted finding is not the whole answer

The counter reports the loss; it does not recover it. Concretely, go-to-k/cdkd#3178
had to make a token whose argument still holds an unsubstituted `${...}` WARN
rather than gate, and a two-argument `Fn::Sub` whose placeholder IS bound is an
under-count for the same reason — in every one of those cases the healthy
sibling tokens in the same leaf are lost too, purely because the loop stops.
With a per-token `try` they would resolve, record their needles, and the leaf
would be certified, so those proxies could be deleted rather than documented.

**Correction (go-to-k/cdkd#3178 spec review).** An earlier revision of this body
claimed the deploy path persists the later secret unredacted. That was not
demonstrated and is probably wrong for the ordinary shape: if the first token
throws, the later token is never FETCHED, so its plaintext exists in no bag,
`result` is discarded at the throw, and the caller keeps the raw
`{{resolve:...}}` token — which is what should be persisted.

The deploy-path exposure needs a DIFFERENT shape to be real, and it is not
proven here: an observed READBACK carrying the already-decrypted value, redacted
against a needle bag this loop failed to populate. `refuseUncertifiedReadbackPositions`
and the empty-map arms in `secret-redaction.ts` are what would have to be traced.
Treat that as an open question, not as a finding.

Severity below is therefore derived from the RESOLUTION LOSS alone — the scrub
verdict is now correct (go-to-k/cdkd#3178), but the leaf still goes unscanned on
both paths, so a legacy plaintext behind a failing token is not rewritten by any
number of re-runs.

## Suggested direction

Wrap the per-token body so a failing token is recorded and skipped rather than
[...]

PR #3275, opened 2026-09-16
fix(resolver): mask the Fn::GetStackOutput state read, chain and all

### LINKED ISSUE #3234: fix(resolver): a producer stack name assembled around a sub-floor secret prints in the clear from the uncaught Fn::GetStackOutput state read
## Summary

`resolveGetStackOutput` masks the producer stack name in every message it builds itself (`loggedStackName`, `src/deployment/intrinsic-function-resolver.ts:8312`), then hands the RAW name to the state read at `:8331-8333` with no `try` / `catch` around it:

```ts
const stateData = roleArn
  ? await this.getCrossAccountStackState(roleArn, stackName, region, context)
  : await this.getSameAccountStackState(stackName, region, context);
```

`getSameAccountStackState` is `return context.stateBackend.getState(stackName, region)` (`:8538`). `S3StateBackend.getState` throws with the name interpolated (`src/state/s3-state-backend.ts:432`):

```ts
`Failed to get state for stack '${shownStackName}' (${shownRegionName}): ${detail}`
```

`shownStackName` comes from `displayName` / `displaySafe`, which are ASCII sanitizers, not secret maskers. So a stack name a template ASSEMBLED around a recorded secret is quoted back in the clear by a module that holds no secrets bag, and the resolver never masks it because the sentence is composed elsewhere and carries no log twin.

The sibling `Fn::ImportValue` path already does the right thing at `:7787`, masking both the stack name and the caught message with a comment citing go-to-k/cdkd#2827 — so this is a site that was MISSED rather than a pattern that does not exist here.

## Shape that reaches it

```json
{
  "Conditions": {
    "Probe": {
      "Fn::Equals": [
        { "Fn::GetStackOutput": {
            "StackName": { "Fn::Sub": ["prod-${P}", { "P": "{{resolve:secretsmanager:s:SecretString:tag}}" }] },
            "OutputName": "X" } },
        "y"
      ]
    }
  }
}
```

with `tag` resolving to a two-character secret `q7`. The pass registers the twin `prod-***` for `prod-q7`. S3 answers `AccessDenied` (NOT `NoSuchKey`, which returns `undefined` and takes the CloudFormation fallback), the `StateError` propagates out of `resolveGetStackOutput` into `evaluateConditions`' catch, and `maskSecretsForLog` finds no registered twin for the composed sentence and falls to the needle mask, whose substring arm has a four-character floor. At DEFAULT verbosity:

```
Failed to evaluate condition Probe: Failed to get state for stack 'prod-q7' (us-east-1): Access Denied, assuming false
```

`evaluateConditions` is one sink; the same uncaught read is reached from every other `Fn::GetStackOutput` caller, so a resource-property resolution surfaces it as a thrown message instead.

## Why it is filed apart from go-to-k/cdkd#3150 and go-to-k/cdkd#3171

- go-to-k/cdkd#3150 closed the names the RESOLVER builds into its own text. This name is built into text `src/state/s3-state-backend.ts` composes, which no twin lookup covers.
- go-to-k/cdkd#3171 is text an AWS SDK authored and the resolver only forwards. This one is cdkd-authored, in another module, and the name it quotes DOES have a twin the throw site simply never consults — which is why the remedy differs: go-to-k/cdkd#3171 needs the needle floor or an SDK-text policy, this needs a catch at the call site.

Not a regression: the same uncaught read and the same throw are on `origin/main`.

## Proposed fix

Wrap the two reads at `:8331-8333` in a catch that rethrows with `this.maskSecretsForLog(error.message, context)`, mirroring `:7787`. `getCrossAccountStackState` has the same gap and the same remedy. Keep the `cause` unmasked so the retry classifiers still see `$metadata`.

## Verification plan

- A unit case in the shape of `tests/unit/deployment/intrinsic-resolver-name-argument-log-twin.test.ts`: an `Fn::Sub`-assembled two-character secret in `StackName`, the state backend mocked to reject with a non-`NoSuchKey` error, asserting the WHOLE warn sentence with `***` at the secret's position.
- The same through a resource property rather than a condition, so the thrown-message sink is pinned too.
- A control where the assembled name carries no recorded secret and prints unchanged.
- The cross-account arm with its own case.
- A mutation probe per arm: dropping the mask reds its case.

Dup-check: searched open and closed issues for `getStackOutput state read mask stack name`, `StateError mask secret stack name in:body`, `state backend stack name plaintext` and `GetStackOutput AccessDenied mask`. Only go-to-k/cdkd#3150 (closed by PR go-to-k/cdkd#3176, which masks the names the resolver itself builds) and go-to-k/cdkd#3171 (SDK-authored text) matched, and neither covers a sentence composed inside `s3-state-backend.ts`.

Session-fit: now (do it in this session) -- found while reviewing PR go-to-k/cdkd#3176 at head `4d290773`; `intrinsic-function-resolver.ts` and the name-argument log-twin suite were read and edited in this session, so the files the fix touches are warm.
Severity: medium -- a default-verbosity warn and a thrown message print a sub-floor secret assembled into a producer stack name; it needs a template that puts a secret in a stack name and an S3 failure that is not a plain miss.
Effort: small (S) -- one catch at each of two call sites plus unit cases; no integ fixture reaches an AccessDenied state read.
Estimate: ~1 h -- most of it is the unit cases and their mutation probes.

go-to-k:
Closes #3234

## Summary

`resolveGetStackOutput` masks every name it prints ITSELF, then hands the RAW `StackName` and `Region` to the state read — they are state-KEY segments, so they have to be raw — and `S3StateBackend` quotes them back through `displaySafe`, an ASCII sanitizer rather than a masker. The read was uncaught, so that sentence reached `evaluateConditions`' default-verbosity warn, and every other caller as a thrown message, with a 1-3 character secret an `Fn::Sub` had assembled into the name still readable.

Found while reviewing go-to-k/cdkd#3176, and pre-existing on `origin/main` rather than introduced there.

## The issue's own remedy was insufficient, in both halves

go-to-k/cdkd#3234 proposed masking `error.message` and keeping the `cause` unmasked "so the retry classifiers still see `$metadata`". A probe pins the first of two problems with that:

- `formatError` renders `` `Caused by: ${cause}` `` for a `CdkdError`, so masking only a fresh top-level message prints the original one link down.
- The S3 key embeds the stack name (`cdkd/<stack>/<region>/state.json`), so the AWS link quotes it too, not just cdkd's wrapper.

So the read is wrapped and the failure re-thrown as a masked **clone of the whole chain**. The clone keeps the class and every own descriptor, so `markNonRetryable`'s marker, `$metadata`, `Code` and `name` still reach the classifiers — which is what threading an unmasked cause was for — and `stack` is masked as well, since its first line embeds the message.

## Design

`maskSecretsInError` takes an optional `extraMask` text transform. The bags mask by VALUE and a sub-floor secret is below `MIN_NEEDLE_LENGTH`, so only the frame holding both the raw name and its masked log text can substitute one for the other. It runs BEFORE the bag pass, and each pair contributes the `displaySafe`-sanitized spelling as well, since that is what the reader is shown. The `DescribeStacks` fallback warn, reached from the same caller, takes the same pass — a real `AccessDenied` quotes the stack ARN back at default verbosity.

**Three review rounds each found a blocker inside the previous round's fix**, every time because the safety property was re-derived per branch and the next round found a shape the enumeration had missed: the sanitized spelling was not substituted at all; then an unmasked pair's entry rewrote the sanitized spelling back to the RAW one; then a masked pair's sanitized key was paired with the RAW twin. So the property is now stated once at the construction site and enforced there:

> a substitution's REPLACEMENT must be at least as sanitized and at least as masked as its KEY

stated PER ENTRY, with the composition explicitly out of scope and its residual named at that comment.

## Tests

Fifteen cases in `intrinsic-resolver-name-argument-log-twin.test.ts` — the thrown message, the cause chain, `stack`, controls, the region, descriptor preservation, a positive throttle classification, positional-before-bag ordering, two substitutions at once, the sanitized spelling, a name embedding the region, the CloudFormation fallback warn, and the cross-account arm — plus six direct cases for `maskSecretsInError`'s new transform. One of the fifteen replaced "add another example" with an assertion of the PROPERTY, over five name shapes and with no character whitelisted at all.

## Verification

Fourteen mutation probes, one change at a time with the file restored between them — eleven red, three survived.

- **Red**: bypassing the mask (5 cases); applying the positional pass to the top link only (the issue's own remedy); swapping the positional/bag order; dropping the sanitized spelling; dropping the CloudFormation positional pass; dropping the clone's descriptors; restoring expand-before-filter; pairing the sanitized key with the raw twin (three times, once per revision of the property assertion — the first two runs are what showed the assertion admitting the shape it existed to catch).
- **Survived, each resolved rather than left standing**: the `raw !== masked` filter is an EQUIVALENT mutant. The longest-key-first sort is defensive and no case here distinguishes it — recorded at the sort as measured, with the shape that WOULD diverge named. `shownMask === ''` is unreachable (any real mask contains `***`, which `displaySafe` preserves) and labelled defensive.

`/check`: `vp run check` 0 errors, `vp run typecheck:test` clean, `vp run build` clean, regenerated matrices produced no diff, full suite 1033 files / 24116 passed / 1 skipped / 0 failed.

No integ fixture reaches this path — it needs an S3 failure that is not a plain miss — so the unit cases carry it. `integ-destroy` / `integ-broad` / `integ-local` are fresh from this session's `lambda` and `local-invoke-from-state` runs and verify rc=0 against this branch.

## Docs

The rule file's `maskSecretsInError` signature is updated so it does not go stale, trimmed to 16 B because a parallel lane had spent part of that path's payload budget on main — the fence's rule is to fund your own addition rather than trim another lane's entry. The mechanism stays on the function's own doc.

## Residuals

Named in the changelog entry: a name the state backend LOGS directly rather than throwing (its legacy-layout warn) cannot be reached from the caller's catch at all; the `Fn::ImportValue` sibling masks its caught message with the bags alone, so the positional half is still open there; and a log twin keeps the template's literal parts verbatim, so the resolver's own lines can still print a control character a name carried — `maskSecretsForLog` masks, it does not sanitize.

A fourth, found by the review of this PR and filed rather than folded in: `recordOutputRead` pushes the RAW names into `state.outputReads`, and `redactStateForPersist` rebuilds only `resources` / `outputs` / `orphans`, so the same assembled secret persists in PLAINTEXT in `state.json` at any length — go-to-k/cdkd#3289. Pre-existing, orthogonal to the message surfaces this PR closes, and it needs the redaction-vs-resolvability question settled (those coordinates are read back, so they cannot simply become `***`).


go-to-k:
Operational notice, not a review finding — and it needs one check from you.

**An earlier draft of this change was sitting UNCOMMITTED in a different worktree** (`triggerfish`), on `src/deployment/intrinsic-function-resolver.ts`. I found it when starting an unrelated lane there (go-to-k/cdkd#3181 / go-to-k/cdkd#3218) and my own edits to the same file collided with it.

What I measured, comparing that worktree's copy against this PR's branch (`de512a6d`):

- 93 lines exist only on this branch;
- 26 lines exist only in the orphaned copy.

So this branch is substantially ahead, and the orphaned copy looks like a superseded earlier structuring of the same masking — but the 26 lines are not purely comment rewording, they include code (`const extraMask`, the `substitutions.length === 0` arm, the two-bag loop).

I have NOT modified, committed or discarded it. It is saved verbatim as a patch at:

```
<scratchpad>/orphaned-3234-draft-in-triggerfish.patch
```

**What I need from you:** confirm this branch already carries everything from that draft that should survive. If it does, say so and I will reset the `triggerfish` copy so that worktree is clean — that is the only thing blocking my lane. If something in it is NOT on this branch, take it from the patch first and tell me when you have.

I am not touching the file until you answer.

Issue #3192, opened 2026-09-16
fix(state): consumers outside the state views and cdkd diff still fabricate from an outputs bag that is not an object


go-to-k:
Sibling-site sweep of go-to-k/cdkd#3189, which guarded `cdkd diff`. One root cause, one sentence at every site: `StackState.outputs` is read out of an unchecked cast (`parseStateBody` validates the root object and the schema version and nothing inside), and every consumer outside the three state views (go-to-k/cdkd#3187) and `cdkd diff` (go-to-k/cdkd#3189) walks or indexes it without the plain-object test `isReadableBag` already exports. `Object.entries` / `Object.keys` enumerate a string or a list as readily as a map, and `in` throws on both.

Population, from `grep -rnE "(state|childStateData\.state|loaded\.state|got\.state)\.outputs" src/ | grep -E "Object\.(entries|keys|values)|in got\.state|hasOwnProperty|\.outputs\["` at `aced052a` plus the two `importableOutputKeys` readers it does not spell — re-run it rather than trusting this list. `src/cli/commands/state.ts:1697` is the one hit already guarded (emptied at the render entry by `repairRenderedContainers`, go-to-k/cdkd#3187).

Two sites are WRITE paths, which is what makes this worse than the read-side class:

- `src/analyzer/orphan-rewriter.ts:536` (`cdkd orphan`) rebuilds `newOutputs` from `Object.entries(state.outputs ?? {})` and the command SAVES the result. A `"abcdef"` bag is written back as a well-formed six-key map — the malformed record is the only signal anything is wrong, and laundering it into a legitimate-looking one is permanent, the exact danger `refuseMalformedState`'s own note records for the `resources` bag.
- `src/types/state.ts:735` `importableOutputKeys` returns `Object.keys(outputs)` for a pre-v9 record, and `importableOutputs` (`:742`) rebuilds the bag from it, feeding `ExportIndexStore.updateForStack` (`src/state/export-index-store.ts:638-678`). A string bag therefore publishes one fabricated export per character into the SHARED `cdkd/_index/<region>/exports.json`, which every consumer's `Fn::ImportValue` resolves against — cross-stack blast radius from one damaged record.

[...]

PR #3276, opened 2026-09-16
fix(state,deployment,local): stop six gate-scoped consumers fabricating from an outputs bag that is not an object

### LINKED ISSUE #3207: fix(state): six gate-scoped consumers still fabricate from an outputs bag that is not an object
Residual of go-to-k/cdkd#3192, measured while working it. Six consumers of `StackState.outputs` still fabricate from a bag that is not an object. Same root cause and same one-sentence fix as the six sites go-to-k/cdkd#3192 closed — a load-site guard through the shared `isReadableBag` / `hasReadableOutputs` predicates, refuse-or-repair per the table in `docs/design/3192-outputs-consumers.md` — but each file sits in a real-AWS integ gate scope, so the edit cannot land without a gate run. They are one issue rather than six for the reason `/work-issues` references/filing.md gives: one root cause, one sentence describes the fix at every site.

Each verified live at HEAD (PR go-to-k/cdkd#3206, `5b8207a9`), with the gate each drags re-derived from `.markgate.yml` and `.claude/hooks/integ-broad-gate.sh` rather than from prose:

- [ ] `src/cli/commands/local-state-loader.ts:614` — `outputName in got.state.outputs`, and `in` on a string is a bare `TypeError` (`'x' in 'abc'` measured in node). Gate: `integ-local` (`src/cli/commands/local-*.ts`). Its sibling `Fn::ImportValue` scan at `:592` is ALREADY covered by go-to-k/cdkd#3192's `importableOutputKeys` guard — only the `Fn::GetStackOutput` arm is open.
- [ ] `src/local/s3-local-state-provider.ts:85` — `Object.entries(loaded.state.outputs ?? {})` fabricates one local env var per character. Gate: `integ-local` (`src/local/**`).
- [ ] `src/provisioning/providers/nested-stack-provider.ts:681,683` — rebuilds a PARENT's `Outputs.<key>` attributes from a child's bag, so a six-character child bag becomes six fabricated parent attributes, persisted and readable by `Fn::GetAtt`. The highest-severity of them: fabricated values here reach live AWS calls. Gate: `integ-destroy` (`src/provisioning/providers/**`).
- [ ] `src/deployment/intrinsic-function-resolver.ts:8307` — the `Fn::GetStackOutput` arm. `Object.hasOwn('abcdef','0')` is TRUE, so `outputs['0']` resolves a fabricated cross-stack value into a consumer's template; `describeAvailableOutputs(Object.keys(outputs))` also lists the fabricated keys back to the user. NOT named by go-to-k/cdkd#3192's own grep, which filtered to lines carrying `Object.entries|keys|values` and so missed the local-variable indirection. Gate: `integ-broad` (`CROSS_CUTTING_REGEX`). Its `Fn::ImportValue` sibling at `:7662` is already covered.
- [ ] `src/deployment/deploy-engine.ts:3125` — `persistedOutputs = currentState.outputs ?? {}` on the no-change merge path. Gate: `integ-destroy` AND `integ-broad`.

What go-to-k/cdkd#3192 DID close, so the residue is unambiguous: `cdkd orphan`, `cdkd import` and `cdkd scrub` now refuse at the load; `importableOutputKeys` / `importableOutputs` fail closed (which also covers the `Fn::ImportValue` halves of the resolver and the local loader, listed above); the exports-index rebuild fails closed and warns; and `exportNames` no longer throws a bare `TypeError` on a non-array.

Dup-check: searched open issues for `outputs bag non-object`, `nested-stack-provider outputs`, `GetStackOutput malformed`, `s3-local-state-provider outputs`, and the body window for `state.outputs`, `isReadableBag`, `importableOutputKeys`. Hits: go-to-k/cdkd#3192 (the parent; PR go-to-k/cdkd#3206 REFS rather than closes it and leaves this residue), go-to-k/cdkd#3193 (a `cdkd diff` real-AWS integ arm, a different gap), go-to-k/cdkd#3202 (the `resources`-bag residual of go-to-k/cdkd#3018 — the same SHAPE one container over, not the same sites), go-to-k/cdkd#3191 (the `properties` bag). None covers these six sites.
Session-fit: next (not this session) — blocked on external input: every one of the six files sits in a real-AWS integ gate scope (`integ-local`, `integ-destroy`, `integ-broad`), and a real-AWS run is the orchestrator's serialized turn, not a lane's. The go-to-k/cdkd#3192 lane was explicitly scoped away from running one. Unblocks the moment a session is free to spend the gate runs; the code context is otherwise warm and the predicates already exist.
Severity: high — `nested-stack-provider.ts` persists fabricated parent attributes that `Fn::GetAtt` then resolves into live AWS calls, and the resolver arm resolves a fabricated cross-stack value into a consumer's template. Reaching any of it needs a hand-edited or truncated record, so no correct deploy is affected.
Effort: medium (M) — the guards themselves are small edits against predicates that already exist; the cost is the gate runs, and `integ-local` plus one broad-set fixture covers them all if the lane batches them.
Estimate: ~2 h — most of it the three real-AWS gate runs and their orphan verification, not the code.

---

**Added after review of PR go-to-k/cdkd#3206 (round 2).**

- [ ] `src/cli/commands/destroy-runner.ts:682` — `const needsStrongRefCheck = !!(state.outputs && Object.keys(state.outputs).length > 0);` fabricates a six-key "this stack has exports" verdict from a six-character bag, flipping the strong-reference check on a record that exports nothing. go-to-k/cdkd#3192 names this as its tenth site and defers it from the FIRST PASS only, not from the issue — so without this row it goes untracked when go-to-k/cdkd#3192 closes. go-to-k/cdkd#3161 is the `resources` container and will not guard it. Gates: `integ-destroy` (`.markgate.yml` include list) AND `integ-broad` (`CROSS_CUTTING_REGEX`), both verified.

**Landed in go-to-k/cdkd#3206, struck from this issue:** the
`hasReadableExportSet` residual this issue briefly tracked (an all-non-string
`exportNames` reading as "exports nothing") was FIXED in that PR's round-3
review, not deferred. The predicate is now
`length === 0 || some(isString)`.

**Withdrawn (review round 5 of PR go-to-k/cdkd#3206).** This issue briefly
carried a bullet deferring scrub's "records no FINDING" arm here. That
deferral rested on the premise that a clean exit over a damaged producer was
PRE-EXISTING; the merge base shows it was not (an array bag raised the
producer-plaintext refusal, a string bag threw an escaping TypeError), so
deferring it would have shipped a false clean introduced by go-to-k/cdkd#3206
itself. It is FIXED in that PR instead.

go-to-k:
## Summary

Six consumers of `StackState.outputs` stopped fabricating from a bag that is not a JSON object. They are the residue go-to-k/cdkd#3206 deliberately scoped out, because each file sits in a real-AWS integ gate scope.

`parseStateBody` validates nothing inside the state record's root, so `outputs` may be any JSON value — from a hand-edited record, a torn write, or anyone holding `s3:PutObject` on the state bucket. Enumerating a string yields one entry per character; `in` throws on a string, a number, a boolean and `null` but answers for a list.

## The site that outranks the rest

`resolveGetStackOutput` was the only reader in this class that **re-applies rather than displays**. `Object.hasOwn('abcdef','0')` is true, so a fabricated character resolved into a consumer's template and the deploy applied it to a live system, while `describeAvailableOutputs` echoed the fabricated keys back to the operator. Every other site fabricated a verdict or a value that was only shown.

## Per-site disposition

The write-capable sites refuse; the read-only ones read the bag as empty and warn. All three refuse helpers delegate to one shared `hasReadableOutputs` predicate — no second spelling of the test.

| Site | Disposition | Why |
| --- | --- | --- |
| `deploy-engine.ts` | REFUSE (`refuseMalformedOutputs`) | persists the record; repairing before a save is the laundering go-to-k/cdkd#3192 exists to stop |
| `destroy-runner.ts` | REFUSE (`refuseMalformedOutputsForDestroy`) | does not rewrite the bag but DELETES the record on the strength of it — `null`/number/boolean/`[]` read as "exports nothing", skipping the cross-stack check |
| `nested-stack-provider.ts` | REFUSE (`refuseMalformedNestedChildOutputs`) | write-capable through its caller; the parent's `Outputs.<Key>` attributes are rebuilt from the child's bag and persisted |
| `intrinsic-function-resolver.ts` | REFUSE (`MalformedProducerRecordRefusalError`) | re-applies; see above |
| `local-state-loader.ts` | read-as-empty + warn | read-only |
| `s3-local-state-provider.ts` | read-as-empty + warn | read-only |

An **absent** `outputs` field is never refused — cdkd writes such records on purpose, and `hasReadableOutputs` exempts `undefined` explicitly.

## Real-AWS gates this arms

Re-derived against `.markgate.yml`'s `include:` lists and the hook regexes at the merge base, not taken from the issue. Corrected in review: the first version of this table read `integ-destroy` off `integ-destroy-gate.sh`'s `filtered_delete` regex, which is a narrowing filter, not the marker's scope — the scope is the `include:` list, whose first entry is `src/provisioning/providers/**`.

| Gate | Armed by |
| --- | --- |
| `integ-broad` | `destroy-runner.ts`, `deploy-engine.ts`, `intrinsic-function-resolver.ts` |
| `integ-destroy` | **does NOT fire** — see below |
| `integ-local` | `local-state-loader.ts`, `s3-local-state-provider.ts` |
| `integ-schema-migration` | not armed — no `StackState.version` literal change |

`integ-destroy` is the one worth stating carefully, because the first two versions of this table got it wrong in opposite directions. All three of `destroy-runner.ts`, `deploy-engine.ts` and `src/provisioning/providers/**` are in its `.markgate.yml` `include:` list — but the hook additionally applies a HUNK filter (`delete|rollback|IMPLICIT_DELETE|hyperplane|DependencyViolation|ENI|detach` on a non-comment changed line). Run over this diff's three files: **zero matches**, so `delete_touch=0` and the hook exits 0. Nothing will block this merge on a destroy run.

That is a reason to schedule one deliberately, not a reason to skip it. This PR adds a hard refusal at the top of `runDestroyForStack`, inherited by every destroy route including a nested child reached through its parent — which is exactly the shape CLAUDE.md's "never merge a PR whose destroy path is unverified" exists for. A real-AWS destroy run is being taken before merge despite the gate being silent.

## Real-AWS and live verification

`local-invoke-from-state` ran clean: rc 0, 7 deleted, 0 errors, no state or lock key surviving, and the container / task-network / svc-network sweeps all empty. It sets `integ-local` and `integ-destroy`. An earlier attempt in the same session exited 1 at step 5 ONLY — a concurrent lane held the destroy lock while assertions 2 through 4c had all passed — and the account was verified clean before the re-run. The `deployments/*.jsonl` keys that survive under the fixture's prefix are event history, which destroy does not remove by design.

The destroy refusal was also live-tested against the shipped binary rather than argued from units, in both directions, with the probe's object versions swept afterwards so nothing readable is left:

- a planted record whose `outputs` is the string `abcdef` → **exit 1**, refused by name with `STATE_RESOURCES_MALFORMED`'s message, and the record NOT deleted;
- the control, the same record with `outputs` an empty object → **exit 0**, destroyed normally.

The control is the half that matters: cdkd writes empty-`outputs` records on purpose, so a guard that refused one would break ordinary destroys.

## Verification

One new test file per guarded site, plus additions to the shared bag suite. Full suite green — no count quoted here on purpose, since a branch that rebases keeps moving it and a stale one reads as a claim about THIS tree.

A later review round found the sibling-site half of that set incomplete: `local-state-loader.ts` warned about a damaged producer once per REFERENCE, where `scrub.ts`'s equivalent arm had already deduped to once per RECORD and says so in its own comment. It now dedupes too, keyed by record through the NUL separator that sibling uses — a printable one lets two records collide into a single key and drop the second warning, which is silence in the class this PR exists to make loud.

Three fences, each measured by a single mutation rather than asserted: three reads of one record warn once (removing the dedup makes it 3), two records in different regions warn twice (keying on the stack alone makes it 1), and one record reached through BOTH the exact and the case-folded recovery arm warns once (passing the caller's spelling instead of the record's makes it 2). The branch is flattened to one commit, so no sha is cited here.

One residue is filed rather than fixed: `local invoke-agentcore` builds two resolvers per boot, so the dedup's scope is the resolver where the honest scope is the run — go-to-k/cdkd#3293.

Two things in this PR were resolved by MEASURING rather than by choosing, and both are recorded at the line:

- **The rebase conflicted on `resolver-mask-coverage`'s triple.** Both sides had moved it for real and independent reasons — go-to-k/cdkd#3150 masked six expressions, this branch adds one throw — so neither side's numbers describe the merge. Main said 136/167/103, this branch said 137/163/109, and the combined tree measures **137/169/103**. Re-measured rather than keep-either, and the paragraph says so.
- **`in` versus `Object.hasOwn`.** Executed rather than reasoned: `in` throws on a string, a number, a boolean and `null`, and answers `true` for a list; `Object.hasOwn('abcdef','0')` is `true`. That difference is why the local reader and the deploy resolver failed in different ways, and the docs now say which is which.

**Claims I cannot reproduce read-only are marked as such rather than asserted**: the per-site red counts from the implementing lane's own probes were not re-measured here, and should be treated as unverified until a review round takes them.

## Docs

A `check-docs` pass over every rule satellite whose `paths:` glob claims these nine source files found four statements the diff contradicts — three of them written by this change itself, one (`cli-diff.md`) not touched by it but made false by it. All four are corrected in `d42c1e098`, with the reasoning in that commit message rather than here.

Closes #3207


go-to-k:
Heads-up on a file collision, not a review of this PR's substance.

This branch rewrites `.claude/rules/architecture.md:74` — the `**A deliberate REFUSAL propagates out of Fn::Sub**` bullet — to add the `MalformedProducerRecordRefusalError` clause for go-to-k/cdkd#3207. PR go-to-k/cdkd#3298 (issue go-to-k/cdkd#3295) MOVES that same line, verbatim, out of `architecture.md` into a new satellite `.claude/rules/intrinsic-refusals.md`, because `architecture.md`'s glob is `src/**/*.ts` and the 3,976 B bullet was being loaded on every `src/` touch in the repo.

Both of us edit only that one line in that file, so git will not report a conflict in the usual place, but a keep-either resolution loses one side's work: merging go-to-k/cdkd#3298 first and rebasing this branch naively re-adds the bullet to `architecture.md` and undoes the split, while merging this PR first and then go-to-k/cdkd#3298's "verbatim move" silently reverts your rewrite.

**The part that needs a decision on this side.** The satellite's `paths:` list is six literal files, so after go-to-k/cdkd#3298 the text loads for three of this PR's nine `src/**` files and NOT for the other six:

| this PR's `src/**` file | after go-to-k/cdkd#3298 |
| --- | --- |
| `src/deployment/intrinsic-function-resolver.ts` | keeps the text |
| `src/utils/error-handler.ts` | keeps the text |
| `src/cli/commands/scrub.ts` | keeps the text |
| `src/deployment/deploy-engine.ts` | loses it |
| `src/provisioning/providers/nested-stack-provider.ts` | loses it |
| `src/state/malformed-resources-bag.ts` | loses it |
| `src/cli/commands/destroy-runner.ts` | loses it |
| `src/cli/commands/local-state-loader.ts` | loses it |
| `src/local/s3-local-state-provider.ts` | loses it |

That matters because the clause this PR adds is ABOUT `malformed-resources-bag.ts`, and the moved text itself names `deploy-engine.ts` and `NestedStackProvider.create`. So simply moving the new clause into the satellite would land it where six of its own modules never load it — the under-loading the payload fence's own failure message warns about. (An earlier version of this comment said "your `src/**` files are unaffected"; that was wrong, and it is corrected here rather than in a follow-up so the thread reads as one instruction. Found by go-to-k/cdkd#3298's spec reviewer.)

**Suggested resolution, whichever order the two land in:**

- If go-to-k/cdkd#3298 lands first: put the new clause in `.claude/rules/intrinsic-refusals.md` AND add whichever of the six files genuinely need it to that file's `paths:` list — the list is wildcard-free and `REACH_FLOORS['intrinsic-refusals.md']` asserts its length EXACTLY, so adding a path is a deliberate one-line table update in the same commit. `src/deployment/deploy-engine.ts` is the one exception already decided: its budgeted payload row leaves 4,254 B of cap headroom against a 5,744 B satellite, so adding it breaches the cap; the mechanism is written at its own throw site instead. For any of the others, check its budgeted row's headroom first. Note that a rebase is needed regardless — this branch's `CORPUS_FILE_COUNT` is 60 against `origin/main`'s 62, and go-to-k/cdkd#3298 takes it to 63.
- If this PR lands first: go-to-k/cdkd#3298 re-takes the bullet from the rewritten line before moving it, and re-runs the same reach question against this PR's file list. That side will check.

No action is needed on this PR right now — this is a note so the rebase is not a surprise. go-to-k/cdkd#3298's body carries the same merge-order paragraph from its side.

PR #3279, opened 2026-09-16
fix(rollback): a replay arm restores or it does nothing, instead of sending an empty desired bag

### LINKED ISSUE #3203: fix(rollback): a replay arm hands a provider an EMPTY desired bag when previousState.properties is absent, stripping every property off a live resource

go-to-k:
## Summary

All three of the rollback executor's replay arms coalesced their desired bag (`desiredProps ?? {}`), and `{}` is not a no-op — it is a COMPLETE desired state saying "this resource has no properties". `JsonPatchGenerator.generatePatch` emits an `op: 'remove'` for every key the previous side holds and the desired side lacks, and the Cloud Control provider calls it with no empty-desired guard, so the rollback the user asked for STRIPPED the live resource. On the reverse-replacement arm the outcome is worse: `{}` CREATES a default-configured resource and the arm then DELETES the live one.

Each arm now refuses to replay such a bag, and **the disposition follows what the record HOLDS** — each half following an existing sibling rather than a new rule:

- **ABSENT** (`undefined`): nothing was ever recorded, so nothing is repairable and no retry can use it. SKIPPED, warned, the pass continues and the segment pops — what `skip-failed-absent` already does.
- **PRESENT but unusable** (`null`, a string, an array): content EXISTS and only its shape is wrong, so the journal is the only durable copy an operator can repair. REFUSED by throwing, which counts a FAILURE and so keeps the segment from popping — what `refuseMaskedReplayBaseline` does for the same class.

Either way the pass continues, and each message names the remedy its own arm supports: `cdkd deploy`, or `cdkd drift` on `--revert-failed`, where the op died mid-flight so `cdkd diff` (which reads `state.properties`, not AWS) would report no change.

**The predicate is a plain-object test, not an absence test, and a 0-red mutation row is what said so.** `=== undefined` was the obvious spelling and it is wrong: `null ?? {}` is `{}` as well, and a string / array bag reaches the provider VERBATIM. go-to-k/cdkd#3149 refuses those shapes at the parser for a JOURNAL-sourced record, but `previousState` is equally often STATE-sourced, and `parseStateBody` validates no inner shape by an explicit decision (go-to-k/cdkd#2947) — so this is the only boundary that sees them on that path. A present but empty `{}` still replays: a resource that really has no properties is a legitimate thing to restore to.

Recorded rather than changed: the `reverse-replacement-readopt` arm writes a `properties`-less record into `state.json`, but sends NO desired bag to a provider, so it is go-to-k/cdkd#3211's class rather than this one's.

## Review rounds

Four rounds, three reviewers each. Round 2 carried a BLOCKER — the throw's retry remedy was arm-blind, so on `--revert-failed` "re-run `cdkd rollback`" would have made the operator destroy the record the refusal had just preserved (a plain re-run replays the COMPLETED ops, finds no failures, and pops the whole segment). The retry phrase is now per-arm.

Round 4 corrected text the round-3 fixes introduced — a fix round's own rationale is its least-probed output, and two of these were measurably false:

- The reachability paragraph was true for two arms and false for the third. `replayFailedOperations` has ONE caller, fed by `parseRollbackJournal`, which already rejects a non-object `previousState.properties` over `failedOperations[]` — so that arm's THROW has no live producer and is kept for PARITY. Its SKIP half stays reachable, because the parser tolerates an ABSENT bag by explicit decision.
- The dead `?? {}`'s stated reason was wrong against tsc. It is kept because `resolveReplayProps`'s DECLARED return type is unconditionally `| undefined`, not because the property access fails to narrow.
- The retry phrase was asserted only on the failed-op arm, so rewriting the reverse-replacement arm's phrase to the `--revert-failed` one was 0 red. Now asserted on both.

Round 6, from the four round-5 reviews (spec / code / test / security — no blocker; spec judged `Closes #3203` earned, security judged it merge-safe on its axis). Every finding was one shape: a per-arm asymmetry fixed on ONE arm and left on its siblings.

- The `revert-failed-update` arm announced the force-revert and THEN refused — the announce-then-refuse round 3 fixed on the `revert` arm, flagged independently by the code, security and spec reviews. Guard moved above the line.
- `remedy` had the SAME arm-blind asymmetry round 4 fixed for `retry`, one field over. Two mutations were predicted 0 red by the test review and **measured 0 red** before fixing: the revert arm's remedy could be swapped for the `--revert-failed` one (both contain `cdkd deploy`, which was the whole assertion), and the reverse-replacement arm's remedy could be deleted outright. Both are now 1 red.
- `requireRestorableBaseline` took six positional arguments, three of them same-typed strings split by a `logger`. Now a named bag — so the swap the round-2 BLOCKER made is a **type error**, not a question of test coverage.
- Two further `?? {}` sites are equally dead since the guard and said nothing while the third carried the note.
- Three prose corrections, two of them to text round 4 itself introduced: the reachability paragraph's closing clause held for the THROW half only, and the suite header's "cdkd never writes this shape" is false — `reverse-replacement-readopt` writes exactly that record, as this PR's own commit body says.

## Follow-up

- go-to-k/cdkd#3338 — a SKIPPED replay op records no deployment event, and on the automatic path `autoRollbackClean` counts warnings as clean, so the journal that was its last trace is deleted. Raised by the round-5 security review (`m1`); it covers `skip-failed-absent` too, which predates this PR. Filed rather than widened here.

## Test plan

- [x] Unit tests pass (1033 files, 23,940 tests, 1 skipped) — one new suite, `rollback-executor-absent-replay-baseline.test.ts`
- [x] 17-row mutation table, one mutation per probe over the whole `tests/unit/`, tree restored byte-exact between probes, every row's count equal to its NAMED failing cases — both halves of the disposition pinned in BOTH directions. Re-measured after the rebase rather than carried over; three rows moved and none weakened. One row was re-taken because its named cases included two unrelated to the mutation (flakes under a loaded tree) — reading a row's count without its case names cannot tell those apart
- [x] Integration test `rollback-command` (the command driving these arms): PASS in 150 s, destroy 3 deleted / 0 errors, 0 orphans, 0 surviving object versions on both prefixes
- [x] Integration test `lambda` for `integ-broad` (`rollback-executor.ts` is cross-cutting, and a narrow feature integ does not satisfy that gate): PASS in 96 s, destroy 9 deleted / 0 errors, 0 orphans
- [x] Both integs RE-RUN after the rebase rather than re-marked — it pulled go-to-k/cdkd#3246 into `rollback-executor.ts`, inside `integ-destroy`'s scope, so the marker correctly went stale and the earlier runs no longer covered the tree being merged
- [x] Documentation: changelog entry added; no `.claude/rules/` note, measured rather than skipped — the clause belongs in `layout-deployment.md`, whose path budget for `src/deployment/secret-redaction.ts` sits 19 B under its cap, so even a 187 B pointer busts it. The mechanism lives in `requireRestorableBaseline`'s doc comment.

Closes #3203


go-to-k:
Merged as 6a6c489c7. Thanks for the review rounds on this one — they carried it.

What they caught, in order: a BLOCKER in round 2 (the throw's retry remedy was arm-blind, so on `--revert-failed` it would have told the operator to run the command that destroys the record the refusal had just preserved), then a run of defects that were each in the PREVIOUS round's own fix — the retry phrase pinned but not the remedy, the guard moved above one announce line but not its sibling's, a rationale comment that stated the inverse of the code it justified. Five of those were predicted as 0-red mutations and every one measured 0 red before it was fixed.

The last round's suggestion to pin the rendered messages as a unit rather than add a seventh needle is what closed the class: the same mutation goes from 0 to 3 red, and the review then verified the pin matches the render character-for-character and is not brittle in the bad direction.

Two follow-ups filed rather than folded in: go-to-k/cdkd#3338 (a SKIPPED replay op records no deployment event, and the automatic path then deletes the journal that was its last trace) and go-to-k/cdkd#3384 (six subprocess-spawning suites declare no per-test timeout — they flaked into two of this PR's mutation rows and inflated a published count, which is how that one was found).

Issue #3306, opened 2026-09-17
fix(secret-redaction): a sub-floor secret in a nested-stack parameter frame #3156 still refuses (nested part, Sub variable, Fn::If) persists in plaintext


nix-tkobayashi:
## Summary

go-to-k/cdkd#3156 made the nested-stack sub-floor carry (`recordNestedStackParameterExpressions` in `src/deployment/secret-redaction.ts`) certify an `Fn::Join` / `Fn::Sub` parameter frame on the resolver's record of that object's own resolution (`substitutedSpellingOf`), which closed the two spellings that issue named: a token spelling `ssm:` or leaving its service to a `Ref`, and a frame with a `Ref` OUTSIDE the token. Several neighbouring spellings are still refused, and a 1-3 character secret in any of them (below `MIN_NEEDLE_LENGTH`, 4) is still persisted in plaintext in the child's `state.json`, and for the outside-the-token shapes in the parent's row too, with no signal. The refusals are listed only in that function's docstring, under residual (e).

## The refused shapes

Each refusal is deliberate, and each falls back to the value scan's answer, which is silent on a sub-floor middle:

1. **A token inside a NESTED intrinsic part**: `Fn::Join ['', ['port:', {Fn::Sub: '{{resolve:ssm:/app/pin}}'}]]`. The outer object's own literal text spells no token, so `frameSpellingOf`'s span count refuses before any record is read; the replacement belongs to the inner object's record.
2. **A token held by a USED `Fn::Sub` string variable**: `Fn::Sub ['port:${V}', {V: '{{resolve:ssm:/app/pin}}'}]`. The template spells a placeholder, not a token.
3. **An `Fn::If` around the frame**: `intrinsicSkeletonSegments` cannot render an `Fn::If` source, so `frameSpellingOf` refuses before any record is read, and the record is stored under the selected branch object anyway.
4. **Two outside-the-token frames of ONE value through DIFFERENT tokens** in one row: counted unframed by the `unwrittenValues` rule, because the parent row's value scan would write one leaf's token onto the other.
5. **An affix holding a plaintext the bag holds** (another secret a nested part resolved): refused so the spelling cannot carry that plaintext as though it were an expression.

## What it costs

[...]

PR #3311, opened 2026-09-17
fix(deployment): carry a sub-floor secret in an ssm or outside-Ref intrinsic frame to a nested stack, on the leaf's own resolution

### LINKED ISSUE #3156: fix(secret-redaction): a sub-floor secret in a nested-stack parameter frame the carry refuses persists in plaintext, and a grandchild's log lines lose the mask
## Summary

`recordNestedStackParameterExpressions` (`src/deployment/secret-redaction.ts`) carries a nested-stack parameter's framed secret to the child as a whole-value entry (`port:q7 -> port:{{resolve:...}}`) only when the parent spells the frame in a way the carry certifies. Two spellings are REFUSED by design, as that function's docstring states under "(e)": an intrinsic frame with a non-literal part OUTSIDE the token (a region `Ref` after it — the nonliteral-frame shape go-to-k/cdkd#2745 defers), and an intrinsic frame whose token spells `ssm:` (or leaves its service to an intrinsic part). For a 1-3 character secret (below `MIN_NEEDLE_LENGTH`, 4) in such a frame, nothing else redacts the value either, so it is persisted in plaintext and printed at `--verbose`. The refusal is recorded only in that docstring; no issue tracks it.

## What it costs

1. **State.** Any child resource property that consumes the parameter (`{ Ref: <Param> }`, or a `Fn::Join` / `Fn::Sub` around it) persists the framed plaintext in the child's `state.json`. `recordInheritedParameterSecrets` records nothing for it (the value is not a whole-value entry and the middle is below the needle floor), and the frame and span arms have no token in the child's source to position. The parent's own record of the `AWS::CloudFormation::Stack` row is positioned by the frame arm for the `ssm` spelling; for the region-`Ref`-outside spelling the parent's row keeps the plaintext too.
2. **A grandchild's log lines.** go-to-k/cdkd#3114 lets a child mask such a value on its debug lines through the log twin its parent registered (the lookup reads the child's `inheritedSecrets`). When a MIDDLE stack passes the value on to a grandchild through an intrinsic, the middle's nested-stack resource bag holds the middle's registered twin but no `plaintext -> expression` pair, and the hand-off passes `inheritedSecrets` only when the bag's `size` is nonzero (`nested-stack-provider.ts` where it builds the child engine's options, and the matching `size > 0` tests in `deploy-engine.ts`). The grandchild then gets no inherited bag and prints the value on its `Parameter X: using user-provided value`, `Resolved Ref to parameter` and `Resolved Fn::Join:` lines. The middle's own nested-stack row also persists the value (point 1).

   That is the RE-WRAP spelling. The PASS-THROUGH spelling (`{ Ref: MiddleParam }` straight into the grandchild's `Parameters`, found in the go-to-k/cdkd#3154 review) is wider: nothing on the inherited-parameter path registers a twin (`recordInheritedParameterSecrets` does not, and neither does `resolveRef`'s parameter arm; `rememberLogTwin`'s only callers are `Fn::Join`, `Fn::Sub`, `splitLogTwins` and the dynamic-reference substitution), so the middle's bag never holds a twin for the value and the grandchild prints it WHETHER OR NOT that bag holds pairs. The bound is: the bag a child hands its nested stack is the child's own, so no twin registered above the child reaches a grandchild.

The `nested-stack-secret` integ's `SubFloorPinSsm` arm (go-to-k/cdkd#3114) avoids point 1 only by consuming the refused-frame parameter in a child CONDITION, which persists nothing.

## Proposed fix

Close point 1 at the carry, which also closes point 2: a whole-value entry for these frames makes the middle's bag non-empty, so the existing hand-off carries both the pair and the twin.

- For the `ssm` spelling: carry the entry when THIS PASS resolved the leaf's own token to the middle and classified it `SecureString`, rather than refusing every `ssm:` token. The process-wide verdict store alone does not establish that (it can hold another pass's verdict); the evidence has to be pass-local, as the frame arm's own pair check (`recordResolvedPair`, per map instance) is, plus the verdict for that same token. The docstring's stated reason for the refusal (keeping the frame arm's wrong-reference residual on the parent's record) has to be re-weighed for that narrower case.
- For a non-literal part outside the token: decide whether the RESOLVED value plus this pass's resolved pair can establish the frame without the source doing so; if not, record the refusal as a won't-fix and make the value's persistence loud (a warning naming the parameter) rather than silent.

If the carry cannot be widened safely, fall back to the narrower point-2 fix: hand the bag down when it holds pairs OR registered log twins (a small predicate beside `LOG_TWINS_BY_PASS`), and keep point 1 as the documented residual. That fallback closes only the RE-WRAP spelling: in the pass-through spelling there is no twin to hand down. The carry fix closes both, since a whole-value pair masks with no `MIN_NEEDLE_LENGTH` floor.

## Verification plan

- Unit, point 1: `recordNestedStackParameterExpressions` over each refused spelling with a 2-character secret, asserting the entry is recorded (or, for a spelling left refused, that the persistence warning fires), and a child `{ Ref }` leaf persisting the framed EXPRESSION.
- Unit, point 2: three real resolver contexts (parent -> middle -> grandchild) handing each nested-stack resource's bag down the way the engine does, asserting that the grandchild's three lines do not print the plaintext, failing on `origin/main`. The expected text depends on which fix lands: with the carry fix the value is a carried whole-value secret, so its lines read `***` (the whole-value mask, and a Join collapses the overlapping masks); only the twin-only fallback gives the positional `x-port:***`. Assert the text the chosen fix produces, whole.
- Integ: a new arm in `nested-stack-3level` on a SEPARATE nested-stack resource whose bag holds no other pairs (its existing `HandoffSecretA` / `HandoffSecretB` row already makes its bag non-empty, which would bypass point 2), with a premise that asserts that isolation from the synthesized template. The arm asserts the middle's and the grandchild's persisted leaves hold the framed expression, and the grandchild's `--verbose` lines present and masked.

Dup-check: searched open and closed issues for `grandchild nested stack secret`, `nested-stack-3level secret`, `inheritedSecrets size`, `log twin grandchild`, `three-level nested stack parameter`, `nonliteral frame plaintext child`, `refused frame nested stack plaintext state`, `ssm token frame nested stack state` and `intrinsic frame secret child state plaintext`. Related, not duplicates: go-to-k/cdkd#3094 (closed) gave `nested-stack-3level` a secret carried through three levels, over frames the carry certifies; go-to-k/cdkd#3062 (closed) added the object-spelled carry and recorded these two refusals in its docstring; go-to-k/cdkd#3114 is the child-level log fix point 2 extends; go-to-k/cdkd#3150 is the name-argument log sites.

Session-fit: now (do it in this session) -- `secret-redaction.ts`'s carry docstring, the resolver's twin lookup and both nested-stack fixtures were read in this session for go-to-k/cdkd#3114; the point-2 half builds on that PR's lookup, so this starts once it merges.
Severity: medium -- a 1-3 character SecureString persisted in plaintext in nested-stack state, reached only through the two frame spellings the carry refuses.
Effort: large (L) -- a change to the carry's refusal rules in `secret-redaction.ts`, possibly the hand-off gates in `deploy-engine.ts` / `nested-stack-provider.ts` (inside the `integ-destroy` / `integ-broad` scopes), and a new isolated arm in `nested-stack-3level`.
Estimate: ~4-6 h -- re-weighing the refusal's wrong-reference residual and the three-level integ arm take most of it.

nix-tkobayashi:
Closes #3156

## What was wrong

`recordNestedStackParameterExpressions` carries a nested-stack parameter's 1-3 character framed secret to the child as a whole-value entry (`port:q7 -> port:{{resolve:...}}`), but for an `Fn::Join` / `Fn::Sub` source it certified the frame only when the token spelled `secretsmanager:` / `ssm-secure:` AND the position pass had written the frame. Two spellings were refused:

- an `ssm:` token, or a service left to a `Ref`;
- a frame with a non-literal part OUTSIDE the token (a region `Ref` after it).

A secret in either was persisted in plaintext in the child, in the parent's own row (the second spelling), and on a grandchild's `Parameter` / `Resolved Ref to parameter` / `Resolved Fn::Join` debug lines, for the pass-through and the re-wrapped hand-off alike.

The `secretsmanager:` / `ssm-secure:` gate existed for a reason: the frame arm picks a candidate among RECORDED secret references, and a public `ssm` parameter records nothing, so a public leaf could be written as a SecureString sibling's reference. The issue's own proposal (carry the `ssm` spelling when this pass resolved the token and classified it SecureString) works for a wholly literal `ssm:` frame, where the recorder can read the leaf's own token off the source skeleton. It does not work for a service left to a `Ref` or a placeholder inside the token: there the only concrete token in view is the one the frame arm WROTE, and in the wrong-reference case that token IS a resolved SecureString. Nor does it work for a frame with a non-literal part outside the token, for a different reason: the frame arm writes nothing there, and although the token is readable off the skeleton, the source cannot supply the RESOLVED affix around it, which only the resolution record's input gives. The design below covers all of them.

## The fix

**The resolver records positive, per-leaf provenance.** `resolveDynamicReferencesWithLogTwin` now also returns every token it replaced with the verdict THAT replacement took (the fresh lookup's, the cache entry's own, or a region-pinned sibling's). `resolveJoin` / `resolveSub` assemble those into an `IntrinsicLeafResolution` stored per pass bag and per intrinsic OBJECT (`recordIntrinsicLeafResolution`, a double `WeakMap`). A second, different resolution of the same object poisons it for the pass.

**The carry certifies on that record where the spelled-secret arm refuses** (`substitutedSpellingOf`; the existing arm runs first, so nothing it carried before changes). The record certifies only when ALL of these hold:

- the resolution replaced every token it met, and exactly one;
- that replacement took a SECRET verdict;
- `singleSpanFrame` over the record's input names that token and value, i.e. the observed equation `value = input[token := plaintext]`;
- the frame's affix holds no plaintext the bag holds.

Where the position pass DID write the leaf, it must have written the same spelling. The record is the leaf's own because it is keyed by the very object the engine hands both the resolver and the carry. So an unrelated resolution with an equal output cannot answer for it, and an unclassifiable-then-public read of one token (equal plaintext) is not mistaken for a secret one.

**One parent-row hazard is refused explicitly.** A frame the pass did not write is read on the parent's record by the value scan alone, so two such leaves of one value through different tokens would both be written with one token. A value holding such a leaf counts as unframed unless every framed spelling of it carries one token.

**Point 2 needs no hand-off change.** With the whole-value entry in the top row's bag, the middle stack's `{ Ref }` copies it into the middle's own resource bag, which is then non-empty and handed down. The grandchild masks each line whole.

Still refused, and tracked as go-to-k/cdkd#3306:

- a token inside a nested intrinsic part, or in a used `Fn::Sub` variable;
- an `Fn::If` around the frame;
- an affix holding another recorded secret;
- two outside-the-token frames of one value through different tokens.

The design went through three Codex design rounds before any code: round 1 broke a public-exclusion design, round 2 an output-keyed provenance design, and round 3 found the two-token parent-row collapse above.

## Tests

- `tests/unit/deployment/nested-stack-refused-frame-carry.test.ts` (29 cases), all through the REAL resolver except five hand-built record cases (four unfaithful records and a faithful control). The resolver cases follow the engine's sequence: resolve the row, run the carry over the same source object and persist the row; the six spelling cases and the public-sibling refusal also persist a child `{ Ref }`.
  - each refused spelling: `ssm` token as a literal part, with a `Ref` inside, a service left to a `Ref`, an `Fn::Sub` placeholder inside the token, and a `Ref` or pseudo parameter outside the token; asserting the entry, the parent row and the child leaf;
  - refusals: a public leaf the frame arm wrote as a sibling's reference; an unclassifiable-then-public read of one token; a public cache hit beside an unclassifiable pair; two outside-the-token frames through different tokens (and the same-token control, which carries); a used variable holding another secret or the SAME token; an affix holding another secret (nested Join part, intrinsic Sub variable); a nested-part token; a used-variable token;
  - verdicts across resolutions: a cache hit into a second bag, an identical re-resolution, and a conflicting re-resolution in both orders;
  - hand-built unfaithful records (`complete: false`, a wrong `output`, a wrong token, a wrong value) refuse, beside a faithful control;
  - point 2: three levels, PASS-THROUGH and RE-WRAP, asserting the middle row's persisted expression and the grandchild's three lines whole (`***`).
- `tests/unit/deployment/nested-stack-refused-frame-carry-delegation.test.ts` (3 cases), with SDK client classes faked by region: a foreign-region SecureString frame carries; a foreign-region public parameter does not; a sibling's public answer beside an unused variable's unclassifiable pair does not.
- 18 of the 32 cases fail with `src/` reverted to this PR's base `65f41cf2` (re-measured with the two test files as they stand at `3f54946f` and after): every carry, the cache and re-resolution certifications, both three-level cases, and the five hand-built record cases (those five because the recorder export does not exist there). The other 14 are refusals that must hold before and after.

Mutation probes, one mutation at a time, tree restored byte-exact between them. Each of the following reds at least one case:

- forcing the delegated, cached or fresh verdict to secret;
- conflict as last-wins, and conflict as first-wins;
- removing the unwritten-value rule;
- dropping `Fn::Sub` used-variable substitutions, either at the flatten or at the per-use lookup;
- dropping `Fn::Join` part substitutions, or taking the Join input from resolved parts;
- removing the secret check, the count check, the affix check, the `complete` check, the `output` check or the token/value check;
- disabling the provenance arm.

The `written === substituted` check survives its removal. Its accepting side runs in the literal `ssm` spelling case; no case reaches its refusing side. The source says why. Two further hand-built rows (added in review) pin the token and value checks: removing that check reds exactly those two.

`tests/unit/deployment/secret-redaction-nested-parameter-source.test.ts` changes a comment: its hand-built bags carry no leaf record, so its refusal cases still hold. `intrinsic-resolver-nested-child-log-twin.test.ts`: its `ssm` case now puts the token in a nested `Fn::Sub` part, a frame the carry still refuses, and runs the carry to show that (value-only redaction stays silent on the value), so the go-to-k/cdkd#3114 twin lookup keeps a production-reachable anchor.

## Integ

- **`nested-stack-3level`**: a new `Framed` branch (root -> Framed -> FramedGrandchild) whose middle owns nothing but its nested-stack row (premises pin the row's exact `Parameters` and `TemplateURL`). The two frames are `pin3156s:{{resolve:ssm:...}}` with the account `Ref` inside, and `pin3156o:{{resolve:secretsmanager:...:pin}}@` + region `Ref`, each handed down pass-through and re-wrapped. It asserts the framed expression on the root row, the middle row and all four grandchild leaves; the live values; that every grandchild `Parameter` / `Ref` line ends in `***` (checked before the deploy output is echoed); and no framed plaintext in any output or state level. PASS (310s): the diff is clean, destroy leaves 10 SSM parameters and 2 SNS topics gone, and 0 state versions survive on all six prefixes. PASS again (239s) after the rebase onto go-to-k/cdkd#3280.
- **`nested-stack-secret`**: its go-to-k/cdkd#3114 `SubFloorPinSsm` lines are now expected masked WHOLE, a live signal that the carry certified that ssm frame. PASS (318s), and again (387s) on `00096ac4` after the rebase and review round 1, including the phase 2c UPDATE deploy.
- **`lambda`** (broad set): PASS (109s), and again (108s) after the rebase.

**Rebase onto go-to-k/cdkd#3280.** That PR wrapped the dynamic-reference token loop in a per-token `try` whose recovery arm, for a caller passing `abandonedResolutions`, leaves a token unreplaced. The conflict was resolved by re-applying this change onto the new loop, and the recovery arm now also marks the pass incomplete. A Codex review of the resolution found nothing, the suite passes on the rebased tree, and `nested-stack-3level` and `lambda` were re-run on it (above). `nested-stack-secret` was re-run on `00096ac4` (above).

The first `nested-stack-3level` run failed at its own premise: under the `cdkd deploy` synth, `Stack.of(this).account` is folded into the literal text, so the frame was a plain string. The fixture now uses `cdk.Aws.ACCOUNT_ID`.

## Review

- Codex delta loop: 5 rounds. Round 1: the count check was wrongly documented as redundant (a used variable holding the same token), and a docstring was misplaced. Round 3: a masked-line check matched by substring. Rounds 2, 4 and 5: NO ISSUES.
- Maintainer-checklist proxy pass: 4 rounds (fixture secret masking, premise invariants, test names, sibling comments, rules and changelog parity), ending NO ISSUES.
- Four reviewer agents (code, test, security, spec), before the PR opened, with no blockers. Their minors were fixed before opening: the affix refusal, the used-variable refusal case, the pinned `complete` / `output` checks, and the fixture checks. The remaining refused shapes were filed as go-to-k/cdkd#3306.
- Maintainer review, three rounds, no blockers in any:
  - Round 1: M0, a re-run of `nested-stack-secret` on the rebased tree, the only live exercise of the carry's UPDATE call site for the `ssm` spelling. Also the body wording, the go-to-k/cdkd#3114 fence repointed to a frame the carry still refuses, `source` made required, the affix scan run once per parameter, and two hand-built rows pinning the token and value checks.
  - Round 2: M1, the frame-spelling memo keyed on the parameter name alone. M2, two body corrections, including the revert count re-measured as 18 of 32. Also `CONFLICTING_PLAINTEXT` named in the reachability comment, and a literal `Control` frame proving the repointed case's carry walked the row.
  - Round 3: merged.
- The `wait-abandoned-guard-population.test.ts` timeouts seen at the default worker count are pre-existing (they arrived with go-to-k/cdkd#3249) and tracked as go-to-k/cdkd#3347.
- Not changed, with the reason:
  - **An unclassifiable `Type` counts as secret.** A single such read is carried as a SecureString would be. This is the resolver's existing fail-closed classification (go-to-k/cdkd#1901), not a choice made here.
  - **No wildcard tie between the skeleton's own token and the recorded token.** It is implied by the count: a token contributed by a non-literal part would be a second replacement beside the literal one.
  - **The unit three-level case mirrors the `size > 0` hand-off gate** instead of driving `NestedStackProvider`. The live `nested-stack-3level` run covers the real gate.
  - **The Framed arm runs a single deploy, so the carry's UPDATE call site** (`deploy-engine.ts`, the same function) is exercised live only by `nested-stack-secret`'s phase 2c, for the `ssm` spelling.


nix-tkobayashi:
Ready for review at head `f7a9b396`. That head is final: no further pushes are planned unless review asks for them.

The two force-pushes after opening were rebases onto `main` as go-to-k/cdkd#3280 and go-to-k/cdkd#3249 merged. Only the go-to-k/cdkd#3280 rebase needed conflict resolution. The branch is now one commit on top of `65f41cf2`.

- **The go-to-k/cdkd#3280 conflict** was in the dynamic-reference token loop. Its recovery arm leaves a token unreplaced, so it now also marks the pass incomplete. Nothing else in this change moved; the body's "Rebase onto go-to-k/cdkd#3280" paragraph has the detail.
- **The go-to-k/cdkd#3249 rebase** had no content conflict. GitHub still reported this PR as conflicting, likely over the integ ledger file, the only file both sides edit (not confirmed), so the branch was flattened and rebased to clear it.

On this head, CI is green, and the local suite, `nested-stack-3level` and `lambda` all pass.

go-to-k:
Reviewed at head `f7a9b3964`, four parallel reviewers (spec, code, test, security). **No blockers.** `Closes #3156` is earned: both costs the issue reports are closed — the `ssm` / service-left-to-a-`Ref` and outside-the-token spellings now carry, and the grandchild's `Parameter` / `Ref` / `Fn::Join` lines mask whole through the middle's copied entry.

Verified rather than taken on trust:

- every non-writing exit of the token loop clears `complete` — the `skipDynamicReferences` skip, the ssm-secure comparison path, the unsupported-service warn and #3280's recovery arm; both throw sites reach the `catch`, which either rethrows (no record at all) or clears it;
- the delegated arm's `foreign.substitutions.length === 1 && delegated!.secret` short-circuits before the assertion, so an empty array yields `secret: false` plus `complete = false`;
- `resolveJoin`'s `input` and `substitutions` both derive from `resolvedParts` in index order, so provenance cannot skew against the output;
- `unwrittenValues` can only add `UNFRAMED_SPELLING` to a value that was already unframed before this PR, so it cannot drop a mask that used to apply;
- the carried spelling is the leaf's OWN `input`, never a sibling's recorded reference, so `positionByIntrinsicFrame`'s wrong-reference residual is not inherited;
- every gate fails closed: missing record, `CONFLICTING_LEAF_RESOLUTION`, `complete: false`, count != 1, public verdict, `output` mismatch, an affix holding a bag plaintext, `written` != the record's spelling, and an absent `recordedSecretValues`.

**M0 — required before merge.** Re-run `nested-stack-secret` on this head and update its ledger row. It is the only live exercise of the carry's UPDATE call site for the `ssm` spelling — the spelling this PR newly certifies — and its recorded PASS (`2026-09-17T05:40:32Z`, tree `8b9e735e`) predates the #3280 rebase. `nested-stack-3level` did re-run on the rebased tree, but the Framed arm runs a single deploy, so it covers CREATE only. ~320 s.

Take-or-leave, none of them blocking:

- **m1** — PR body: "the only token the recorder sees is the one the frame arm WROTE" is stronger than the code. For a wholly literal `ssm:` frame the recorder reads the leaf's own token off the source skeleton (`src/deployment/secret-redaction.ts:2129-2137`, `ownToken`), exactly as the docstring at `:2062-2064` says. So #3156's own proposal was workable for that one sub-shape; it is genuinely unworkable for the service-left-to-a-`Ref` and outside-the-token shapes. Wording only — the design you substituted is a strict superset.
- **m2** — PR body: "`written === substituted` … unreached in this lane's tests" is imprecise. Its ACCEPTING arm is exercised by spelling case 1, where the frame arm does write that wholly-literal `ssm` frame. Only the refusing arm is unreached, and it looks structurally unreachable: a one-match pattern derived from the leaf's own skeleton can only match the leaf's own recorded token.
- **m3** — `tests/unit/deployment/intrinsic-resolver-nested-child-log-twin.test.ts:390`: its input shape is now carried in production, so the #3114 twin fence no longer has a production-reachable anchor. Pointing it at a shape #3306 still refuses — a nested-part token, or a used `Fn::Sub` variable — restores it.
- **m4** — small ones:
  - `recordLeafResolution`'s `source === undefined` branch is unreachable (both call sites at `intrinsic-function-resolver.ts:3837` / `:3841` always pass one), so `source` can be required;
  - `certifiedSpellingOf` re-invokes `frameSpellingOf`, re-running `substitutedSpellingOf`'s full `secrets.keys()` affix scan a second time per parameter;
  - the token/value check is pinnable in ~6 lines — two more hand-built rows with a mismatched `token` / `value`, beside the existing `complete` / `output` ones;
  - `tests/integration/nested-stack-3level/verify.sh:240-248`: `scan_output` greps only the framed plaintexts, so a line carrying a bare `Jx` / `Zq` is still echoed. `masked_whole` covers the parameter lines; a bare-value arm in `scan_output` would cover the rest.

Out of scope here: the five shapes on #3306. I re-derived each against this tip — the issue body is accurate on all five, and none of them got worse (each falls back to the pre-PR value-scan answer). #3306 also still holds #3156's "warn loudly on a residual" question.

Push M0's re-run and ledger row, plus whichever take-or-leave items you want, and I will merge.

nix-tkobayashi:
Thanks for the review. Pushed as `3f54946f` (code changes in `00096ac4`, ledger row on top). No other changes on this branch.

**M0.** `nested-stack-secret` re-run on `00096ac4`: PASS in 387s, including the phase 2c UPDATE deploy (`SubFloorPinSsm` lines masked whole on both arms), destroy with 0 errors and 0 surviving state versions. The ledger row is updated.

**m1** -- taken; the body now says the issue's proposal works for a wholly literal `ssm:` frame (the recorder reads that leaf's own token off the skeleton), and fails only for a service left to a `Ref`, a placeholder inside the token, and the outside-the-token frame.

**m2** -- taken for the body. It now says the accepting side runs in the literal `ssm` spelling case and no case reaches the refusing side. On reachability I kept a narrower claim in the source comment than "structurally unreachable". For a wholly literal token the frame arm's pattern matches only that token, as you say. But for a token with a placeholder, the arm drops a candidate whose map reverse index ties it to another plaintext, and it does that even while the pair is intact (the inherited-entry shape refusal 4 names). So the leaf's own token can be dropped, leaving a sibling as the one match. I have no test that reaches it.

**m3** -- taken. The `ssm` case in `intrinsic-resolver-nested-child-log-twin.test.ts` now puts its token in a nested `Fn::Sub` part, a shape go-to-k/cdkd#3306 still refuses. It runs the carry over that source and asserts value-only redaction stays silent on the value (no whole-value entry, no substring needle), so its masked child lines can only come from the parent's log twin. The file docstring names it as the one case that runs the carry.

**m4**
- `source` is required on `resolveJoin` / `resolveSub` / `recordLeafResolution`.
- `frameSpellingOf` is memoized per parameter name, so the affix scan runs once per parameter.
- Two hand-built rows pin the token and value checks: a wrong token, and a wrong value. Removing that check reds exactly those two (measured).
- **Not taken: the bare-value arm in `nested-stack-3level`'s `scan_output`.** `Jx` / `Zq` are two characters, and a grep for them over a `--verbose` deploy log can match unrelated text (a hash, an id, a key), which would fail a correct run. What would print a bare value is a parameter line not masked whole, and every grandchild `Parameter` / `Ref` line is already required to end in `***` before the output is echoed. Other lines stay covered only by the framed-value scan -- the residual you point at.

Checks on `00096ac4`:
- Codex on the delta, then a maintainer-checklist proxy pass. The proxy caught that the repointed test's premise excluded only a whole-value entry, which is now fixed.
- `vp run check`, typecheck and build are clean.
- The full suite passes (1057 files) with `--maxWorkers=50%`. At the default worker count, three self-probe cases in `wait-abandoned-guard-population.test.ts` timed out at 5000 ms twice (about 5.5 s each under load; about 1.5 s run alone). That file came in with go-to-k/cdkd#3249 and nothing in this change touches it.

go-to-k:
Round 2 at head `3f54946f6`, same four reviewers over the delta `f7a9b3964..3f54946f6`. **No blockers.** Round 1's items all landed, verified individually rather than taken from the reply:

- **M0** — the ledger row matches what the script asserts: `scan_verbose_output` requires every `Parameter SubFloorPinSsm:` / `Resolved Ref to parameter: SubFloorPinSsm ->` line to end in `***`, and it is called on both the phase 1 CREATE and the phase 2c UPDATE. `00096ac4` is the last commit touching `src/`, and `00096ac4..3f54946f6` is the TSV alone, so the run covers this head's code and its verify.sh. Naming the run's tree rather than the ledger commit matches the file's other rows.
- **m3** — the nested-part refusal is live on this tree (`intrinsicSkeletonSegments` renders the nested `Fn::Sub` as `UNKNOWN_PART`, so `spans.length !== 1` fires before the provenance arm), the premise now excludes both arms (`m8` is sub-floor, `port:m8` is not a key), the carry's argument shape matches `deploy-engine.ts`, and the masked lines still can only come from the twin.
- **m4** — the two new rows each trip a different half of the token/value check and nothing earlier refuses them; `source` is non-optional with both call sites provably passing a non-null object, so no route can reach `WeakMap.set(undefined)`.

**A correction on my side.** Round 1's m2 called the `written === substituted` refusing side "structurally unreachable". That was wrong, and your narrowing is right. `plaintextIndexOf` stores `CONFLICTING_PLAINTEXT`, a symbol, so the candidate filter drops the leaf's own token while `middle -> ownToken` is still in the map; a sibling then passes both the filter and `resolvedPlaintextOf`, which reads the pass-local pair table. The refusing arm is reached. Your source comment is accurate as written.

**M1 — required before merge.** `frameSpellingOf`'s memo is keyed on `name` while the function still takes `resolvedValue`. Sound today — both call sites iterate the same `resolvedParameters`, and `secrets.set` runs after both loops, so the memo is behaviour-identical to recomputation. But the second parameter is redundant, and keeping it leaves the invariant true by convention on a security-relevant certification: a future third caller passing a different value gets the stale answer, and the spelled-secret arm's `written !== resolvedValue` test is not re-validated downstream the way `singleSpanFrame` + `resolvedPlaintextOf` are. Drop the parameter and read `resolvedParameters[name]` inside `computeFrameSpelling`, narrowing to `string` there. Behaviour-identical, and the key becomes provably total. Two of the four reviewers reached this independently.

**M2 — required before merge.** Two PR-body corrections:

- The m1 rewrite gives one reason for all three unworkable shapes — "the only concrete token in view is the one the frame arm WROTE". That is right for a service left to a `Ref` and for a placeholder inside the token, but wrong for the outside-the-token frame: there the frame arm writes nothing (your own comment says so, `written === resolvedValue`), and the token IS readable off the skeleton. What the source cannot supply there is the resolved AFFIX, which only the record's `input` gives. Scope the shared reason to the first two and state the affix gap for the third.
- "16 of the original 30 cases fail with `src/` reverted … and the three hand-built record cases" is stale. The file is 29 cases now, the suite 32, and by the body's own reason the two new hand-built rows fail on reverted `src/` too — so 18 of 32. Either re-state it, or say plainly the measurement was taken at `f7a9b3964` and not re-run.

Take-or-leave:

- **m5** — `secret-redaction.ts:2119-2120`: "the MAP's reverse index ties that token to another plaintext" reads oddly beside "while its pair is intact", since the index yields `CONFLICTING_PLAINTEXT`, not another plaintext. The paragraph above it names the symbol; naming it here too keeps the two from reading as different shapes.
- **m6** — in the repointed #3114 case, the carry's refusal is currently indistinguishable from an early bail on the resource type or an empty bag. One `expect(parentBag.get(PIN_SSM))` style assertion would pin that the walk actually ran.
- **m7** — on the `scan_output` decline: agreed, but the bound is not the one you gave. `masked_whole` covers only the eight `Gc*` prefixes, not the `Resolved Fn::Join:` / `Resolved Fn::Sub:` lines nor the root and middle `Parameter MidPinSsm:` lines. What actually bounds it is that every value-printing site emits the assembled frame, which `scan_output`'s `PIN_*_FRAMED` needles already grep — and the script's own comment says that. If you want the belt anyway, restricting the 2-character grep to the value-printing lines rather than the whole log removes the false-FAIL surface you object to. Not required.

Not yours: the `wait-abandoned-guard-population.test.ts` timeouts. Confirmed pre-existing — the file arrives in your base `65f41cf2a` (#3249), imports nothing this PR touches, and passes alone in 3.0 s; its four `SELF-PROBE` cases re-parse all of `src/` per case with no declared timeout under Vitest's 5000 ms default. Filed as #3347.

Push M1 and M2, plus whichever of m5-m7 you want, and I will merge.

nix-tkobayashi:
Pushed as `4da6f79a`. It changes only `src/deployment/secret-redaction.ts` and `tests/unit/deployment/intrinsic-resolver-nested-child-log-twin.test.ts`, on top of `3f54946f`.

**M1** -- taken. `frameSpellingOf`, `computeFrameSpelling` and `certifiedSpellingOf` now take only the parameter name. The value is read as `resolvedParameters[name]` inside `computeFrameSpelling`, which narrows it to `string` and returns `undefined` otherwise. This is behaviour-identical: both loops only asked for string values, and for those the lookup returns the same value the callers passed.

**M2** -- both body corrections are in.
- The m1 sentence now scopes "the only concrete token in view is the one the frame arm WROTE" to a service left to a `Ref` and a placeholder inside the token. For the outside-the-token frame it says the frame arm writes nothing, the token is readable off the skeleton, and what the source cannot supply is the resolved affix, which only the record's `input` gives.
- The count is re-measured, not just re-stated: with `src/` reverted to the base `65f41cf2`, 18 of the 32 cases fail (the five hand-built rows among them, for the missing export), and 14 pass.

**m5** -- taken. The comment now says the reverse index holds `CONFLICTING_PLAINTEXT` for that token.

**m6** -- taken. The repointed case adds a literal frame, `Control: 'lit:{{resolve:ssm:pinssm}}'`, in the same row, and asserts its whole-value entry is recorded. Measured: making the carry return early for a nested-stack row reds that case.

**m7** -- agreed, and thanks for the correction: my bound was wrong. `masked_whole` covers only the eight `Gc*` prefixes. What bounds the rest is that every value-printing site emits the assembled frame, which the `PIN_*_FRAMED` needles grep. I'm leaving the 2-character grep out.

And thanks for re-deriving the `written === substituted` reachability.

Checks on `4da6f79a`:
- Codex on the delta and the maintainer-checklist proxy pass: both clean.
- Codex on the updated body: clean.
- `vp run check`, `typecheck:test` and build: clean.
- Full suite: 1057 files pass (`--maxWorkers=25%`, because of go-to-k/cdkd#3347).

go-to-k:
Merged as `8d8b131b4`. Thanks — three rounds and every item came back answered precisely, with the measurement rather than the assertion: the 18/32 revert count re-measured instead of re-stated, the `Control` row shown to red when the carry returns early, and the two new hand-built rows shown to red on the check they claim to pin.

Two things stood out. You pushed back on the `written === substituted` "structurally unreachable" claim instead of taking the reviewer's word, and you were right — `CONFLICTING_PLAINTEXT` is a symbol, so the candidate filter drops the leaf's own token while its pair is intact and a sibling becomes the one match. And on `scan_output` you declined the suggestion, then accepted the correction to your own stated bound; the residual is now described accurately in the script rather than papered over.

Housekeeping on my side: I ran `lambda` (83 s, 9 deleted, 0 errors, orphan scan clean) for the `integ-broad` gate, since markgate markers are per-worktree, then flattened and rebased the branch onto `main` — `docs/_generated/integ-last-run.tsv` had picked up a duplicate `lambda` row from another lane, which the normalizer resolved to the newer one.

One leftover you may want to fix at leisure, not a blocker: the body's "Review" section still cites `8b9e735e`, which no longer resolves after the flattens, and stops at the first reviewer pass.

Issue #3336, opened 2026-09-17
test(rules): split layout-scripts.md into routing + satellites — the payload fence's own remedy is blocked wherever an index sits at its cap


go-to-k:
## The corpus has no room left for a new area

`rule-file-payload.test.ts` prescribes one remedy when a rule file outgrows a budget: *"Move the detail into a NEW `.claude/rules/<area>.md` satellite whose `paths:` glob is as narrow as the content, and leave a one-line pointer behind."*

**CORRECTED 2026-09-18.** One of the three figures below was wrong when this issue was filed, and the other two have since moved by an order of magnitude as `main` advanced — so the framing "every file that could hold the pointer is within ~100 bytes of a cap" does not hold. What is true is narrower and, for one index, worse: `testing.md`'s tightest loader has **zero** headroom, not 66 B.

| index, by its tightest budgeted loader | at `6601e8bfa` (as filed) | re-measured at `6601e8bfa` | today |
|---|---|---|---|
| `layout-scripts.md` (`scripts/refresh-cfn-schemas.mjs`) | 5 B | 5 B ✓ | **869 B** |
| `code-layout.md` (tightest loader) | 43 B | 43 B ✓ via `src/deployment/secret-redaction.ts` | **1,326 B** via `src/synthesis/synthesizer.ts` (see the note below) |
| `testing.md` (tightest loader) | 66 B | **0 B**, via `tests/unit/scripts/rule-file-payload.test.ts` | **0 B** |

Two notes on the corrections themselves, since a correction is the least-checked
text there is:

* **Each row is the MINIMUM across every budgeted path that loads the index**, and which path is binding CHANGES as `main` moves. `code-layout.md` was bound by `src/deployment/secret-redaction.ts` at the base commit, by `src/state/malformed-resources-bag.ts` a few hours later, and by `src/synthesis/synthesizer.ts` at the time of writing. **Three different paths in one day** — quoting a fixed representative path is what made the original table age badly, and this note itself named a stale one until it was corrected. Re-derive the minimum; do not cite a remembered path, including these.
* **Where the 66 B came from is unknown.** It matches neither the binding loader (0 B) nor `tests/setup.ts`, the path it named (580 B at that commit). An earlier revision of this correction asserted it was the `tests/setup.ts` figure; that is not supported either.

Consequences for this issue, stated rather than left implicit:

[...]

Issue #3282, opened 2026-09-16
ci: profile and shorten hook-suites, so its timeout can come back down from 60


go-to-k:
## What is slow

`hook-suites` in `.github/workflows/hooks.yml` is the longest job in the repository and the only one whose bound has had to move. Measured over the closed range `2026-09-09..2026-09-15` (n=49): median **1040 s**, max **1587 s** — 26.5 minutes.

Against its original `timeout-minutes: 30` that is **1.1x headroom**: a real run came within 3.5 minutes of being killed, and nobody noticed because a timeout kill reads like a flake. go-to-k/cdkd#3272 raised it to 60 (2.3x), which is the right immediate move and the wrong long-term one.

## Why raising it is the wrong direction

The job is `pull_request`-triggered with a `paths:` filter on `.claude/hooks/**`, so a **fork PR touching those paths runs fork-authored bash on a `macos-latest` runner**. It is the only job that executes fork-authored **shell**, and it is the one the timeout PR had to loosen. (Qualified deliberately: `check-build-test` also runs fork-supplied code on `pull_request`, and go-to-k/cdkd#3272's own comment calls *that* the job its cap matters most for. Two unqualified superlatives about the same property would contradict each other.) Containment is real (read-only token, explicit `contents: read` since go-to-k/cdkd#3272), so this is latency, not a vulnerability.

**Correction to an earlier revision of this issue:** it argued from macOS runners being billed at 10x the Linux rate. That is wrong here — cdkd is a **public** repository, so GitHub-hosted runners of every platform are free, and the round-3 security review of go-to-k/cdkd#3272 caught the error. The case rests on latency and on the bound alone: a 26-minute job with a 60-minute ceiling is the longest feedback loop in the repo, and `hook-suites` is a required check on any PR touching the hooks.

Raised in the security review of go-to-k/cdkd#3272: "45 is defensible ... making the suite faster would have been better."

## What to look at

The job runs two things: `bash .claude/hooks/run-tests.sh` and `bash .claude/hooks/lib/command-match-mutants.sh`. Neither has been profiled. Questions worth answering before optimising anything:

* which of the two dominates, and within it which suite;
* whether the mutation harness re-runs work it could cache;
* whether `macos-latest` is required at all, or only for the `/bin/bash` 3.2 vs bash 5 distinction the job's own "Report available shells" step prints — if only a subset needs macOS, splitting the matrix would cut both wall-clock and cost.

[...]

PR #3340, opened 2026-09-17
test(ci): refuse a workflow timeout that is too tight, not just one that is absent

### LINKED ISSUE #3283: test(ci): the workflow-timeout fence catches an absent bound but not one that is too tight
## The gap

`tests/unit/scripts/workflow-job-hardening.test.ts` (go-to-k/cdkd#3272) refuses four shapes of `timeout-minutes`: absent, non-integer, below 1, and **at or above the Actions default of 360**. That last one exists because a bound equal to the default is indistinguishable from no bound at all.

It does not refuse a bound that is too **tight**, and that is the failure mode the same PR had to fix by hand.

## Why it matters, with the measurement that prompted it

`hooks.yml` carried `timeout-minutes: 30` for its whole life. Measured over `2026-09-09..2026-09-15` (n=49) its max is **1587 s** — 26.5 minutes, **1.1x headroom**. A real run came within 3.5 minutes of being killed. The fence was green throughout: 30 is an integer, at least 1, and below 360.

A too-tight bound fails in the direction that hides: the job is killed, the run reads as a flake, someone re-runs it, and it passes. Nothing accumulates evidence. That is strictly worse than the absent-bound case the fence already catches, which at least fails loudly when it fails at all.

## The approach, decided

The fence is offline and reads only the YAML. Deciding whether a bound is too tight needs the job's observed durations, which come from the Actions API — a network call, per job, in a unit test, which this suite does not allow. Three options were put to the maintainer; **option 1 was chosen**:

**A committed duration snapshot.** `docs/_generated/workflow-job-durations.json`, refreshed by a `vp run gen:*` task like every other matrix in this repo, with the fence asserting `bound >= k * max` against the snapshot plus a staleness guard. This matches the existing codegen'd-critic family (see `.claude/rules/layout-scripts.md`) and keeps the fence offline and deterministic. The cost is accepted and stated: one more generated artifact, and a refresh someone has to run.

Not chosen: a scheduled workflow that re-derives durations and opens an issue when a ratio drops (rejected as outward-facing automation), and accepting the gap (rejected — it is the gap that hid `hooks.yml` at 1.1x).

### What the implementation must get right

* **The snapshot must be taken over a CLOSED date range with the population, not a `--limit` sample.** go-to-k/cdkd#3272 paid for this twice: a positional window slides, and `--limit 50` silently caps (the range held 865 `ci.yml` runs), which is how a 902 s `docs-deploy` run was missed and a 15-minute bound shipped below it.
* **Durations must be per JOB, not per run.** A run includes queue time and every other job in it; `timeout-minutes` bounds neither. A run-level 235 s figure in that PR was a job that took 11 s.
* **Every unreadable input is a finding, never a skip** — a truncated snapshot, a job present in the tree but absent from the snapshot, and a snapshot entry whose job no longer exists must each fail rather than pass quietly.
* Per this repo's checker convention, prove each blocking verdict by breaking REAL code, not only a synthetic fixture.

## The four fields

Session-fit: now — the design question above was the only thing awaited, and it has been answered (option 1). Neither deferral reason now holds: nothing external is pending, and the file the fence lives in was authored in the session that filed this, so it is not cold.
Severity: low — no bound in the tree is too tight today (every one is at least 2.3x after go-to-k/cdkd#3272). This prevents the next one from going unnoticed for as long as `hooks.yml` did.
Effort: medium (M) — the assertion is trivial; the generator, the snapshot format, the staleness guard and the probes that prove the checker fails against real code are the work.
Estimate: 3h — roughly 1h for the generator and snapshot, 2h for the fence plus real-code failure probes, per this repo's checker conventions.

Dup-check: searched open and closed issues for `timeout-minutes`, `too tight`, `headroom` and `workflow duration`; nothing covers the lower bound. go-to-k/cdkd#3229 added the timeouts and go-to-k/cdkd#3272 implements them plus the upper-bound refusal; go-to-k/cdkd#3282 is the separate request to make `hook-suites` faster.

go-to-k:
## The half that was missing

`workflow-job-hardening.test.ts` (go-to-k/cdkd#3272) refuses a `timeout-minutes`
that is **absent**, non-integer, below 1, or **at or above** the Actions default
of 360. That is all of it, and it left the failure that actually happened wide
open.

`hooks.yml` carried `timeout-minutes: 30` against a job observed at **1587 s** —
**1.1x** — for its entire life. The fence was green throughout, because 30 is an
integer, at least 1, and below 360. A real run came within three and a half
minutes of being killed and nobody noticed.

**A too-tight bound fails in the direction that hides.** The job is killed, the
run reads as a flake, someone re-runs it, it passes, and no evidence
accumulates. That is strictly worse than the absent-bound case the sibling
catches, which at least fails loudly when it fails. And it is not hypothetical:
go-to-k/cdkd#3272 itself shipped a revision setting `docs-deploy` / `build` to
**15 minutes** while a real **902 s** build sat inside the range its own comment
cited.

A bound must now be at least **2x** the longest duration ever observed for that
job, read from a committed snapshot.

**Two revisions of this paragraph were wrong about where the `>=` came from, so
here it is from the issue text.** go-to-k/cdkd#3283 writes `bound >= k * max`
itself — the inclusive operator is its own, not borrowed — and leaves `k`
unfixed; **`k = 2` is this PR's decision**. go-to-k/cdkd#3282's target is "a max
comfortably **under** 15 minutes, at which point the bound goes back to 30",
which is *above* 2x rather than exactly on it, and go-to-k/cdkd#3283's Dup-check line does
mention `hook-suites`. Each of the three earlier claims — that `>=` was owed to
go-to-k/cdkd#3282, that go-to-k/cdkd#3282 asks for a state sitting on the floor, and that go-to-k/cdkd#3283 never
mentions `hook-suites` — was false. None is load-bearing: the operator is the
issue's, the constant is mine.

## Why a committed snapshot

Answering "too tight for what?" needs observed durations, which live in the
Actions API — a network call, per job, in a unit suite that forbids one. So
`scripts/gen-workflow-job-durations.ts` captures them into
`docs/_generated/workflow-job-durations.json` and the fence reads that, offline
and deterministically. Three designs were weighed on the issue and the
maintainer chose this one over a scheduled workflow that files issues by itself.

It is deliberately **not** in `gen:all-matrices` and **not** a CI staleness
guard: it needs a network round-trip and a credential, and the underlying data
changes on every CI run, so a byte-diff guard would fail constantly.

## Three rules the generator encodes, each paid for in go-to-k/cdkd#3272

- **Per JOB, never per run.** A run's wall clock carries queue time and every
  other job in it, and `timeout-minutes` bounds neither. A figure briefly
  published there as a 235 s job had actually taken **11 s**.
- **The POPULATION of a closed range, never a capped sample.** A positional
  window slides and never reproduces; `--limit` silently caps — the range used
  there held 865 `ci.yml` runs while the sample took 50, which is how the 902 s
  build was missed. A capped listing is **refused**, not truncated. The range is
  **per workflow**, because run volume spans two orders of magnitude.
- **Only `max` is recorded.** The walk prunes runs that cannot contain a longer
  job, which leaves any count or percentile a statistic of an arbitrary prefix.

## What review found

Thirteen rounds, four axes each. Round 2's code pass independently re-derived the whole
snapshot — **1859 runs, 6746 job records, every `max` matching to the second**,
figures that describe the revision current at that round rather than the
committed one (re-walking today's range gives **1,167 runs and 4,590 job
objects** — 222 more of each than before go-to-k/cdkd#3082 landed, because a
declared job with no successful run can never satisfy `allSeen`, so `hooks.yml`
must now be walked in full) —
and found no blocker. The other axes found seven.

**Round 1's three were one defect wearing three faces.** The generator keyed on
the Actions API's *display name*; the fence keys on the YAML job id. The only
four jobs in the tree that override `name:` are exactly the four I had listed as
"gated on an issue or comment event" — and `issue-conventions.yml` is the
**highest-volume workflow in the repo** (106/71/274/389/496 successful runs per
day over 2026-09-12..09-16). `english-pr` fires on every pull request. All
four carry `timeout-minutes: 10` and sat outside the fence behind a fabricated
justification. **An exemption list absorbing a generator defect and giving it a
plausible story is the exact failure this fence exists to catch, reproduced
inside the fence.**

The prune was **sound for a maximum and unsound for coverage** — a distinction
the old comment blurred by saying "the reported maxima are exactly the
population's" and letting the reader infer the rest. `release.yml` has 387
successful **runs**, in 45 of which `publish` itself succeeded (299 skipped);
the walk stopped after **one**, in which it was skipped. (An earlier revision
wrote that as "`release.yml/publish` has 387 successful runs" — a run total
restated as a job figure, which is measurement rule 1 above.) A
reviewer's brute force over 4000 populations found **0 understated maxima and 28
whole-job omissions**.

And `max: 0` made every bound pass: headroom is `Infinity`, and `Infinity < 2`
is **false** — reachable without an adversary, because `Math.round` recorded 0
for any sub-500 ms job. It is also why deleting the absent-bound guard was
*unkillable*.

With all three fixed the snapshot covered **21 of 21** jobs and the exemption
list went **empty**. It holds one entry again — `hooks.yml/mutation-harness`,
added to `main` by go-to-k/cdkd#3082 mid-review — with a reason that is
checkable and self-retiring ("no successful run on main yet"), which is the
property the first cut's five entries lacked. 21 of 22 today.

**Round 2 found the brand shipped without its enforcement half.**
`Finding.job` and `Finding.range` were typed `string` where the sibling types
them `Safe`, so dropping *both* sanitisers produced **zero `tsc` errors** and
left all 42 cases green. The whole argument for the brand is that a forgotten
sanitiser becomes a compile error rather than a review finding; typing the
fields `string` quietly declines that.

**The forgery class reached a sixth venue** — three module-scope throws, all
carrying *fork* text: `declaredJobs` parsed YAML outside a `try`; `parseSnapshot`
interpolated a snapshot object **key** raw (arbitrary attacker text, no length
limit); `JSON.parse` echoed the offending source. The tell was in the file
already: `FindingKind` declared `'unreadable-workflow'` and **nothing
constructed it**.

**Round 2: three published figures were wrong and one rationale was inert.**
"~885 runs in a day" (twice) was a range total restated as a daily rate; "headroom to over
100x" — the loosest is 69.23x; and the snapshot had been generated against an
**in-progress** UTC day, making its own "closed range" note false. Worst: I
justified including `failure`/`cancelled` conclusions as capturing "a job killed
at its own bound", but the run-level `--status success` filter means such a job
never arrives — **0 records**, in that round's 6,746 and again in the committed
artifact's 4,590. That is the third consecutive round in
which this pair of files shipped a rationale that does not hold, so the comment
now states the *limit*: a job killed at its bound never completes, so the
snapshot cannot hold the observation that would show a bound is too tight most
clearly.

**The floors could not have caught the defect they exist for** — `MIN_SNAPSHOT`
was 12 against a 21-entry snapshot, so four vanishing jobs cleared it. Pinned to
`MIN_DECLARED` structurally now.

**Round 3 measured what I had only reasoned about, and inverted it.** I had
recorded that breaking the generator's entry-point guard *hung* the suite. Run
to completion, all five mutants of that guard report **57 passed / 0 failed** —
and leave `docs/_generated/workflow-job-durations.json` rewritten in place. So
importing the generator for its pure helpers runs the whole network walk,
**overwrites the fence's own committed input**, and the fence then passes
against data the same run just fetched from the live API. A hang is loud; this
is silent, and it makes the one case that validates the committed artifact
self-referential.

Two locks answer it, on **independent signals**: `invokedDirectly` takes
`(entry, self)` as arguments and is pinned by six cases that never touch
`process.argv`; and `main` opens with `refuseInsideTestRunner(process.env)`,
keyed on the runner rather than the entry point. The same mutant is now a
**red at import**, with the snapshot byte-identical afterwards. Deleting the
lock's *call* from `main` was itself a survivor — registration is not execution
— so a subprocess case pins the wiring, using a bogus flag so that neither
outcome can reach the network or the writer.

**The forgery class reached a seventh venue, and it is not a rendered line.**
`declaredJobs` returned `string[]`, and `expect(unreadable).toEqual([...])`
prints that array in vitest's failure diff at column 0 — pretty-format escapes
`"` and `\` and nothing else. The render path was sanitised; the **failure path
of the case proving a fork cannot forge a line** was not. Both the return type
and `auditHeadroom`'s fourth parameter are `Safe[]` now, which turned a
hand-built `['evil.yml']` into a compile error.

**The brand catches a missing sanitiser, not a wrong one.** Swapping `safeName`
for `safeText` typechecks — both return `Safe` — and left every case green. A
hostile FILE half now discriminates them. **And the floors permitted the defect
they were raised for**: lowering all three together to 14, or to 11, stayed
green, so `MIN_DECLARED` is now pinned from below at the measured size of the
round-1 defect.

**Four more figures were wrong, three of them in prose that had replaced a wrong
figure**: the `publish` run total above; a per-day series ending on an
in-progress day; an undated "~3300 runs"; and the `job <= run` check described
as "self-checking" when it is a tripwire — the loop sees only runs that were
*fetched*, so the one place a violation would do damage is the one place it
cannot look.

**Round 4 found the eighth venue next door to the seventh.** `parseSnapshot`
interpolates the offending key into **two** refusal messages and only one was
pinned — the hostile-key case supplies no `max`, so it never reached the
sibling. `max: 0` routes it there, and 0 is the value that motivated that
refusal in the first place. Also: the **job half** of a key was never
*constrained*, only flattened and clamped, so a pure-ASCII job id rendered as a
complete finding about another job; and the key splitter was wrong for a key
with **no `/`** — `slice(0, -1)` dropped the last character, so `evil` rendered
as `"evi"/evil`.

**The extraction that shared the sanitisers left half the contract behind.**
`boundedList` and `MAX_RENDERED_FINDINGS` stayed in the sibling — whose own
docstring calls the cap "the second half of that contract" — so this fence
shipped a bare `.map`. Per-field clamping bounds a *line*; nothing bounded the
*count*, and 5,000 fork-added snapshot keys bury the finding a reader must act
on.

**The shared module moved to `scripts/`**, because the generator was the one
file in this pair with **no sanitiser at all** — eight of its own messages
interpolated raw workflow names and raw parse errors — and a module under
`tests/` cannot be imported by a script without reversing the dependency. The
`.js` import spelling typechecks and then **dies at runtime under `node`**;
`vp run check` cannot see that, and a live probe caught it.

**A count floor cannot see a vacuous snapshot.** Every floor here counted
*entries*: setting every `max` to 1 left all 72 cases green, which makes each
`too-tight` verdict meaningless while passing every count. `MIN_LONGEST_MAX`
now bands the longest job (1587 s today — the job this fence was written for)
from both sides.

Also closed, each measured green first: the matrix-leg `$` anchor (only a
*renamed* job with a leg separates the two patterns); `isMapping(node)` against
a null job node, which would take the whole file's collection down at module
scope; the floor in the rendered line; a retry that encoded "three attempts" in
three literals and classified on `String(error)`, so a workflow named `EOF.yml`
turned every 404 into three attempts; a `--to` inside a day still in progress,
now refused; and a repeated or inverted range flag.

**Round 5 found two blockers, and one of them I had just written.** The
generator's halving loop — the thing that lets a busy workflow's window shrink
until its listing is not capped — was **dead**: round 4 removed a `Math.max`
floor from the window and left the loop's exit comparing `days` against the
value `days` is initialised from, so it was true on the first iteration always.
The first capped listing threw, and the generator could no longer reproduce the
snapshot this PR commits (**six** of its workflows carry a narrowed range:
`ci.yml`, `pr-content-checks.yml` and `pr-title-check.yml` at 3 days,
`docs-deploy.yml` and `pr-inherit-issue-labels.yml` at 7, `issue-conventions.yml`
at 1 — an earlier revision said four). Verified
with a fake capped `gh`: it now walks 14 → 7 → 3 → 1 days and then refuses,
with the window in the message matching the one in the range.

**And `safeJobId` reached one of its two consumers.** Round 4 wrote it for
exactly this threat and applied it only in the new fence; the sibling still had
`safeText`, so a fork declaring a job named
`x: no-timeout - and ci.yml / check-build-test` rendered a complete fabricated
finding against a real critical job — pure ASCII, nothing for `flatten` to
remove, in a run that is genuinely red so the line is genuinely printed. That is
the "a fix lands in one place and not the other" failure the shared module's own
header gives as the reason it exists, committed in the round that wrote the
header. `safeKey` and `safeRange` are now in the shared module too, because the
same round left four more key-shaped values going through `safeText`.

**The staleness guard go-to-k/cdkd#3283 asked for was never built.** The issue
says `bound >= k * max` "plus a staleness guard", and three *structural* guards
were built — job in the tree but not the snapshot, entry whose job is gone,
exemption no longer needed — while the one the phrase most plainly means, AGE,
was not; the body argued only that a CI byte-diff guard is impossible, which
answers a different question. `auditSnapshotAge` now refuses a snapshot older
than 90 days, one with an unreadable stamp (`Date.parse` → `NaN`, and every
comparison against `NaN` is false — the same shape this whole fence exists
because of), and one stamped in the **future**, which is how a stale snapshot
would be made to look fresh.

Also: `MIN_LONGEST_MAX` was 900 with **two seconds** of real margin — the
assertion takes the max over all jobs, and the second-longest is
`docs-deploy.yml/build` at 902 s, so once go-to-k/cdkd#3282 succeeds the floor would be held
by an unrelated variable job and an honest snapshot would red. It is 600. The
cap was **order-evadable** (20 unbounded jobs in an `aaa.yml` push the real
finding past it), so `too-tight` sorts first. The inverted-range guard fired
only when *both* flags were given, so a lone `--from` after the default `--to`
walked an empty range and would have written `"jobs": {}`. And `gh`'s rethrow
carried `Command failed: gh … --workflow <name>` with the fork's name in it.

**Round 6 found the tenth venue and a case passing for the wrong reason.**
`detail` was the only rendered fork-controlled field still on `safeText`, and it
is the **last** field on its line: for an unparseable workflow it is the YAML
parser's own message, echoing fork bytes. Measured end to end, the `)` closes
the parenthesis early and the tail reads as a finding against a real job — and
the padding is **U+2800**, which `\s` does not match and which survived
`flatten` entirely, as does U+3164. Quoting is on *shape*, not unconditional:
quoting everything put quotes around fourteen machine-generated strings like
`360 >= 360`, which is the noise that gets a rule switched off.

**And the SECOND refusal's case was passing for exactly the wrong reason** — it
asserted only the *absence* of LF and RLO, which flattening removes, so
`safeKey` → `safeText` survived with all 186 cases green while rendering a
forged line. Its twin got the quoting assertion a round earlier; this one did
not. Three findings this round were that same shape — a fix landing in one place
and not its neighbour — in a **sanitiser**, in an **assertion**, and in
**prose** (round 5's retraction of the go-to-k/cdkd#3282 and go-to-k/cdkd#3283 claims reached the PR body
and not the fence's own docstring).

**Sorting is not sharing.** Round 5's `too-tight`-first order does not stop a
fork emitting 25 findings *of the winning kind* from its own file; the cap is
now shared **round-robin by workflow** inside `boundedList`, so both fences
inherit it.

**The staleness guard keyed on the wrong field.** `generatedAt` is refreshed by
every run, so re-deriving the snapshot over its own recorded range — the normal
way this artifact gets re-checked — writes a fresh stamp over an arbitrarily old
population and the guard answered `ok`. It now reads the **oldest** `to` across
entries — the maximum was tried first, and its justification turned out to be
unproducible, since the generator writes one uniform `to` for every workflow;
the two differ only on a partially refreshed or hand-edited snapshot, and there
the minimum is the one that notices. The generator also refuses a `--to` more
than 45 days old, so the failure is named where it is caused.

**And `nextWindow` is extracted**, so the arithmetic that produced rounds 4 and
5's blockers is finally reachable from a case — eight of them, including the
progress property the loop's termination depends on.

**Round 7 fixed a clamp done in the wrong order.** `safeText(JSON.stringify(v))`
clamps *after* quoting, so a field over the cap loses its **closing quote** and
everything after it on the line reads as part of it — live in `safeJson`, and
written into `safeDetail` by the round that was fixing the sibling defect. Both
go through `quoteClamped` now. Fixing it then broke a test helper whose filter
assumed the quoted form *contains* the unquoted one: it silently selected
**nothing**, turning a forgery case green. Also closed: `DETAIL`'s exclusion was
pinned by nothing (widening it to admit `:`, `/` and parens left 212 cases
green), and the blank class had reached for the exotic code points a review had
named while omitting ZERO WIDTH SPACE and its neighbours.

**Round 8's blocker was mine, in prose.** Round 7 demoted the `too-tight`-first
sort from burial-prevention to "ordering", generalising from a measurement of 20
findings inside *one* file. With 21 distinct **workflows** against a cap of 20,
one group loses its line entirely and order decides which — measured: unsorted,
the genuine finding is absent from the kept lines *and* from the five names the
summary can fit. The round-5 text was right; round 7 inverted it in a comment
and a commit message.

**Round 8 found no twelfth forgery venue, and one blocker that was mine in
prose.** The summary that names crowded-out workflows named only the groups that
lost *every* line, so a fork could leave a real workflow one line and push the
interesting one out of its own group — it now names any group that lost a line.
The grouping key also merged quoted names sharing a prefix (`"zz alpha.yml"`,
`"zz beta.yml"` and `"zz gamma.yml"` all cut to `"zz`), so a quoted name is
grouped by its whole JSON token now.

**Round 9 found the anti-burial summary was fork-selectable**, and two of its
three blockers were mine from round 8. Round-robin gives every group a share
only while the groups FIT, and when they do not, both which groups keep a line
and which five are NAMED were decided by insertion order — directory order,
which a fork picks by naming its files `aaa*.yml`. Widening the filter to
"lost any line" made it worse in one direction: starved groups are last in
insertion order and got pushed out of the five names by groups that merely lost
one. `incomplete` is ranked by kept-count ascending now, so a starved workflow
leads the list. What no ranking can promise is that a particular LINE survives —
with a fork-chosen file set and a fixed budget, some group loses — and the case
says that rather than claiming the guarantee.

**Three of my own cases claimed more than they checked**, each measured: the
many-workflows flood case was vacuous (`auditHeadroom` emits the declared loop
first, so the genuine finding was always the first group); the `oldestRangeEnd` rows put
the oldest value at position 0, where "minimum" and "first entry" coincide; and
the pure-ASCII detail case pinned a **disjunction**, since its fixture carried
`:`, `(` and `)` at once — admitting any one of them stayed green. There is one
fixture per excluded character now, each with a control that is *not* quoted.
Groups also hold **indices**: a `Set<string>` is value equality, so two job ids
clamping to the same text made a dropped line read as kept — and the label
calling that fix "by identity" was itself wrong.

**And the fence caught a real job in live CI.** go-to-k/cdkd#3082 added
`hooks.yml/mutation-harness` to `main` mid-review; CI went red on this branch
with `not-in-snapshot` plus the `MIN_DECLARED` ratchet, exactly as their comments
predict. The two-step flow written as prose in round 5 ran for real for the first
time: the three floors move together 18 → 19, and the new job takes a
`RARELY_RUN` entry with a true, self-retiring reason.

**Round 11 refuted the premise round 10's fix rested on.** The kind ranking was
justified by "a fork cannot manufacture these kinds for someone else's
workflow" — but a fork **owns `.github/workflows/` in its own PR**, so
twenty-five files it deliberately breaks are twenty-five findings of the
top-ranked kind, and the genuine line was neither kept nor named. Measured. The
cap is now split between the **kinds present** before it is split between
workflows: a fork can flood one kind, but it cannot stop another from existing.

Also this round: five of nine ranks in one table and four of six in the other
could be lifted to first place with the whole suite green, because every fixture
carried only two kinds at once; `KIND_RANK` is `satisfies`-checked so an added
kind is a compile error rather than an `indexOf` of −1 that silently sorts first;
the `MIN_LONGEST_MAX` band admitted 901, which its own docstring condemns for the
same reason as 900; and the last unguarded `JSON.parse` in the generator is
wrapped.

**And a commit message claimed two fixes its diff did not contain** — an edit
batch had aborted on an unrelated anchor failure and I wrote the message as
though it had landed. Both edits are in now. The ratchet guidance was
recommending the exact reason string the same file retires as having been false,
and a figure recorded as "28 successes" grew to 40 while three rounds argued
about it, so the count is gone and only "non-zero" is recorded.

**Rounds 12 and 13 are where this stopped being about the fence and started
being about the failure MESSAGE.** Five rounds went into `boundedList`, the
helper that caps how many finding lines a failure prints — and each round's
remedy became the next round's defect: fork-chosen insertion order, then one
lifted kind, then kinds a fork can manufacture (it owns `.github/workflows/` in
its own PR), then a per-kind share that under-filled the cap and spent one
kind's allocation on a single workflow.

**That is now closed by decision rather than by more mechanism.** With a fork
controlling both the workflow directory and the committed snapshot, no fixed
print budget can guarantee a particular LINE is visible. What the code does
guarantee, and what the comments now say, is that the group order interleaves
the kinds, that every workflow which lost a line is NAMED, and that a count of
starved workflows — which a fork cannot crowd out by adding files — is always
printed. None of this affects DETECTION: both auditors return every finding and
the suite asserts `toEqual([])`.

**Round 13 aimed at the detection path instead, and found two real defects
there.** A snapshot entry with no `to` made an ancient population read as fresh:
`parseSnapshot` coerced it to `''`, which is `oldestRangeEnd`'s "nothing seen
yet" sentinel, so one such entry reset the accumulator and discarded every older
date — a six-year-old maximum then compared against a live bound, and
`too-tight` never fires. And the twins' parallel `out`/`outKinds` arrays could
desync, pooling a genuine finding into a fork's bucket; they are pairs now.

Also closed there: `/\s/` was fenced for its existence and not its extent (U+2028
was the only member any case carried, while NBSP and the BOM survived); the
`TRANSIENT` retry classifier — whose own comment records two measured defects —
had no case at all and is exported and fenced; and the generator's 45-day floor
and the fence's 90-day window were related only by a comment.

## Probes

Against **real code** on a scratch copy, per `.claude/rules/testing.md`.

- Round 5: **21 arms probed** — 14 first, of which **6 survived**, every one a
  missing case rather than a broken fix, then 7 re-probes once the cases
  existed: **all killed**. Tree and snapshot byte-exact throughout. Two arms are
  labelled rather than dressed up: `auditSnapshotAge`'s `typeof` guard is
  *equivalent* (every non-string reaches `Date.parse` → `NaN`, which the next
  line already handles), and deleting an assertion is not a mutation of
  production code.
- **Neither round-5 blocker was found by mutation.** The dead halving loop is
  invisible to a suite that never runs the network half, and the sibling's
  missing `safeJobId` was pinned by a case asserting the **weaker** property —
  flattened, not quoted — so it stayed green. Both have cases now, and the
  sibling's two were rewritten to assert quoting.
- Round 4: **15 arms probed, 15 killed**, tree and snapshot byte-exact after
  each. One first read as a survivor and was a **harness defect, not a result**
  — the mutation was a syntax error, the run failed to load, and the summary
  regex picked up the sibling file's passing count. Re-probed validly, it reds.
- Round 3: **17 arms probed, 16 killed and 1 equivalent**, tree and snapshot
  restored byte-exact — including both first-pass survivors once their cases
  existed (the lock's wiring into `main`, and a raw file name reaching the
  assertion path), and the arm that used to go green-and-rewrite, which is now
  a red at import.
- **17 of 18** arms killed in round 2 across all three files; the survivor is
  *equivalent* (optional chaining handles `null` exactly as the explicit guard).
  Earlier passes were 15/15 and 21/21 on smaller arm sets — each enlarged
  because a reviewer probed arms my table had not thought of.
- **Two things are NOT pinned, and are labelled rather than dressed up.**
  `declaredJobs`'s `.sort()`: `readdirSync` returned already-sorted names in 40
  of 40 probes on this host, so any case would pass with the call deleted — it
  stays for determinism on a filesystem that does not. And
  `invokedDirectly`'s `entry === undefined` guard is an **equivalent** arm,
  since `realpathSync` throws into the `catch` that answers the same `false`;
  an earlier revision of its comment claimed the opposite.
- The prune is extracted as `walkMayStop` and pinned by four cases: deleting the
  `allSeen` line killed nothing before, though 200,000 populations confirm the
  logic is right. **Correct and pinned are different properties.**
- The `job <= run` inequality the prune rests on carries a **tripwire** — 9,526
  of 50,000 violating populations give a wrong result — but it is not a proof,
  and the file says so: the loop sees only runs that were *fetched*, and a
  pruned run is never fetched. (An earlier revision of this line called it
  "self-checking", the very word the same body elsewhere retires.)
- **8 real-code probes** including the boundary pair: 53 min (2.0038x — "exactly
  2.00x" in an earlier revision was a two-decimal artifact of 3180/1587)
  passes, 52 min (1.97x) fails, tree restored byte-exact.

**The first version of this fence was unfenceable.** Written as assertions
reading the real tree directly, **all thirteen of its arms survived mutation** —
a fence asserting properties of a tree that *satisfies* them returns the same
empty list broken or working. It is a pure function of
`(declared, snapshot, exemptions, unreadable)` for that reason.

## `layout-scripts.md` entry — added, once the figures excusing its omission were re-measured

An earlier revision of this section declined the convention-standard entry on
three measurements, taken at the base commit `6601e8bfa`:
`scripts/refresh-cfn-schemas.mjs` **5 B** under its 90,000 B cap,
`code-layout.md` **43 B** of headroom, `testing.md` **66 B**. Re-measured, at
that commit and at today's `main`:

| index, by its tightest budgeted loader | at `6601e8bfa` | today |
|---|---|---|
| `layout-scripts.md` (`scripts/refresh-cfn-schemas.mjs`) | 5 B | **869 B** |
| `code-layout.md` (tightest loader) | 43 B (`src/deployment/secret-redaction.ts`) | **1,326 B** (`src/synthesis/synthesizer.ts`) |
| `testing.md` (`tests/unit/scripts/rule-file-payload.test.ts`) | **0 B** | **0 B** |

Two of the three moved by an order of magnitude as `main` advanced, and the
third was wrong when written: `testing.md`'s tightest loader has **zero**
headroom, not 66 B. The 66 B matches neither the binding loader nor
`tests/setup.ts`, the path the original figure named (which measures 580 B at
that commit), so where it came from is unknown and an earlier revision of this
paragraph asserted a provenance it could not support. So the corpus is *not*
uniformly saturated, one index is genuinely at its cap, and the conclusion
rested entirely on numbers that no longer hold.

So the entry is **in**, covering both new `scripts/` files in one bullet and
written tersely on purpose: it costs 525 B and leaves that path **344 B**, in
line with the corpus's other tight paths (`vite.config.ts` 436 B,
`tests/setup.ts` 580 B) rather than an outlier. `rule-file-payload.test.ts`
passes, 445 cases. go-to-k/cdkd#3336 still stands — the corpus is tight — but
its body carried the same wrong figures; it has been corrected — three times,
as each correction was itself re-measured — and retitled.

<details><summary>The superseded reasoning</summary>

The budgeted path
`scripts/refresh-cfn-schemas.mjs` loads that index plus `layout-schema-refresh.md`
and sits **5 B** under its 90,000 B cap. Splitting into a satellite — what
`rule-file-payload.test.ts` prescribes — failed too, because no index could take
even a one-line pointer (`code-layout.md` 43 B of headroom, `testing.md` 66 B).
The fence funds an addition only from what your own change made stale, and a new
area makes nothing stale.

</details>

Suite 1080 files / 25 593 tests (after rebasing onto today's main). No changelog entry: CI and tests ship no
user-visible behaviour delta.

Closes #3283


go-to-k:
Closing: PR 3407 adopted a new policy for the tooling layer (CLAUDE.md, docs/tooling-backlog.md) under which fences on CI prose / workflow config are added only on the second occurrence of the same failure. Issue 3283 is recorded in the backlog; if the gap bites again, this branch is the starting point.

Issue #3353, opened 2026-09-17
fix(cli): `cdkd orphan`'s state and outputs refusals render an empty region as `<unrenderable>`


go-to-k:
## What

`cdkd orphan`'s two other refusals render an empty region as `<unrenderable>`, producing a remedy command that names no record.

`src/cli/commands/orphan.ts` passes `targetRegion` — the result of `pickStackRegion`, which returns `Promise<string>` and yields `''` for a v1-legacy record — to three refusal builders. One of them normalises `''` to "absent" and degrades gracefully. The other two do not:

- `refuseMalformedState` (`orphan.ts:232`)
- `refuseMalformedOutputs` (`orphan.ts:245`)

Both take `region: string` and render, measured:

```
State for MyStack ('<unrenderable>') holds ...
Inspect the record with: cdkd state show MyStack --stack-region '<unrenderable>' --json
```

`displaySafe('')` is `<unrenderable>` by design — an empty string is not an identity — so the flag is emitted with a value no one can act on, and the single quotes collide with the wrapping ones in the sentence.

## Why it is a nit rather than a defect

The command it produces is READ-ONLY, and `parseStackRegion` rejects an empty value, so a user who pastes it gets a clean argument error rather than a wrong record. Nothing is deleted and nothing is mis-aimed. What is lost is the remedy: a user already holding a damaged record is handed a command that cannot run, with no hint of what to substitute.

## Shape of a fix

go-to-k/cdkd#3346 settled the pattern for the third builder and is worth copying rather than re-deriving: normalise `''` to `undefined` ONCE at the message builder's entry, so every helper it feeds — the opening clause, the inspect command, and any remedy — is protected together. Guarding inside one helper fixes a fraction of the message; that was measured there (the first cut did exactly that, and the other two helpers still rendered the placeholder).

The alternative is to normalise at the call site in `orphan.ts` instead, which fixes all three builders at once but leaves the builders themselves accepting a value they render badly. Prefer the builder-side hoist for the same reason go-to-k/cdkd#3346 chose it: the builders are exported and a fourth caller need not know the rule.

Note while fixing: `''` and `undefined` must produce the SAME output, because which one arrives depends on a caller these modules do not own. go-to-k/cdkd#3346's tests pin exactly that non-divergence and are worth mirroring.

[...]

Issue #3318, opened 2026-09-17
fix(cli): `cdkd orphan` persists a resource `properties` map it could not read, with no diagnostic


go-to-k:
## What

`cdkd orphan` reads each surviving resource's `properties` map through `rewriteValue`, re-assigns the result with a bare cast, and SAVES it. A map that is not an object — a string, a list, a number, or absent — is carried through untouched and persisted back into the record cdkd writes.

`src/analyzer/orphan-rewriter.ts:484-531`:

```ts
for (const [logicalId, resource] of Object.entries(state.resources ?? {})) {
  if (orphanSet.has(logicalId)) continue;

  const rewrittenProperties = await rewriteValue(
    resource.properties as unknown,   // <- no readability check
    ...
  );
  ...
  newResources[logicalId] = {
    ...resource,
    properties: rewrittenProperties as Record<string, unknown>,   // <- bare cast
    ...
  };
}
```

`rewriteValue`'s first line is `if (typeof value !== 'object' || value === null) return value;`, so a string / number / `undefined` bag is returned verbatim; `[]` takes the array branch and comes back as `[]`. Either way the cast at `:526` re-declares it a `Record<string, unknown>` and the rewritten record is saved.

## Why the existing guard does not cover it

`src/cli/commands/orphan.ts:231` calls `refuseMalformedState(state, ...)`, which checks the ROOT `resources` bag only. The per-ENTRY predicate that answers this question — `unreadableResourcePropertyBags` / `refuseMalformedResourceProperties` in `src/state/malformed-resources-bag.ts` — is not called from this path. That predicate's own doc states the rule directly: it returns `[]` when the root bag is unreadable, and "what a caller must not do is take only this one". `cdkd orphan` takes only the other one.

## Consequence

`cdkd orphan` is a WRITE-CAPABLE command: it saves the rewritten record. So a torn `properties` map survives an operation the user ran to repair the record's references, with no diagnostic. The next `cdkd deploy` then hits go-to-k/cdkd#3191's refusal — correct, but one command later than the evidence was last in cdkd's hands — and a `cdkd diff` in between previews every property of that resource as an addition and a create-only one as a replacement.

There is a second, smaller edge on the same lines: `rewriteValue` walks a non-object bag without recursing, so an orphan reference nested inside a torn map is never rewritten and never reported as `unresolvable` either. The command's `--force`-less hard-fail on unresolvable references cannot fire for it.

## Suggested direction

[...]

PR #3226, opened 2026-09-16
fix(state,drift,diff,export): refuse a malformed state record where the flow can write it, repair and report where it cannot

nix-tkobayashi:
Closes #3018.

go-to-k/cdkd#3159 closed the unreadable-BAG half of that issue for `cdkd diff`, `import`, `orphan`, `rollback` and `scrub`. This closes what its own follow-up comment records as remaining: the null-ENTRY class, the NON-OBJECT bag reaching the one WRITER in `state.ts`, the code-unit `slice` in `describeOverlayValueShape`, and the `drift.ts` sibling folded on afterwards.

## The shape, and why it was still open

`hasReadableResources` tests the BAG. `{"resources": {"R": null}}` passes it, so the `null` surfaced one dereference later as a bare `TypeError` naming no stack, no key and no remedy — from the commands a user reaches for when the state is already broken. `parseStateBody` deliberately does not validate inner shape (go-to-k/cdkd#2947's placement decision), so every shape below really does arrive at a command.

## Refuse where the flow can WRITE, repair where it cannot

The rule go-to-k/cdkd#3159 established, applied per FLOW rather than per file:

| Flow | Malformed bag | Unreadable entry |
| --- | --- | --- |
| `cdkd state refresh-observed` | REFUSE | REFUSE, naming the ids |
| `cdkd drift --accept` / `--revert` | REFUSE | REFUSE, naming the ids |
| `cdkd drift` (detection) | repair to empty + warn + report | drop + warn + report, naming the ids |
| `cdkd diff` (plain and `--recursive`) | repair to empty + warn + report | drop + warn + report, naming the ids |
| `cdkd export`'s baseline report | warn, skip the report | name them, leave them out of the tally |
| `cdkd import`, in the sibling context a provider's `readCurrentState` receives | REFUSE (unchanged) | skipped **silently** — it names no resource a sibling lookup could match, and the shapes that reach it there (a string, a number, a list, an object carrying no `resourceType`) used to enter the map as a sibling record naming no resource — a `null` entry never gets that far under `cdkd import`, throwing earlier in its property resolution; the entry half of `import.ts` is recorded on go-to-k/cdkd#3202 |

[...]

PR #3363, opened 2026-09-17
fix(cli): point cdkd orphan's properties refusal at the region a legacy record is listed under

### LINKED ISSUE #3359: fix(cli): cdkd orphan's properties refusal points a legacy region-less record at a synthesized region its remedies cannot select
`cdkd orphan`'s `properties` refusal (go-to-k/cdkd#3346) names two commands built from the stack and region it was handed: a `cdkd state orphan <stack> --stack-region <region>` remedy in its prose (`dropRecordCommand` in `src/state/malformed-resources-bag.ts`), and the `cdkd state show <stack> --stack-region <region> --json` it ends on (`inspectCommand`). For a stack whose only record is a legacy one (`cdkd/<stack>/state.json`, no region in the key) that carries no `region` field, and a CDK app that synthesizes a region, both commands name that region and both then find no record: the way out the refusal hands a stuck operator is a dead end twice over.

## Why

`dropRecordCommand`'s own note says an absent region means a v1 record, for which the bare command is the correct one, and it omits the flag when `region` is `undefined` or `''`. The value it is handed is `pickStackRegion`'s answer (`src/cli/commands/orphan.ts`), and on this path that answer is not absent:

- `listStacks()` yields one ref for the stack, whose `region` comes from `readLegacyRegion` — `undefined` when the legacy body has no `region`.
- `pickStackRegion` finds no ref whose region equals the synthesized region, falls to `refs.length === 1`, and returns `refs[0]!.region ?? synthRegion ?? ''` — the SYNTHESIZED region.

So the refusal prints `--stack-region '<synthesized region>'` in both commands. `cdkd state orphan` filters `stackRefs.filter((r) => r.region === options.stackRegion)` and `cdkd state show` goes through `resolveSingleRegion`'s `matches.find((r) => r.region === requestedRegion)` (both `src/cli/commands/state.ts`); the legacy ref's region is `undefined`, so neither matches, and each throws a `No state found for stack '<stack>' in region '<region>'` error listing `(legacy)`. Nothing is deleted — the failure is the missing way out, not a wrong target.

What this does not establish: it was traced from the code, not driven through the CLI, and whether a legacy record with no `region` field is reachable on a current record depends on how old the writer was.

## Proposed fix

Hand the refusal the region the record is actually keyed under, not the resolved region used to load it: from `cdkd orphan`, pass `undefined` when the selected ref is legacy (its `region` is absent), so `dropRecordCommand` takes its documented bare-command arm — `cdkd state orphan <stack>` with no flag, which the region filter in `state.ts` then applies to every record for that name, the legacy one included. The narrower alternative is to have `pickStackRegion` return `''` for a legacy single ref regardless of `synthRegion`, but that changes what every other consumer of its answer loads, so the refusal-local fix is preferred.

That fixes the DESTRUCTIVE remedy only. The inspect command cannot be fixed the same way: `cdkd state show` refuses a region-less legacy record outright (`stateShowCommand` in `src/cli/commands/state.ts`, "has only a legacy state record without a region ... migrate it"), with or without the flag. So for this record shape the refusal should not end on a `cdkd state show` it cannot serve; the lane decides whether that line points at the migration `state show` itself names or whether `state show` should learn to read a legacy record, and says which.

## Verification plan

- A `cdkd orphan` unit case with a single legacy ref whose body has no `region`, a synthesized region set, and a torn `properties` map on a surviving record: the refusal's `cdkd state orphan` remedy is `cdkd state orphan <stack>` with no `--stack-region`, and driving `cdkd state orphan` with exactly that text selects the legacy record (not merely "the flag is absent"). The inspect line is asserted to be one that works for that shape under whichever decision the lane records.
- The same with a legacy body that DOES carry a region: the flag is present and equals it (so the fix does not drop a region that selects the record).
- A region-keyed record keeps `--stack-region <region>`.
- Mutation probe: restoring the synthesized region into the refusal reddens the first case.

Dup-check: searched open and closed issues for `orphan refusal stack-region legacy synth region`, `dropRecordCommand` and `pickStackRegion legacy region synth`. No hits. Nearest: go-to-k/cdkd#3350 (`cdkd orphan` saving an unreadable entry reshaped; a different defect in the same command) and go-to-k/cdkd#3328 (the destroy runner's body-region divergence).
Session-fit: now (do it in this session) — found by the review of go-to-k/cdkd#3226's rebase onto go-to-k/cdkd#3346, which read `orphan.ts`, `state.ts` and the module; nothing external blocks it.
Severity: low — the refusal's remedy is a dead end on a legacy region-less record, and the command it prints deletes nothing.
Effort: small (S) — one argument at the refusal call site, the inspect-line decision for a legacy record, and unit cases.
Estimate: 1-2 hours

### LINKED ISSUE #3360: fix(cli): cdkd orphan's properties refusal substitutes an unvalidated stack name into its destructive remedy with no exactness gate
`cdkd orphan`'s `properties` refusal (go-to-k/cdkd#3346) prints a destructive remedy, `cdkd state orphan <stack> --stack-region <region>`, built by `dropRecordCommand` in `src/state/malformed-resources-bag.ts`. It renders the stack name through `safeStackName` (sanitize, then shell-quote) and substitutes it with no exactness gate. When this issue was filed its note justified that by the identity's source — the synthesized stack and `pickStackRegion`'s answer — but the synthesized stack name is not guaranteed to survive sanitizing, so the command can name a DIFFERENT record than the damaged one.

## Why the name is not guaranteed

`cdkd orphan` takes the stack name from `resolveConstructPaths(...).stack.stackName` (`src/cli/commands/orphan.ts`), which comes from the synthesizer. `aws-cdk-lib` validates the stack names it synthesizes, but cdkd also accepts a PREBUILT cloud assembly: `Synthesizer.synthesize` treats an `--app` that points at an existing directory as a pre-synthesized assembly and reads its manifest directly (`src/synthesis/synthesizer.ts`), and `AssemblyReader.extractStackInfo` takes `props?.stackName || artifactId` with no validation (`src/synthesis/assembly-reader.ts`). A manifest naming a stack `Prod ` (trailing space), or one containing a control or non-ASCII character, reaches the refusal unchanged, while `safeStackName` trims or substitutes it.

So for a record keyed `Prod ` beside a healthy `Prod`, the refusal about the damaged `Prod ` record ends its removal advice on `cdkd state orphan Prod --stack-region …`, which `cdkd state orphan`'s exact name match resolves to the healthy record. That is the misdirection class `malformedDestroyResourcesRefusalMessage` closes with its exactness gate (go-to-k/cdkd#3332) and `reportDriftBaselineGaps` closes for `refresh-observed`.

What this does not establish: it was traced from the code, not driven through the CLI; it needs a hand-built or hand-edited cloud assembly, so it is not reachable from an ordinary `cdk synth`.

## Proposed fix

Gate the substituted command the way `malformedDestroyResourcesRefusalMessage` does: substitute only when `safeStackName(stackName) === stackName` and (when a region is present) `safeRegion(region) === region`; otherwise fall back to the template form `cdkd state orphan <stack> --stack-region <region>` that `dropRecordCommand` already emits when it has no trusted stack. Then update the notes that describe the missing gate: go-to-k/cdkd#3226 has already rewritten `dropRecordCommand`'s note and the `safeStackName` note to cite this issue rather than to call the substitution safe, so once the gate exists they should say it is gated.

## Verification plan

- A refusal built with stack `Prod ` (trailing space) and region `us-east-1` prints the template form, not `cdkd state orphan Prod …`.
- The same with a control character and with a non-ASCII character in the stack name.
- An ordinary exact name (`My-App-Stack`) still prints the substituted command, so a gate that always withholds is caught.
- A region that sanitizing alters falls back to the template as well.
- Mutation probe: removing the gate reddens the first case.

Dup-check: searched open and closed issues for `dropRecordCommand exactness gate`, `orphan refusal sanitized stack name different record` and `prebuilt assembly stackName unvalidated`. No hits. Nearest: go-to-k/cdkd#3359 (the same refusal's region for a legacy record — a dead end, not a different target) and go-to-k/cdkd#3350 (`cdkd orphan` saving unreadable entries reshaped).
Session-fit: now (do it in this session) — found by the review of go-to-k/cdkd#3226's rebase onto go-to-k/cdkd#3346, which read `dropRecordCommand`, `orphan.ts` and the synthesis readers; nothing external blocks it.
Severity: medium — a destructive command can name a healthy record, but only from a hand-built or hand-edited cloud assembly.
Effort: small (S) — one gate in `dropRecordCommand`, two notes and unit cases.
Estimate: 1-2 hours

nix-tkobayashi:
Closes #3359
Closes #3360

## Summary

`cdkd orphan`'s `properties` refusal (go-to-k/cdkd#3318) ends on two commands built from the region it is handed. For a stack whose only record is a legacy `<prefix>/<stack>/state.json` listed with no region, `pickStackRegion` loads it under the SYNTHESIZED region (`getState` falls back to the legacy key for any region), and the refusal printed that region into both `cdkd state orphan <stack> --stack-region <region>` and `cdkd state show <stack> --stack-region <region> --json`. Both commands select a record by the region it is LISTED under, so neither found it.

- `src/cli/commands/orphan.ts`: `pickStackRegion` now returns `{ region, recordRegion }`. `recordRegion` is the selected ref's own region, which is `undefined` for a legacy ref listed with none. The properties refusal takes `recordRegion` plus a `LockRecoveryContext` (profile / resolved bucket / prefix, the object the lock-contention hint already built). The load, the lock and every other use keep `region`.
- `src/state/malformed-resources-bag.ts`:
  - With `undefined`, the drop remedy takes `dropRecordCommand`'s documented bare arm (`cdkd state orphan <stack>`). `cdkd state show` refuses a region-less legacy record with or without the flag, and the `cdkd deploy` it suggests for migration refuses this torn record, so for a known stack with no region the new `orphanInspectClause` names the S3 object instead, with the real prefix and the bucket when the caller threads them — on trailing `Object key:` / `State bucket:` lines, never inside the sentence, because a quoted value after English prose is only as safe as the apostrophes before it. Decision recorded in its doc: point at the object rather than teach `state show` to read a legacy record.
  - `dropRecordCommand` now substitutes the stack name and the region only when each renders EXACTLY at `STACK_REF_MAX_CODE_POINTS` (the new private `rendersExactly`); otherwise a template, and `withheldIdentityClause` says to take the identity from `cdkd state list --json` (raw names via `JSON.stringify`; `--long` renders through `displayIdent`, which trims). That is go-to-k/cdkd#3360, which this closes.
  - The `cdkd state show` line of the same refusal is built by a new private `orphanInspectCommand` under the same gate, so the two commands cannot disagree about a name: an altered name makes the name a hole, an altered region makes both holes. The shared `inspectCommand` keeps its substituted forms; only its no-identity template changes (quoted holes, below), and this builder no longer calls it.
  - The three commands this refusal prints — drop, `cdkd state list --json` and `cdkd state show`, substituted, template and no-identity template alike — carry the run's `--profile` / `--state-bucket` / non-default `--state-prefix`, so filling a hole can never produce an unqualified command.
  - Those three commands are printed LAST and UNWRAPPED, one per line after a fixed label (`Drop the record:`, `Find the exact name:`, `Inspect the record:`), rather than inside the prose's `'...'`. The prose still quotes fixed command names and templates that interpolate no value (`'cdkd orphan <its construct path>'`). Every placeholder in those commands is QUOTED through the exported `commandHole` (`'<stack>'`, `'<profile>'`, ...), since a bare `<profile>` is two shell redirections that, with flags after it, swallow `--state-bucket` when pasted; the shared `inspectCommand` no-identity template takes it too, so the deploy/diff properties refusal and warning and the destroy and divergent-region refusals now end on `cdkd state show '<stack>' --stack-region '<region>' --json`. Measured before the change: pasting the prose-quoted drop command WITH its quotes turned `shellQuote`'s quoting inside out, and `--state-bucket 'b; printf INJECTED; #'` ran `printf INJECTED` in bash. The RELEASED text already wrapped the stack name and region that way, and `displaySafe` keeps `'` and `;`, so a prebuilt-assembly name or a planted key segment broke out on the shipped binary too; the `cdk.json`-plantable bucket this PR added was the newest carrier. `.claude/rules/lock-contention-message.md` gains a sub-bullet recording it.
- `src/state/lock-contention-message.ts`: the flag rule is extracted from `buildForceUnlockCommand` into the exported `recoveryCommandFlags`, so the two commands cannot disagree about which flags a recovery command carries. It carries go-to-k/cdkd#3390's sanitize + exactness pair (issue go-to-k/cdkd#3377, merged to main while this was in review) and returns `{ flags, exact }`. `buildForceUnlockCommand` suppresses on `!exact`, as go-to-k/cdkd#3390 has it. The orphan refusal prints an inexact fragment as a `<profile>` / `<bucket>` / `<prefix>` hole, and its object path omits an altered bucket or prefix. One behaviour change for `force-unlock`: an explicit `--state-prefix ''` (accepted — the option has no `argParser` — and keying every record under `/`) is emitted as `--state-prefix ''` instead of being dropped. `.claude/rules/lock-contention-message.md`'s sentence saying an empty fragment emits nothing is amended for the prefix. Every other input renders as go-to-k/cdkd#3390 left it.
- `docs/state-management.md`: a paragraph on the legacy exception and on what the printed remedy carries. `.claude/rules/state-malformed-properties-orphan.md`: a pre-existing sentence corrected (a forced run would cost a write, not a lock, since the lock is taken before the refusal). The region decision itself is recorded in the `PickedStackRegion` and `orphanInspectClause` docs, not in the rule file, because touching `orphan.ts` already loads close to the 60,000 B rules-payload cap. Changelog entry added.

Out of scope: the sibling `state` / `outputs` refusals in the same command still receive the loaded region, so for this record shape their `cdkd state show --stack-region <synthesized>` inspect line selects nothing. That is go-to-k/cdkd#3388 (filed from the M1 finding; go-to-k/cdkd#3353 is the EMPTY region rendering as `<unrenderable>`, a different value and a different fix). Also filed rather than folded: the same prose-wrapped / bare-placeholder shape across the rest of `src/` — 59 candidate lines in 21 files by one grep, plus bare placeholders by another, including the destroy and divergent-region refusals' spelled-out `cdkd state orphan <stack> --stack-region <region>` (go-to-k/cdkd#3436); the three withhold arms that still point at `cdkd state list --long`, which has the `displayIdent` trim the `--json` choice above avoids (go-to-k/cdkd#3420); and routing the two older inline exactness tests in this module through `rendersExactly`, folded into go-to-k/cdkd#3388's body because go-to-k/cdkd#3226 rewrites one of those lines.

## Verification

Issue go-to-k/cdkd#3359's plan, bullet by bullet:

- Single legacy ref with no region, synthesized `us-east-1`, torn survivor: the drop remedy is exactly `cdkd state orphan MyStack --state-bucket test-bucket`, no `us-east-1` anywhere in the text, no `cdkd state show MyStack` / `--json`, and the object is located by `Object key: cdkd/MyStack/state.json` and `State bucket: test-bucket` lines; `getState` is still called with `us-east-1` (`tests/unit/cli/orphan.test.ts`). Driving the real `cdkd state orphan` with the command the real builder renders for a legacy identity (`It's Legacy`, with a full recovery context, split the way a shell would) calls `deleteLegacyState("It's Legacy")` and not `deleteState` (`tests/unit/cli/state-orphan.test.ts`).
- Legacy body that names `eu-west-1`: the remedy carries `--stack-region eu-west-1` and the inspect line is `cdkd state show MyStack --stack-region eu-west-1 --json`.
- Region-keyed record keeps `--stack-region us-east-1` in both commands. With two listed regions and `--stack-region eu-west-1`, the remedy carries `eu-west-1`. `--profile prod --state-prefix custom` reach the drop command, and `custom/MyStack/state.json` plus the bucket reach the object line.

Issue go-to-k/cdkd#3360's plan (`tests/unit/state/malformed-resources-bag.test.ts`):

- `'prod-api '` (with and without a region), `' prod-api'` and an over-cap name (without), and an ESC-bearing and a non-ASCII name (with `us-east-1`) each get the template with the `cdkd state list --json` hint, keeping an exact region; a padded region (`'us-east-1 '`) beside an exact name withholds both, giving `cdkd state orphan '<stack>' --stack-region '<region>'`; `My-App-Stack` is still substituted, so a gate that always withholds is caught. A 166-code-point name keeps its full spelling in the command (at the 128 default it ended in `...`, which `state orphan` resolves to zero records at exit 0). A quote-bearing region is shell-quoted in both arms; a quote-bearing bucket too. An unknown or `''` name gets no hint. `undefined` and `''` regions both give the bare command, the object path and the fill-in note; an empty prefix gives `--state-prefix ''` and `/S/state.json`.
- The inspect line agrees with the drop line for a trailing-space, an ESC-bearing and an over-cap name (name hole, exact region kept), both become holes for a padded region, a quote-bearing name and region are shell-quoted on it, and a non-default or empty prefix rides on it. The no-identity drop and inspect templates carry the recovery flags. `''` and `undefined` regions give byte-identical WHOLE messages across three names; the helpers' own `''` floors, unreachable through the builder, are held by a source fence.
- `buildForceUnlockCommand` with `statePrefix: ''` emits `--state-prefix ''`. The premise is fenced as an invariant: `--state-prefix` has no `parseArg`, four values reach `opts()` verbatim, and `S3StateBackend` keys each verbatim in both layouts (`tests/unit/state/lock-contention-message.test.ts`).
- `tests/unit/cli/state-orphan.test.ts` drives the real `cdkd state orphan` with the command the builder renders for `It's Legacy` plus a full recovery context, split the way a shell would, and checks the record deleted and the prefix the backend was built with.

Mutation probes, one at a time, tree restored byte-exact between them, each red on the case named above:

| Mutation | Result |
| --- | --- |
| refusal call passes `targetRegion` again | red: the legacy runtime case and the source fence |
| inspect clause always emits `cdkd state show` | red |
| inspect-path exactness guard removed / measured at 128 | red / red |
| flag arm returns `recordRegion: undefined` | red |
| `shellQuote` on the path replaced with naive quoting | red |
| drop gate removed (always substitute) / measured at 128 | red / red |
| region exactness gate removed | red |
| recovery flags dropped from the remedy / from the hint | red / red |
| real prefix not threaded into the path / fill note removed | red / red |
| withheld-identity hint never printed / absent-name guard removed | red / red |
| bucket quoting removed / region quoting removed in either arm | red / red |
| `orphan.ts` does not thread `recovery` | red |
| prefix rule back to bare truthiness (`''` dropped) | red |
| m3 wording reverted | red |
| inspect line ungated again (M2) / altered region no longer holes the name | red / red |
| inspect line drops the recovery flags / inspect stack or region unquoted | red / red |
| template drops the recovery flags (M3) / no-identity drop or inspect template drops them | red / red |
| builder's `''` entry normalisation removed | red |
| hint's `''` floor removed | red (source fence only; unreachable through the builder) |
| a trimming `--state-prefix` argParser / a backend that trims the prefix | red / red |
| drop stack name unquoted (shell-split drive test) | red |
| inexact profile / bucket omitted instead of a hole, including values that sanitize to nothing | red / red |
| `force-unlock` ignores an inexact fragment | red (go-to-k/cdkd#3390's own cases) |
| object path takes an altered bucket or prefix / fill note shown for a supplied-but-altered prefix | red / red |
| drop / list / inspect line re-wrapped in quotes | red / red / red |
| all command lines removed / a value-carrying command put back in the prose | red / red |
| `commandHole` back to a bare `<name>` (bash-paste test: argv and decoy files) | red |
| listing line gated off without a region / a value flag in the prose | red / red |
| `identityWithheld`'s empty-stack-name floor removed, or made a no-op | red / red |
| legacy sentence restored with its apostrophe and the bucket inside it (prose-paste test: sentinel file) | red |
| withhold hint back to "(shell-quote them)" | red |
| fill-in note printed unconditionally | red |

- `vp run check`, `vp run typecheck:test`, `vp run build` pass; `gen:all-matrices` left no drift.
- Full unit suite: 1079 files, 25407 passed / 1 skipped (`--maxWorkers=25%`) at `9ae91640`. Under heavy host load (load average up to 50, other sessions' suites running) earlier runs timed out in `tests/unit/scripts/*` spawn-heavy files unrelated to this change; each re-ran green in isolation, and the recorded run above is a clean full pass.
- Live CLI run: not done. A region-less legacy record plus a torn `properties` bag cannot be produced by a current writer without hand-editing the state bucket, and an inexact stack name needs a hand-built cloud assembly. The unit cases drive the real message builder and the real `cdkd state orphan` command instead.
- Codex review: the delta loop closed on NO ISSUES for each stage — before the first review, on each of the three reviews' responses, on the go-to-k/cdkd#3390 merge resolution, and on the unwrapped-commands fix — and a maintainer-checklist proxy pass ran after each until NO ISSUES. Findings those passes added beyond the reviews: the hint's `--long` had the `displayIdent` trim; an explicit `--state-prefix ''` was dropped by the shared flag rule; the hint and then the no-identity templates lost the account context; a fragment that sanitizes to nothing was omitted rather than holed; the prose-quote wrapper let a quoted value execute when pasted; two comments over-promised what a quoted hole does; and several clauses were asserted rather than pinned. The PR body was reviewed separately.
- Rebased onto `fedd09fd` (go-to-k/cdkd#3369 touches the same three files in disjoint hunks; its `tryGetLegacy` still reads a region-less legacy record from any region, which is the premise here), then flattened to one commit (`3f9ad68c`) on `87250f74`, resolving the conflict with go-to-k/cdkd#3390 described above, with the unwrapped-commands fix on top as `0fc40c9f` the round-3 response as `c5f07636`, and the round-4 response as `9ae91640`. Main has moved since with no overlap and merges cleanly. go-to-k/cdkd#3390's tests in `tests/unit/state/lock-contention-message.test.ts` and `tests/unit/cli/local-profile-display-population.test.ts` pass unchanged. Trial merge against go-to-k/cdkd#3226's head: `git merge-tree --write-tree HEAD f5b13ad9` (its head at this push; `80df6d02` before it gave the same result) CONFLICTS in `src/state/malformed-resources-bag.ts`, in two hunks: `dropRecordCommand`'s doc, and its body together with the new `rendersExactly` this PR adds beside it; the test file merges cleanly. go-to-k/cdkd#3226's side there renames `safeIdentifier` to its new `safeStackName` / `safeRegion` and adds a doc sentence saying the command substitutes WITHOUT an exactness gate, citing go-to-k/cdkd#3360 — the gate this PR adds. Resolution for whichever PR merges second: take this PR's side of both hunks (the gate is `safeStackName(v) === v` spelled as `rendersExactly`, the same cap), and update the `safeStackName` doc sentence go-to-k/cdkd#3226 adds, which calls `dropRecordCommand` ungated. (At `b37570b6`, before go-to-k/cdkd#3226's latest push, the trial merge was clean.)

No integ run: message text on a refusal path plus a return-shape change in a module-private helper and a flag-rule extraction, with no AWS call and no deletion logic touched.


go-to-k:
Reviewed at head `3f116f7a7` — 4 axes in parallel (spec / code / test / security), plus my own verification of the two findings the code and security axes converged on.

**No blocker.** The verification story holds up: I re-ran the mutation table and every probe the body calls red is red, on the test it names. The `''`/`undefined` normalisation, the `shellQuote` composition, and the `safeIdentifier` exactness gate on the new path all survive hostile input — 31 name classes driven (LF, CR, NUL, ESC, C1 CSI, NEL, U+2028, RLO/isolates, ZWJ/ZWSP/BOM, combining marks, lone surrogates, astral, leading/trailing space, over-cap) and not one raw control character reaches any clause. `Closes #3359` is earned: the issue's named approach was implemented, its named alternative demonstrably not taken, and the inspect-line decision it demands ("and says which") is recorded in the tree rather than only in the PR body.

Two things to fix before this merges, then the rest is take-or-leave.

---

### M0 — required before merge: the new exactness gate has a destructive sibling one sentence away that has none, and this PR removed its last narrowing token

`orphanInspectClause` withholds its path unless `safeIdentifier(stackName, STACK_REF_MAX_CODE_POINTS) === stackName`. `dropRecordCommand` two functions up substitutes the same name with no such test, at `IDENT_CAP_DEFAULT` (128). Before this PR that mattered less, because the command it built still carried `--stack-region`. It no longer does for this record shape, so the substituted name is now the ONLY thing narrowing a command that deletes. Two live renderings, both from `malformedOrphanResourcePropertiesRefusalMessage` at this head:

**(a) the 166-code-point nested name your own test pins.** One message, two spellings of one name:

```
… drop it whole with 'cdkd state orphan 'Root~NNNN…(cut at 128)…CCC...''
…
… it is '<prefix>/Root~NNNN…(all 166)…CCCC/state.json' in the state bucket
```

The path is right. The pasteable destructive command above it ends in a literal `...`, resolves to zero refs, and `stateOrphanCommand` prints `No state found for stack: …, skipping` at exit 0 — a silent no-op remedy printed beside a correct path, in the same paragraph. `tests/unit/state/malformed-resources-bag.test.ts` asserts the full path for exactly this name, so the PR pins one half of the contradiction.

**(b) a name that sanitizes to a different, healthy one.** `'prod-api '` (one trailing space) renders:

```
… drop it whole with 'cdkd state orphan prod-api' …
… it is the legacy 'state.json' under this stack's name in the state bucket …
```

The path clause correctly withholds — the name did not render exactly. The delete command substitutes the trimmed name anyway, unquoted, and the bare form targets every record named `prod-api` in every region. `recordRegion === undefined` is only reachable from the single-ref arm, so the record the refusal is *about* is one record; the command it hands the operator is not about that record at all. `displaySafe` trims before `PLAIN_IDENT` runs — the same padding spoof `displayIdent`'s rule-3 note and `lock-contention-message.ts` both close by comparing against the raw text.

Reachability for (b): the name comes from the synthesized assembly, and nothing in `src/` validates a stack name, so it needs a hand-crafted `cdk.out` rather than a CDK app. That is why this is not a blocker. It is still the case that the one gate this PR adds is on the read-only clause and not on the destructive one beside it, and `state orphan`'s confirm prompt is the only thing left between the paste and the delete.

One fix covers both: gate `dropRecordCommand`'s substituted form on the same `safeIdentifier(name, cap) === name` test the new clause uses, falling back to its existing template arm, and use one cap for the stack name across the message. If widening `dropRecordCommand`'s cap is too broad a blast radius for this lane — it is shared with the destroy / state / outputs builders — say so and gate on the 128 cap instead; the asymmetry is what needs closing, not any particular cap.

### M1 — required before merge: the out-of-scope residual is sent to an issue that does not cover it

The body says the sibling `state` / `outputs` refusals "still receive the loaded region. That is go-to-k/cdkd#3353." #3353 is a different defect — an EMPTY region rendering as `<unrenderable>` — and its fix is normalising `''` → `undefined` at the builder entry. That does nothing for the residual this PR leaves: `orphan.ts:232` / `:245` still pass `targetRegion`, which for a legacy region-less record with a synthesized region is the SYNTHESIZED region, not `''`, so `malformedOutputsRefusalMessage` prints `cdkd state show MyStack --stack-region us-east-1 --json` and `resolveSingleRegion` rejects it with `Available regions: (legacy)`. That is #3359's defect class on a sibling call site 40 lines up, and nothing tracks it — a lane sent to #3353 closes it without touching this.

Please file it (or widen #3353's body to cover both) and correct the PR-body sentence to point at whichever issue ends up owning it.

---

### Take-or-leave — merge without these if you prefer

- **m2** — `<prefix>` stays a placeholder the operator must fill in, and the sentence then has to explain how to fill it. `orphan.ts` holds the resolved prefix (`options.statePrefix`, used at `:143` / `:214`) and `src/state/state-prefix.ts` exports `DEFAULT_STATE_PREFIX` precisely so a `src/state/**` message module can decide the wording — `lock-contention-message.ts` does exactly this one file away. Threading it prints the real key and lets the trailing parenthetical go away. Same call: the drop remedy carries no `--profile` / `--state-bucket` / `--state-prefix` either, though `cdkd orphan` accepts all three, so after `cdkd orphan --profile prod …` the pasted command resolves the default profile's bucket. `buildForceUnlockCommand` is the in-repo shape for qualifying it.
- **m3** — "The record is a legacy one with no region" asserts something the code cannot know: the same arm fires when `readLegacyRegion`'s probe FAILED (403, 503, non-string `region`), as that function's own JSDoc says. "listed with no region" is true in both cases, and the remedy stays correct either way.
- **m4** — the cap-raise justification in `orphanInspectClause`'s doc ("an ordinary multi-level nested-stack name truncates") is borrowed from `malformedDestroyResourcesRefusalMessage`, whose caller reads names off S3 keys. `cdkd orphan`'s name comes from the assembly, where a `Parent~Child` record name cannot appear. The gate is still right — cite the hand-crafted-manifest reachability instead, which is also what makes M0(b) real.
- **m5** — two small test asymmetries: the `''` spelling asserts `not.toContain('--json')` but never the `'<prefix>/S/state.json'` its `undefined` twin asserts, and the hostile-name loop covers ESC and over-cap but not the trailing/leading-space class, which is the one M0(b) turns on.

---

Not merging this round. Push M0 and M1 and I will re-review; the take-or-leave items need no reply if you are skipping them.

nix-tkobayashi:
Thanks — pushed as `9566b5a5`. Both required items and all four take-or-leave ones are in.

**M0** — `dropRecordCommand` now substitutes only when the stack name AND the region render exactly, at `STACK_REF_MAX_CODE_POINTS`, through one private predicate (`rendersExactly`) that the inspect line uses too, so the two clauses cannot disagree about a name. Widening the cap had no blast radius: `dropRecordCommand` has exactly one caller (the orphan message); it is `inspectCommand` that is shared, and that one is unchanged. Your (a): the 166-code-point name now keeps its full spelling in the command — pinned, with the `...` form asserted absent. Your (b): `'prod-api '` gets the template `cdkd state orphan <stack>` (no flag when the record is listed with none, the exact region kept when there is one, a `<region>` hole when it is the region that is off), and the message says to take the identity from `cdkd state list --json` rather than from its own prose. `--json` rather than `--long` on purpose: `--long` goes through `displayIdent`, which trims, so it would hand back the very spelling the gate refused. The same holds for the destroy refusal's `cdkd state list --long` pointer, which I have left alone — yours to call. This is also go-to-k/cdkd#3360 as filed, so the PR now carries `Closes go-to-k/cdkd#3360`.

**M1** — you are right that go-to-k/cdkd#3353 is the empty-region rendering, a different value with a different fix. Filed go-to-k/cdkd#3388 for the two sibling call sites (`refuseMalformedState` / `refuseMalformedOutputs` still receive the loaded region, and their `cdkd state show --stack-region <synthesized> --json` selects nothing for this record shape; the fix needs the object-path arm as well, since `cdkd state show <stack> --json` fails on that record too). The PR body's out-of-scope sentence now points there and says why go-to-k/cdkd#3353 does not cover it.

**m2** — threaded. `orphan.ts` builds one `LockRecoveryContext` (profile / resolved bucket / prefix, the object the lock-contention hint already built) and hands it to the refusal; the substituted drop command, and the `cdkd state list --json` hint, carry `--profile` / `--state-bucket` / non-default `--state-prefix` through `recoveryCommandFlags`, extracted from `buildForceUnlockCommand` so the two rules cannot drift. The object line prints the real prefix and names the bucket, and the fill-in parenthetical only appears when no context was threaded. One behaviour change in `force-unlock`'s hint fell out of that extraction: an explicit `--state-prefix ''` — accepted (no `argParser`) and keying every record under `/` — was dropped as falsy and is now emitted as `--state-prefix ''`; the rule is spelled as the old truthiness test plus that one value, so nothing else moves.

**m3** — the arm now says the record is *listed* with no region ("a legacy `state.json` whose body names none, or one the listing could not read"), pinned.

**m4** — the cap justification now cites the hand-crafted-manifest reachability (an assembly name cannot acquire a nested child's key spelling; nothing validates it either), in the doc and in the test name.

**m5** — the `''` twin asserts the object path; the hostile loop has the leading- and trailing-space class; and every clause above is mutation-probed (the table in the PR body lists each, one mutation per probe, tree restored between).

Rebased onto `fedd09fd` on the way (go-to-k/cdkd#3369 shares the three files in disjoint hunks; the legacy read premise is unchanged there). Suite: 1072 files, 25249 passed.

One thing for merge order: go-to-k/cdkd#3226's latest head (`80df6d02`) now conflicts with this branch in `src/state/malformed-resources-bag.ts`, in two hunks — `dropRecordCommand`'s doc, and its body together with the new `rendersExactly` beside it. Its side renames the helpers and adds a doc sentence saying the substitution is ungated (go-to-k/cdkd#3360), which is what this PR changes. Whichever merges second takes this side of both hunks and updates the `safeStackName` doc sentence go-to-k/cdkd#3226 adds. Its `safeStackName(v)` is `safeIdentifier(v, STACK_REF_MAX_CODE_POINTS)`, the same test `rendersExactly` makes, so nothing about the gate has to change.

go-to-k:
Round 2, reviewed at head `9566b5a5` — same four axes, plus my own rendering of the two findings that landed inside this round's fix.

**M0 and M1 are both closed, and closed well.**

- **M0** — the gate holds. The real builder was driven over 32 hostile identities (leading/trailing space, BEL, ESC, U+009B, U+0085, U+2028, U+202E, U+2066, ZWJ, ZWSP, BOM, LRM, ALM, combining mark, lone surrogate, astral, 1153 code points, newline), on the stack axis and the region axis, and every one withholds. `My-App-Stack` and a 1152-code-point name still substitute, so it is not always-withhold — and that case is pinned, which is the half a gate usually loses. The mutation table re-ran 24/24 red at this head, including an isolated probe that blinds `rendersExactly` to the space class only and reddens the DESTRUCTIVE assertion, not just the path withhold.
- **M1** — #3388's repro checks out line by line, it is genuinely not #3353, and it carries the four fields and the dup-check.
- `Closes #3359` and `Closes #3360` are both earned; #3360's five verification bullets each have a case.

Two things this round's fix introduced. Both are at the same site and both are small; after these I have nothing else required.

---

### M2 — required: the exactness gate covers the drop command but not the inspect line beside it, so one message now contradicts itself

`rendersExactly`'s doc says it is "One predicate for the destructive remedy and the inspect line, so the two cannot disagree about a name." `orphanInspectClause`'s first branch returns `inspectCommand(stackName, region)` whenever `region !== undefined`, and `inspectCommand` is ungated and still at the 128 cap. Rendered from the real builder at this head, stack `'prod-api '` (one trailing space), region `'us-east-1'`, one message:

```
DROP : cdkd state orphan <stack> --stack-region us-east-1
TAIL : Inspect the record with: cdkd state show prod-api --stack-region us-east-1 --json
```

The message tells the operator the identity did not render exactly and to take it from `cdkd state list --json` — and then hands them the trimmed spelling anyway, naming the healthy sibling. Before this PR both clauses substituted ungated, so they agreed; gating one of the two is what created the disagreement, and the doc asserting they cannot disagree is now false.

Read-only, so not a blocker. But it is the M0 asymmetry with the sides swapped, and it makes the new predicate's own documented guarantee untrue.

`inspectCommand` has five callers, so do not widen it. Lines 382 and 531 already show the shape for this arm: when the identity is not exact, return the template. Whatever you choose, `rendersExactly`'s doc needs to describe what it actually covers.

### M3 — required: the withhold arm drops the account qualification the clause next to it carries

m2 threaded `--profile` / `--state-bucket` / `--state-prefix` onto the substituted command. The withheld arm did not get them. Rendered with recovery fully populated:

```
exact name   : cdkd state orphan MyStack --stack-region us-east-1 --profile prod --state-bucket cdkd-state-… --state-prefix custom
inexact name : cdkd state orphan <stack> --stack-region us-east-1
… take them from 'cdkd state list --json --profile prod --state-bucket cdkd-state-… --state-prefix custom'
```

So the operator reads the raw name out of a fully qualified listing, fills the one hole, and runs an UNQUALIFIED delete — which resolves the default profile's bucket. `stateOrphanCommand` then prints `No state found for stack: …, skipping` and exits 0, so it reads as done; if a same-named record exists in the default account, that is the one the prompt offers. The identity is the hole, not the account — emit `recoveryCommandFlags(recovery)` on the template arm too.

---

### Take-or-leave

- **m4** — `rendersExactly` is introduced as "the one test", but two pre-existing sites still spell `safeIdentifier(x, cap) === x` inline at the same cap (`:357`, `:499`). Three spellings of one security predicate is the drift shape this repo fences elsewhere.
- **m5** — `withheldIdentityClause` omits the `region === ''` normalisation `dropRecordCommand` duplicates on purpose "so a fourth caller added later inherits the floor". With `region: ''` the two disagree.
- **m6** — the bucket and prefix are `shellQuote`d but never `displaySafe`d, unlike every other value in this module; quoting does not neutralise an ESC. Operator-supplied, so self-injection only, and pre-existing in `buildForceUnlockCommand` — but this PR newly puts the unsanitized prefix inside a printed S3 key. Related: with `--state-prefix ''` the command renders inside the prose wrapper as `… --state-prefix '''`, which is the LAST-and-UNWRAPPED composition `safeIdentifier`'s own note warns about. Not injectable — I fed it to bash and got `unexpected EOF`, never execution.
- **m7** — `(prefix || prefix === '')` is `prefix !== undefined`; and `parts` in `buildForceUnlockCommand` is now a one-element array built only to be spread.
- **m8** — `--state-prefix ''` reaching the key verbatim is an unfenced premise. I confirmed it independently (`src/cli/options.ts:118` has no `argParser`; `getStateKey` yields `/S/…`), so it is right today — but an `argParser` added later would make both the emitted flag and the printed object path wrong with zero reds.
- **m9** — the one "the remedy actually works" drive test uses `cdkd state orphan LegacyStack`, a form the binary never emits (production always adds at least `--state-bucket`), and its `.split(' ')` means no quoted identity is ever driven.

### Not yours to fix in this PR, but worth recording

Your `--json` over `--long` call is right and is recorded in the tree — thank you for putting it there rather than only in the reply. It is under-enumerated though: `--long` goes through `displayIdent`, which trims *and* JSON-quotes, and three sites end on a pointer with that property (`malformed-resources-bag.ts:381`, `:531`, `src/cli/commands/export.ts:3474`), not just the destroy refusal you named. By the same standard I applied to M1, that residual wants an issue rather than a PR comment. File it whenever — it does not gate this PR.

Also worth a comment on #3388: its "Not #3353" section is one-directional. `recordRegion` can never be `''`, so landing #3388 removes #3353's measured repro and leaves #3353 as builder-side hardening only. Without that note a lane will re-measure #3353 and find nothing.

I re-ran the #3226 trial merge myself — one conflicted file, `src/state/malformed-resources-bag.ts`, exactly as you described. Your merge-order note is accurate; I will take this side of both hunks.

---

Not merging this round. Push M2 and M3 and I will re-review and merge; the take-or-leave items need no reply if you are skipping them.

nix-tkobayashi:
Pushed as `0fc40c9f`. M2 and M3 are in, and of the take-or-leave items m5, m6, m7, m8 and m9 are too; m4 is routed rather than folded, for the reason below.

**M2** — the `cdkd state show` line is now built by a new private `orphanInspectCommand`, gated and qualified exactly like the drop command two sentences above it: an altered name makes the name a hole, an altered region makes both holes, and the run's `--profile` / `--state-bucket` / non-default `--state-prefix` ride along. Your rendering, `'prod-api '` with `us-east-1`, now gives `cdkd state show <stack> --stack-region us-east-1 --json`, and a case asserts that the drop line and the inspect line agree for each hostile name. `inspectCommand` is untouched: the orphan builder no longer calls it at all. Its no-identity arm uses its own qualified template too. `rendersExactly`'s doc now says what it covers (the drop command, the inspect line and the object path of this one refusal) and what it does not (the two inline spellings of m4, and `inspectCommand` for its four other callers).

**M3** — the template arm of `dropRecordCommand` keeps `recoveryCommandFlags`, and so does its no-identity template (the proxy pass found that one after I had fixed only the arm you named). So filling the hole can no longer produce an unqualified delete.

**m5** — the hint's region check carries the `''` floor now; it lives in `identityWithheld`, which `withheldIdentityClause` and the new listing-line helper both call. Worth saying plainly: it is unreachable through the builder, which normalises `''` at entry, so no rendered message can observe it (removing it left every rendered case green). It is held by a source fence that requires the floor in `dropRecordCommand` and `identityWithheld`. The rendered-message case compares `''` against `undefined` as WHOLE messages instead, which is what pins the entry normalisation.

**m7** — `prefix !== undefined`, and `buildForceUnlockCommand` builds one string instead of a one-element array.

**m8** — fenced as an invariant rather than one value: the option's `parseArg` must be undefined, four values (`''`, `' padded '`, `a/b`, `custom`) must reach `opts()` verbatim, and `S3StateBackend` must key each of them verbatim in both layouts. A trimming `argParser`, and a backend that trims the prefix, each redden it.

**m9** — the drive test now renders with a full recovery context and the name `It's Legacy`, whose space and quote make an unquoted rendering name a different stack. It asserts the rendered quoting, splits the command the way a shell would, drives `cdkd state orphan` with that argv, and checks the backend was built with the pasted prefix.

**m4** — not folded here. go-to-k/cdkd#3226 rewrites the first of those two inline lines into `safeStackName(...) === ... && safeRegion(...) === ...`, so touching it would add a third conflict hunk to the two you are already resolving. I have put it on go-to-k/cdkd#3388, which works in this module after both land, together with your note that its "Not go-to-k/cdkd#3353" section is one-directional.

**m6** — both halves are in, and the second turned out to be more than a paste nicety. You measured `--state-prefix '''` giving `unexpected EOF`, but a crafted value does execute. With `stateBucket: 'b; printf INJECTED; #'`, the real builder rendered `drop it whole with 'cdkd state orphan S --state-bucket 'b; printf INJECTED; #'', which leaves ...`, and pasting that quoted span into bash ran `printf INJECTED`. The wrapper closes at the value's opening quote. The command between the wrapper quotes is safe on its own. The bucket is the plantable one (`context.cdkd.stateBucket` in a cloned repo's `cdk.json`), and it is m2 of this PR that put it in the wrapped command. So I fixed it here rather than filing it. Every pasteable command now prints LAST and UNWRAPPED, one per line after a fixed label (`Drop the record:`, `Find the exact name:`, `Inspect the record:`). That is the rule `.claude/rules/lock-contention-message.md` already records for the force-unlock hint; that rule now has a sub-bullet with this measurement. The prose still quotes fixed command NAMES (`'cdkd orphan'`, `'cdkd deploy'`), which interpolate nothing. A test pins that no command carrying a value is in the prose, across the exact, withheld, legacy and no-identity arms. The exact label sequence is asserted per case, and the builder-parametrized line-forgery tests now compare the label sequence against a benign input. The object path in the legacy arm still names the bucket, as one shell-quoted location token.

The other half landed on main while this was in review. go-to-k/cdkd#3390 (issue go-to-k/cdkd#3377) sanitizes all three recovery fragments in `buildForceUnlockCommand` and conflicted with this PR's extraction. I rebased onto current main and moved that rule into `recoveryCommandFlags`, which now returns `{ flags, exact }`. `buildForceUnlockCommand` suppresses on `!exact` exactly as go-to-k/cdkd#3390 has it, and go-to-k/cdkd#3390's tests pass unchanged. This refusal prints an inexact fragment as a `<profile>` / `<bucket>` / `<prefix>` hole, never as the altered text and never omitted, and the object path names neither an altered bucket nor an altered prefix. `--state-prefix ''` is still emitted. go-to-k/cdkd#3390's rule sentence said an empty fragment emits nothing, so I amended it for the prefix.

**`--long` pointers** — filed as go-to-k/cdkd#3420 for the three sites you listed (`malformedDestroyResourcesRefusalMessage`, `divergentRecordRegionRefusalMessage`, `export.ts`'s `orphanCommandFor`). It carries the `displayIdent` measurement: `'prod-api '` and `' prod-api'` both render `"prod-api"`.

Each new clause was mutation-probed, one mutation at a time, and the PR body's table lists them. The three reviewed commits are now one (`3f9ad68c`) on `87250f74`, with the unwrapped-commands fix on top as `0fc40c9f`; main has moved since with no overlap and merges cleanly. The trial merge against go-to-k/cdkd#3226's head `80df6d02` still conflicts in the same two hunks of `src/state/malformed-resources-bag.ts` (`dropRecordCommand`'s doc, and its body with `rendersExactly`), so your resolution note stands. Suite: 1079 files, 25405 passed.

go-to-k:
Round 3, reviewed at head `0fc40c9f` — four axes again, plus my own reproductions.

**First, my error.** In round 2 I wrote that the prose-wrapped command was "not injectable — I fed it to bash and got `unexpected EOF`, never execution". That was one value, not the class, and the class is live. Reproduced independently on the pre-fix tree (`9566b5a5` built out into a scratch dir, real builder, the span a human would select): `'cdkd state orphan MyStack --stack-region us-east-1 --state-bucket 'b; printf INJECTED; #''` → `INJECTED`. The wrapper closes at the value's own opening quote, `;` separates, `#` eats the tail. `--state-prefix '''` is a syntax error, which is the wrong shape, not evidence of safety. Thank you for not taking my measurement at face value.

**M2 and M3 are closed, and the injection fix closes the CLASS.** The same harness over 324 arms (six message shapes × hostile values in stack / region / profile / bucket / prefix × 12 payloads including `$( )`, backticks, `|`, `&&`, embedded quote, raw newline), with execution observed by a PATH stub writing a sentinel rather than by string matching: **0 executions at this head, 1,576 on `9566b5a5`** — so the probe is live rather than vacuous. The `#3390` restructure was differenced over 656,250 input tuples: 2,352 diffs, every one of them `statePrefix: ''`, which is the declared change. Holes render as holes for an altered fragment and for one that sanitizes to nothing, never as altered text and never omitted. `Closes #3359` and `Closes #3360` both still earned. #3420's three sites and its `displayIdent` measurement check out.

Two required items, both from this round's own change. I will say plainly that this is the second round running where the fix carried something new — if a fourth round does it again I will take the remainder myself as a follow-up PR rather than keep you in the loop.

---

### M4 — required: the placeholders are shell redirections, and with flags now after them the paste silently loses `--state-bucket`

`<stack>` / `<region>` / `<profile>` / `<bucket>` / `<prefix>` are `<` and `>`. Until this round the hole was last, so pasting was a syntax error and nothing ran. `recoveryCommandFlags` now appends after it, which gives `>` a target. Measured with a stub `cdkd` on PATH and an unrelated file named `profile` in cwd:

```
$ cdkd state orphan 'MyStack' --stack-region us-east-1 --profile <profile> --state-bucket realbucket
rc=0, nothing on the terminal
$ cat -- '--state-bucket'
ARGS: state orphan MyStack --stack-region us-east-1 --profile realbucket
```

`<profile` takes stdin from the cwd file, `>` creates a file named `--state-bucket` and swallows stdout, and `realbucket` becomes the value of `--profile`. So the pasted line runs a delete with **no** `--state-bucket` — the exact harm M3 added those flags to prevent — against a bucket the operator cannot see, having shown them nothing. `confirmOrRefuse` refusing the redirected stdin is the only backstop, and `-y` removes it.

Quoting the holes fixes it (`--profile '<profile>'`, `--stack-region '<region>'`, `cdkd state orphan '<stack>'`); I confirmed the quoted form passes them through as literal arguments. Bracing them instead is fine if you prefer `{stack}`. Worth sweeping the pre-existing `inspectCommand` template in the same edit — same shape, and it is now printed unwrapped too.

### M5 — required: the vulnerability record understates its reach

The changelog entry, the rule sub-bullet and the doc comment all describe the vector as the `--state-bucket` this PR's own m2 introduced, which reads as "nothing shipped was affected". That is not so. On `origin/main` today:

```js
`repair the record by hand, or drop it whole with ` +
`'${dropRecordCommand(stackName, region)}', which leaves ` +
```

`dropRecordCommand` already `shellQuote`s the stack name and the region, and `displaySafe({asciiOnly:true})` keeps `'` and `;` — so the same break-out is reachable on the **shipped binary** through a prebuilt-assembly stack name or a plantable region, with no `--state-bucket` involved. Your fix closes that too; only the record is wrong. Please widen the sentence in all three places so a later reader can judge whether this needs anything beyond a normal release.

One correction to something my own reviewers flagged and I checked before passing on: `src/deployment/destroy-runner.ts` does not exist in this tree, so there is nothing to file for it. Rather than assert this is the only such site, a grep for other prose-wrapped value-carrying commands is worth doing while you are in there — and if you find any, they are an issue, not this PR.

---

### Take-or-leave

- **m10** — `withheldIdentityListCommand`'s withheld-name + absent-region arm has no case. Gating that line off left all 341 tests in the three files green; the prose then points at a `Find the exact name:` line that is not there, while the drop command is a `<stack>` hole, so the operator has nothing to fill it from. One row in the existing label loop (`['S ', undefined]` → `['Drop the record', 'Find the exact name']`, and `['S', undefined]` → `['Drop the record']`) pins both unasserted label sets.
- **m11** — the prose invariant greps `cdkd state (orphan|show|list) `. A value-carrying command with another verb could be re-wrapped without reddening it; asserting the prose contains no `--state-bucket` / `--profile` / `--state-prefix` would close it by property rather than by verb list.
- **m12** — `identityWithheld` floors `region === ''` but not `stackName === ''`, while `dropRecordCommand` floors both, so a later direct caller passing `''` gets the withheld clause while the drop command takes its no-identity arm. Unreachable through the builder — but the stated point of the floor is the later caller.
- **m13** — m4 and the one-directional-#3353 note landed on #3388 as a comment, while its title, proposed fix, verification plan and Effort are still about the two region arguments alone. A lane closing #3388 against its own plan leaves m4 undone. Fold it into the body or give it its own issue.

---

Not merging this round. Push M4 and M5 and I will re-review and merge.

nix-tkobayashi:
Pushed as `c5f07636`. M4 and M5 are in, and so are all four take-or-leave items (m10–m13). The grep turned up the same shape repo-wide, filed as go-to-k/cdkd#3436 rather than folded in.

**M4** — every placeholder the orphan refusal prints is now quoted, through one exported helper in `lock-contention-message.ts`, `commandHole(name)` → `'<name>'`. That covers the drop and inspect templates, the no-identity templates, and `recoveryCommandFlags`' fragment holes. Per your note, `inspectCommand`'s shared no-identity template is swept too. That changes the `Inspect ... with:` template printed by the deploy/diff properties refusal and warning and by the destroy and divergent-region refusals: each now ends on `cdkd state show '<stack>' --stack-region '<region>' --json`. Two existing assertions needed updating for it. One in `tests/unit/analyzer/diff-calculator-malformed-properties.test.ts` pinned the bare spelling (the full suite caught it; my targeted set did not include that file). The destroy refusal's absence check had assumed a quote after `cdkd state show ` always meant a substitution, so it now strips that template before looking. The pin is your measurement, turned into a test. Every hole-bearing command line this refusal prints, across six message shapes, is pasted into real bash under a stub `cdkd`, in a scratch directory holding decoy files named `stack` / `region` / `profile` / `bucket` / `prefix`. The test asserts that bash's argv equals a quote-aware split of the line and that no file appears. Reverting `commandHole` to a bare `<name>` reddens it on bash's evidence: argv comes back empty because stdout was redirected. I had first filtered the lines on `'<`, which would have reddened for the wrong reason, so the filter now matches holes with or without quotes. The helper's doc says only what quoting buys, literal argv. An unfilled hole is not refused; it runs as that literal value.

The destroy and divergent-region refusals still spell `cdkd state orphan <stack> --stack-region <region>` in their prose. They are the same shape B, outside this refusal, so they are in the grep issue below.

**M5** — widened in all three places: the builder comment, the `.claude/rules/lock-contention-message.md` sub-bullet and the changelog. The released text already put the STACK NAME and REGION inside the wrapper, and `displaySafe` keeps `'` and `;`. So a prebuilt-assembly name or a planted key segment broke out on the shipped binary. The `cdk.json`-plantable bucket was only the newest carrier.

**Grep** — the shape is repo-wide, larger than I first listed. My first draft of the issue named about twenty sites, and Codex's review of the draft found as many again, including `destroy-runner.ts`'s `hintFor`, which builds `'${command} ${t}'` with neither token on the line. So go-to-k/cdkd#3436 DERIVES the population instead of listing it. One grep gives 59 candidate lines across 21 files for the wrapped-value shape at `c5f07636`, a second covers bare placeholders followed by more words, and the issue names each grep's known blind spots and exclusions (e.g. `deploy-engine.ts`'s import hint already escapes the wrapper via `quoteSafe`). It proposes one `pasteableCommand` helper plus a source fence whose fixtures include a helper result re-wrapped by hand. Both shapes are measured in bash on a representative. No candidate is asserted exploitable until classified, since none was driven end to end. Filed `severity:high` / `effort:large`.

**m10** — two region-less rows in the label table: `['S', undefined]` gives `Drop the record` only, and `['S ', undefined]` gives `Drop the record, Find the exact name`. Gating the listing line off without a region reddens the second.

**m11** — the prose check is now by property as well as by verb: no `--state-bucket` / `--profile` / `--state-prefix` / `--stack-region` followed by a value, in any arm.

**m12** — `identityWithheld` floors an empty stack name too, and the structural fence matches the guard STATEMENT (block or `return` form) in both helpers. A substring check was satisfied by a contrived `(stackName === '' && false)`, which now reddens it.

**m13** — folded into go-to-k/cdkd#3388's body. The one-directional-go-to-k/cdkd#3353 note is now a paragraph there, and m4 is its own section, with a verification bullet, a fence and the Effort / Estimate updated.

Each new clause was mutation-probed one mutation at a time, and the PR body's table lists them. Main has not moved onto these files. The go-to-k/cdkd#3226 trial merge against its current head `f5b13ad9` still conflicts in the same two hunks (`dropRecordCommand`'s doc, and its body with `rendersExactly`); `inspectCommand` merges cleanly. Suite: 1079 files, 25406 passed.

go-to-k:
Round 4, reviewed at head `c5f07636` — four axes, plus my own reproductions of both items below.

**The redirect class is closed, measured rather than read.** 138 labelled command lines from the real builder (8 names × 6 regions × 6 recovery contexts) pasted into real bash under a stub `cdkd`, in a directory seeded with decoys named `stack` / `region` / `profile` / `bucket` / `prefix` / `json`: argv equals a quote-aware split on every line, nothing created, nothing truncated. The same harness against `0fc40c9f` gives **74 failures**, including lines that exit 0 having silently created `--state-bucket` / `--profile`, so the probe is live. Your bash test is genuine on all four counts I asked about — it really spawns bash, the decoys exist before the snapshot, reverting `commandHole` reddens it on bash's argv rather than on a string match, and the widened filter does select unquoted holes. `Closes #3359` and `Closes #3360` both still earned. #3436's greps reproduce exactly — 59 lines / 21 files, every per-file count and every named blind spot and exclusion checked.

Two items, both in text this round changed, both reproduced here. **These are the last things I will ask you for**: everything else the round turned up I am taking myself in a follow-up PR, as I said I would in round 3.

### B1 — an English apostrophe in the prose flips quote parity and un-quotes the plantable bucket

`malformed-resources-bag.ts:1964`:

```ts
return { sentence: `${lead}: it is the legacy 'state.json' under this stack's name ${where}.` };
```

`'state.json'` balances; `stack's` then OPENS a quote which closes at `shellQuote(bucket)`'s own opening quote, leaving the bucket value bare. Reproduced:

```
$ bash -c "… the legacy 'state.json' under this stack's name in state bucket 'evil; touch OWNED; #'."
bash: line 1: The: command not found ; rc=0
$ ls → OWNED
```

`exactOrUndefined` does not stop it: `sanitizeRecoveryValue` uses the denylist `displaySafe`, and `evil; touch OWNED; #` contains nothing it strips, so it renders exactly and is shell-quoted into the sentence. Reachable on the arm this PR added — a torn `properties` bag on a legacy region-less record whose name does not `rendersExactly` (the #3360 prebuilt-assembly arm), in a cloned repo whose `cdk.json` sets `context.cdkd.stateBucket`. The sibling branch at `:1974` has no stray apostrophe and is safe.

Your fence at `:3182` asserts the bucket appears only as `'<hostile>'` — syntactic quote PRESENCE, never quote CONTEXT — so its matrix renders this very message and passes.

I rate this below round 3's break-out, because it needs the operator to paste an English sentence rather than the command the message tells them to paste. I am still asking for it: it is your own LAST-and-UNWRAPPED rule, and the fix is the one that rule already prescribes — move the object location onto its own labelled trailing line. Deleting `stack's` alone leaves the class open. Worth extending the fence to paste every PROSE line into bash, not only labelled command lines.

### B2 — the withhold hint now tells the operator to do the thing that re-opens #3360

`malformed-resources-bag.ts:1838` still says `(shell-quote them)`. That was right when the hole was bare `<stack>`. It is now `'<stack>'`, so an operator following the parenthetical literally substitutes inside the existing quotes — `''prod-api ''`, which bash splits into `prod-api` and `''`, and `cdkd state orphan` then targets the healthy sibling. That is exactly the misdirection #3360 exists to close, arriving through the instruction instead of the command. `(replace the whole quoted hole, keeping its quotes)` or similar.

---

### What I am taking myself, so you do not have to

A follow-up PR after this merges, no action from you: the changelog sentence naming the four refusals whose `Inspect … with:` text this changes (the PR body declares it precisely; the entry does not, and it is at 1964/2000 so it needs a trim); the destroy-refusal absence check, whose `includes("cdkd state show '")` misses a `shellQuote`-bare name like `prod-api` and so was inert for two of its three cases; `Inspect … with:` pins for `divergentRecordRegionRefusalMessage` and `malformedResourcePropertiesWarning`, neither of which has one; the `''` floor on `orphanInspectClause`, which has it on the other two helpers; `toBeGreaterThanOrEqual(messages.length)` being a floor of 6 against 10 actual; m11's regex needing a trailing space; `layout-state-types.md`'s export list, now missing `commandHole` / `recoveryCommandFlags` / `sanitizeRecoveryValue`; and adding to #3436 the shape neither of its greps matches — a shell-quoted VALUE in prose whose context an apostrophe has already flipped.

Two notes for #3388 whenever it is convenient, not blocking: its title and `## Proposed fix` heading still scope to the two region arguments even though m4 now has its own body section.

For the record on why I am not asking for more: `origin/main` today carries the round-3 break-out in a COMMAND span, which is far likelier to be pasted than a prose sentence. Merging this is a net improvement to that surface, not a regression, and #3436 owns the rest.

---

Push B1 and B2 and I will merge.

nix-tkobayashi:
Pushed as `9ae91640` — B1 and B2, nothing else.

**B1** — the legacy region-less arm no longer puts any value in its sentence. The object's location prints on trailing labelled lines like the commands: `Object key: <shell-quoted key>` (only when the name renders exactly) and `State bucket: <shell-quoted bucket>` (only when it renders exactly). The sentence points at them and carries no apostrophe of its own; it reads "it is the legacy state.json under this stack name". So a quoted value never sits after English prose.

The fence is your suggestion, widened for one reason. A new test renders the real builder with hostile values in every slot (stack name, region, profile, bucket, prefix), pastes into real bash every line, every prose sentence and every clause of a sentence, and asserts that a sentinel file never appears. Whole prose LINES alone would have passed against the bug, because `record(s)` is a syntax error and bash then runs nothing on that line. Sentences and clauses are what an operator selects. Restoring the old wording (the apostrophe plus the bucket in the sentence) reddens it on bash's evidence: the sentinel is created. Moving the location also made two negative assertions on the fill-in note stale, since they searched for the old quoted wording. They now search for the new wording, and forcing the note to print unconditionally reddens them.

**B2** — the hint now says "replace each quoted hole, quotes included, with the shell-quoted value", pinned by the withhold cases. Putting back "(shell-quote them)" reddens them.

Thank you for taking the rest; I have not touched any of it. Suite: 1079 files, 25407 passed. The go-to-k/cdkd#3226 trial merge against `f5b13ad9` is unchanged: the same two hunks.

go-to-k:
Merged as `f371dd28`. Thank you — five rounds and it got better every one of them.

Three things I want to name, because they are the reason this merged rather than stalled. You did not take my round-2 measurement at face value: I had cleared the prose wrapper as not injectable, and you went and found the shape that actually executes. When I asked for a fence that pastes prose lines into bash, you noticed that line-level pasting would have passed vacuously — `record(s)` is a syntax error, so bash never reaches the payload — and split to sentences and clauses instead; I re-measured that and it is exactly right (line-only: 0 hits against the vulnerable build; sentence: 2; clause: 2). And when the fix for one item made a neighbouring assertion stale, you re-pointed it at the new behaviour and said so, rather than deleting it.

The rest is mine, as agreed — a follow-up PR for the changelog blast-radius sentence, the destroy-refusal absence check's `shellQuote`-bare gap, the two missing `Inspect ... with:` pins, `orphanInspectClause`'s `''` floor, the fence floors and the `(shell-quote them)` negative, `layout-state-types.md`'s export list, the prose-value shape for go-to-k/cdkd#3436, and the `lock-contention-message.md` clause generalising LAST-and-UNWRAPPED to a shell-quoted VALUE in prose. I will also fix go-to-k/cdkd#3420's proposed fix, which still tells a lane to say "shell-quote it" — the wording B2 removed.

Issue #3301, opened 2026-09-16
chore(hooks): a combined short-flag cluster carrying R is not read as a repo slug, so the merge gates still judge the cwd repo for it


go-to-k:
## What happens

`gate_gh_repo_slug` (added by go-to-k/cdkd#3273) reads the four spellings gh
documents and this repo writes -- `-R <slug>`, `-R<slug>`, `--repo <slug>`,
`--repo=<slug>` -- and does NOT read a COMBINED SHORT-FLAG CLUSTER carrying
`R`. `ci-green-gate` and `pr-review-gate` then judge the cwd repo for those,
exactly as they did before go-to-k/cdkd#3273:

```
gh pr merge 42 -sR go-to-k/cdk-local --admin     slug=[]  -> judges the cwd repo
gh pr view  -cR go-to-k/cdkd 3214                (the measured proof gh honours it)
```

gh honours the cluster: measured 2026-09-16 on gh 2.92.0 from a NON-repo
directory, `gh pr view -cR go-to-k/cdkd 3214` and the glued `-cRgo-to-k/cdkd`
both resolve the cdkd PR, and `verify-pr-gate.sh` records that measurement in
its own repo-naming guard, which refuses the shape rather than reading it.

## Why it was left open rather than folded into go-to-k/cdkd#3273

The cluster is NOT DECIDABLE FROM THE TEXT. Only a per-flag ARITY table
separates `-sR <slug>` (`--squash --repo <slug>`, where the next token is the
slug) from `-tRelease` (`--subject Release`, where `elease` is part of a title
and there is no slug at all), and a per-flag table is the enumeration
`.claude/rules/hooks-class-fences.md` exists to refuse -- it is the instrument
that produced the four-spellings-in-four-rounds treadmill recorded there.

The alternative -- over-REFUSE any `-[!-]*R*` token, which is what
`verify-pr-gate` does -- was considered and rejected for these two gates.
`verify-pr-gate`'s over-refusal falls back to a binding requirement the caller
can satisfy; on a MERGE gate a wrong refusal blocks the merge outright, and
`gh pr merge 42 -tRelease v2 --squash` is an ordinary command.

## Blast radius

None beyond today: these spellings get the same cwd-relative verdict they got
before go-to-k/cdkd#3273, so nothing regresses. It bounds how much of the
wrong-repo class that change closed rather than opening anything. The honest
statement is that four spellings are closed and the cluster is not.

## What would close it

[...]

Issue #3256, opened 2026-09-16
chore(hooks): decide verify-pr-gate's repo-override question on the SELECTOR POSITION, not on every token's text


go-to-k:
## What happens

`verify-pr-gate.sh`'s `__vpg_names_no_other_repo` decides "does this command name another repository" by scanning EVERY token of the matched gh segment and testing each one's TEXT. It has no notion of which token gh will actually read as the PR selector, so it must also model prose — today by testing the first whitespace-delimited word, so that a URL inside `--body` stays exempt.

That text-only shape produced a blocker in each of two consecutive review rounds of go-to-k/cdkd#3214, both in the same predicate and both because the token's text was a poor proxy for its ROLE:

- round 3: a QUOTED URL selector (`gh pr merge "https://github.com/go-to-k/cdkd/pull/5"`) took the relaxed path, because the quoted-token arm was reached before the `://` test;
- round 4: a URL with TRAILING WHITESPACE did, because gh prefix-matches `^/OWNER/REPO/pull/(\d+)` after `url.Parse` and so resolves `…/pull/42 `, `…/pull/42?x=a b`, `…/pull/42/files x` and a trailing NBSP.

The narrow fix that shipped is complete for text-in-argv shapes, and its completeness is proved rather than sampled: gh's `ParseURL` requires `http`/`https` + `:` at index 0 and a scheme cannot contain whitespace, so every selector gh accepts carries `://` in its first whitespace-delimited word. What remains is the SHAPE, not a missing spelling.

## The structural fix

The hook already has the position. `gate_verb_span` returns the offset past the verb and `gate_pr_selector` (`lib/command-match.sh:~2924`) already walks flags from `rest="${segment:$_gate_span}"` to find the first non-flag token after the verb. Deciding the question on the SELECTOR POSITION instead of on every token makes both shipped blockers structurally impossible and removes the prose heuristic — and with it the `--body "https://x is the link"` / `--body=https://x` over-refusals the current predicate has to accept.

[...]

PR #3372, opened 2026-09-17
chore(hooks): decide verify-pr-gate's foreign-target relaxation on what gh would resolve in the target checkout, not only on the command text

### LINKED ISSUE #3235: chore(hooks): verify-pr-gate reads the COMMAND for a repo override, but gh also resolves one from the target checkout (an upstream remote wins over origin)
`verify-pr-gate.sh`'s foreign-target relaxation (go-to-k/cdkd#3209) decides "does this command name a repo other than the target directory's?" by reading the COMMAND. gh does not only read the command: it resolves the repo from the target checkout's git config, and that channel leaves nothing in the text for the guard to see.

Measured 2026-09-16 in a fresh throwaway repo with `origin` = `https://github.com/go-to-k/cdk-local.git`, `upstream` = `https://github.com/go-to-k/cdkd.git`, NO `remote.*.gh-resolved` key and no interaction:

```
$ git -C <repo> config --get-regexp gh-resolved
(none)
$ cd <repo> && gh repo view --json nameWithOwner -q .nameWithOwner
go-to-k/cdkd
$ git -C <repo> remote remove upstream && gh repo view --json nameWithOwner -q .nameWithOwner
go-to-k/cdk-local
```

gh prefers an `upstream` remote over `origin`. So a plain `git remote add upstream <other repo>` — an ordinary agent operation, and the standard fork setup — silently re-points every later bare `gh pr merge <N>` in that checkout. `gh repo set-default <slug>` writes `remote.<name>.gh-resolved` and does the same deliberately. In both cases the allowlist added by go-to-k/cdkd#3214 sees an ordinary `gh pr merge 42 --squash`, relaxes, and a sibling repo's fresh `verify-pr` marker clears the merge of a pull request in ANOTHER repository — the exact 2 -> 0 shape go-to-k/cdkd#3214's review round found for `-R`, arriving through a channel the command text does not carry.

## The fix this needs, and why it is not in go-to-k/cdkd#3214

A remote-slug COMPARISON: resolve what gh would resolve in `$target_dir` and require it to equal that checkout's own repo, refusing the relaxation otherwise. A presence test is NOT a substitute and was rejected on measurement — refusing any `gh-resolved` holding a slug would refuse a sibling that ran `gh repo set-default` on ITSELF, which is what gh tells a multi-remote checkout to do, and that re-breaks the flow go-to-k/cdkd#3209 exists to enable.

That is a structural fix at the end of a three-round cascade (`-R` denylist -> allowlist -> this), and `.claude/skills/work-issues/references/verify.md` 8-a says to take the narrow fix and FILE the structural one rather than land new entrypoint code late in a cascade. go-to-k/cdkd#3214 states it as a KNOWN BOUND in the hook and points here.

Landing it would also retire the second-clone KNOWN BOUND in the same block: a second clone of cdkd is "foreign" only because the identity test is the git common dir, and a slug comparison answers the real question instead.

Two design constraints for whoever takes it:
- The resolution must NOT execute anything from the target repo (`.claude/rules/hooks.md` records why delegation was abandoned in PR 1970). Reading `git config` / `git remote` in the target is the same class of read the gate already does; running `gh` there is a live network call inside a PreToolUse hook and needs its own justification.
- It must fail CLOSED: an unresolvable slug keeps the binding requirement, as the identity test already does.

Dup-check: searched open issues for `verify-pr-gate repo resolution remote slug`, and ran a body-window scan for `gh-resolved` / `repo override` / `nameWithOwner` / `remote slug` / `GH_REPO`. The only hit is go-to-k/cdkd#2354 (a quoted command LEADER evading the shared matcher), a different root cause — none covers gh's own repo resolution.
Session-fit: next (not this session) -- reason (a), external input: the comparison has to agree with what gh ACTUALLY resolves across the remote-precedence rules and `gh-resolved`, and the only honest way to pin that is a differential fence against gh itself over a matrix of remote configurations. That is a fixture this session has not built and cannot borrow; go-to-k/cdkd#3214 closes the command-text half now and states this as a bound. Next session starts with: bash .claude/hooks/verify-pr-gate.test.sh
Severity: medium -- a sibling checkout carrying an `upstream` remote can clear the merge of a pull request in a different repository on a marker that attests to something else. It needs an `upstream` remote pointing at the other repo, so it is not reachable by an arbitrary command, and every other gate on that merge still runs
Effort: medium (M) -- one resolution helper plus a differential fence over a remote-configuration matrix, in a hook whose suite is already 90 cases
Estimate: ~2 h -- most of it the matrix (origin only, origin+upstream, `gh-resolved` = base, `gh-resolved` = a slug, no remotes) and re-running the whole hook harness under both bashes

go-to-k:
## What was wrong

`verify-pr-gate.sh`'s foreign-target relaxation (go-to-k/cdkd#3209) decided "is it safe to skip cdkd's `.markgate-verify-pr-sha` binding here" by reading the COMMAND. gh does not only read the command — it resolves a base repo from the TARGET CHECKOUT's git remotes, and that channel leaves nothing in the text for the guard to see.

Measured on gh 2.92.0: a checkout whose `origin` is `go-to-k/cdk-local` and which also carries `upstream = go-to-k/cdkd` answers **go-to-k/cdkd** for a plain `gh pr merge 42 --squash`. So cdk-local's own fresh `verify-pr` marker cleared the merge of a CDKD pull request — the same 2 -> 0 shape go-to-k/cdkd#3214's review round found for `-R`, arriving through a channel no reading of the command closes. `git remote add upstream <other repo>` is an ordinary agent operation and the standard fork setup; `gh repo set-default` does it deliberately.

## The fix

Resolve what gh would resolve in the target, and require it to equal that checkout's own `origin`.

- **`git config` reads only.** No `gh` invocation (a network round trip inside a PreToolUse hook) and nothing executed FROM the target — `.claude/rules/hooks.md` records why delegation was abandoned in PR 1970.
- **A COMPARISON, not a presence test.** Refusing any checkout carrying a `gh-resolved` key would refuse a sibling that ran `gh repo set-default` on ITSELF, which is exactly what gh tells a multi-remote checkout to do, and that re-breaks the flow go-to-k/cdkd#3209 exists to enable.
- **Fails closed**, matching the identity test beside it: remotes that exist but cannot be read as GitHub, and a checkout with no unambiguous `origin`, both keep the binding. A checkout with NO remotes AT ALL is the one wave-through, and it is decided by asking git whether any remote exists rather than by our parser returning nothing — gh answers `no git remotes found` there too, so the command cannot act on any repository.
- **The refusal carries the matching remedy.** A remote-config retraction prints the remote fix (drop the stray remote, or `gh repo set-default <own slug>`); the command-shape advice would clear nothing. `__vpg_retract_kind` is what keeps the two apart.

### It also retires the second-clone KNOWN BOUND

A second CLONE of cdkd has a different git common dir, so the identity test answered FOREIGN and the relaxed path dropped the go-to-k/cdkd#2686 binding in a checkout that IS cdkd. The common dir was only ever a proxy for repo identity; with the slug available the gate asks the real question. This only ever makes the gate STRICTER (foreign -> same, which ADDS the binding requirement), so it cannot open a path.

## gh's resolution order, measured rather than read off its source

upstream > github > origin > the rest **alphabetically**, with a `gh-resolved` key on the first remote that has one winning outright. The cells that changed the implementation:

- `zzz` added before `aaa`, no `origin` -> **aaa's** repo. The tail tie-break is alphabetical, not config order — a first implementation took git's config order and this row refuted it.
- `gh-resolved` on BOTH `origin` (`base`) and `upstream` (a slug) -> **upstream's**. The key does not win by being first in the config.
- `gh-resolved` holding a full URL is accepted; holding junk makes **gh itself error**, so refusing agrees with gh rather than guessing.
- `origin` spelled `GO-TO-K/CDK-Local` -> gh answers `go-to-k/cdk-local`. gh normalises case through the API and this hook cannot, so the comparison is case-insensitive.

## Verification

- `verify-pr-gate.test.sh`: **137 cases, 0 failures**, under bash 5.x and bash 3.2.
- A **19-row checked-in matrix** of remote configurations. Column 4 is gh's own measured answer, so the table is re-checkable: `VPG_REMEASURE_GH=1 bash .claude/hooks/verify-pr-gate.test.sh` rebuilds every row and diffs live gh against it — **all 19 agree**. The ordinary offline run is the enforcement, because a check that skips when its oracle is unreachable is a vacuous pass.
- A row-count floor, so a mangled heredoc cannot report green over an empty table.
- **Mutation probes, one mutation per probe, tree restored byte-exact between them** (receipts taken with an explicit occurrence count, after one probe reported green because the substitution never landed):

  | mutation | cases red |
  |---|---|
  | rank sort removed (config order decides) | 6 |
  | rank table flattened | 4 |
  | `gh-resolved` ignored | 4 |
  | slug comparison disabled | 9 |
  | slug case-folding removed | 1 |
  | no-remotes early return deleted | 5 |
  | non-GitHub `origin` no longer refuses | 1 |
  | slug-identity override removed (second clone) | 2 |
  | empty-slug guard dropped | 19 |

  The case-folding arm was UNFENCED when first measured — the upper-case row took both slugs from the same URL, so the folding cancelled out. `case_differs_across_sources` was added for it and is the only case that reds that mutation.

### One suite-level fix this needed

`verify-pr-gate.test.sh` never honoured `HOOK_BASH`, so running it under bash 3.2 exercised the SUITE under 3.2 and left the HOOK on whatever was first on PATH — the trap `.claude/rules/hooks.md` names, and go-to-k/cdkd#2715 lists this suite among those missing the shim. `install_hook` now rewrites the installed copy's shebang, which is the single point covering every invocation. Proven live: `HOOK_BASH=/nonexistent` takes the run from 137/0 to 3/131; `HOOK_BASH=/bin/bash` (3.2.57) is 137/0.

## Residuals, stated rather than closed

- go-to-k/cdkd#3256 (decide the command-text question on the SELECTOR POSITION) is untouched — a different instrument on a different predicate.
- go-to-k/cdkd#3301 is **not** subsumed, despite its own body offering that. It is about `ci-green-gate` / `pr-review-gate` reading a clustered `-R` out of the command to choose which repo to QUERY; this change gives `verify-pr-gate` an identity comparison and hands those two nothing. Said on that issue rather than closing it.
- A `GH_REPO` / `GH_HOST` assembled at run time and a `gh alias` expanding to a `-R` remain open and a comparison cannot see them either: they steer gh at invocation time, from state the hook is not given.
- A genuine FORK checkout (`origin` = `<you>/cdk-local`, `upstream` = `go-to-k/cdk-local`) compares unequal and is refused. That is the guard's declared failure direction — an over-refusal falls back to the binding, never below `origin/main` — and it does not touch the prescribed flow, whose sibling checkouts point `origin` straight at `go-to-k/<repo>`.

Closes go-to-k/cdkd#3235


go-to-k:
Closing this unmerged: **go-to-k/cdkd#3235 was closed on `main` while this PR was in review**, by a better design, and landing this would undo part of it.

PR go-to-k/cdkd#3357 (for issue go-to-k/cdkd#3351) moved the foreign-target decision out of this gate into `gate_target_is_foreign` in `lib/command-match.sh` — deliberately "the only copy" — and its own comment says that function "is now the home of" go-to-k/cdkd#3235's record. It settles identity on the repo SLUG **across every remote** rather than emulating gh's rank order, and inverts an unreadable remote to "not foreign". That sidesteps the entire ordering question this PR spent three rounds getting right, and it is the better instrument.

Measured against `origin/main` at 59b1e1be3, with the markgate verdict mocked `fresh` (0 = relaxed / binding dropped, 2 = refused):

```
#3235 repro  origin=cdk-local + upstream=cdkd   -> 2   closed
control      origin=cdk-local only              -> 0   sibling flow preserved
second clone origin=cdkd                        -> 2   bound retired
no remotes   (the #3209 fixture shape)          -> 0   #3209 preserved
ssh alias    origin=cdk-local + alias upstream  -> 2
insteadOf    origin=cdk-local + gh: upstream    -> 2
```

So every defect this PR fixes is already fixed on `main`, except one — and that one is a LIVE BYPASS I am filing and fixing separately rather than dragging this branch along:

```
git@ssh.github.com:go-to-k/cdkd.git        -> 0   gh resolves go-to-k/cdkd
https://www.github.com/go-to-k/cdkd.git    -> 0   gh resolves go-to-k/cdkd
ssh://git@ssh.github.com/.../cdkd.git      -> 0   gh resolves go-to-k/cdkd
https://github.com/go-to-k/cdkd.git        -> 2   (correct)
```

`gate_slug_from_url` keeps the host verbatim, so a remote naming THIS repo through one of gh's `*.github.com` aliases produces a slug that compares unequal, the checkout reads as foreign, and the go-to-k/cdkd#2686 binding is dropped in a checkout that is cdkd.

What this PR found that is worth keeping is recorded there rather than lost: gh's resolution order and its `gh-resolved` precedence, that `git remote get-url` (not the raw config) is what applies `insteadOf`, that gh shells out to `ssh -G <host>` to expand an ssh alias, and the measured refutation of "being more generous than gh can only over-refuse" — over-acceptance feeds the EQUALITY, so an over-accepted remote outranking gh's real choice relaxes.

No part of this branch is being merged. Thanks to the three review rounds on it; two of the four fail-opens above were found by review, not by me.

PR #3391, opened 2026-09-18
chore(hooks): refuse an UNREADABLE pr selector in the three integ-* gates, not an empty one

### LINKED ISSUE #3365: chore(hooks): three integ-* gates hand-roll the PR-number walk, so a URL selector makes them judge the wrong pull request
Three `integ-*` gates hand-roll the `gh pr merge` PR-number walk instead of calling `gate_pr_selector`, the library helper written for exactly this. A URL selector yields an empty number, and each gate then judges the WRONG pull request or fail-opens.

## The gates

`grep -l 'cmd#\*merge' .claude/hooks/*.sh` returns:

- `.claude/hooks/integ-broad-gate.sh`
- `.claude/hooks/integ-local-gate.sh`
- `.claude/hooks/integ-schema-migration-gate.sh`

`ci-green-gate.sh` and `pr-review-gate.sh` already call `gate_pr_selector` (`.claude/hooks/pr-review-gate.sh:277`).

## What goes wrong

The hand-rolled walk matches `^[0-9]+$` only, so a URL selector never matches:

```
gh pr merge https://github.com/go-to-k/cdkd/pull/1234 --squash
```

`pr_number` comes back empty and the gate falls to `gh pr view --json files` with no argument, which resolves the CURRENT BRANCH's PR. Two outcomes, both wrong:

- the current branch has a PR -> the gate scope-checks a DIFFERENT pull request's files and passes or blocks on its verdict;
- it has none -> `gh pr view` fails and the gate takes its `infra fail-open` arm and exits 0.

Either way `gh` merges the URL'd PR. `gate_pr_selector` handles the selector form and carries its own reasoning about empty-vs-wrong (`.claude/hooks/lib/command-match.sh`, the comment block above the valueless-flag list).

## Why this is filed now

`integ-schema-migration-gate` never fired at all until go-to-k/cdkd#3351 corrected its activation regexes, so the defect was unreachable there. Correcting them makes it live. The other two gates have been reachable all along.

Deliberately NOT fixed in go-to-k/cdkd#3357: changing how a gate TARGETS a pull request is a behaviour change none of that PR's four reviewers examined, and the right fix is one shared helper across all three gates rather than a third hand-rolled variant. Recorded there as a known residual rather than left as an oversight.

## Proposed fix

- Replace the hand-rolled walk in all three gates with `gate_pr_selector "$cmd" "$GATE_RE_GH_PR_MERGE"`.
- Decide, once and in one place, what an EMPTY selector should mean for a BLOCKING gate. `gate_pr_selector`'s own comment records that empty means "resolve the current branch's PR", which is right from the PR's own worktree and wrong anywhere else, and that a sibling repo measured an empty selector reaching a fail-open arm and merging past red CI. For these three the safe answer is probably to REFUSE rather than fall back, but that is a behaviour change deserving its own measurement.
- Add a case per gate driving the URL form, plus a mutation probe that reverting the call reds it.

## Verification plan

- For each gate: a case with `gh pr merge <URL>` over a diff that SHOULD block, asserting exit 2 rather than 0.
- A case where the current branch has an unrelated PR, proving the gate does not judge it.
- Run both bashes via `run-tests.sh`; `gate_pr_selector` uses `[[ =~ ]]` and `GATE_EMBEDDING_TOKEN`, so the 3.2 arm matters.

Dup-check: searched open and closed issues for "gate_pr_selector PR number", "pr number extraction hand-rolled", "URL selector gate". The nearest is go-to-k/cdkd#3273, which is about `ci-green` and `pr-review` judging the PR number against the cwd repo rather than the `-R` slug -- a different question (WHICH REPO, not WHICH NUMBER) about the two gates that already use the helper. No match.

Session-fit: next (not this session) -- external input: the three gates it edits are held by go-to-k/cdkd#3357, an open PR in this session's own lane, and one of them is the file that PR is repairing. Rebasing this onto a moving fix-back round would conflict on every push. Start once go-to-k/cdkd#3357 merges.
Severity: medium -- a merge gate can be bypassed by a spelling a maintainer might plausibly type, but every one of these gates is self-enforced discipline rather than an adversarial boundary, and the prescribed spelling in CLAUDE.md is `gh pr merge <N> --squash --delete-branch`.
Effort: small (S) -- one call site per gate plus cases; the helper already exists and is exercised by two other gates.
Estimate: 1-2 hours -- most of it deciding and measuring what an empty selector should mean for a blocking gate, not the substitution.

go-to-k:
Closes #3365.

## What the issue asked, and what the measurement said

The issue asked the three `integ-*` gates to stop hand-rolling the PR-number
walk and call the shared `gate_pr_selector`, on the theory that a URL selector
was being mis-read. Measured first, per the issue's own open question:

| command | hand-rolled walk | `gate_pr_selector` |
| --- | --- | --- |
| `gh pr merge 3365 --squash` | `3365` | `3365` |
| `gh pr merge -s 3365` | *(empty — flag ate it)* | `3365` |
| `gh pr merge --squash` | *(empty)* | *(empty)* |
| `gh pr merge <URL>` | *(empty)* | *(empty)* |
| `gh pr merge feat/x` | *(empty)* | *(empty)* |
| `gh pr merge -sd 3365` | *(empty)* | *(empty)* |

Identical under bash 5.x and 3.2. So the substitution buys exactly ONE spelling
and closes **nothing** about a URL selector — that is empty through both walks.

The real defect is that "no selector" and "a selector that is present but
unreadable" were conflated. Both yield the empty string, and every gate read
empty as "fall back to the current branch's PR" — which is right for
`gh pr merge --squash` from the PR's own worktree (the spelling CLAUDE.md
prescribes) and wrong for `gh pr merge <URL>`, where gh merges the URL'd PR
while the gate scope-checks a different one, or finds none and takes its infra
fail-open at exit 0.

## The change

`gate_pr_selector_unreadable` (new, in `lib/command-match.sh`) makes the
three-way split, and the three gates consume it:

- selector read → judge that PR (unchanged)
- no selector at all → current-branch fallback (unchanged)
- selector present, unreadable → the scope check is skipped as **UNKNOWN** and
  the marker decides

**The refusal sits AFTER the marker question, not before it**, and that is the
design decision rather than an implementation detail. With the marker FRESH the
gate permits the merge whatever the PR touched, so the PR's identity never
mattered; refusing there would falsely refuse a PR the gate exempts by design
(a docs-only PR merged by URL). The refusal is spent only where the answer is
load-bearing — a stale marker, where falling back could exempt the PR gh is
actually merging. Each suite carries that price as a case
(`fresh marker + unreadable selector does NOT refuse`), which the first cut of
this change reddens.

The refusal message is its own arm and does not borrow the ordinary one: the
ordinary text asserts the PR touches the gate's scope, which is exactly what
could not be checked. Every suite pins that with a NEGATIVE needle.

## Two live defects found in self-review, both green under the suites

1. **`integ-local` failed OPEN on `git merge`.** `__selector_unreadable` was
   initialised inside the `gh pr merge` branch; this gate also fires on
   `git merge`, which never enters it, so under `set -u` the read killed the
   hook — exit 1, which PreToolUse treats as a non-blocking error. Measured
   live: `rc=1`, `unbound variable`. Fixed by hoisting the initialisation.
2. **The message arm landed inside `if [ -n "$reason" ]`** in `integ-local` and
   `integ-schema-migration`, so it was silent for a marker that is merely
   MISSING, and the other branch then asserted the scope it could not check.

Defect 2 MASKED defect 1 — the arm never evaluated the flag on the no-reason
path — which is why the suite read 55/0 with both present. Both suites gained a
`stale_noreason` markgate verdict and a case per defect; each defect was
re-introduced SINGLY and measured red.

## Verification

- `integ-broad` 37/0, `integ-local` 58/0, `integ-schema-migration` 35/0,
  `lib/command-match` 1002/0, `command-match-differential` 27/0,
  `unresolved-target-class` 14/0, `markgate-gate-name-class` 4/0
- Full `run-tests.sh` under BOTH bashes (5.3.9 and system 3.2)
- Mutation probes, each applied singly and the tree restored byte-exact:
  arm disarmed → the three blocking cases go red on the needle while still
  exiting 2 (so an exit-code-only assertion would have passed); refusal moved
  ahead of the marker question → the fresh-marker control goes red; each
  self-review defect re-introduced → its own new case goes red.

No changelog entry: `.claude/hooks/**` changes no shipped-binary behaviour.


go-to-k:
Closing: PR 3407 removed the three integ-* gates this PR patches (integ-broad / integ-local / integ-schema-migration), so the target files no longer exist on main. The finding (issue 3365) is recorded in docs/tooling-backlog.md under the new policy for tooling items.

go-to-k:
Closing as overtaken: go-to-k/cdkd#3407 deleted all three gates this PR fixes
(`integ-broad-gate`, `integ-local-gate`, `integ-schema-migration-gate`), so the
files it edits no longer exist on main.

The work is not lost. The LIBRARY half — `gate_pr_selector_unreadable`, the
structural redirection classifier, unquote-once, and the per-segment probe —
carried four review rounds of findings, and its one surviving caller is
`ci-green-gate`, which had the same conflation. That is go-to-k/cdkd#3417,
opened against current main, and it closes go-to-k/cdkd#3395.

What is dropped: the gate-side diff and its cases, because their subject is
gone.

Worth recording for whoever reads this later — each of the four rounds found its
blocker INSIDE the previous round's fix, and every one was a wrong-PR read:

1. the union verb ERE let `git merge 42 && gh pr merge <URL>` scope-check PR 42
2. a fixed two-character operator strip left `<<<` glued, so its word became the PR
3. enumerating a LENGTH is still enumerating, and a per-character peel was quadratic
   enough (9.07 s on a 20 KB token) to blow these hooks' 10 s timeout, which fails OPEN
4. two fixes of the same helper cancelled — a whole-command "did anything resolve"
   test made the per-segment scan unreachable

The redirection rule in go-to-k/cdkd#3417 is structural for that reason, and both
arity directions are pinned.

Issue #3351, opened 2026-09-17
fix(hooks): integ-schema-migration-gate matches a shape src/types/state.ts has never had, so every schema bump since the gate shipped merged ungated


go-to-k:
`.claude/hooks/integ-schema-migration-gate.sh` blocks `gh pr merge` when a PR bumps the state schema version, so a bump cannot merge without a real-AWS `schema-v<N>-to-v<N+1>-migration` run proving the round trip. It matches neither shape a real bump edits, and it never has.

## Measured

The gate walks the `src/types/state.ts` block of the PR diff and matches `+`/`-` payloads against two regexes (lines 256-264):

```
version:[[:space:]]*[0-9]+([[:space:]]*\|[[:space:]]*[0-9]+)+
STATE_SCHEMA_VERSION[[:space:]]*=[[:space:]]*[0-9]+
```

A v10 -> v11 bump edits three lines. Running the gate's OWN two regexes against each, verbatim:

| line a real bump edits | regex 1 | regex 2 |
| --- | --- | --- |
| `export type StateSchemaVersion = 1 \| 2 \| ... \| 10 \| 11;` | no match | no match |
| `export const STATE_SCHEMA_VERSION_CURRENT: StateSchemaVersion = 11;` | no match | no match |
| `  version: StateSchemaVersion;` | no match | no match |

Regex 1 wants `version:` followed by a digit union; the declaration spells `StateSchemaVersion = `, with `=` not `:`, and `StackState.version` carries the named type and no digits at all. Regex 2 wants `STATE_SCHEMA_VERSION` adjacent to `=`; the constant is `STATE_SCHEMA_VERSION_CURRENT:`, so `_CURRENT:` breaks the adjacency.

Both `plus_match` and `minus_match` stay 0, the gate takes its "state.ts changed but the version constant didn't — non-bump edit" pass-through at line 272, and exits 0.

## Root cause: the regex was written against the DOC, not the file

At the gate's own commit (`23ce6642b`, go-to-k/cdkd#521), `src/types/state.ts` line 73 already read `version: StateSchemaVersion;` and line 32 `export const STATE_SCHEMA_VERSION_CURRENT: StateSchemaVersion = 5;`. The named type and the `_CURRENT` suffix both predate the gate — they landed in go-to-k/cdkd#57.

`git log -S'  version: 1 | 2' -- src/types/state.ts` returns NOTHING: the shape regex 1 matches has never existed in the file the gate greps.

[...]

PR #3399, opened 2026-09-18
fix(hooks): case-fold the repo slug before the trims, not after

### LINKED ISSUE #3387: fix(hooks): gate_repo_slug trims `.git` before it case-folds, so a case-variant remote reads as a FOREIGN repo
## What

`gate_repo_slug` trims the `.git` suffix BEFORE it case-folds, so an upper- or mixed-case remote URL keeps the suffix in its slug and compares UNEQUAL to the same repository spelled in lowercase. `gate_target_is_foreign` then answers FOREIGN for a checkout of this very repo.

`.claude/hooks/lib/command-match.sh`:

```bash
path="${path%.git}"          # <- trims a LOWERCASE suffix only
...
host=$(printf '%s' "$host" | tr 'A-Z' 'a-z')   # <- fold happens here
path=$(printf '%s' "$path" | tr 'A-Z' 'a-z')
```

The fold's own comment says "Case-fold LAST, over the final spelling", and that is exactly the ordering that breaks: everything above it is spelling-sensitive, and the `.git` trim is one of those things.

## Measured

In a checkout whose `origin` is `https://github.com/go-to-k/cdkd.git`:

```text
own slug:  github.com/go-to-k/cdkd
cv  slug:  github.com/go-to-k/cdkd.git     # origin = HTTPS://GITHUB.COM/GO-TO-K/CDKD.GIT
gate_target_is_foreign ... -> rc 0 (FOREIGN)
```

`command-match.test.sh`'s own case for this — `a CASE-variant spelling of THIS repo -> NOT foreign` — FAILS there (`want rc 1, got 0`), and the suite reports `Pass: 1188 Fail: 1`.

## Why CI does not see it

The case's fixture derives its slug from the repo the SUITE sits in (`git config --get remote.origin.url`). GitHub's `actions/checkout` sets `origin` to `https://github.com/<owner>/<repo>` with **no `.git` suffix**, so upper-casing it leaves nothing for the trim to miss and the case passes. Every local clone made with `git clone <url>.git` — the spelling GitHub's own "Code" button offers — fails it.

So this is not a flake and not a worktree oddity: it is decided by whether the remote URL carries the suffix.

## Consequence

`gate_target_is_foreign` is what lets `integ-schema-migration-gate` exit 0 for a target that is provably a different repository (go-to-k/cdkd#3351). A wrong FOREIGN answer therefore turns that gate off for a checkout of cdkd itself, reached by nothing more exotic than a remote URL in a different case. `.claude/rules/gate-sibling-repos.md` states the relaxation is sound because "at `none` there is no marker to ask for" — that argument holds only while `foreign` is answered correctly.

The same helper keys `GATE_MARKER_ALIASES`, so a case-variant remote also misses its alias row and takes the unclearable refusal instead — the opposite polarity, loud rather than silent, but also wrong.

## Fix shape

Fold the host and path FIRST, then apply the spelling-sensitive trims — or make the trims case-insensitive. Folding first is the smaller change and matches what the comment intends; the comment then needs correcting too, since "fold last" is the defect.

Two things a fix must keep, both of which the current order gets right for a different reason: the decode must stay ahead of the trims (go-to-k/cdkd#3351 round 4), and a fold can only make two spellings compare EQUAL, which is the refusing direction.

Fence it with the existing case plus a second whose fixture URL carries the suffix explicitly, so the verdict stops depending on how the runner cloned the repo. That is the load-bearing half: the case exists today and passes vacuously wherever `origin` has no suffix.

Session-fit: now — measured live this session, and `command-match.sh` is the file the go-to-k/cdkd#3082 lane has had open for thirty-two review rounds.
Severity: high — a case-variant remote spelling makes `integ-schema-migration-gate` exit 0 on a cdkd checkout, silently, and the same helper mis-keys the sibling-repo alias table.
Effort: small (S) — a two-line reordering plus a comment correction and one suite case.
Estimate: 45 minutes — the fix is minutes; the cost is re-running `command-match.test.sh` under both bashes and checking no other consumer depended on the pre-fold spelling.
Dup-check: searched open issues for `gate_repo_slug`, `foreign`, `slug`, `case-fold`; go-to-k/cdkd#3351 is the change that introduced the helper and its review rounds, and neither it nor any open issue carries this ordering.

go-to-k:
Closes go-to-k/cdkd#3387.

## What

`gate_repo_slug` trimmed the `.git` suffix **before** it case-folded, so an upper- or mixed-case remote URL kept the suffix in its slug and compared UNEQUAL to the same repository in lowercase. `gate_target_is_foreign` then answered FOREIGN for a checkout of cdkd itself — and a FOREIGN answer is what lets `integ-schema-migration-gate` exit 0 (go-to-k/cdkd#3351). A blocking gate, turned off by the case of a remote URL.

Measured in a clone whose `origin` is `https://github.com/go-to-k/cdkd.git`, against a second clone whose `origin` is that URL upper-cased:

| | slug | `gate_target_is_foreign` |
| --- | --- | --- |
| this repo | `github.com/go-to-k/cdkd` | — |
| case variant, before | `github.com/go-to-k/cdkd.git` | rc 0 (FOREIGN) |
| case variant, after | `github.com/go-to-k/cdkd` | rc 1 (not foreign) |

## The fix

The fold moves above the spelling-sensitive trims. Its own comment prescribed the wrong order — "Case-fold LAST, over the final spelling" — while the *argument* for folding is unchanged and still right: folding can only make two spellings compare EQUAL, and equal means "same repo, refuse to relax", the safe direction. The decode stays above both, for the reason go-to-k/cdkd#3351 round 4 recorded: it has to see the encoded spelling.

## Why CI never saw it

The suite has carried a case for this since go-to-k/cdkd#3351, and it passed. Its fixture derives the URL from the repo the **suite** sits in, and `actions/checkout` writes an `origin` with no `.git` suffix — so in CI there was nothing for the trim to miss. Every clone made the way GitHub's own "Code" button offers failed it, silently, on the developer machines where the gate actually runs.

That is the transferable half: **a case whose subject is a URL spelling has to pin that spelling**, not inherit whatever the runner cloned with.

## Fences

- A new case pins `.git` explicitly.
- A lowercase control ships with it and passes under **both** orderings. The pair is the point — the suffix-less spelling cannot discriminate, which is exactly how the original case went vacuous.
- The block's cardinality guard goes 36 → 38, so neither can be dropped silently.

Probed: restoring the old order reds the new case **and** the pre-existing one, while the control stays green.

## Verification

- `command-match.test.sh` **1191/0** on a `.git`-suffixed clone, where the same suite was 1188/1 before this change — so this PR also turns the local suite green for anyone who cloned that way.
- `bash .claude/hooks/run-tests.sh` green under bash 5.x and 3.2.
- Mutation matrix exits 0 over 102 mutants; `vp run check` and `vp test run` (25,279 passed) green.

## Not changed

`gate_repo_slug`'s other consumers take the same slug, so `GATE_MARKER_ALIASES` keying improves in the same direction: a case-variant remote used to miss its alias row and take the unclearable refusal. No alias row's spelling changes.


go-to-k:
Closing without merging: go-to-k/cdkd#3386 landed the same fix while this was being written, and the half of this PR that was left is something that change explicitly decided against.

**The defect is gone.** go-to-k/cdkd#3386 strips the suffix case-insensitively (`*.[Gg][Ii][Tt]`) instead of reordering the fold — a different remedy, the same effect. Re-measured on `origin/main` just now, in a clone whose `origin` carries the suffix:

```text
cv slug on main: github.com/go-to-k/cdkd     (was github.com/go-to-k/cdkd.git)
gate_target_is_foreign -> rc 1 (not foreign)
```

So the reordering here buys nothing, and shipping it would churn a function two lanes have just rewritten.

**The test half is declined by that change, in writing.** This PR proposed pinning a `.git`-suffixed URL in the fixture, because the existing case derives its URL from the repo the suite sits in and is therefore green in CI for a reason unrelated to what it asserts. go-to-k/cdkd#3386's own comment answers that directly:

> Do not "fix" that case by pinning a URL; the environment difference is what it detects.

That is a deliberate call recorded at the site, and a lane arriving afterwards should not re-litigate it. I still think a case that is green in CI cannot gate a regression there — but the counter-argument is real (the local/CI split is itself a signal), it is written down, and one more opinion is not a reason to reopen it.

Nothing is lost by closing: go-to-k/cdkd#3387 is closed as fixed by go-to-k/cdkd#3386 with the measurement attached, and this branch is deleted.

PR #3418, opened 2026-09-18
chore(hooks): a redirection may lead a simple command, in the strip AND in the segmenter

### LINKED ISSUE #3204: hooks: a leading redirection defeats every blocking gate (gate_strip_prefix strips no redirection)
## Summary

`gate_strip_prefix` in `.claude/hooks/lib/command-match.sh` strips assignments, `case` arm labels, groups, keywords and the `env` / `command` / `nohup` / `time` / `timeout` / `exec` wrappers before the verb is matched in command position — but **not a redirection leader**. bash allows redirections anywhere in a simple command, including before the command word, so one token defeats every blocking gate.

Measured on this head and byte-identically on `origin/main` (so pre-existing, found during round 20 of go-to-k/cdkd#3082):

```
$ source .claude/hooks/lib/command-match.sh
git commit -m x                          MATCH
>/dev/null git commit -m x               NOMATCH
2>/dev/null git commit -m x              NOMATCH
</dev/null git commit -m x               NOMATCH
x=1 >/dev/null git commit -m x           NOMATCH
```

All four of the refused-by-nothing spellings RUN the commit — verified with a stub `git` on PATH under bash 5.3.9, bash 3.2.57 and zsh 5.9 (each logged `commit -m x`). The same holds for every verb the matcher guards, so `branch-gate`, `check-gate`, `verify-pr-gate`, the four `integ-*` gates, `pr-review-gate`, `ci-green-gate` and `dirty-path-restore-gate` all go silent on a command the shell executes.

This is the class `.claude/rules/hooks-class-fences.md` calls out: "over-approximate the TRIGGER and stay strict on RESOLUTION". A leading redirection is a spelling nobody enumerated.

## Fix shape

In `gate_strip_prefix`, strip a leading redirection token the way the assignment prefix is stripped: an optional digit, one of `<` `>` `>>` `<<<` `<&` `>&` `&>` `&>>`, then its target word (quoted or not), repeated. Keep it in the STRIP list rather than the verb regex, so every gate gains it at once. `<<` (a heredoc opener) must not be consumed as a target — the heredoc machinery owns that.

Cases: one per spelling above plus a control that a redirection AFTER the verb still matches (`git commit -m x >/dev/null`, already MATCH today) and one that a redirection in an ARGUMENT is not stripped (`echo ">/dev/null git commit"`). Mutants beside the existing prefix ones in `command-match-mutants.sh`.

## Verification

```
bash -c 'source .claude/hooks/lib/command-match.sh; gate_matches ">/dev/null git commit -m x" "$GATE_RE_GIT_COMMIT" && echo MATCH || echo no'
```

prints `no` today and must print `MATCH`.

## The four TODO fields

Session-fit: now (this session) -- the fix lands in `.claude/hooks/lib/command-match.sh`, the file this session has had open for twenty review rounds, and the survey and differential fence it needs are the two harnesses this session has been running all along. Worked as a follow-up PR after go-to-k/cdkd#3082 merges.
Severity: high -- one token in front of any guarded verb silences every blocking PreToolUse gate, in all three shells, on `origin/main` today.
Effort: medium (M) -- the strip itself is small, but the differential fence prices every changed cell and the false-refusal survey has to be re-run; both are the cost.
Estimate: 3 hours -- the survey and the differential cells, not the parse.
Dup-check: searched `is:issue redirection leader`, `gate_strip_prefix`, `>/dev/null git commit` -- nothing open or closed; go-to-k/cdkd#2333 (quoted structural tokens) and go-to-k/cdkd#2650 (`cd\ /tmp`) are the neighbours in this class, both closed and neither covering a redirection.

go-to-k:
Closes go-to-k/cdkd#3204.

## What

bash allows a redirection anywhere in a simple command, the command word included, so **one token in front of a guarded verb silenced every blocking gate at once**. Measured with a stub `git` on PATH under bash 5.3.9, bash 3.2.57 and zsh 5.9 — each RUNS the commit — against `gate_matches` on `origin/main`:

```text
git commit -m x                    MATCH
>/dev/null git commit -m x         NOMATCH
2>/dev/null git commit -m x        NOMATCH
</dev/null git commit -m x         NOMATCH
x=1 >/dev/null git commit -m x     NOMATCH
```

`branch-gate`, `check-gate`, `verify-pr-gate`, the four `integ-*` gates, `pr-review-gate`, `ci-green-gate` and `dirty-path-restore-gate` all go silent on a command the shell executes.

## The fix is TWO layers, and the issue's own shape is only the first

**1. `gate_strip_prefix` gains a leading-redirection strip.** In the STRIP list rather than a verb regex, so every gate gains it from one place — the shape `.claude/rules/hooks-class-fences.md` prescribes ("over-approximate the TRIGGER, stay strict on RESOLUTION").

`<<` is deliberately **not** an operator there. A heredoc opener belongs to the heredoc machinery, and consuming its delimiter as a redirection TARGET hands the body back as commands — the fail-open go-to-k/cdkd#3040 spent thirty review rounds closing. `<<<` is a real here-string and IS stripped. Separating them needs an explicit guard, because POSIX ERE alternation is leftmost-**longest** rather than ordered, so a bare `<` alternative otherwise matches the first character of `<<X`.

**2. An `&` or `|` that belongs to a redirection operator is no longer a segment separator.** Layer 1 alone left two spellings broken, and only measuring found them: the segmenter split `2>&1 git commit -m x` into `2>` and `1 git commit -m x`, and `>|out git commit -m x` into `>` and `out git commit -m x`, so the verb lost its command position **even though the strip was correct**. The decision is made from the character beside the operator, so a bare `a & b` and `a | b` still split — pinned by two controls.

The two layers fail independently, which is why each has its own cases and its own mutant.

## Fences

- **Fifteen cases** in `command-match.test.sh`: the four spellings from the issue, a here-string, a quoted target carrying a space, an assignment and a redirection nested (the strip loop runs to a fixpoint), the `>&` / `>|` / `&>` operator spellings, and five controls — a bare `&`, a bare `|`, a redirection AFTER the verb, one inside an argument, and one with no verb after it.
- **Mutants** `redir-leader` and `redir-operator-split`, measured separately: the first reds 10 named cases, the second 2.
- **The differential fence gains eight inputs** and a `REDIR_LEADER` class of twelve declared cells, floored at that count. The three control inputs declare **no** cell, which is what a control has to do in that fence.

## Verification

- `command-match.test.sh` 1233/0; differential 30/0 over 273 inputs.
- `bash .claude/hooks/run-tests.sh` green under bash 5.x and 3.2; the mutation matrix exits 0.
- `false-refusal-survey.sh` — the measurement this issue's own Estimate named as the cost, since widening a trigger trades a silent bypass class for a loud false-refusal one. Figures in the comments below.
- `vp run check`, `vp run build`, `vp test run` green.


go-to-k:
Closing without merging, under the policy recorded on `main` in [docs/tooling-backlog.md](https://github.com/go-to-k/cdkd/blob/main/docs/tooling-backlog.md):

> **Hook fail-open on exotic shell shapes is ACCEPTED.** ... they are not a security boundary ... Finding one more such shape is therefore **not issue-worthy** ... **The hook layer is NOT being moved off bash. ... The layer is being SHRUNK instead.**

go-to-k/cdkd#3204 was closed as `not planned` under that decision while this branch was in verification. This PR is squarely the class it names, and it moves the layer in the opposite direction: 15 new cases, 2 new mutants (roster 102 -> 104 in a matrix already sharded four ways because it outgrew a CI job), 8 new differential inputs, and a third field added to `gate_segments_marked`'s output contract that every consumer would then have to handle.

Recording what was measured, so the work is not re-done from scratch if the policy is ever revisited:

- The defect is real in all four spellings the issue lists, plus `2>&1` / `>|` / `&>` / `<&0` / two-digit and `{fd}` file descriptors. Both bashes and zsh run the verb; `gate_matches` answers NOMATCH.
- The fix needed **three** layers, not one: the strip in `gate_strip_prefix`; a segmenter arm so an operator's own `&`/`|` is not a separator (with a glue record, without which an escaped `>` regressed four shapes); and a raw-segment third field, because layer 1 broke `main-tree-edit-gate` -- it reads the segment for write TARGETS, and the target was being stripped away with the operator (19 shapes went rc 2 -> 0 with the tracked file really written).
- Verification was green: `command-match.test.sh` 1245/0, differential 30/0 over 544 generated inputs, `run-tests.sh` green under bash 5.3.9 and 3.2, `command-match-mutants.test.sh` 21/0, `vp run check` / `typecheck:test` / `build` / `test run` (1079 files, 25383 tests) all green.
- **Layer 3 is the transferable finding**: it is not about redirections. Any future change that teaches `gate_strip_prefix` to remove more of a segment silently blinds every consumer that scans the segment for something other than the verb. That is worth knowing whether or not this PR lands.

The two residues filed from this lane, go-to-k/cdkd#3421 and go-to-k/cdkd#3422, are the same class and are closed alongside it.

Issue #3428, opened 2026-09-18
test(scripts): the wc-trim classifier's two declared un-backstopped bounds have no pinning case


go-to-k:
## Summary

One residual item from the go-to-k/cdkd#3304 review, in `tests/unit/scripts/integ-verify-wc-trim.test.ts`. (A second item was filed here and turned out to be already fixed in the merged tree; it is struck below with the measurement that confirms the remedy that landed was the right one.) Neither changes a verdict today; both are the "a comment asserts something the code does not do" class that PR corrected three times during its review, so leaving them unrecorded is what this issue prevents.

## 1. ~~`leadStdout`'s doc claims a clearing the code does not do~~ — ALREADY FIXED, struck 2026-09-19

This was filed against a stale read. The merged tree already carries the narrowed wording: `leadStdout` is documented as "Cleared at `;`, `&`, `|` and a newline, the separators that can come before a later command word (`>f (` and `>f )` are not valid bash, so `(` and `)` need not)." go-to-k/cdkd#3304's final commit folded it in.

Recording the measurement, because it settles which of the two remedies was right rather than merely noting the wording changed. Making `(` and `)` clear `leadStdout` DOES change results — 48 of 600 generated cases — but `bash -n` rejects all 48, and the direction is the wrong one: the current code reports a violation on those inputs (fail-closed, loud) while the clearing accepts them. So narrowing the doc, which is what landed, is the correct remedy and the clearing should NOT be applied later.

## 2. The two un-backstopped bounds are declared but not pinned

The classifier's header names exactly two bounds that can pass an untrimmed `wc` with nothing else noticing — a `$'...'` escape spelling the command (`$'\x77c'`), and a quoted `<<-` delimiter that itself starts with a tab (`cat <<-"<TAB>EOF"`) — and `docs/integ-fixture-conventions.md` lists the same two. Both reproduce today, and both were verified during review to be fail-open AND unreported by the tree invariant.

Nothing pins either. If a later change closes one, or opens a third, the header and the docs keep claiming the old set with no test disagreeing. The rest of that file's claims are pinned by cases; these two — the strongest claims it makes — are not.

Remedy: a case per bound asserting the current behaviour (zero invocations, zero violations, and the tree invariant silent), named so it reads as pinning a DECLARED bound rather than endorsing it. Then a change that closes one reds the case, and whoever closes it updates the header and the docs in the same commit.

## Verification plan

[...]

PR #3304, opened 2026-09-17
test(integ): fence that every verify.sh wc count is trimmed, so a fixture written on GNU coreutils cannot fail unconditionally on macOS

nix-tkobayashi:
Closes #3213

## What was wrong

Nothing checked that a `wc` result in a `tests/integration` shell file is trimmed. BSD `wc` — the one on a stock macOS host — right-aligns its count in a field eight characters wide, GNU `wc` does not, and `$(...)` strips only the trailing newline. So

```bash
N="$(printf 'a\n' | wc -l)"
[ "${N}" = "1" ]
```

is true on a GNU host and false on a Mac. A fixture written on GNU coreutils passes review, CI and its own real-AWS run, then fails every time on macOS. go-to-k/cdkd#3182 was the second time the class landed, and review caught it only because a reviewer ran the idiom verbatim on a Mac.

## What changed

A classifier, `scripts/check-integ-wc-trim.ts`, enforced by `tests/unit/scripts/integ-verify-wc-trim.test.ts` — the unit test IS the CI enforcement, the same shape as its sibling `check-integ-capture-shape.ts`, so there is no `vp run` task and no `ci.yml` step. The writeup is a new section in `docs/integ-fixture-conventions.md`, with a table of what is and is not counted.

**The rule: every `wc` invocation must pipe straight into `| tr -d ' '` or `| tr -d '[:space:]'` as its very next stage, and the `tr` stage must end right after that argument** (`tr -d ' ' -c` keeps the spaces and deletes the digits); an output redirection written after a blank (`> count.txt`, `2>/dev/null`) is not an argument and is allowed, while a descriptor joined to the argument (`tr -d ' '2>x`, which deletes ` 2`) and any redirection that replaces `tr`'s stdin are not. A trim that is not the next stage, a `||` fallback, a tab-only trim, extra arguments after the trim, a trim sitting in a trailing comment, and a redirection that replaces the trim's own stdin (`<file`, `<<EOF`, `<<<x`, `<&3`, `<&-`, so the count never reaches `tr`) are all refused. `# allow-untrimmed-wc: <reason>` (colon required, reason of at least 10 characters) is the escape hatch — on the `wc` line itself or as a full-line comment directly above; a comment naming the marker in any other shape is reported as malformed, and the test pins how many are in use (zero).

[...]

Issue #3213, opened 2026-09-16
test(integ): nothing lints that a verify.sh wc result is trimmed, so a fixture written on GNU coreutils fails unconditionally on macOS


nix-tkobayashi:
## Summary

Nothing checks that a `wc` result in a `verify.sh` is trimmed before it is string-compared, so a fixture written on a GNU-coreutils host passes review, passes CI and passes its own real-AWS run, then fails unconditionally on a stock macOS host. BSD `wc` pads its count to width 8 and `$(...)` strips only the trailing newline, so the captured value is `"       0"` and every `=` comparison against it is false.

This is the second time the class has landed. It was raised as a take-or-leave note on the go-to-k/cdkd#3182 review ("Nothing lints the `| wc` trim, unlike the sibling capture-shape rule. Not this PR's job -- noting it because M0 is the second time that class has bitten"), and filed here rather than fixed there.

## Why a lint rather than review

Three sibling defects in `verify.sh` are already mechanically enforced. They do NOT share a single root cause -- naming what each actually catches, since the shared property is narrower than it first looks:

- `tests/unit/scripts/integ-verify-probe-not-found.test.ts` -- a blind probe reading any AWS failure as "gone"
- `tests/unit/scripts/integ-verify-capture-shape.test.ts` -- a capture that loses the diagnostic under `pipefail`
- `tests/unit/scripts/integ-verify-version-literals.test.ts` -- an assumption about Lambda version counters holding across runs

What they share is only the enforcement argument, and it is the one that applies here: the author cannot observe the defect on the run they just did. For this one the hidden variable is the HOST's `wc`, which is why review on go-to-k/cdkd#3182 caught it only because a reviewer executed the idiom verbatim on macOS, and why the fixture's own green real-AWS run said nothing.

## Proposed fix

A classifier under `scripts/`, fenced by a unit test, flagging a `wc` result that reaches a string comparison without a trim. The forms present in the tree today, which a shape list has to cover:

- the piped capture, `V="$(... | wc -w)"` / `V=$(... | wc -l)` -- the go-to-k/cdkd#3182 shape
- an INLINE here-string comparison rather than a capture: `tests/integration/local-ecs-service-connect/verify.sh:247` and `:251` are `[[ "$(wc -l <<<"${X}" | tr -d ' ')" -ne 1 ]]`, i.e. the value never lands in a variable
- `<file` redirection forms, several of which appear inside echoed text rather than in a comparison

[...]

Issue #3385, opened 2026-09-18
chore(hooks): gate_slug_from_url keeps the host, so a remote naming this repo through one of gh's github.com aliases reads as another repo and drops the verify-pr binding


go-to-k:
## The defect

`gate_slug_from_url` (`.claude/hooks/lib/command-match.sh`) keeps the URL's HOST verbatim in the slug — deliberately, per its own note at `gate_repo_slug` ("The slug carries the HOST on purpose"). gh does not: it normalises its own host aliases, so `ssh.github.com` and `www.github.com` resolve the SAME repository as `github.com`.

So a checkout that names THIS repository through one of those aliases produces a slug that compares unequal to the hook's own, `gate_target_is_foreign` answers FOREIGN, and `verify-pr-gate` drops the go-to-k/cdkd#2686 sha binding in a checkout that IS cdkd.

Measured 2026-09-18 against `origin/main` at 59b1e1be3, gh 2.92.0 / git 2.49.0, running the real `verify-pr-gate.sh` with the markgate verdict mocked `fresh`. Each fixture has `origin = github.com/go-to-k/cdk-local` plus an `upstream` naming cdkd through the spelling shown. **0 = relaxed (binding dropped), 2 = refused.**

```
upstream spelling                            gate   gh resolves
git@ssh.github.com:go-to-k/cdkd.git           0     go-to-k/cdkd
https://www.github.com/go-to-k/cdkd.git       0     go-to-k/cdkd
ssh://git@ssh.github.com/go-to-k/cdkd.git     0     go-to-k/cdkd
https://github.com/go-to-k/cdkd.git           2     go-to-k/cdkd   (correct, the control)
```

The control is what makes the other three a defect rather than a preference: the same repository, named four ways gh treats identically, gets two different verdicts.

`ssh.github.com` is not exotic — it is GitHub's documented SSH-over-443 host, which is the standard workaround on a network that blocks port 22, so it appears in ordinary `~/.gitconfig` / remote setups.

## Why this is not go-to-k/cdkd#3235

go-to-k/cdkd#3235 was "the gate reads the COMMAND but gh also resolves from the checkout's REMOTES", and it is fixed: go-to-k/cdkd#3357 moved the decision into `gate_target_is_foreign` and settles identity on the slug across every remote. This is a defect in the SLUG NORMALISER that fix depends on, and it survives it. Verified on the same tree in the same run: the go-to-k/cdkd#3235 repro refuses while these three relax.

[...]

PR #3429, opened 2026-09-18
docs(tooling-backlog): record the wc-fence bounds finding and the two near-timeout unit cases, and mark #3213 shipped

go-to-k:
Three rows in `docs/tooling-backlog.md`, no code.

## Why this is rows and not a fix

The wc-trim fence merged in #3304 declares two bounds that can ship an untrimmed `wc` with nothing else noticing — a `$'...'` escape spelling the command, and a quoted `<<-` delimiter that itself starts with a tab. Both reproduce, both are named in the classifier header and in `docs/integ-fixture-conventions.md`, and neither has a case pinning it, so closing one or opening a third would leave both documents asserting a stale set with no test disagreeing.

I had written the two pinning cases and then reverted them. CLAUDE.md: a test-of-prose is added only on the SECOND occurrence of the same failure, and the first is recorded here with nothing built. This failure has occurred once. The same rule is why the finding is a row rather than an issue — #3428 was filed under the previous rule and is closed with a pointer here.

## The rows

- **#3213** — marked SHIPPED by #3304, so nobody picks it up twice.
- **The two un-pinned bounds** — the finding above.
- **Two near-timeout unit cases** — `withretry-interrupt.test.ts` and `gen-handled-property-wiring.test.ts` each hold one case running at ~4 s against vitest's 5 s default under the full suite, and both timed out on runs that overlapped a build. Neither file is touched by #3304; observed while verifying it.

## Convention added

The Issue column now reads `—` for a finding recorded before anyone worked it — the table had no shape for one — and a shipped row says which PR took it.

## Test plan

- `vp run check` — 0 errors.
- `vp test --run tests/unit/scripts/docs-error-strings.test.ts tests/unit/scripts/rule-file-payload.test.ts` — 486 passed; these are the two fences that read `docs/**`.


go-to-k:
All three review findings fixed in `4d5ffd1c`, each re-measured here rather than taken from the report.

**Blocker — the timing row was wrong by ~8x for one of its two files.** Measured isolated on this tree: `gen-handled-property-wiring.test.ts` 4788 ms, `withretry-interrupt.test.ts` 510 ms (its top three are 510 / 468 / 446). The row now names only the case that is actually near the ceiling — `every malformed FIELD gets a STRUCTURED refusal, not a caught crash` — and drops the other file, whose residual is the "spawning suite with no declared bound" class rather than this one.

I had taken the `~4 s` figure from a comment on #3304 instead of measuring it. That is the mistake worth recording: a row exists so the number survives, so a row carrying someone else's unverified number is worse than no row.

**Minor — three shipped rows contradicted the sentence this PR added.** Confirmed `#3040`, `#3066` and `#3385` are closed `COMPLETED` while the other 94 are `NOT_PLANNED`, and traced their merges: `33bff0e3a (#3082)` for the first two, `a6505af43 (#3386)` for the third. All three now carry the suffix.

**Nit — no pointer to `#3384`.** Correct, and it is a widening rather than a duplicate: that row's population is six spawning suites and this file is not among them. The row says so.

`vp run check` 0 errors; the two fences that read `docs/**` pass 486.

Issue #3277, opened 2026-09-16
fix(synthesis): AssemblyReader's absolute-asset-path refusal prints template-controlled values unsanitized


go-to-k:
## What happens

`AssemblyReader`'s nested-template absolute-path refusal interpolates three attacker-controlled values into its message without sanitizing them, so a hand-modified Cloud Assembly can forge terminal output inside the error cdkd prints when it rejects that assembly.

`src/synthesis/assembly-reader.ts:310-316`:

```ts
throw new SynthesisError(
  `Stack '${stackName}' nested-stack '${logicalId}' has ` +
    `Metadata['aws:asset:path']='${assetPath}' which is absolute. ` +
    `CDK emits relative asset paths for nested templates; an absolute ` +
    `path indicates the synth output was hand-modified or generated by a ` +
    `non-CDK toolchain. Refusing to load.`
);
```

All three come from the synth output: `logicalId` is a template key, `assetPath` is raw `Metadata['aws:asset:path']`, and `stackName` comes from the manifest. `displaySafe` is not imported in that file.

The message's own text states the threat model — "the synth output was hand-modified or generated by a non-CDK toolchain" — so by construction it fires only on input someone else wrote.

## Why the message is not sanitized downstream

`formatError` (`src/utils/error-handler.ts`) sanitizes only `error.cause`; a thrown error's own `message` is rendered as-is. C1 bytes (`U+0080`-`U+009F`) and the bidi overrides (`U+202A`-`U+202E`) pass straight through ordinary string building, which is what lets a crafted value append what looks like a second, cdkd-authored line to the refusal.

## Scope — the same class, two of which are already closed

go-to-k/cdkd#3243 closed the two sibling refusals in `src/cli/commands/diff-recursive.ts`: the cycle refusal (three interpolations) and `indexNestedChildTemplates`' absolute-path refusal (two), which is a near-duplicate of the message quoted above. That PR deliberately stopped at its own file rather than widening into the synthesis layer.

This is the third copy and the one the CLI reaches FIRST: `AssemblyReader` runs during synth, before any diff walk, so on a hostile assembly this is the message a user sees.

Worth checking while there whether other `SynthesisError` throws in that file interpolate template- or manifest-derived values the same way — the fix is per-call-site, so a sweep of the file is cheaper than a second issue.

## Proposed fix

[...]

PR #3448, opened 2026-09-19
fix(providers): refuse a cyclic nested-template tree before any level of it deploys

### LINKED ISSUE #3247: fix(deploy): NestedStackProvider deploys every level of a nested-template cycle, creating real resources until the S3 key limit stops it
## What happens

`NestedStackProvider` walks the nested-template chain on `create` and `update` with no visited set and no self-reference check, so a relative `Metadata['aws:asset:path']` that resolves back onto the same nesting chain is followed — **creating real AWS resources and a child state record at every level** until S3's key limit stops it.

`grep -c 'visited\|ancestor\|seenTemplates\|maxDepth' src/provisioning/providers/nested-stack-provider.ts` returns **0**.

The walk:

- `create` (`src/provisioning/providers/nested-stack-provider.ts:193-196`) and `update` (`:254-257`) both derive `childStackName` via `deriveChildStackName` (`:733`) and index the next level with `indexGrandchildTemplates` (`:896`);
- each level runs a whole child `DeployEngine` (`:557`), which provisions that level's resources and writes its state record before descending.

`indexGrandchildTemplates` mirrors `indexNestedChildTemplates` in the diff path: it refuses an ABSOLUTE `aws:asset:path` and joins a RELATIVE one unconditionally. A cycle takes the second branch.

## Why this is worse than the diff-side twin

go-to-k/cdkd#3239 is the same defect in `cdkd diff --recursive`, and it is now fixed in go-to-k/cdkd#3243. That one is read-only: it burns ~190 pointless state reads and ends with a confusing `Your key is too long`, and nothing is created.

Here the identical input **deploys**. Each level of the cycle is a real stack: resources created, a state record written at `cdkd/<parent>~<Loop>~<Loop>.../<region>/state.json`, assets published. The run ends on the same key-length error, but by then it has provisioned an unbounded-until-the-key-limit number of duplicate resource sets, and the rollback path has to unwind a tree the user never declared.

`deriveChildStackName` appends one `~<logicalId>` per level, so the bound is the same ~190 levels — roughly 190 nested deploys of the same template.

## Reachability

The same threat model `indexGrandchildTemplates`'s own absolute-path refusal already names: "the synth output was hand-modified or generated by a non-CDK toolchain". CDK emits an acyclic nested-template tree, so a well-formed `cdk.out` never reaches this.

Surfaced by the code review of go-to-k/cdkd#3243, which fixed the read-only half and explicitly did not widen into the provisioning layer.

## Proposed fix

Port the go-to-k/cdkd#3243 shape: thread the resolved absolute template paths of the ancestor chain through `create` / `update`, and REFUSE the first repeat before any child engine runs, naming the logical id, the owning stack and the path that closed the cycle. Two properties carry over and are the reason to copy rather than re-derive:

- the set must be the ROOT-TO-NODE chain, not a global visited set, or two sibling rows legitimately sharing one child template are refused as a cycle;
- a refusal, not a depth cap — a cyclic assembly has no correct deploy, and a truncated one leaves a half-built tree.

**Refuse in the PRE-FLIGHT, not mid-walk, if that is reachable.** Unlike the diff path, stopping at level N here means levels 0..N-1 are already deployed. If the chain can be validated from the assembly before the first child engine runs, that is strictly better than throwing partway.

While there: both refusals in that file interpolate template-controlled text (`logicalId`, the asset path) into bare `Error` messages unsanitized — the same class go-to-k/cdkd#3243 closed for its own new throw with `displaySafe`. The two pre-existing ones in `src/cli/commands/diff-recursive.ts` (`:251`, `:1247`) are in the same bucket.

## Verification plan

- A unit case whose child `aws:asset:path` resolves back to the parent, asserting the refusal and that NO child engine was constructed.
- A longer cycle (A -> B -> A), which a self-reference check alone would miss.
- A diamond (two sibling rows naming one template), which must still deploy.
- A legitimate multi-level nested deploy, unchanged.
- If the refusal lands in pre-flight, a case asserting nothing was provisioned; if mid-walk, one asserting the rollback path unwinds what was.

Dup-check: searched open and closed issues for `NestedStackProvider cycle`, `nested stack asset path cycle deploy` and `indexGrandchildTemplates`. The only related hit is go-to-k/cdkd#3239, which is the diff-side twin and explicitly scoped to `cdkd diff --recursive`; go-to-k/cdkd#556 covered unrelated `NestedStackProvider` test gaps and is closed.
Session-fit: next (not this session) — `src/provisioning/providers/nested-stack-provider.ts` was not read or edited in this session beyond confirming the missing guard, and the fix is heavy in a way go-to-k/cdkd#3243 was not: it lands in the `integ-destroy` gate scope (`src/provisioning/providers/**`), so it needs a real-AWS deploy + destroy run, and the pre-flight-vs-mid-walk choice needs the provisioning layer's context this session has not built.
Severity: high — an input class cdkd already refuses elsewhere instead provisions real, billable AWS resources at every level of a cycle, and the failure arrives only after ~190 nested deploys.
Effort: medium (M) — the guard is small, but the pre-flight-vs-mid-walk decision, the rollback interaction, and an integ fixture for a cyclic assembly are not.
Estimate: ~3-4 h plus one real-AWS integ cycle — most of it in deciding where the refusal lands and proving nothing is provisioned before it.

go-to-k:
## Summary

`NestedStackProvider` followed `Metadata['aws:asset:path']` from one nested template to the next with no check that the path stayed off its own nesting chain. Every level of that walk is a whole child deploy, so a hand-modified or non-CDK assembly whose relative path resolved back onto the chain **deployed** the same template again and again -- real resources and a `cdkd/<parent>~<Loop>~<Loop>.../state.json` record per level -- until S3's key-length limit stopped it at roughly 190 levels. This is the provisioning twin of the read-only `cdkd diff --recursive` defect fixed in #3243.

Closes #3247

## Design: refuse up front, not at the first repeat

The issue proposed porting #3243's shape (an ancestor chain threaded through the recursion) and added: refuse in a pre-flight instead if the chain can be validated before the first child engine runs. It can, so that is what this does.

- `create()` and `update()` call `findNestedTemplateTreeDefect` on the row's template **before** `readChildTemplate` / `runChildDeploy`. It statically walks every template reachable from the row and returns the first defect. A chain threaded through the recursion would refuse a cycle that sits N levels down only after levels `0..N-1` had deployed; the static walk sees the whole cycle from the top-most row's call, so no level of a cyclic tree is provisioned and no child state record is written.
- The refused shape is a repeat along **one root-to-node chain**, never a repeat anywhere: two sibling rows sharing one template (a diamond) still deploy. A clean-subtree memo keeps a tree of diamonds linear. It is keyed on the LEXICAL path, because a template's content and its children's paths are functions of that and nothing else; the chain is keyed on identity, which is what guarantees termination.
- The walk and `NestedStackProvider.indexGrandchildTemplates` share ONE row reader, `listNestedTemplateRows`. The walk is only a guard while it follows exactly the rows the deploy follows; two hand-written copies disagreed about an array-valued `Resources` (review finding, see below).
- It is a refusal, not a depth cap, for the reason #3243 gives.
- Template identity is the **real directory plus the file name**. A symlinked directory (`d -> .`) makes every level's joined path a new, longer string for one file, so comparing resolved path strings (what the diff guard does) never sees that repeat. The file's own symlink is deliberately not followed, because a template's children resolve against the directory it was reached in.
- A row's `Condition` is not evaluated, matching the diff walk, so the two commands agree on which assemblies are well-formed.
- The error is `markNonRetryable`, renders every hop (`'Child' (path) -> 'Loop' (path)`) with the middle of a long chain elided, and names the stack that declares the closing row. Every interpolation goes through `displaySafe`.
- A chain of more than 512 DISTINCT templates is refused with a readable message rather than overflowing the recursion. That is not a cycle heuristic: each level adds at least two bytes to the child's state key and S3 caps a key at 1024, so no tree that deep can deploy.

The walker lives in a new leaf-ish module, `src/utils/nested-template-cycle.ts`, so another layer can reuse it without importing the provisioning layer.

### Also in this PR (same class, same file)

- The walk reports an **absolute `aws:asset:path` at any depth** before anything deploys. `indexGrandchildTemplates` already refused one, but only on reaching that level, after the levels above it had deployed. That check stays as the per-level backstop and is now `markNonRetryable`.
- Template-controlled text in this provider's own thrown messages (logical ids, the parent stack name, asset and template paths, the fs / `JSON.parse` message, a child Parameter name, an attribute name), its skip `reason` and its log lines goes through `displaySafe`, as the issue asked.

## What this does NOT do

- **The parent stack's own resources can already be in flight** when its nested-stack row is dispatched; they are unwound by the parent's ordinary rollback. Refusing before the root engine starts needs a call in `src/cli/commands/deploy.ts`, which parallel lanes hold. Filed as #3449; the helper already takes the seeded-ancestor argument that call would use.
- **`src/cli/commands/diff-recursive.ts` is untouched** (held by #3226). Its guard keys on `path.resolve`, so the symlinked-directory spelling still walks to the key limit on the read-only path: filed as #3450. Its "Nested template file not found" throw still renders the logical id and stack name raw: folded into #3277 as a checklist row.
- The refusal is deliberately **not** placed in `AssemblyReader`, although that would cover every command: `docs/cli-diff.md` (also held) documents the diff command's own refusal text, and an assembly-load refusal would pre-empt it and make that page false.
- `nestedStackChildFailureMessage` (the delete-path throw) still renders `childStackName` raw. Its remedy is meant to be pasteable and the module must stay import-free, so it needs its own decision: folded into #3277 as a checklist row.

## Sibling sweep

Walkers that follow `aws:asset:path` from template to template, i.e. whose depth is driven by the TEMPLATE graph: `NestedStackProvider` (this PR) and `diff-recursive.ts` (#3243). 2 of 2 guarded after this PR.

The other recursive nested paths are bounded by a finite real tree and cannot loop on a cyclic assembly: `NestedStackProvider.delete` and `cdkd destroy` / `state destroy` (the cdkd state tree), `cdkd import --migrate-from-cloudformation` (the CloudFormation stack tree from `DescribeStackResources`), `cdkd export` (`walkCdkdStateStackTree`). `cdkd local` and `cdkd scrub` do not descend nested templates by path.

## Tests

- `tests/unit/utils/nested-template-cycle.test.ts` -- self-cycle, A -> B -> A, a cycle below an acyclic lead-in, a diamond, one template at two depths, a cycle hidden behind a memoised sibling, a 2^20-path diamond tree (fails by timeout in ~40 s without the memo), sub-directories, `sub/../` spellings, the symlinked-directory cycle, a symlinked FILE reached in two directories (must NOT be refused), seeded and unseeded ancestors, a deep absolute path, unreadable / unparseable templates, non-stack rows, and the rendering (hops, elision, control characters).
- `tests/unit/provisioning/nested-stack-provider-cycle.test.ts` -- uses a `DeployEngine` double that **recurses** into the provider for every nested row, as the real engine does, so "no child engine was deployed" is a real assertion. Self-cycle, A -> B -> A, a cycle three levels down, `update()`, a deep absolute path: each refuses with zero engine deploys. A diamond and a three-level tree still deploy one engine per level.
- Mutation probes (run, tree restored byte-exact): removing both guard calls turns 6 of 8 provider cases red with `runaway: 13 child engines deployed` (the two legitimate-tree cases stay green, as they should); removing only the `update()` call turns only the `update()` case red; removing the memo turns only the diamond-tree case red.

## Review round (4 reviewers on 581a99cab, fixes in the second `fix(providers)` commit)

- **Security, blocker, fixed**: an ARRAY-valued `Resources` bypassed the guard (the deploy follows rows `'0'`, `'1'`, ...; the walk returned no rows). One shared row reader now; unit cases at both the walker and the provider.
- **Code + security, minor, fixed**: the memo was keyed on identity, which is unsound with a symlinked directory plus a `..` row. Keyed on the lexical path now; the reviewer's layout is a unit case.
- **Security, nit, fixed**: a 20000-template acyclic chain ended in a raw `RangeError`. Bounded at 512 with a reasoned refusal.
- **Spec, minor, fixed**: residuals had no issue. Filed #3449 and #3450, folded two rows into #3277, cited in the changelog entry.
- **Spec, nit, fixed**: the message now names the stack declaring the closing row.
- **Test, fixed**: the fixture's S3 checks cannot discriminate in this fixture (a level's record is written when its first resource completes, and each level's only resource waits on the row below). Step 0 now also asserts no `Deploying nested stack` line was printed, with Step 1 as the positive control for that wording. Also added: the unreachable per-level backstop driven directly, every newly sanitized throw site, a second entry row, a row into a missing directory, the 8/9-hop elision boundary.
- **Code, nit, fixed**: `state destroy` inside the cleanup `while read` loop takes `</dev/null`.
- **Declined**: Windows-style path cases (`C:\x`, `sub\a.json`). `path.isAbsolute` is platform-dependent and the suite runs on POSIX, where those are relative names; the predicate is unchanged from main and keeps its existing cases.
- Mutation probes for the round (run, restored byte-exact): memo back on identity turns only the symlink + `..` case red; re-adding the array skip turns the three array cases red.

## Review round 2 (delta re-check by the same four reviewers on 2d32993cc; fixes in the third commit)

- Code, test and spec: round-1 findings confirmed resolved. Spec found one doc sentence the round had made false ("nesting depth itself is never limited"): rescoped to the cycle rule. Code nit: the owning stack name is now elided like the chain.
- **Security, blocker, fixed**: in asset-redirect mode `readChildTemplate` ran the asset-reference rewrite over every string in the child template, `Metadata['aws:asset:path']` included, BEFORE the grandchild paths were indexed. A path segment spelling a bootstrap bucket name was rewritten, so the deploy followed a different file than the one the walk had validated from the raw template. The index is now taken before the rewrite, and `readChildTemplate` returns both so a caller cannot undo the order. Unit case with a real `buildAssetRedirectMap`; reverting the order turns only that case red.
- **Security, minor, fixed**: with the memo on the lexical path, two self-pointing directory symlinks multiply a finite tree per level without repeating on any chain (measured 2.1x per level). The walk now refuses after following 10,000 rows (`too-large`); unit case with 3^40 rows.
- `cdkd diff --recursive` has the same rewrite-then-index order in `diff-recursive.ts` (held by #3226); its cycle guard is inline so it is not bypassed, but diff can now follow a different file than deploy in that contrived shape. Recorded on #3450.

## Review round 3 (delta re-check on 2ef4f1d0a, fixes in the fourth commit)

- Code and security: clean. Security confirmed no other transformation sits between the raw file the walk reads and what the deploy follows (`ctx.nestedTemplates` has two writers, both pre-rewrite), and that the 10,000-row budget accepts a 7,985-row legitimate tree.
- **Test, fixed**: the symlink-multiplied case is now sized just past the budget, so a lost budget finishes and fails the assertion (12 s) instead of spinning; a wide template pins the boundary (10,000 accepted, 10,001 refused) and that a memo hit costs one row.
- **Spec, fixed**: the `too-large` refusal and the docs sentence no longer claim such a tree must be hand-modified; the changelog entry names the 10,000-row refusal.
- **Security nit + spec round 4, fixed**: a nested-stack row named `__proto__` looked up `Object.prototype` in the plain-object index. On main that surfaced as the readable `Failed to read nested template` error; this PR's walk reached the value first and let a bare `TypeError` out. The walk now ignores a non-string path, and the provider's index is null-prototype, so such a row is followed like any other. `AssemblyReader`'s root index keeps the plain-object shape (readable error, fails closed): recorded on #3277.

## Live test

`tests/integration/nested-stack-deep/verify.sh` gains **Step 0**, before the existing deploy: synth to a temp dir, rewire the Child template's `Grandchild` row onto the Child template itself (row and file are discovered from the assembly, with asserts), `cdkd deploy --app <that dir>`, and assert a non-zero exit, the `contains a cycle` wording with a second independent marker as the drift sentinel, the `'Child' (...) -> 'Grandchild' (...)` path, that **no child engine was started** (no `Deploying nested stack` line, with Step 1 asserting an ordinary deploy does print it), and **zero** objects under `cdkd/NestedStackDeep~` plus a strict `head-object` gone-probe on the child record. `cleanup` also sweeps runaway `NestedStackDeep~...` levels deepest-first, so a regression does not leak ~190 stacks.

Not run by this lane (no real-AWS access). Verified locally without AWS: the synth, the rewire script as extracted from `verify.sh`, and that the walker refuses the resulting real assembly with the message the fixture greps. This diff is in `integ-destroy` scope (`src/provisioning/providers/**`), so a `/run-integ nested-stack-deep` run is owed before merge.


go-to-k:
Real-AWS verification on the final src sha: `nested-stack-deep` (the new Step 0 refusal arm fired with rc=1, no child engine started, no state under `cdkd/NestedStackDeep~*`; Step 1 is its positive control for the log marker), `nested-stack`, `nested-stack-3level`, `nested-stack-secret` and `recreate-nested-logical-id-collision` all passed with a clean destroy and an empty orphan scan. Decided NOT to run the guard-reverted arm against real AWS: it would deliberately start an unbounded chain of real child deploys in the account, and the same discrimination is already pinned at unit level (both guard calls removed turns 6 of 8 provider cases red with a runaway of 13 child engines).

Issue #3453, opened 2026-09-19
fix(diff): a row about to be replaced is annotated with the OLD record's sticky Cloud Control route


go-to-k:
`cdkd diff` annotates a row that is about to be REPLACED with the routing of the resource that is about to be DELETED. `collectCcApiRoutes` in `src/cli/commands/diff-recursive.ts` reads `state.resources[logicalId].provisionedBy` and, when it is `'cc-api'`, tags the row `[via CC API: sticky]` (or the returning-to-SDK token, computed by `wouldReturnToSdkProvider` from the TEMPLATE's type paired with the OLD record's `properties`). The deploy does not route a replacement that way: `deploy-engine.ts`'s `replaceDecision` passes no `provisionedBy` and no `previousProperties` for the new physical resource (only the `--recreate-via-*` hint), and the engine's own live-progress label mirrors that on purpose (`peekRoutingForLabel`, the `needsReplacement` arm). So for any replaced row whose old record is on Cloud Control, the preview names a route the create will not take.

It is sharpest on a resource `Type` change (go-to-k/cdkd#2668), where the old record's layer and property bag belong to a DIFFERENT type than the one the annotation is printed beside, but it is the same read for an ordinary create-only property change. The row itself is already truthful about the replacement: a Type change renders `Type: <old> -> <new> [requires replacement]`. Only the routing annotation is wrong.

A second preview gap in the same file, recorded here because go-to-k/cdkd#2668 (where it was first noted) closes with the routing fix: `cdkd diff` does not consult the nested-stack Type-change guard (`src/deployment/type-change-guard.ts`, `TYPE_CHANGE_NESTED_STACK`). A Type change into or out of `AWS::CloudFormation::Stack` is previewed as an ordinary `[~]` row that `cdkd deploy` (and `cdkd deploy --dry-run`) then refuses before touching anything. The preview should list such a row as blocking, the way it already does for a refused rollback-orphan adoption. Nothing is mis-provisioned; the preview just promises a plan the deploy will not run.

Found while fixing go-to-k/cdkd#2668, which deliberately left this file alone.

Fix direction: for a row the diff marks as a replacement (any `propertyChanges[].requiresReplacement`, or a recorded type that differs from the template's), compute the annotation the way `replaceDecision` routes — silent-drop properties of the NEW template bag only, no sticky read, no `previousProperties` — and keep the current read for in-place rows.

[...]

PR #3454, opened 2026-09-19
fix(deploy): route each half of a Type-change replacement on its own type

### LINKED ISSUE #2668: fix(deployment): a resource Type change routes the OLD resource's delete at the NEW type's provider, so changing a Type to AWS::CloudFormation::Stack destroys a child stack

### LINKED ISSUE #3036: fix(deploy): the UPDATE arm's no-op skip swallows a resource Type change whose property bags compare equal
The UPDATE arm's no-op skip in `src/deployment/deploy-engine.ts` can swallow a resource `Type` change. When the property bags compare equal, the deploy completes, calls no provider, and the state record keeps the OLD type. Found during the review of go-to-k/cdkd#3020.

The path, each link read in the code:

- `src/analyzer/diff-calculator.ts` has a branch for `currentResource.resourceType !== desiredResource.Type`. It emits the type change as `changeType: 'UPDATE'`, carrying `propertyChanges: [{ path: 'Type', requiresReplacement: true }]`. It sets the NEW type as `resourceType` and the old record's properties as `currentProperties`.
- In `deploy-engine.ts`, the UPDATE arm's no-op re-check (`JSON.stringify(desired...) === JSON.stringify(currentPropsAsWritten)`) compares properties only. It runs before `propertyDrivenReplacement` reads the `Type` change's `requiresReplacement`. Earlier in the arm, `resourceType` is used for narrowing, secret handling and asset auditing, but the arm never compares the recorded type with the new one before it skips.
- When the bags match, the skip breaks out of the arm. With an attribute change present, the attribute-only branch nested inside it spreads `...currentResource`, so the record keeps the old type. Without one, the resource is skipped outright.

Measured at unit level, with a mocked diff reporting that change, in one run on the branch of go-to-k/cdkd#3020. The record was on `AWS::SQS::Queue` and the template declared `AWS::CloudWatch::Alarm`, with identical properties that no narrowing changes. The deploy completed with no error and made zero provider calls (`update`, `create` and `delete` all 0). The saved record's `resourceType` stayed `AWS::SQS::Queue`.

What that one run does not establish:

- It was not run against real AWS.
- It was not run on `main`. By the code path, `main` behaves the same for bags that neither side's narrowing changes. For a bag holding a removable silent drop, `main` narrows only the stored side while go-to-k/cdkd#3020 narrows both, so the two differ there.
- It did not repeat across deploys. As long as state, template, resolution and the allow set stay unchanged, the next deploy computes the same diff and takes the same skip, but that is derived from the code, not measured.

go-to-k/cdkd#3020 widens the case slightly. With it, a record not on `'cc-api'` is also swallowed when desired-side narrowing applies and the two bags serialize equal after it: that is, when they differ only by an allow-listed removable silent drop of the NEW type.

Reachability needs a `Type` change whose bags compare equal after resolution and narrowing. A change into or out of `AWS::CloudFormation::Stack` is already refused at plan time by the go-to-k/cdkd#2668 stopgap in `doDeploy`, so the nested-stack pair cannot reach this.

Related, but a different root cause: go-to-k/cdkd#2668. That issue routes the OLD resource's delete, and its rollback, on the NEW type. This one discards the replacement intent before any routing happens, so fixing the routing alone does not close it.

Fix direction: the no-op skip should not fire when `currentResource.resourceType !== resourceType`. That alone is not a complete fix for type changes. It sends the change into the replacement path, whose `oldDeleteProvider` still routes on the new type (go-to-k/cdkd#2668). So it should land with or after that routing decision, or be limited to type pairs where the mis-route cannot delete an unrelated resource.

Verification plan:

- A unit case of the measured shape must show the replacement being attempted, not zero provider calls, and must fail on the pre-fix engine.
- If the fix lands with the routing decision, the case must also show the old resource's delete going to the OLD type's provider.
- A real-AWS arm that changes a `Type` across two deploys with an unchanged property bag. Two green deploys prove nothing here, because the bug also produces two green deploys. The arm must assert:
  - the saved record's `resourceType` is the NEW type;
  - a resource of the NEW type exists in AWS, read from AWS rather than from state;
  - the old resource is deleted or retained as its `UpdateReplacePolicy` says;
  - those assertions FAIL against the pre-fix engine.

Dup-check: searched open issues (titles and bodies) for "type change", "Type change", "resourceType !==" and "path: 'Type'" -- only go-to-k/cdkd#2668 is close; it is the routing half of Type-change handling, not the no-op skip
Session-fit: next (not this session) -- the fix has to be ordered against go-to-k/cdkd#2668's routing decision, and the lane that found it (go-to-k/cdkd#3020) is at its final review and was asked only for a comment correction
Severity: medium -- an intended Type change is silently not applied: the deploy reports success, and the state record and AWS keep the old type; rare, because it needs equal property bags across two types
Effort: medium (M) -- the condition is one line, but it sits in deploy-engine.ts, so it drags the broad integ and a unit matrix, plus the ordering with go-to-k/cdkd#2668
Estimate: ~2-3 h -- mostly the interaction with the replacement routing and a real-AWS arm that changes a Type

go-to-k:
Closes #2668
Closes #3036

## The defect

A resource whose `Type` changes on an existing logical id is diffed as an `UPDATE` carrying the TEMPLATE's (new) type. The engine bound that one type and routed BOTH halves of the replacement on it, paired with the state record's layer:

- the OLD resource's delete went to the NEW type's provider: an API error, a leak, or, where two types share a physical-id namespace, the deletion of an unrelated live resource;
- type-keyed guards evaluated the wrong type, so a stateful old type escaped `--force-stateful-recreation` whenever the new type was not stateful;
- the rollback journal named only the new type, so reversing the replacement re-created the old resource through the new type's provider;
- (#3036) the UPDATE arm's properties-only no-op skip swallowed a Type change whose two bags compare equal: green deploy, zero provider calls, record still on the old type.

## What this PR does

Rule: **a replacement has two types.** Everything aimed at the OLD resource routes on the state record's `resourceType` + `provisionedBy`; everything aimed at the new one on the template's. A recorded type that differs from the template's is always a replacement. Design: `docs/design/2668-type-change-routing.md`.

**Deploy engine** (`case 'UPDATE'`): `oldResourceType` / `typeChanged`. The no-op skip (and the attribute-only branch nested in it) is gated on `!typeChanged`; `propertyDrivenReplacement` is `typeChanged || ...`, read off the record so a change shape without the `Type` row cannot reach the in-place arm; `isStatefulRecreateTargetForReplace`, `oldDeleteProvider`, every `prepareFinalSnapshotForDelete` and every old-side `delete()` (the `--recreate-via-*` arm, `replaceDeleteFirstAndRecreate`, the post-create cleanup) take the old type. The two name-idempotent-create guards (`createResult.physicalId === currentResource.physicalId`) are skipped across a Type change, where an equal id is two resources (an SSM parameter and a log group, an IAM user and a group, can share a bare name) — EXCEPT between two custom-resource types (`Custom::*`, `AWS::CloudFormation::CustomResource`) on the SDK layer, whose handler picks the id: there the equal id IS the existing resource, so the guards stay live and refuse with `NAMED_REPLACEMENT_IDEMPOTENT_CREATE` instead of sending `Delete` for what the create just built. One helper, `equalIdNamesSameResource` in `type-change-guard.ts`, serves the engine and the rollback's adopted-live-resource shortcut. The create-first collision messages say, on a Type change, that the name may be held by an unrelated resource of the new type. The live-progress verb mirrors the dispatch.

**Rollback journal**: additive `CompletedOperation.previousResourceType`, stamped on every UPDATE with a previous record. No `journalVersion` bump (same precedent as `oldResourceRetained` / `failedOperations`), no state-schema change. `parseRollbackJournal` refuses a non-string.

**Rollback executor**: `resolveReplacementOldType(op)` is the single reader. Stamped field, else `previousState.resourceType` (journaled by every binary, so a legacy journal replays correctly), else REFUSE. It also refuses when the two sources disagree and for a nested-stack pair. A refused op classifies `refuse-replacement-routing` AFTER the idempotent skips, and the replay THROWS `ROLLBACK_REPLACEMENT_UNROUTABLE`: a failure, so the segment is kept (a warning would pop it); nothing is called in AWS; `--orphan <id>` lets the rest proceed. `isReplacementOp` is true for a changed type even with an unchanged physical id, and both "already reverted" recognitions (equal id, equal property bag) additionally need the record to be the old type. The re-create, its Cloud Control fallback name and the stateful-data warning use the old type; the new resource's delete keeps `op.resourceType`.

**`--revert-failed`**: a failed Type change classifies `skip-failed-type-change`. That arm is an in-place `update()` routed on the new type against the old physical id.

**`cdkd rollback` plan**: shows `NEW -> OLD` for such a replacement, `(REFUSED)` for an unroutable one.

## The #2674 stopgap: KEPT

A Type change into or out of `AWS::CloudFormation::Stack` is still refused (`TYPE_CHANGE_NESTED_STACK`). With the delete routed on the old type the original mis-aim is gone, but the replacement's cleanup delete is warn-and-continue, which for a nested row would strand a whole child stack and its state record under a green deploy; the into-nested direction runs a whole child deploy as one row of a replacement, which nothing has exercised; and neither direction has run against real AWS. The guard's doc comment, message and docs section were rewritten, because the old text described a mis-route that no longer happens. Lifting the refusal is a separate decision.

## `cdkd diff`: not changed, filed

The row already renders `Type: old -> new [requires replacement]`. Its routing annotation still pairs the template's type with the OLD record's sticky layer for a replaced row; that is `src/cli/commands/diff-recursive.ts`, held by open PR #3226, so it is filed as #3453 (`Session-fit: next`, external input).

## Sweep (work-issues 5-b)

Population: `getProviderFor(` call sites in `src/deployment` + `src/cli`, 24 before and 24 after. Sites pairing a state-side layer with a template-side type for the OLD half:

| site | before | after |
| --- | --- | --- |
| `deploy-engine.ts` `oldDeleteProvider` | mismatched | old type |
| `rollback-executor.ts` reverse-replacement re-create | mismatched | old type |
| `rollback-executor.ts` `revert-failed-update` | mismatched on a Type change | op skipped |
| `deploy-engine.ts` in-place `updateDecision` / its fallback `replDecision`; `rollback-executor.ts` `revert` | reachable by a Type change only through the no-op / in-place arm | unreachable: a changed type is always the replacement arm / a replacement op |
| DELETE arms, CREATE-rollback deletes, readopt / new-resource deletes, `state.ts`, `drift.ts`, `destroy-runner.ts`, `import.ts`, `diff.ts` | single source (both from one record) | unchanged |
| `diff-recursive.ts` `collectCcApiRoutes` (not a `getProviderFor` site) | mismatched | #3453 |

## Tests

- `deploy-engine-type-change-routing.test.ts`, `rollback-executor-type-change-routing.test.ts`, and plan-label cases in `tests/unit/cli/commands/rollback.test.ts`. The registry double hands out one provider PER TYPE; with a shared provider "delete was called" is satisfied by the mis-route itself. Both layers (`sdk`, `cc-api`), both polarities of each gate.
- Measured at the first review sha: against `origin/main`'s `deploy-engine.ts` 10 of the then-14 engine cases failed (the 4 that passed are the controls and the two cases that were already correct); mis-routing the replay re-create failed 3 rollback cases; removing the adopted-resource gate failed 1; reverting the namespace predicate to `!typeChanged` fails the custom-resource case; unwrapping `safe()` on the old type fails the forged-old-type log-injection case.
- Existing rollback suites paired an op of one type with a `previousState` of another (`'T'` over an S3 record). That shape now means "Type change", so those fixtures were given matching types; two count pins moved for the new renders, each with its reason.

## Real-AWS arm (written here, NOT run: lane agents do not run integs)

`tests/integration/type-change-replacement/` moves three rows in one deploy: `Subject` (stateful SSM parameter -> topic), `Overlap` (SSM parameter -> log group of the SAME name, so equal physical ids), `EqualBags` (topic -> log group with an identical `Tags`-only bag, the #3036 shape). Steps: v1; v2 refused without `--force-stateful-recreation`; v2 + injected failure with automatic rollback; the same under `--no-rollback` (journal must name both types per op) then `cdkd rollback`; v2 for real; no-op redeploy; destroy with every physical name from both variants confirmed gone. Both templates were synthesized locally and every jq / JMESPath expression and not-found signature was executed (read-only). Assertions that stay unvalidated until the real run: the wording greps (`--force-stateful-recreation` + `Subject (AWS::SSM::Parameter)` on one line, `Reversing replacement of <id>` appearing in a non-verbose automatic rollback, `No changes detected`), the exit code of the plain `cdkd rollback` (expected 0), and cdkd's generated names for the nameless topic / log group (cross-checked against state, so a drift fails loudly).

Review (code + spec + test + security, one dispatch, fix rounds re-checked by delta with the same reviewers). Round 1: a security blocker — the equal-id guards were skipped for EVERY Type change, which for custom resources deletes what was just created. Round 2: the round-1 fix keyed on "both halves resolve to one provider instance", which the code review showed is too broad (one instance serves `AWS::IAM::User` and `AWS::IAM::Group`); it is now keyed on the custom-resource family. Also fixed in round 1; minors fixed (stale rationale in `recreate-targets.ts`, collision wording, type-aware equal-bag recognition, docs table row, the unpinned arms the test review listed). Declined: removing the `routing.ok ?` fallback expression at the two refusal renders (kept as type narrowing; the classifier and the replay are separately callable).

The #3036 bullet "those assertions FAIL against the pre-fix engine" is met by derivation only for the real-AWS arm in this lane (step 1 asserts the two bags really are equal, steps 3 and 5 assert the old topic gone and the new log group present); the discrimination run is the parent's.

This touches deletion logic, so `integ-destroy` applies; the fixture list for the run is in the lane report.


go-to-k:
Real-AWS verification on the final src sha (ebb803c41's src; later commits are fixtures and the ledger only): the new `type-change-replacement` fixture passed end to end, and is RED at step 2 on the pre-fix engine (the same run reproduced the mis-routed old-half delete and the wrong-type rollback re-create live). Broad (`lambda`, `drift-revert`, `remove-protection`, `export`), replacement (`update-replace`, `replacement-immutable-name`, `replacement-fanout`, `efs-immutable-replacement`, `elbv2-same-name-replacement`), rollback (`rollback-command`, `rollback-sqs-cooldown`, `rollback-replay-effective-props`, `rollback-replay-cc-generated-name`, `rollback-replacement-retain`, `rollback-failure-injection`, `rollback-deletion-policy-snapshot`) and recreate (`recreate-via-cc-api`, `recreate-via-sdk-provider`, `recreate-mixed-direction`) all passed with a clean destroy and an empty orphan scan. Two notes: `rollback-deletion-policy-snapshot` failed once on an intermittent Cloud Control volume-delete hang; it passed on re-run and is unrelated to this diff (A/B against main's engine, CloudTrail evidence in go-to-k/cdkd#3455); and the two `recreate-*` fixtures were failing at their baseline on main since the SDK provider began handling `RuntimeManagementConfig`, so this PR re-seeds their Cloud Control baseline with `--recreate-via-cc-api` (tests only).

Issue #3456, opened 2026-09-19
fix(diff): cdkd diff does not re-read a stale attribute map, so its preview of a Fn::GetAtt over one differs from what deploy resolves


go-to-k:
`cdkd deploy` now re-reads a state record's attributes from AWS when a `Fn::GetAtt` on it is about to take the physical-id fallback (issue go-to-k/cdkd#1852: a record written before its provider recorded `AWS::SSM::Parameter.Arn`, a `--no-wait` `DBInstance` with no `Endpoint.*` yet, a pre-#1681 placeholder ARN). The heal is supplied through `ResolverContext.attributeHealer`, and only `DeployEngine.buildResolverContext` supplies one. `cdkd diff` builds its resolver context in `src/cli/commands/diff-recursive.ts` and supplies none, so until a deploy has healed the record the preview and the apply disagree:

- an Output (or a new resource's property) reading the missing attribute is previewed as unresolvable -- the resolver's refusal is caught by the best-effort diff resolution and the raw intrinsic is kept -- while the deploy resolves it to the real value;
- for a non-`*Arn` attribute (`Endpoint.Address`) the preview resolves to the physical id with a warning, while the deploy serves the real endpoint.

The refusal text on a healer-less context already points at `cdkd deploy`, so nothing is wrong on screen; the gap is preview fidelity. The fix is a READ-ONLY healer for `cdkd diff` (and, if wanted, `cdkd drift`): the same provider `import()` read, memoized per run, merged into the in-memory record only and never saved. `DeployEngine.readStaleAttributes` / `src/deployment/stale-attribute-heal.ts` hold everything except the wiring; `diff.ts` already builds a `ProviderRegistry` to route with. A unit test must pin both polarities: the diff reads, and no state write happens. The `stale-attribute-heal` integ fixture's phase 3 already asserts the no-write half against real AWS.

Session-fit: next (not this session) -- external input: open PR go-to-k/cdkd#3226 holds `src/cli/commands/diff-recursive.ts` and `src/cli/commands/drift.ts`, the only files the wiring can go in; start once it merges.
Severity: low -- until the first healing deploy, `cdkd diff` previews an unresolved or physical-id value for a reference the deploy resolves correctly; nothing is written and no deploy is affected.
Effort: small (S) -- unit tests only, plus one re-run of the existing `stale-attribute-heal` fixture for the no-write assertion.
Estimate: ~1 h -- extracting the engine's read into a helper both callers share, and the two-polarity diff test.

[...]

Issue #1824, opened 2026-08-13
fix(provisioning): AWS::RDS::DBSubnetGroup.DBSubnetGroupArn and AWS::SSM::Parameter.Arn are uncached, so an output Fn::GetAtt hard-fails (1800 re-capture residual)


go-to-k:
Surfaced by the [#1800](https://github.com/go-to-k/cdkd/issues/1800) fixture re-capture (us-east-1, 2026-08-13), which refreshed
every `tests/fixtures/cfn-schemas/*.json` from a 2026-05-16 capture. AWS has since
added an ARN read-only attribute to two types, and neither provider caches it under
its CFn name — so `gen-sdk-attr-coverage` reports two `gap`s where it reported none.

## The two gaps

| type | attribute | SDK availability |
|---|---|---|
| `AWS::RDS::DBSubnetGroup` | `DBSubnetGroupArn` | `CreateDBSubnetGroup` returns `DBSubnetGroup.DBSubnetGroupArn` directly |
| `AWS::SSM::Parameter` | `Arn` | `PutParameter` returns only `Version` / `Tier`; the ARN comes from `GetParameter`'s `Parameter.ARN` (or must be constructed) |

## Why it matters

This is the [#1179](https://github.com/go-to-k/cdkd/issues/1179) GetAtt-key class. Output / cross-resource `Fn::GetAtt` reads the
cached `resource.attributes[<CFnName>]` in `IntrinsicFunctionResolver.constructAttribute`,
which never calls a provider's `getAttribute`. For a `*Arn` / `*Url` name the
resolver's shape guard HARD-FAILS rather than degrading, so a template doing

```ts
new cdk.CfnOutput(this, 'ParamArn', { value: param.attrArn });
```

fails the deploy instead of silently producing a wrong value. That is the correct
loud behavior — but the attribute is resolvable and should simply be cached.

The two differ in effort, which is why they are worth splitting when picked up:

- **`DBSubnetGroupArn`** is a direct read off the create response — the ordinary
  "record the ARN under its CFn name" fix.
- **`SSM::Parameter.Arn`** has no create-response source. It needs either a
  follow-up `GetParameter` (an extra call on every parameter create) or an ARN
  constructed from account / region / name — and the constructed form must go
  through the `fabricated` account guard (#1730 / [#1746](https://github.com/go-to-k/cdkd/issues/1746)), which refuses to record
  an ARN built from the placeholder account.

[...]

Issue #1681, opened 2026-08-12
fix(analyzer): four composite-physicalId types resolve {Ref} to cdkd's compound id (AppSync ApiKey / DataSource / Resolver, Route53 RecordSet)


go-to-k:
Found by the composite-physicalId `Ref` audit run alongside issue #1667 (the
`AWS::Glue::Table` half, which IS fixed in that PR).

## Context

`cfnRefValueFromPhysicalId` (`src/deployment/intrinsic-function-resolver.ts`)
translates cdkd's compound physicalId back to the value CloudFormation's `Ref`
returns, via two Sets — `REF_RETURNS_SEGMENT_AFTER_PIPE` and
`REF_RETURNS_SEGMENT_BEFORE_FIRST_PIPE` — plus a handful of hand-written
special cases (WAFv2 WebACL ARN reconstruction, S3Tables / Backup / CodeCommit
`stateLookup` recoveries).

The #1667 PR re-checked EVERY type in the composite-id table of
`docs/state-management.md` against its docs-verified `Ref` return value. Four
types are WRONG and **cannot be fixed by adding a Set entry**, so they were
recorded in the resolver's AUDIT RECORD comment and filed here rather than
bundled into that PR.

## The four

| Type | cdkd physicalId | documented `Ref` | why no Set entry works |
|---|---|---|---|
| `AWS::AppSync::ApiKey` | `<apiId>\|<apiKeyId>` | the **ARN** of the API key (`arn:aws:appsync:<region>:<acct>:apis/<apiId>/apikey/<keyId>`) | `Ref` is not a segment of the id — it must be RECONSTRUCTED (the WAFv2 WebACL shape) or read from a stored attribute |
| `AWS::AppSync::DataSource` | `<apiId>\|<name>` | the **ARN** of the data source (`.../apis/<apiId>/datasources/<name>`) | same |
| `AWS::AppSync::Resolver` | `<apiId>\|<typeName>\|<fieldName>` | the **ARN** of the resolver (`.../apis/<apiId>/types/<type>/resolvers/<field>`) | same |
| `AWS::Route53::RecordSet` | `<hostedZoneId>\|<name>\|<type>` | "The name of the record" | the value is the **MIDDLE** segment; after-LAST-pipe yields the record TYPE (`A`), before-FIRST-pipe yields the hosted zone id |

Sources (all "Return values / Ref", read 2026-08-12):
[...]

PR #3457, opened 2026-09-19
fix(deploy): heal a stale attribute map on a Fn::GetAtt miss instead of refusing forever

### LINKED ISSUE #1852: fix(deployment): an issue-1824-era stale attribute map is never healed, and the resolver "not enriched for this resource type" message is now false for the two fixed types

go-to-k:
## Summary

A state record's `attributes` map is written by the provider's `create()` / `update()`, and the deploy engine's no-change skip never re-runs the provider. A record written before its provider recorded an attribute (`AWS::SSM::Parameter.Arn`, `AWS::RDS::DBSubnetGroup.DBSubnetGroupArn` -- issue #1824 fixed new deploys only), written under `--no-wait` while a `DBInstance` had no `Endpoint.Address` / `Endpoint.Port` yet (#3077, the checklist row), or holding a pre-#1681 placeholder ARN (#1727), stayed wrong for as long as the resource's own properties did not change. Adding `new CfnOutput(this, 'ParamArn', { value: param.attrArn })` therefore hit the resolver's `*Arn` refusal on every deploy, with a message ("not enriched ... file an issue") that was false.

This takes the issue's direction 1 plus direction 3.

## What changes

- **Heal lazily, on a miss.** `resolveGetAtt` now goes through `constructWithStaleRecordHeal`: a PROBE pass in which `guardedPhysicalIdFallback` raises an internal miss signal instead of deciding, then -- only on that signal -- one call to `ResolverContext.attributeHealer`, then either the healed value or a SETTLE pass in which the fallback decides exactly as before and words its refusal from the heal outcome. A cached attribute, and every per-type arm that constructs or live-reads its answer, costs no AWS call. The phase rides a DERIVED context, never the resolver instance, which concurrent resolutions share.
- **The read primitive is the provider's `import()` with `knownPhysicalId`** (what `orphan-adoption.ts` already verifies a record with): read-only by contract, returns the map `create()` records, and is routed by the RECORD's `resourceType` + `provisionedBy`, so a Cloud-Control-routed record reads through Cloud Control. `getAttribute()` is implemented by about half the providers and by neither named type.
- **Only `DeployEngine` supplies a healer**, on every context it builds (diff pass, both provisioning arms, outputs). It is single-flight and memoized per deploy on the ENGINE instance (`Map<logicalId + physicalId, Promise>`): one read per record per deploy, success or failure, no retry, never throws. A record this deploy (re)wrote, a custom resource and a nested stack are never re-read.
- **Persistence** is at `redactStateForPersist`, the choke point every `saveState` passes through, BEFORE `scrubResourceRecord`: add-only into `attributes` (a recorded value always wins; the one exception is a recorded pre-#1681 placeholder ARN), never an empty value, nothing else in the record touched, dropped if the record's physical id / type changed this run. So it survives every exit path that saves (the no-change path gained a trigger) and `--dry-run` persists nothing.
- **Eligibility is checked on every ask, ahead of the memo, and again at the persist merge** (`isHealEligible`): the memo key survives an in-place UPDATE, so a read the diff pass took is neither served after this deploy rewrote the record nor merged under the rewritten record.
- **A masked read-back is never a value.** `CloudControlProvider.import` masks every leaf it cannot certify as a read-only attribute (all of them when `cloudformation:DescribeType` is denied), on the premise that a masked read is refused downstream. The engine drops such keys before returning or storing them, the resolver refuses a masked value again (`usableHealedAttribute`), a served value passes through `noteAttributeSecrecy`, and the refusal says the value was WITHHELD and names the permission (`withheldKeys` on the `read` outcome).
- **Messages.** A failed / not-found read says the record is stale, that cdkd tried, the error CLASS and HTTP status (AWS's text stays behind `--verbose`) and the real remedy. "not enriched ... file an issue" remains only where the re-read completed without the attribute or was never attempted. All stay `markNonRetryable` (the outcome is memoized for the deploy).
- **#1727 decision:** `rejectPlaceholderArnAttribute`'s remedy ("deploy again so the next update heals it") had the same no-change gap, so the heal COVERS it and the sentence is replaced. The DBProxy / DBProxyEndpoint `VpcId` refusal had it too and takes the same heal.
- **#3077 row:** `importDBInstance` in the RDS / DocDB / Neptune providers returned `attributes: {}`; it now returns `definedAttributes({Endpoint.Address, Endpoint.Port, Arn})` from the describe it already issued. A still-`creating` instance heals nothing and keeps today's warn + fallback. Side effect: `cdkd import` of a `DBInstance` records those attributes.
- `cdkd diff` / `drift` / `export` supply no healer: no read, no write, and their refusal now points at `cdkd deploy`. Wiring a read-only healer into `cdkd diff` is #3456 (its files are held by open PR #3226).

Design: `docs/design/1852-stale-attribute-heal.md` (includes the sibling-site sweep table).

## Tests

- `tests/unit/deployment/stale-attribute-heal.test.ts` -- resolver: asked only on a miss, per-outcome wording, `Fn::Sub`, the Cloud Control nested shape, the DBInstance row in both polarities, the placeholder and DBProxy arms, concurrency (a slow heal must not turn a sibling's fallback into a probe), leaf helpers.
- `tests/unit/deployment/deploy-engine-stale-attribute-heal.test.ts` -- real resolver + mocked registry: `import()` arguments, `cc-api` routing, single flight, per-deploy memo, merge rules, failure degradation (AccessDenied / not found / mismatched id / no `import()`), custom resource exclusion, `--dry-run` both polarities, a deploy that FAILS after the heal still persisting it, a re-written record not being re-read.
- `tests/unit/provisioning/dbinstance-import-endpoint-attributes.test.ts` -- three providers, available vs `creating`.
- Review round 1 added: true concurrent single flight (three parallel misses parked on one read), heal-then-UPDATE, a replaced record, the metadata-only arm, the persisted placeholder-ARN overwrite, a masked Cloud Control read-back (not served, not persisted, worded as withheld), the `healedAttributes` per-deploy reset, an unroutable type, nested-stack exclusion.
- Mutation probes run against the committed tree (each restored byte-exact): unwiring the healer (12 red), dropping the persist merge (7), the memo (3), the eligibility gate (1), the no-change trigger (1), the per-deploy reset (1), the mismatched-id guard (1), the type exclusion (1), the probe throw (8+), the stale-wording switch (4), sharing the context instead of deriving it (1). Removing the engine-side empty-value normalization alone stays green: the merge and the resolver read drop empty values too (three layers, pinned at the two outer ones).
- New real-AWS fixture `tests/integration/stale-attribute-heal` (verify.sh): deploy v1, strip `attributes.Arn` from the state record out of band, `cdkd diff` writes no state (object `VersionId` + `ETag`), deploy v2 (adds ONLY the output) exits 0 on the no-change path with the output and the record equal to the ARN AWS reports and the parameter's `Version` / `LastModifiedDate` unchanged, a second no-change deploy writes nothing, destroy + gone-probes. Both variants synthesized and every jq / JMESPath expression executed locally; NOT yet run against AWS.

## Verification still owed before merge

Real-AWS runs (the engine and resolver are cross-cutting): the new `stale-attribute-heal` fixture, plus the broad set. Until that run, the fixture's assertions on cdkd's own output (`No changes detected`, absence of `Cannot resolve Fn::GetAtt`) and on S3 `VersionId` behaviour are unvalidated.

Closes #1852


go-to-k:
Real-AWS verification on the final src sha (77e1b316f's src; later commits are a fixture repair and the ledger only): the new `stale-attribute-heal` fixture passed end to end (a record stripped of `attributes.Arn` heals on a deploy that only adds the output; the parameter is untouched; `cdkd diff` and a no-change re-deploy write nothing) and is RED at phase 4 on the pre-fix engine with the exact false "attributes are not enriched" message from go-to-k/cdkd#1852. Broad (`lambda`, `drift-revert`, `remove-protection`, `export`) and GetAtt / outputs / cross-stack fixtures (`getatt-fallback-guard`, `output-never-resolved-diff`, `outputs-only-export`, `deep-getatt-chains`, `cc-getatt-readback`, `sdk-ccapi-crossref`, `custom-resource-getatt-data`, `cross-stack-cfn-fallback`, `nested-stack`, `appsync`, `sdk-to-cc-autoroute`) all passed with a clean destroy and an empty orphan scan. One note: `sdk-ccapi-crossref` was failing at its baseline on main since the Kinesis SDK provider began handling `DesiredShardLevelMetrics` (go-to-k/cdkd#1700), so this PR rebuilds its Cloud Control side on `AWS::Events::Archive` (no SDK provider) plus a flag-seeded Lambda (tests only).

Issue #3458, opened 2026-09-20
fix(dynamodb): a Table StreamSpecification.ResourcePolicy and StreamSpecification.Tags are dropped on the wire and unreadable on drift


go-to-k:
Found by the nested-member sweep of `AWS::DynamoDB::Table` done while fixing go-to-k/cdkd#1782 (the #1370 nested-key silent-drop class).

CloudFormation's `AWS::DynamoDB::Table.StreamSpecification` has THREE members — `StreamViewType`, `ResourcePolicy` (`{ PolicyDocument }`, the resource-based policy of the STREAM) and `Tags`. Evidence: `tests/fixtures/cfn-schemas/AWS-DynamoDB-Table.json` (`definitionShapes.StreamSpecification` = `StreamViewType`, `ResourcePolicy`, `Tags`) and `aws-cdk-lib` 2.270.0's `convertCfnTableStreamSpecificationPropertyToCloudFormation`, which renders all three.

`DynamoDBTableProvider` handles only the first. `create()` builds `{ StreamEnabled: true, StreamViewType }` and every `update()` arm (enable, view-type change, disable) does the same, so a template declaring `StreamSpecification.ResourcePolicy` or `StreamSpecification.Tags` deploys green with neither applied, and `readCurrentState` emits only `{ StreamEnabled, StreamViewType }`, so neither can be compared for drift. The SDK's `StreamSpecification` has no such members: a stream policy is applied with `PutResourcePolicy` against the STREAM ARN (the table-level `ResourcePolicy` branch already calls that API against the table ARN). Whether a stream accepts tags through `dynamodb:TagResource`, and what CloudFormation does with `StreamSpecification.Tags`, needs measuring before the write path is designed.

Not fixed in the #1782 PR because it is a different write path with its own lifecycle: a `StreamViewType` change disables and re-enables the stream, which mints a NEW stream ARN, so the policy has to be re-applied there, removed on a block removal, swapped on rollback, and read back with `GetResourcePolicy` against the current `LatestStreamArn`.

[...]

Issue #1812, opened 2026-08-13
feat(dynamodb): declare the new drift seams for AWS::DynamoDB::Table (consume the #1799 mechanism)


go-to-k:
The (#1799) PR shipped the two drift declaration seams as MECHANISM ONLY and
named `AWS::DynamoDB::Table` as the intended first consumer. No provider
declares either yet. This issue is that consumption.

Both seams were surfaced as residuals by the (#1767) work, filed as (#1783) /
(#1784), and closed by (#1799) when the mechanism landed — so the remaining
work has no issue of its own until this one.

## 1. Declare the leaf-only ordering form

`getDriftUnorderedPaths` in `src/provisioning/providers/dynamodb-table-provider.ts`
does not declare the index lists. The reason it could not was that every entry
was a SUBTREE declaration and the unordered walk descends into array elements,
so `'GlobalSecondaryIndexes'` would also sort each index's `KeySchema`, where
HASH must precede RANGE — hiding a real key change instead of removing a
phantom one.

`LEAF_ONLY_PATH_SUFFIX` now exists: `'GlobalSecondaryIndexes[]'` claims the
path alone. Declaring it removes the remaining phantom for the `properties`
-baseline population (the one a reverse-replacement rollback leaves behind),
where an index list AWS returns in a different ORDER than the template declared
still reports as drift.

## 2. Use `canonicalizeDriftProperties` for the array-member residual

`getDriftUnknownPaths` currently ignores each index list only when the template
declares none. For a table that DOES declare indexes, an `observedProperties`
record written by a pre-(#1767) binary reports a one-sided difference until the
next deploy or `cdkd drift --accept`, because `isIgnoredPath` is never asked
about a path crossing an array (`diffAt` compares arrays wholesale).

`ResourceProvider.canonicalizeDriftProperties(resourceType, properties)` is the
seam for exactly that: applied to BOTH comparison sides, stripping the
AWS-managed member from each bag converges an already-written record with a
post-fix readback, with no ignore-path and no lost detection. Note the
constraints the interface imposes — pure, synchronous, non-mutating,
identity-returning when inapplicable, and NO `side` argument (one-sided
normalization manufactures drift on the `properties`-fallback baseline).

## Also update the stale in-code rationale

The two JSDoc blocks in `dynamodb-table-provider.ts` that explain WHY each half
was not shipped now assert limitations that no longer hold. They were corrected
to point here when (#1799) landed, but the explanation of what to do is this
issue.

## Sibling

[...]

Issue #1783, opened 2026-08-13
feat(drift): getDriftUnorderedPaths has no leaf-only form, so an unordered object list containing an order-significant array cannot be declared


go-to-k:
Found while fixing #1767, which proposes declaring the `AWS::DynamoDB::Table` index lists in `getDriftUnorderedPaths` alongside the member reverse-map (the issue calls the two separable). The reverse-map shipped; the ordering half could not, and the obstacle is the declaration mechanism rather than the lists.

## The wall

Every `getDriftUnorderedPaths` entry is a SUBTREE declaration — there is no leaf-only form — and, unlike `getDriftUnknownPaths`, this walk DESCENDS INTO ARRAY ELEMENTS, giving each element its parent path (`src/analyzer/drift-normalize.ts`, `walkUnordered`). So declaring

```ts
getDriftUnorderedPaths() { return ['AttributeDefinitions', 'GlobalSecondaryIndexes']; }
```

sorts the index list AND then reaches `GlobalSecondaryIndexes.KeySchema` inside every element, sorting the per-index key schema too. `DynamoDBTableProvider.getDriftUnorderedPaths` states the opposite in the same docstring, deliberately (#1760):

> `KeySchema` is deliberately NOT declared, at the table level or the index level: it is order-SIGNIFICANT (HASH before RANGE), so sorting it would HIDE a real key change rather than remove a phantom one.

Declaring the index list would reverse that decision at the index level only — an inconsistency no reader could be expected to derive — so #1767 left the lists undeclared and recorded the reason in-code.

## Consequence being carried

An index list AWS returns in a different ORDER than the template declared is still phantom drift against a `properties` baseline (the post-reverse-replacement-rollback / pre-observed-capture population — the same reachability #1760 documents for `AttributeDefinitions`). It is NOT reachable on the ordinary `observedProperties` path, where both comparison sides come from the same readback.

The same wall applies to any type wanting an unordered object list whose elements contain an order-significant array of their own; `AWS::DynamoDB::GlobalTable` is the obvious next one.

## Candidate answers

1. **A leaf-only form.** e.g. a trailing marker (`'GlobalSecondaryIndexes[]'`) or a second list, meaning "sort the array AT this path, do not descend". Smallest change, and it makes the two provider path-lists diverge in spelling, which the shared `matchesPathPrefix` currently prevents.
[...]

Issue #1767, opened 2026-08-13
fix(dynamodb): the Table GSI / LSI readback is emitted verbatim, so ItemCount alone drifts every in-use table with an index


go-to-k:
Found by the code review of the (#1760) PR, in the three lines below the fix it was reviewing.

`DynamoDBTableProvider.readCurrentState` emits the GSI / LSI blobs straight off the SDK response:

```ts
// src/provisioning/providers/dynamodb-table-provider.ts (readCurrentState)
if (table.GlobalSecondaryIndexes && table.GlobalSecondaryIndexes.length > 0) {
  result['GlobalSecondaryIndexes'] = table.GlobalSecondaryIndexes;
}
if (table.LocalSecondaryIndexes && table.LocalSecondaryIndexes.length > 0) {
  result['LocalSecondaryIndexes'] = table.LocalSecondaryIndexes;
}
```

The method's own docstring says the comparator "filters them", and it does not — that is only true for a TOP-LEVEL key absent from the baseline (`calculateResourceDrift` walks `Object.keys(stateProperties)` at the top level). These are members of an ARRAY under a key the baseline DOES carry, and arrays are compared by positional `deepEqual`, so every AWS-managed member participates.

Measured us-east-1 2026-08-13 — a table created with one GSI and no capacity settings reads back:

```json
"GlobalSecondaryIndexes": [
  {
    "IndexName": "gsi1",
    "KeySchema": [{ "AttributeName": "gsipk", "KeyType": "HASH" }],
    "Projection": { "ProjectionType": "ALL" },
    "IndexStatus": "ACTIVE",
    "ProvisionedThroughput": { "NumberOfDecreasesToday": 0, "ReadCapacityUnits": 0, "WriteCapacityUnits": 0 },
    "IndexSizeBytes": 0,
    "ItemCount": 0,
    "IndexArn": "arn:aws:dynamodb:us-east-1:...:table/.../index/gsi1",
    "WarmThroughput": { "ReadUnitsPerSecond": 12000, "WriteUnitsPerSecond": 4000, "Status": "ACTIVE" }
  }
]
```

[...]

PR #3459, opened 2026-09-20
fix(dynamodb): apply and read back a Table GSI's per-index ContributorInsightsSpecification

### LINKED ISSUE #1782: fix(dynamodb): a Table GSI per-index ContributorInsightsSpecification is dropped on the wire AND unreadable on drift
Found while fixing #1767 (the `AWS::DynamoDB::Table` GSI / LSI readback reverse-map).

CFn's `AWS::DynamoDB::Table.GlobalSecondaryIndex` carries a `ContributorInsightsSpecification` member. cdkd handles it at neither end:

- **Write side.** `DynamoDBTableProvider.create()` forwards `properties['GlobalSecondaryIndexes']` verbatim into `CreateTableCommandInput.GlobalSecondaryIndexes`. The SDK's `GlobalSecondaryIndex` shape has no `ContributorInsightsSpecification` member, so the AWS SDK v3 serializer discards it silently — the #1370 nested-key silent-drop class. A template asking for per-index Contributor Insights deploys green with the feature OFF. (The TABLE-level `ContributorInsightsSpecification` IS wired, via `UpdateContributorInsights`; only the per-index one is not. `UpdateContributorInsights` takes an `IndexName`, so the wiring exists one argument away.)
- **Read side.** `DescribeTable`'s index description has no counterpart either, so the #1767 reverse-mapper cannot emit it. A template declaring it therefore cannot converge against a `properties` drift baseline: the recorded index entry carries a member the AWS side can never produce.

The two halves want fixing together — emitting the key from `readCurrentState` without sending it would record a value cdkd does not hold, and sending it without reading it back leaves permanent phantom drift.

## Suggested shape

Per index, call `UpdateContributorInsights({TableName, IndexName, ContributorInsightsAction})` after the index is ACTIVE (the same call the table-level branch already makes, with `IndexName` added), and read it back with a per-index `DescribeContributorInsights({TableName, IndexName})` in `readCurrentState`, emit-when-present and gated on the desired index declaring the block — the shape #1767 shipped for the three throughput blocks. Note the readback costs one API call per index, so the gate matters.

Not fixed in the #1767 PR: that change is a pure reverse-map with no new AWS calls, and this needs both a new write path and a new per-index read.

go-to-k:
Closes #1782

## The defect

CloudFormation's `AWS::DynamoDB::Table.GlobalSecondaryIndex` carries a `ContributorInsightsSpecification` member. The SDK's `GlobalSecondaryIndex` shape has none, so:

- **write**: `create()` forwarded the index entry to `CreateTable` and the serializer discarded the member; `applyGsiUpdates` never sent it either. A template asking for per-index Contributor Insights deployed green with the feature OFF.
- **read**: `DescribeTable`'s index description has no counterpart, so the #1767 reverse-map could not emit it and such a template could never converge against a `properties` drift baseline.

## The fix (both halves, as the issue asks)

New pure module `src/provisioning/dynamodb-contributor-insights.ts` holds ONE reader / planner / reverse-map for the table-level and the per-index block, so the write plan and the read-back gate cannot disagree about what "declares the block" means.

**Write** — `applyIndexContributorInsights`, called from `create()` (after the ACTIVE wait) and from `update()` (after `applyGsiUpdates`, so an index the update adds exists by then):

- one `UpdateContributorInsights({TableName, IndexName, ContributorInsightsAction, ContributorInsightsMode?})` per index whose block CHANGED against the recorded previous entry (matched by `IndexName`, never by position): index new, block added, `Enabled` flipped, `Mode` changed while enabled. `Mode` is never sent with `DISABLE`.
- **removal**: a block removed from an index that STAYS is `DISABLE`d — the rule the table-level block, PITR, TTL and Kinesis already follow in this provider. An index the update REMOVES gets no call (its rules go with it, and the call would fail with `ResourceNotFoundException`).
- **rollback**: the planner is symmetric in its two sides, so `update()` replayed with the sides swapped restores the old per-index setting (pinned by a unit test).
- **timing**: the command doc says `UpdateContributorInsights` answers `ResourceNotFoundException` when the index "status might not be ACTIVE", which `retryOnTransientControlPlane` does not retry. So `waitForTableAndIndexesActive` runs first — only when there is at least one call to make, so an ordinary update pays no extra `DescribeTable`. Each call then goes through `retryOnTransientControlPlane`, exactly like the table-level one. Like the table-level branch, the deploy does NOT wait for ENABLING/DISABLING to settle.
- the member is **stripped** from the `CreateTable` forward (`stripIndexContributorInsights`), so nothing relies on the serializer dropping it. The `Create` GSI action already builds a fresh object and never carried it. The caller's bag is not mutated (it is what the engine records).
- an **unreadable** block (a scalar, an array, an unresolved intrinsic, a non-boolean `Enabled`, a non-string `Mode`), at either level: a template CREATE refuses it before `CreateTable` (naming the index by position, never by its resolved name); an UPDATE, and a state-replay create, issue nothing and warn through the masked sink, leaving the live setting alone. The warning's path never embeds the index name, so a short secret-derived name is printed only where it is masked as a whole value.

**Read** — in `readCurrentState`, after the index list is reverse-mapped:

- one `DescribeContributorInsights({TableName, IndexName})` per index, ONLY when the bag is informative AND the desired entry for that `IndexName` declares a usable block (the same reader the write plan runs). An uninformative bag reads nothing, which is the pre-fix answer.
- `Mode` is emitted only when the desired block declares one (by the reader's own answer, so `Mode: ''` counts as undeclared). While `ENABLING`, a mode AWS does not report yet reads as the declared one; once `ENABLED`, only AWS's own report is used. A stringly `Enabled: 'true'` is read back in its declared spelling, since the comparator does not coerce. The block sits inside an ARRAY entry, which the comparator matches by exact key set, and AWS reports a mode for every enabled rule set — an ungated `Mode` would be permanent one-sided drift.
- `ENABLING` / `DISABLING` are read as their TARGET. This deliberately differs from the table-level read (which omits the block on a transient status): the deploy does not wait for the toggle, so the post-deploy `observedProperties` capture normally sees `ENABLING`, and an omitted block would freeze an array entry the settled index can never equal. `FAILED` and a failed read (AccessDenied, not-found) omit the block and log at debug, like the sibling reads.

**Table level, same reader** — `applyContributorInsights` now reads through the shared reader. Two behaviour changes: `Enabled: 'false'` (a CloudFormation-legal string) used to ENABLE the feature (`Boolean('false')`), and now disables; an unreadable block is refused at create / announced and skipped on update instead of being read as a default. The table-level read-back is unchanged (it now goes through the shared reverse-map with its old options).

## Parity notes

- `cdkd import`: `import()` records no properties for this type (table-level block included), so there is nothing to add.
- `cdkd drift --revert`: it replays `update()` with the recorded baseline as desired and the read-back as previous, so an out-of-band per-index toggle of a DECLARING index is reverted by the same planner (covered by the fixture's I3 arm). An out-of-band toggle of an index that declares nothing is not detected — the read-back gate's trade-off, the same one #1767 made for the throughput blocks.
- Cloud Control route (`provisionedBy: 'cc-api'`): untouched. The CC handler applies the member natively, and this change does not alter routing: routing keys on TOP-LEVEL properties, and `GlobalSecondaryIndexes` was already handled. The two fixtures asserting `provisionedBy=sdk` on a DynamoDB table (`dynamodb-gsi-update`, `dynamodb-ondemand`) keep their premise.
- No new IAM surface is documented anywhere in the repo for this provider; the calls are `dynamodb:UpdateContributorInsights` / `dynamodb:DescribeContributorInsights`, which the table-level branch already needed.

## Not verified here

- That CloudFormation also disables on a per-index block removal was NOT A/B-measured; the removal arm follows this provider's existing table-level rule.
- What `DescribeContributorInsights` reports for a never-configured index, whether it reports `ContributorInsightsMode` while `ENABLING`, and whether a read immediately after the toggle can be stale, are unmeasured — the fixture's negative control, I2 and I5 arms are what settle them on real AWS.

## Tests

- `tests/unit/provisioning/dynamodb-contributor-insights.test.ts` — the reader (both polarities, stringly booleans, every unusable shape), the planner (enable / disable / mode change / removal / swapped sides / unusable), by-name matching, strip identity, reverse-map options.
- `tests/unit/provisioning/dynamodb-table-provider-index-contributor-insights.test.ts` — through the provider with the SDK mocked by command: every write carries `IndexName` (and the table-level one does not), the member is NOT forwarded to `CreateTable` nor to a `Create` GSI action, ordering after the index Create, no call and no extra `DescribeTable` when nothing changed, the rollback swap, the masked warning, the read-back gate (one read per declaring index only), convergence and out-of-band detection through the real `calculateResourceDrift`, transient / FAILED / AccessDenied handling.
- Integ `dynamodb-gsi-update`: a new Phase 2e (the block on an index the update ADDS) and a dedicated `CdkdDynamodbGsiInsightsExample` stack (phases I1-I5) whose two indexes are created with the table and never rebuilt: enable at create with an undeclared sibling as the negative control, observed-baseline shape + drift clean, out-of-band DISABLE detected and reverted, template-baseline clean, removal + enable-with-Mode on update, destroy. Both variants synth; every jq / JMESPath expression was executed on sample data. Not run against AWS in this lane.

## Sweep (#1370 class, this provider only)

Every nested member of the `GlobalSecondaryIndex`, `LocalSecondaryIndex` and table-level shapes in `tests/fixtures/cfn-schemas/AWS-DynamoDB-Table.json` was classified. The only other dropped members are `StreamSpecification.ResourcePolicy` and `StreamSpecification.Tags`, which need a different write path (the stream ARN, re-applied after a view-type change) — filed as #3458, not fixed here.

## Round 2: the index lists are compared as unordered sets (a PRE-EXISTING phantom)

The first real-AWS run of the extended fixture passed 2e, 2b and I1-I3 and failed at I4. `DescribeTable` returned the insights table's indexes as `[giB, giA]` for a `[giA, giB]` template; every member round-tripped, but arrays are compared positionally, so the TEMPLATE baseline reported whole-list `GlobalSecondaryIndexes` drift. (The observed baseline passed at I2 because it was captured in AWS's order.) This pre-dates the PR — the #1767 arm only ever had one index — but `Closes #1782` promises a converging read-back, so it is fixed here.

`getDriftUnorderedPaths` now declares `'GlobalSecondaryIndexes[]'` and `'LocalSecondaryIndexes[]'`: the existing LEAF-ONLY form (#1783 / #1799), which sorts the list on BOTH comparison sides by a key-order-independent serialization of the entry and does NOT extend the claim to each index's order-significant `KeySchema`. Chosen over re-ordering the read-back to the desired order because it needs no edit outside the provider, converges for an `observedProperties` baseline an earlier binary captured in AWS order (both sides are sorted, pinned by a unit test), and cannot hide a real difference (entries that differ cannot be made equal by reordering — also pinned). This is the ordering half of #1812; the rest of #1812 (the `canonicalizeDriftProperties` seam, the key-order-independent `Projection` compare) is untouched.

## Review round 1 (4 reviewers on dc2dfe416, delta re-checked on the fix commit)

Fixed: the index name embedded in the refusal path (security m1); stringly `Enabled` and `Mode`-while-`ENABLING` convergence (code m1 / m2, spec nits); the `Mode: ''` gate (code n1); unpinned index-ACTIVE wait and table-level unusable warning (test M7 / M17); unreadable block on create now refused; fixture I4 floors and the exact-name teardown guard.

Declined, with reasons:
- Threading a masker into `retryOnTransientControlPlane` (its debug line prints AWS's raw message): pre-existing, eight call sites, its own change.
- Merging `indexEntriesByName` with the provider's `desiredIndexEntriesByName`: they agree on every array, and the latter is #1767's shared predicate with `getDriftUnknownPaths`.
- A rollback `DISABLE` landing while the index is still `ENABLING`, and `DISABLE` on a never-configured index at create: unmeasured AWS behaviour with table-level precedent (`rollback-replay-effective-props` creates a table with `enabled: false`); the integ run is what settles them.
- The same clean/drifted floor on the pre-existing #1767 template-baseline arm: outside this change.


go-to-k:
Real-AWS verification on the final src sha (a019d6598's src, rebased; the last commit is the ledger only): the extended `dynamodb-gsi-update` passed end to end — per-index Contributor Insights lands on a GSI the update adds (2e) and on indexes created with the table (I1), reads back without phantom drift on both baselines (I2, I4), an out-of-band DISABLE is reported and `drift --revert` re-enables it (I3), and removing one block while enabling another with `Mode: THROTTLED_KEYS` converges (I5), with a clean destroy. The FIRST live run was red at I4: `DescribeTable` returned the two indexes in the opposite order to the template, and the template-baseline comparison was order-sensitive — a phantom that pre-dates this PR and that its two-index arm exposed; round 2 declares the index lists leaf-only unordered and the re-run is green. Siblings `dynamodb-ondemand`, `rollback-replay-effective-props`, `lambda`, `dynamodb-streams`, `dynamodb-sse`, `dynamodb-tableclass-switch`, `dynamodb-ttl-attr-change` and `dynamodb-autoscaling` all passed with an empty orphan scan.

PR #3464, opened 2026-09-20
fix(dynamodb): apply and read back a Table StreamSpecification.ResourcePolicy and Tags

### LINKED ISSUE #3458: fix(dynamodb): a Table StreamSpecification.ResourcePolicy and StreamSpecification.Tags are dropped on the wire and unreadable on drift

go-to-k:
## Summary

`AWS::DynamoDB::Table` `StreamSpecification` has three CloudFormation members (`StreamViewType`, `ResourcePolicy`, `Tags`), and the SDK's `StreamSpecification` has only the first. The provider built `{StreamEnabled, StreamViewType}` on every path, so a template granting access to a table's stream through `StreamSpecification.ResourcePolicy` deployed green with no policy attached (and no tags), and `readCurrentState` could not report either.

This PR applies both members against the STREAM arn and reads them back, following the real-CloudFormation measurements recorded on the issue (both land on `LatestStreamArn`, both are re-applied to the new arn a `StreamViewType` change mints, both are removed when they leave a block whose stream stays).

Closes #3458

## Design

New pure module `src/provisioning/dynamodb-stream-members.ts` (reader, create-time refusal, planner, read-back gates, reverse map), same shape as `dynamodb-contributor-insights.ts` from #3459. The provider owns only the API calls.

- **Write timing.** After the stream exists and the table is ACTIVE: at create (the arn `waitForTableActive` reports), on an update-time enable, when either member changes, and again after a `StreamViewType` change. Tags go through `TagResource` / `UntagResource` (old-vs-new key diff), the policy through `PutResourcePolicy`, all wrapped in the provider's existing transient-control-plane retry. On an arn minted by the same operation a `ResourceNotFoundException` is also retried (bounded); on a stream that stays it is an answer and is neither retried nor tolerated.
- **A minted stream is never confused with the one before it.** A `StreamViewType` change is applied as disable / re-enable, which mints a new arn, while DynamoDB keeps reporting a disabled stream's arn as `LatestStreamArn` -- the same state a table is in when an earlier deploy disabled its stream and this one enables it again. Both arms share one resolver: the arn echoed by the enabling `UpdateTable` is used only when it differs from the one read before the update, otherwise `DescribeTable` is re-read; and if a plan has calls to make and the only arn in hand is that pre-update one (or none), the update throws rather than writing to a dead arn and recording success.
- **"The stream stays" is keyed on the LIVE stream.** When the desired view type is what the live stream already has, no `UpdateTable` goes out (DynamoDB rejects enabling an enabled stream) and the members are reconciled on that arn. That covers a block differing only in the two members or in a spelled-out `StreamEnabled` (which `drift --revert` produces; this used to fall into the enable arm), and a re-run after a deploy that enabled the stream and then failed before state recorded it -- there the previous side does not describe the live stream, so its contents are UNKNOWN rather than empty (the failed run may have put a policy): a policy the desired side does not declare is deleted, a declared one is always put, and tags cdkd cannot name are left alone. Only a stream this operation minted is planned as holding nothing. A recorded stream that was disabled out of band is re-enabled and treated as minted.
- **Nothing forwarded.** `CreateTable` / `UpdateTable` receive a block rebuilt from `StreamViewType` alone, so neither member reaches the serializer; the unit tests assert that on both commands.
- **Removal.** A member that leaves a block whose stream stays is removed: `DeleteResourcePolicy` (`PolicyNotFoundException` is success) and an untag of every previously declared key. A disabled stream / removed block gets no call against the dead arn.
- **Order is a contract.** `deletePolicy` first, then untag, then tag, `putPolicy` last. A policy can condition on `aws:ResourceTag`, so a policy that CHANGES together with the tags is deleted before the first tag call and put back after the last: neither the old nor the new policy is ever evaluated against the other side's tag set, also when a call in between fails. The price is a window with no policy, which drops that policy's `Deny` statements with its grants; nothing here is atomic, and that window is the narrower hazard. An unchanged policy stays in place across a tag-only change. A failed `create()` deletes the table it made, as before.
- **Rollback.** The planner is symmetric in its two sides, so the rollback executor's replay (sides swapped) restores the previous policy and tags through the same arms; across a view-type change the rollback mints another arn and the previous members are applied to it.
- **Unreadable member** (unresolved intrinsic, wrong container, missing `PolicyDocument`, a tag entry without a string `Key` / `Value`). Refused before `CreateTable`, naming the entry by POSITION and never by a resolved key or value; stands down under `replayingState` like the #3459 refusal. **On a template-path UPDATE it is refused too, before any call** -- a deliberate departure from #3459's warn-and-skip, taken on the security review: the member is an access grant, and a warn-and-skip deployed green while a policy the template meant to narrow stayed as broad as it was, recorded the new block as applied, and was never looked at again (the read-back gate skips an unreadable member). The two state-borne callers (`replayingState`, `desiredFromAwsReadback`) keep the warning through the masked sink, since neither has a template-side remedy: the live setting is left alone, and after a view-type change "left alone" means the PREVIOUS readable member is carried onto the new arn. An unchanged unreadable block is not refused. One unreadable tag entry makes the whole list unreadable (filtering it would send a list nobody declared). The policy text is never logged at any level, and the retry line is masked.
- **Read-back.** `GetResourcePolicy` and `ListTagsOfResource` (paginated) on the current `LatestStreamArn`, each only when the desired block declares a member cdkd would send, since each is one API call and an undeclared member must stay out of an `observedProperties` baseline (compared by key union). Emitted in the template's shape inside `StreamSpecification`; the policy document comes back in the DECLARED spelling (object, or the declared JSON string when it is semantically equal). `aws:` tags are dropped; an empty answer emits nothing, except a declared empty list. The tag list needs no `getDriftUnorderedPaths` entry: the comparator already sorts every `{Key, ...}` list on both sides (unit-tested). Both reads are documented as eventually consistent -- right after a write they can answer "absent" or with the PREVIOUS value -- and the deploy captures `observedProperties` right after the writes, so an answer that does not match the declared member is re-asked (0.5 s, 1 s, 2 s) before it is believed. A member that really drifted costs those seconds per drift run. A failed read omits the member and logs at debug, like the sibling reads.
- **`drift --revert`** replays through `update()` with the read-back as the previous side, so a deleted policy is restored by the same planner (unit test and fixture phase M3).
- **`cdkd import`.** `DynamoDBTableProvider.import()` records no properties for the table-level policy or tags either (it verifies the table and returns the physical id), so there is nothing to mirror.
- **Cloud Control untouched.** CC applies both members natively. `handledProperties` is unchanged (`StreamSpecification` was already handled at top level), so no fixture's `provisionedBy` premise moves; the four fixtures asserting it for DynamoDB (`appsync`, `dynamodb-ondemand`, `dynamodb-gsi-update`, `tags-propagation`) were checked.
- Also fixes the comment above `readCurrentState` that claimed the CFn shape of `StreamSpecification` "includes only `StreamViewType`".

## IAM

Needed only by a template that declares the member, all on the stream arn (`arn:aws:dynamodb:<region>:<account>:table/<name>/stream/*`): `dynamodb:PutResourcePolicy`, `dynamodb:GetResourcePolicy`, `dynamodb:DeleteResourcePolicy`, `dynamodb:TagResource`, `dynamodb:UntagResource`, `dynamodb:ListTagsOfResource`. No doc in the repo enumerates per-type DynamoDB permissions, so the list lives in the changelog entry.

## Residuals

A table deployed by an earlier binary has the members recorded in state but not on AWS. It gets them on its next change to the block. A record with no deploy-time capture reports the missing members as drift and `drift --revert` applies them; a record WITH a capture compares clean (the capture never held them), the same upgrade residual every silent-drop fix has.

After a `StreamViewType` change the REPLACED stream stays readable for 24 hours with whatever policy it had. cdkd makes no call against it, and the issue holds no CloudFormation measurement for that case.

## Sweep

Other consumers of the stream arn (`getAttribute`'s live `LatestStreamArn` read, the intrinsic resolver's `StreamArn` arm, the CC provider's enrichment, the ESM providers) read the recorded / live arn and are unaffected; `update()` still re-records `StreamArn` after a view-type change (asserted in the unit tests). One sibling of the class exists and is already recorded, not widened here: `AWS::DynamoDB::GlobalTable` `ReplicaStreamSpecification.ResourcePolicy` (listed as deliberately unmapped in `dynamodb-globaltable-provider.ts`, issue #1387).

## Tests

- `tests/unit/provisioning/dynamodb-stream-members.test.ts`: reader shape table, refusals by position without resolved names, planner (fresh vs same stream, order, removal, rollback symmetry, unusable handling, carry-over), gates, reverse map.
- `tests/unit/provisioning/dynamodb-table-provider-stream-members.test.ts`: a stateful DynamoDB double in which enabling mints a new arn and a disabled stream's arn keeps being reported. Asserts every policy / tag write carries the STREAM arn, the arn after a view-type change is the NEW one for all three `UpdateTable` echo behaviours, the members never ride `CreateTable` / `UpdateTable`, removal and rollback, the loud failure, masked warnings, no policy text in any log sink, and both drift baselines converging. Measured on the first revision: 22 of its then 30 cases failed against the pre-fix provider, and the test review ran 25 single-line mutations; the 8 that survived each have a case now.
- Fixture: `tests/integration/dynamodb-streams` gains a second stack, `DynamodbStreamMembersStack`, with its own table (the view-type change replaces the stream arn the first stack's EventSourceMapping is wired to). Phases M1 to M5 in `verify.sh`. Every variant was synthesized and every jq expression executed on sample data; it has NOT been run against AWS from this lane.

## Review round 1

Four reviewers (code, spec, test, security) ran on the first revision; no blockers. Taken: the dead-arn guard on the enable arm (raised by code, spec and security alike), the live-keyed "stream stays" arm (code N1, security S3), refusing an unreadable member on a template-path update (security S2), deleting a changing policy before the tag calls and correcting the order claim (security S1), re-asking a read-back that still holds the previous value (code S1), dropping the not-found tolerance on delete (code N2, test gap 4), masking the retry line (security N2), and the test gaps (both polarities of the not-found retry, the tag-call retry, the no-arn-at-all failure, the vacuous negative assertion).

Declined, with the reason: `effectiveProperties` for a skipped member (code N3) -- the template path now refuses, and the replay path follows the #3459 precedent in this provider; an early break on credential errors in the fixture's by-name reap loop (code N4) -- it mirrors the helper `dynamodb-gsi-update` already runs; a best-effort policy delete on the replaced stream (security N1) -- recorded as a residual above instead, since CloudFormation's behaviour there is unmeasured; a live rollback arm and a tag-only drift injection in the fixture (test nits) -- the rollback is the planner's symmetric arm, unit-tested on the double's end state.

## Review round 2

Delta re-check by the same four reviewers: code, spec and test clean; security raised one should-fix inside the round-1 fix, taken -- the live-keyed arm planned an undescribed live stream as EMPTY, which would have kept a grant that a failed earlier run applied and the template then withdrew. Also taken: comparing policy documents by meaning in the planner (code n1), and pinning the `StreamEnabled` clause of the live-keyed arm (test A). The `once-leak-detect` failure on the round-1 head was the repo's swallowing-handler fence on `refusalOf`'s `String(error)`; fixed.

Known and left as is (code n2): a live view type changed OUT OF BAND to something other than the recorded and desired one reaches the enable arm, and DynamoDB rejects enabling an enabled stream -- a loud failure, and `drift --revert` for it takes the view-type arm.


go-to-k:
Real-AWS verification on the final src sha (f8c26afa6's src, rebased; the last commit is the ledger only): the extended `dynamodb-streams` passed end to end — the policy and the tag land on the STREAM arn at create with both drift baselines clean (M1); a `StreamViewType` change mints a new arn that carries both (M2); an out-of-band `delete-resource-policy` is reported on `StreamSpecification.ResourcePolicy` and `drift --revert` restores it (M3); removing both members removes them on the same arn (M4); clean destroy (M5). It is RED at M1 on the pre-fix provider (no policy on the stream arn). The behaviour matches the CloudFormation A/B recorded on go-to-k/cdkd#3458. Siblings `dynamodb-stream-filter`, `lambda`, `dynamodb-gsi-update` and `dynamodb-ondemand` passed with an empty orphan scan. Acknowledging the one departure from the lane's brief stated in the body: a template-path UPDATE with an unreadable member REFUSES rather than warns — accepted, since for an access policy a warn-and-skip can leave a broader grant in place under a green deploy.

Issue #3462, opened 2026-09-20
fix(deploy): an UPDATE clears an import parameter-divergence baseline refusal and captures the decrypted readback against the placeholder Default


go-to-k:
## The shape

`cdkd import` refuses the `observedProperties` baseline for a resource whose properties depend on a template parameter that was not provably deployed at its `Default` (go-to-k/cdkd#2854, PR go-to-k/cdkd#3461), and records `observedBaselineRefused` on it. The marker's contract (go-to-k/cdkd#2944) is that a CREATE / UPDATE / replacement clears it, because a deploy "holds the template evidence the import lacked". For this refusal class that premise is false: `cdkd deploy` has no parameter input either, so the deploy binds the SAME placeholder `Default`.

`provisionResource` rebuilds the record from the template resolution on any UPDATE, which drops the marker (`src/deployment/deploy-engine.ts`, the note beside the post-provisioning capture around the `markSameGenerationBag({ ...observed })` line). The post-UPDATE capture is then positioned against the literal placeholder, pairs the AWS readback as an ordinary drifted literal, and persists it. When the UPDATE does not rewrite the placeholder-bound property, AWS still holds what the CloudFormation deploy put there - the decrypted secret.

```
state (after import):  Fn.properties.Environment.Variables.DB_PASSWORD = 'CHANGEME'
                       Fn.observedBaselineRefused = true
next deploy:           only Fn.Code changes  -> UPDATE
                       (the Lambda provider skips UpdateFunctionConfiguration on a code-only change)
saved record:          observedProperties.Environment.Variables.DB_PASSWORD = <decrypted secret>
                       observedBaselineRefused: gone
```

Measured at unit level during the security review of PR go-to-k/cdkd#3461, by copying the "a real UPDATE CLEARS the refusal" case of `tests/unit/deployment/deploy-engine-auto-refresh.test.ts` with a code-only change and a readback returning the secret.

## Why it is not fixed in PR 3461

Nothing `cdkd import` writes on the record survives the rebuild, so the import side's only lever is to persist nothing, which it already does. The fix belongs in `deploy-engine.ts`.

## Remedy direction

[...]

Issue #2854, opened 2026-09-09
fix(import): a parameter bound to its placeholder Default while the DEPLOYED value was a secret reference persists the plaintext, with no throw and no warning


go-to-k:
## The shape

`cdkd import` captures an `observedProperties` baseline for a resource whose
property resolved to a PLACEHOLDER, while the value AWS actually holds is a
decrypted secret. Nothing throws, no `{{resolve:` opener exists in any template
the import walk is handed, and no warning fires — so the refusal added by
(#2842) does not fire either, and the plaintext is persisted.

```
Child template (--migrate-from-cloudformation):
  Parameters: { DbPassword: { Type: String, Default: 'CHANGEME', NoEcho: true } }
  Resources.Fn.Properties.Environment.Variables.PW: { Ref: DbPassword }

Parent stack, as DEPLOYED:
  Parameters: { DbPassword: '{{resolve:secretsmanager:prod/db:SecretString:password}}' }
```

The child walk calls `resolveImportedProperties(childStackState,
childStateTemplate, ...)` with the TEMPLATE only, and binds parameters via
`resolveParameters(defaultOnlyParameterTemplate(template))` — so only `Default`
is visible and the parent's `Stack.Properties.Parameters` is never plumbed
through. `properties` become `{...PW: 'CHANGEME'}`. Both opener counts read 0
and `threw` is false, so the resource is captured; `readCurrentState` returns
the decrypted password; the source spells no reference, so
`subtreeHasDynamicReference` is false, `redactByPath` no-ops, and there is no
certified position for `deriveReadbackNeedles` to learn from. The plaintext
lands in `cdkd/<parent>~<child>/<region>/state.json`.

The ROOT walk reaches the same shape with no nesting: any CloudFormation stack
deployed with `--parameters DbPassword='{{resolve:...}}'` over a placeholder
`Default`. Since `cdkd import` accepts no parameter values, this is the
ORDINARY case for `--migrate-from-cloudformation`, not an exotic one.

## Why it is its own class

It is NOT (#2850): no `Fn::If`, no downgraded condition, the `Default` is not a
token, and the reference is absent from every template the walk is handed, so
(#2850)'s remedy (treat `Parameters.Default` / `Mappings` as opener SOURCES)
enumerates nothing here. It is NOT (#2852): the reference never reaches the
persisted bag at all, so there is no position to pair.

The three classes, named at the disclosure surfaces in `import.ts`:
[...]

PR #3471, opened 2026-09-20
fix(deploy,import): read a reason-less observed-baseline refusal fail closed

### LINKED ISSUE #3468: fix(deploy,import): a reason-less ARM 4 baseline refusal written by v0.290.35 is still cleared by an UPDATE or a re-import
v0.290.35 shipped go-to-k/cdkd#3461 (the ARM 4 observed-baseline refusal) WITHOUT the fix for go-to-k/cdkd#3462 (PR go-to-k/cdkd#3467). A stack imported with exactly that version therefore carries `observedBaselineRefused: true` on its parameter-dependent resources with NO `observedBaselineRefusalReason`. PR 3467 reads a reason-less marker as "an UPDATE may clear it", which is the correct reading for every marker written by 0.290.34 and older (those can only come from the other three refusal arms) - and the wrong one for these records. For them the leak of go-to-k/cdkd#3462 remains after PR 3467: a `cdkd deploy` whose update leaves the placeholder-bound property alone clears the marker and writes the decrypted readback to `observedProperties`. A re-import cannot repair the record either: after `--migrate-from-cloudformation` the CloudFormation stack is gone, so the run has no deployed parameter values, ARM 4 does not run, and the reason-less marker is cleared and a baseline captured against the placeholder.

Evidence: `gh release view v0.290.35 -R go-to-k/cdkd` lists #3461; the unit case "a REASON-LESS marker clears on the same code-only update" in `tests/unit/deployment/deploy-engine-parameter-refusal.test.ts` pins the behaviour.

Candidate fix (deliberately not built in PR 3467, where the design premise was that no released binary could write such a marker): treat a reason-less marker as an unverifiable-parameter refusal whenever the template at hand still names a declared parameter at that resource - in the deploy rebuild and in the re-import carry. That needs the dependence walk in `src/cli/commands/import-deployed-parameters.ts` (`collectParameterDependencies`) moved to a layer `src/deployment/` may import. Its cost is false stickiness for a pre-existing arm 1-3 refusal on a resource that also names a parameter; its known hole is a template edited to replace the `Ref` with the literal placeholder before the first deploy. The alternative is to accept the one-version window and document it.

Session-fit: next (not this session) -- external input: whether to build the fail-closed reading or accept and document the one-version window is the maintainer's decision; it was explicitly deferred during the design review of PR go-to-k/cdkd#3467, on a premise (no released binary writes such a marker) that the v0.290.35 release has since made false.
Severity: high -- a stack imported with 0.290.35 still persists a decrypted secret into `state.json` on an ordinary `cdkd deploy`, even with the fixed binary.
Effort: medium (M) -- moving the dependence walk across the layer boundary, both call sites, unit cases in both polarities, and an integ arm seeded with a reason-less marker.
Estimate: ~3 h -- the layer move and the seeded-record integ arm.

go-to-k:
Closes #3468

## Problem

v0.290.35 shipped the unverifiable-parameter observed-baseline refusal (#2854) without the reason field that lets it survive an in-place UPDATE (#3462 / #3467). A stack imported with that version carries `observedBaselineRefused: true` with NO `observedBaselineRefusalReason`, and #3467 reads a reason-less marker as "an UPDATE may clear it". A code-only Lambda update therefore still cleared the marker and wrote the decrypted deployed parameter value to `observedProperties`; a source-less re-import did the same.

Failing repro on main (unit): reason-less marker, template `Fn.Properties.Environment.Variables.DB_PASSWORD = {Ref: DbPassword}`, code-only UPDATE, readback answers a sentinel -> the sentinel is in the saved `observedProperties` and the marker is gone.

cdkd 0.290.36 does not change this: it ships #3467 (it has the reason field and writes it for ARM 4), but it still writes the other three arms' markers without a reason, still reads a reason-less marker as clearable, and therefore still leaks on a stack imported with 0.290.35. A stack already updated by 0.290.35 / 0.290.36 after such an import has lost the marker and no binary can detect that afterwards; `docs/import.md` now says so and names the remediation (rotate, purge noncurrent state versions, re-baseline through a spelled reference or edit `state.json` by hand; `state refresh-observed` is not a remedy).

## Fix (maintainer decision: build the fail-closed reading)

- **Deploy**: `DeployEngine.deploy` runs `stampReasonlessParameterRefusals` right after the state load, before the diff and before anything can take a readback. Every reason-less marker whose RAW template definition names a declared parameter is stamped `unverifiable-parameter` in memory (record replaced, not mutated). The in-place rebuild, the metadata-only spread, the NO_CHANGE carry and the auto-refresh then see the explicit form with no rule of their own, and any state write records it, so the record heals. Nothing is saved for the stamp itself (`--dry-run` / no-change deploys re-derive). A nested child engine gets its own raw template, so the rule runs per child. A logical id the template no longer defines is being deleted and gets no verdict.
- **Verdict**: `resourcesNamingDeclaredParameter` in the new `src/analyzer/parameter-dependence.ts` reuses the #2854 walk WHOLE (`reachableDivergentParameters` + `computeParameterTaint` with every declared parameter divergent): `Ref`, both `Fn::Sub` forms, `Fn::If` -> `Conditions` transitively, intrinsic operands, resource-level `Condition`, unclassifiable `Fn::*`, closure over attribute reads. Pseudo parameters and CDK's Rules-only `BootstrapVersion` do not count. It fails closed (every id is a yes) on a missing / unreadable template, a non-map `Resources` or `Parameters`, or a throw.
- **Layering**: the walk moved byte-identical out of `src/cli/commands/import-deployed-parameters.ts` (verified with `diff` against `git show`), because `src/deployment` may not import `src/cli`. `import.ts` and the existing walk tests changed their import specifier only.
- **Re-import**: `buildStackState` carries a reason-less prior marker as `unverifiable-parameter` on an unchanged physical id when the verdict is yes; downstream is unchanged (no source -> kept, proving source -> ARM 4 re-judges and clears).
- **Second reason value**: every writer of the marker now records a reason. The other three import arms (resolve threw / a `{{resolve:` opener was lost / a non-inert subtree was discarded) write `incomplete-resolution`; `unverifiable-parameter` wins when both fire. This bounds the fail-closed reading to older records. No reader branches on the new value (the only two readers of the field are `hasUnverifiableParameterRefusal` and `hasReasonlessBaselineRefusal` in `src/types/state.ts`); a binary that predates it sees "not unverifiable-parameter" and clears on UPDATE, which is today's behaviour. No schema bump. A reason value this binary does not know (`null`, a later binary's value) is read like an absent one: fail closed.

Writers of `observedBaselineRefused` in `src/`: `import.ts` capture refusal (reason always written), `import.ts` `buildStackState` carry (always `unverifiable-parameter`), `deploy-engine.ts` in-place rebuild (always `unverifiable-parameter`), `deploy-engine.ts` stamp (adds the reason). Sites that only PRESERVE a record (metadata-only spread, NO_CHANGE carry, selective-merge preserved rows, the v9->v10 migration, which writes no marker) keep whatever the record had: a legacy non-parameter marker stays reason-less there by design and is re-judged per deploy.

Discharge rules are unchanged from #3462 (`wasReplaced` AND a changed physical id, CREATE, proving re-import).

## Accepted cost and stated residual

- A refusal of the other class written by an OLDER cdkd, on a resource whose definition names a parameter, becomes sticky until replacement or a proving re-import.
- Residual: replacing the parameter reference with its literal before the first fixed-binary deploy that saves state hides the dependence, and the marker clears. One sentence each in `docs/import.md` and the type doc.

## Tests

- Unit, both polarities: six naming shapes + attribute closure + unclassifiable -> kept, reason stamped, no readback, sentinel in no save / provider call / log; pseudo-parameter-only, BootstrapVersion-only and "another resource names it" -> cleared exactly as before (the pre-existing "REASON-LESS clears" case is untouched); `incomplete-resolution` clears; replacement discharges; metadata-only and NO_CHANGE save the stamped form; a preserved legacy non-parameter marker stays reason-less; failed update never captures; DELETE path unaffected; `--dry-run` saves nothing. Helper fail-closed arms in `tests/unit/analyzer/parameter-dependence.test.ts`; reader contract in `tests/unit/types/baseline-refusal-reason.test.ts`; re-import carry cases in `import-deployed-parameters.test.ts`.
- Three existing #3462 cases were updated because they pinned the behaviour this PR changes (arm 1-3 now writes `incomplete-resolution`; a reason-less prior marker that names a parameter is now carried).
- Mutation probes (byte-exact restore, each went red at the intended case): the stamp predicate, the helper's catch / no-template / non-map `Resources` / non-map `Parameters` arms, the reachable-parameter narrowing, the import carry disjunct, the same-physical-id guard, the second reason writer.
- Integ (`import-deployed-parameter-secret`): step 6a strips `observedBaselineRefusalReason` from every refused record in root and child state (scope-guarded, `--if-match` rewrite, count > 0 asserted from a re-read, no body printed). `CDKD_TEST_UPDATE` changes the inline code of the root's `SecretEnvFn` AND of a new `ChildSecretEnvFn` in the nested child, so in 6b a CHILD engine UPDATEs a secret-bearing resource whose update does not rewrite the leaf (asserted: child code sha moved, AWS still holds the decrypted value, child state rewritten); 6b requires root AND child records to come back `refused` WITH the reason; 6c strips the root again so the re-import carry runs on a reason-less prior. Step 2b now waits (polled strict gone-probe) for the secret name to be free before `create-secret`. Literal floor 43 -> 52. Real-AWS results: see the comment below.

Full suite: 1091 files / 25555 tests green; `vp run check`, `typecheck:test`, build, generators clean.

## Review round 1 (spec / code / test / security, on f2e7ca0cc; all four: no blocker)

Applied: the "that writes state" qualifier on the known gap (docs + changelog); the docs now name the two cannot-tell cases (unknown intrinsic, unreadable template); three stale comments; `hasReasonlessBaselineRefusal` fails closed on an unknown reason value; the FAILED-update case now asserts the stamped reason; the replace-not-mutate contract is pinned (probe: mutating in place goes red); a child-engine redeploy arm on the integ (both the test and the security reviewer asked for it).
Declined: driving the engine itself into the unreadable-template arm in a unit test -- the engine breaks on such input before the arm matters, so the arm is pinned on the helper, which is where it lives.

Delta re-check on aec7fff62 (same four reviewers): spec clean, code clean, test clean, security clean. One nit declined: a `-z` guard on the child `lastModified` proof in `verify.sh` -- an empty read on both sides compares equal and FAILS (the safe direction), and an empty read on one side is caught by the `assert_record` on the same JSON two lines later.

## Orchestrator fix round

- F1: step 2b polls for the secret name to be free (a run right after a failed run died on "already scheduled for deletion").
- F2: the child's UPDATEd secret-bearing resource is now a code-only Lambda, with a child live premise after 6b; floor 52.
- F3: kept today's behaviour and corrected the docs. An unclassifiable bag counts only when something in the template names a declared parameter -- the same rule ARM 4 has always had. The verdict only has to cover what an older import's ARM 4 can have marked, and ARM 4 never ran on a template in which nothing names a declared parameter (`divergentParameters.size > 0` gates it), so such a marker is provably from the other class. Counting the bag whenever ANY parameter is declared would make every CDK stack's old refusals of the other class sticky (each declares `BootstrapVersion`) and protect nothing. Helper doc, type doc, `docs/import.md`, `docs/state-management.md` and the changelog now state the condition; the unit case that pins it carries the argument.
- F4: the stamp re-runs after `adoptRollbackOrphans` splices records in (unit case + probe).
- F5: one debug line when the verdict came from the fail-closed arm -- count and cause class (`unreadable-template` / `walk-threw`) only -- on the deploy and the import side (unit cases assert no template value reaches the log; probes red).
- F6: docs (a)-(c), this body (d), changelog at 2000 chars with the PR number (e).
- Declined, per the orchestrator: test G2-G4, security nits 3-5, code nit 2.

Lane reviewers on the fix round (code / test / security, same agents): clean, no blocker. Taken: the child-arm comments in the fixture now say what the arm proves (a child engine ran and stamped -- the `refused` verdicts discriminate; the child plaintext scans backstop the redaction, since under `cdkd deploy` the parent hands the child the literal reference); two wording fixes in the `docs/import.md` exposure note (a source-less re-import also lost the marker; "no command removes a baseline IN PLACE"); two comment fixes. Declined: passing the state bags into the deploy-time verdict (a deploy's bags are resolved literals; the one raw-bag shape is a selective-merge leftover whose stale leaf differs from the desired value, so the update rewrites it) -- the helper doc now says "on template evidence".


go-to-k:
Real-AWS results (us-east-1, on 305a3d30a — the src at the current head is unchanged since; the later commit is docs + the integ ledger):

| Fixture | Result |
|---|---|
| `import-deployed-parameter-secret` | PASS, 52 assertions, destroy 0 errors, 0 orphans, 0 surviving S3 object versions under both state prefixes |
| `schema-v9-to-v10-migration` | PASS (its live CLEAR phase still holds, i.e. the source-less import writes `incomplete-resolution`, not a reason-less marker) |
| `lambda` | PASS |
| `nested-stack` | PASS |
| `import-nested-stack` | PASS |

Discrimination proofs, each with one expression neutralised, the binary rebuilt, the fixture run on real AWS, then the source restored byte-exact:

- deploy stamp disabled (`if (namesParameter(logicalId) || true) continue;` in `stampReasonlessParameterRefusals`): RED at `the DECRYPTED secret is in the root state.json after the redeploy`.
- import carry disjunct disabled (`(false &&` in `buildStackState`): RED at `the DECRYPTED secret is in the root state.json after the re-import`.

Both red runs wrote the plaintext of the fixture's synthetic secret into state by design; every object version under both prefixes was purged and asserted zero afterwards.

An independent review round (spec, code, test, security) ran once on e073a624e and re-checked the fix-round delta on 305a3d30a: no blocker or major. The one finding taken after that re-check is the remediation ordering in `docs/import.md` (purge last), which is the docs commit on top.

Issue #1935, opened 2026-08-18
fix(state): a MIXED leaf holding a {{resolve:...}} token is spliced by the value scan, breaking rollback replay


go-to-k:
## Context

Found by the security review of the #1917 / #1915 lane (GHSA-p5qg-v9gv-hc7w
follow-up set). **Pre-existing on `main`** -- not a regression of that PR, which
closed the whole-value half of the same class.

## The gap

`src/deployment/secret-redaction.ts`'s sourceless walk guards a leaf that IS a
complete `{{resolve:...}}` token, but a MIXED leaf holding a token inside a
larger string still enters the substring arm and gets spliced:

- Deploy 1 persists an embedded secret as
  `jdbc://appdb:{{resolve:secretsmanager:appdb/creds:SecretString:password}}@host`
  (ordinary substring-arm output).
- Deploy 2 records an ssm SecureString whose plaintext is `appdb`.
- The sourceless walk on the journal's `previousState` then produces
  `...{{resolve:secretsmanager:{{resolve:ssm:/app/dbname}}/creds:...}}`.

`resolveReplayProps` then asks Secrets Manager for secret id
`{{resolve:ssm:/app/dbname` (its `/\{\{resolve:([^}]+)\}\}/` stops at the first
`}`) -- **rollback blocked, or garbage applied**. The same output comes from
`cdkd scrub` into both `properties` and `observedProperties`.

## The obvious fix is wrong -- read this before taking it

The first suggestion was "mask only OUTSIDE `{{resolve:...}}` spans". That is
**not** a safe generalisation on its own: a token-shaped secret PLAINTEXT
embedded in a larger leaf -- the #1917 shape -- is itself such a span, so
span-skipping would stop redacting it. That trades a mangling bug for a
disclosure.

The correct form is **two rules, not one**:

1. for each `{{resolve:...}}` span: replace it if it IS a recorded plaintext,
   otherwise leave it alone;
2. mask normally OUTSIDE the spans.

## Why it was deferred rather than folded in

It touches the fallback of all six `PathSourceRules` constants and every
sourceless walk, it has its own edges, and it is pre-existing -- so it did not
belong in a lane already on its fourth review round with a real-AWS integ
queued. Recorded here rather than carried silently.

## Verification

Unit cases per rules constant for: a mixed leaf whose token must survive intact;
a mixed leaf whose span IS a recorded plaintext (must still be replaced); and a
[...]

PR #3475, opened 2026-09-20
fix(resolver,redaction): refuse a secret assembled into an unresolvable dynamic reference, and stop sparing such a span in state

### LINKED ISSUE #2743: fix(resolver): the unsupported-dynamic-reference-service arm echoes an assembled plaintext unmasked at default verbosity, and leaves a literal {{resolve:...}} span that redaction spares into state.json
## What is wrong

`IntrinsicFunctionResolver.resolveDynamicReferences` has an `else` arm for a
`{{resolve:<service>:...}}` token whose service is none of `secretsmanager` /
`ssm` / `ssm-secure`:

```ts
// src/deployment/intrinsic-function-resolver.ts:7095-7096
} else {
  continue;
}
```

`service` is `inner.split(':')[0]`, and `resolveSub` / `resolveJoin` re-enter
`resolveDynamicReferences` with the ASSEMBLED string. So a template that lets a
resolved value land in the SERVICE position produces a token whose service IS
that value:

```yaml
Fn::Sub:
  - "{{resolve:${Pw}}}"
  - Pw: "{{resolve:secretsmanager:MySecret:SecretString:password}}"
```

The `Pw` variable resolves first and records the password into
`context.recordedSecretValues`. The re-entry then matches
`{{resolve:<password>}}`, and this arm fires. Two consequences:

**1. The warn echoes the plaintext, unmasked, at DEFAULT verbosity.**
Every sibling echo in this same function goes through `maskSecretsForLog`
(`:7234`, `:7253`, `:7486`, `:7501`, `:7547`), and `context` is in scope here —
it is used two statements later at `:7107`. This one line was missed.

**2. `continue` leaves the literal token in the value, and redaction SPARES
it.** The arm does not substitute, so the resolved string keeps
`{{resolve:<password>}}` verbatim. `redactOutputs`' substring rule
(`src/deployment/secret-redaction.ts:4610-4625`, issue #1935) replaces every
recorded-plaintext match EXCEPT one lying "strictly inside a complete
`{{resolve:...}}` span" — and `<password>` is strictly inside, and shorter
than, `{{resolve:<password>}}`. The carve-out exists so a plaintext that
coincides with text inside a REAL token is not spliced into that token; here
the span is a bogus token, so the carve-out hands the plaintext straight
through to `state.json` outputs and the exports index.

That is the GHSA-p5qg-v9gv-hc7w class (dynamic-reference plaintext persisted to
state), reached through a path the #1935 carve-out was not written against.

## Where it was found

Independently reported by the spec and security reviewers of PR #2742 (the
`fix(deploy-engine)`/#2728 masking PR), which closed the same echo class at the
four other sites in this function but not at this arm. Consequence 2 is out of
that PR's scope.

## Suggested fix

- Mask the warn: `this.logger.warn(this.maskSecretsForLog(..., context))`.
- Decide what the arm should DO with an unrecognized service. Leaving the
  literal token in the value is what makes redaction's span carve-out
  load-bearing on attacker-shaped text. Either refuse (an
  `IntrinsicResolutionRefusalError`, matching the `ssm-secure`-on-a-public-type
  arm at `:7086`), or substitute a placeholder so no `{{resolve:...}}` span
  survives into the persisted value.
- Whichever is chosen, add the negative to the redaction side too: a span whose
  service is not a KNOWN service should not qualify for the "strictly inside a
  complete `{{resolve:...}}` span" carve-out.

## Verification

`/run-integ secrets-dynamic-ref` with a new gated arm (same shape as the
`CDKD_TEST_OUTPUT_LEAK` arm PR #2742 added): declare an output whose `Fn::Sub`
puts the resolved password in the service position, then assert the warn is
masked AND that the persisted `state.json` carries no plaintext. Plus unit
cases in the resolver and `secret-redaction` suites.

## Classification

Session-fit: next (not this session) - the session that found this was reviewing
PR #2742; this is a pre-existing defect in a different arm, and closing it needs
a new integ arm written plus a redaction-side decision, neither of which rides
the review's verification.
Severity: medium - a secret plaintext reaches a default-verbosity log line and
`state.json`, but only for a template that places a resolved dynamic reference
in another reference's SERVICE position, which no ordinary CDK app produces.
Effort: medium (M) - an existing fixture (`secrets-dynamic-ref`) gains a new
gated arm, and the redaction carve-out change needs its own unit cases.
Estimate: ~1-2 h - most of it is deciding the refuse-vs-substitute semantics and
proving the redaction negative, not the run (the fixture's median run is minutes).

Dup-check: searched open + closed issues for "Unsupported dynamic reference
service" (hits #2166, #2482 - neither is this arm) and for the persistence half
(#2485, #1917 - #1917 is the inverse, a plaintext that IS a `{{resolve:...}}`
string; #2485 is the shared-plaintext sibling class). Also read the code around
`:7095` and around the `redactOutputs` carve-out (`secret-redaction.ts:4610`) for
an issue number cited in a comment: the carve-out cites #1935, which is the
carve-out's own defect, not this path. No duplicate.

go-to-k:
## What

An `Fn::Sub` / `Fn::Join` that puts a resolved secret in a dynamic reference's SERVICE position assembles `{{resolve:<plaintext>}}`:

```yaml
Fn::Sub:
  - '{{resolve:${Pw}}}'
  - Pw: '{{resolve:secretsmanager:MySecret:SecretString:password}}'
```

The resolver's unsupported-service arm warned (masked since #2728) and left that token in the value. Reproduced on `main` with the real engine, resolver and redaction: the saved state held `"outputs":{"Leak":"{{resolve:cdkd-known-pw-123}}"}`. Readers of that value: `state.outputs` / the exports index, the deploy summary's `Outputs:` lines (default verbosity), and in a resource property the PROVIDER (AWS stores the token, plaintext included, under a green deploy), then `properties`, `observedProperties` and the rollback journal.

Consequence 1 of #2743 (the unmasked warn) was already closed by #2728 and is re-verified here; this PR closes consequence 2.

## Fix, two layers

1. **Resolver** (`intrinsic-function-resolver.ts`): on the TEMPLATE route the arm refuses a TAINTED token with a masked, non-retryable `IntrinsicResolutionRefusalError`. Tainted = the log twin masks part of the token (floor-free; an unpaired twin is the bare mask and fails closed) OR the needle mask changes the raw token (covers a plaintext no twin describes, e.g. the inherited bag). An untainted look-alike (`{{resolve:notaservice:/x}}`) keeps warn-and-leave. **State-driven paths are unchanged:** the public `resolveDynamicReferences(leaf, ctx)` entry point is what every PERSISTED-text caller already uses (`cdkd drift` / `--revert`, the rollback replay, scrub's per-token delegation, the cross-stack re-resolution), and it never refuses in this arm; template leaves go through a private refusing twin of it. No edit to `drift.ts`, `rollback-executor.ts` or `scrub.ts`. Why: AWS already holds persisted text, so a refusal buys no confidentiality, and it would be deterministic -- `resolveReplayProps` shares one secrets map per op, so a journal `previousState` holding `{{resolve:<P>}}` beside the reference resolving to `P` would fail its rollback op on every retry (found by the orchestrator's code review). Both predicate halves are skipped there explicitly; only the needle half could fire, since that entry passes the value as its own twin.
2. **Redaction** (`secret-redaction.ts`): a needle inside a `{{resolve:...}}` span is spared only when the span names a service cdkd resolves (the existing `SECRET_BEARING_REFERENCE_PREFIXES`). Narrowed at every site that kept a token-shaped leaf holding resolved text: the value scan's whole-token early-out and `scanLeaf`, the path pass's cross-generation whole-token arm, and `refuseUncertifiedSubtree`. This covers what layer 1 cannot: a readback, a record an older release leaked into, and `cdkd scrub`'s repair (it calls `redactSecretsForState`; `scrub.ts` is not edited). The result `{{resolve:{{resolve:secretsmanager:...}}}}` is unresolvable text but not plaintext, and the reference nested in it stays protected from later splices as long as it is an expression of the pass's own map (`resolvableReferenceSpans` re-runs the shared token scan from each resolvable-service opener; no new pattern is built). An UNVOUCHED nested token is not protected, because it is also what a secret assembled one level deeper looks like (`{{resolve:x-{{resolve:ssm:<plaintext>}}}}`), and redacting wins over keeping already-unresolvable text intact. The stray-opener residual narrows as a side effect: it now needs the opener to spell a resolvable service.

Refuse-all was rejected: `cdkd drift` relies on warn-and-leave for untainted look-alikes (its `unresolvedToken` reporting is built on it). "Substitute a placeholder so no span survives" was rejected: the placeholder would still be SENT to AWS as the property's value, and an inline mask in persisted text is indistinguishable from a user literal that `drift --revert` / the replay would then push. "Redaction only" was rejected: it leaves the provider call in place, so the plaintext still reaches AWS, and a sub-floor secret has no needle for redaction to match.

Known limits, each pinned by a test or stated in a code comment:
- A look-alike token whose text merely coincides with a recorded secret of 4+ characters is refused when the secret resolved BEFORE the token; when it resolves after, the token is not refused and the value scan rewrites that text in state like any other text (the same trade the substring arm already makes for ordinary text).
- A plaintext holding a `}` forms no token under the shared `[^}]+` grammar, so the resolver has nothing to refuse; the value scan redacts it in state as ordinary text.
- A legacy PRODUCER output still holding `{{resolve:<plaintext>}}` is handed to a consumer as written through `Fn::ImportValue` / `Fn::GetStackOutput`: that site reads persisted text and cannot tell the token from a look-alike (the consumer usually holds no needle). Pinned as intentional; the docs say to scrub the producer first.
- The healed text `{{resolve:{{resolve:secretsmanager:...}}}}` does not round-trip (the resolver reads it as one unresolvable token), so the documented cleanup order is strict: deploy the corrected template, THEN `cdkd scrub`. Measured in `drift-secret-redaction.test.ts`: a record scrubbed first is reported `unresolvedToken` at that leaf (no drift row, nothing for `--accept`, the live value never displayed); a rollback or `drift --revert` touching the resource writes the unresolvable text. Reasoned, not run: on the persisted route, a leaked token walked BEFORE the reference that records its needle is echoed by the masked-against-the-bag warn with nothing yet in the bag, as on `main`.
- Pre-existing and out of scope: a plaintext in the KEY position of a RESOLVABLE service (`{{resolve:ssm-secure:/x/${Pw}}}`) is sent as the lookup name (SSM API / CloudTrail) and, if it resolves, the expression is persisted and spared by #1935.
- What CloudFormation does with an unsupported dynamic-reference service is unmeasured; the design does not depend on it.

## Behavior change

A deploy that used to go green while sending `{{resolve:<plaintext>}}` to AWS now fails that resource before its provider is called (and rolls back like any other failure). For an Output it warns, skips the output and exits 0 -- except under `--strict-getatt`, where a failed output fails the deploy.

## Tests

- `deploy-engine-unresolvable-service-span.test.ts` (new; real engine + resolver + redaction): the provider is never called for the refused resource; sentinel absent from the thrown chain, every logger spy, saved state and the journal; the completed sibling settles identically to a control non-retryable failure, with the point of failure (provider vs resolution) asserted apart from the compared object.
- `deploy-engine-output-failure-masking.test.ts`: the "SERVICE position" case flipped from masked warn to masked refusal (default + strict arms), asserting the per-output warn verbatim, `DeployResult.outputs` as `deploy.ts` renders it, saved state, and all spies.
- `dynamic-references.test.ts`: twin half alone (sub-floor), needle half alone (recorded + inherited bag), untainted control, a negative control for sub-floor substrings, an unpaired twin failing closed, per-token-recovery callers still get the refusal.
- `intrinsic-resolver-name-argument-log-twin.test.ts`: the two unsupported-service cases flipped, masking assertions kept.
- `secret-redaction-unresolvable-service-span.test.ts` (new): whole-leaf + embedded shapes, both polarities, idempotence, GOLDEN outputs produced by running `origin/main`'s `redactSecretsForState` over the #1935 and mangled-legacy inputs, scrub/journal readers, a whole-token source under every rules constant, a later needle occurring in the written reference (idempotence), near-miss service spellings, a `}`-holding plaintext.
- `secret-redaction-uncertified-fail-closed.test.ts`, `scrub-deleted-output.test.ts`: one case each for the unresolvable-service token.
- `secret-redaction-mixed-leaf-spans.test.ts`: the two stray-opener residual rows now carry an unknown-service (replaced) and a known-service (kept) leaf each.

Mutation probes (one per arm, byte-exact restore, tree clean after each): predicate -> `false` (11 red), needle half dropped (3 red, all needle cases), twin half dropped (3 red, all sub-floor cases), early-out gate dropped (whole-leaf red, embedded green), `scanLeaf` gate dropped (embedded red).

- Orchestrator round: `rollback-executor-log-masking.test.ts` (the reviewer's scenario through the real `resolveReplayProps`, both key orders: op succeeds, literal re-sent unchanged, record redacted afterwards), `drift-secret-redaction.test.ts` (persisted token with the needle already recorded is `unresolvedToken`, not `refused`; the scrubbed-first measurement), `cross-stack-secret-reresolve.test.ts` (the producer limit), the deployment-events sink assertion, a `}}`-holding plaintext, the exact expected output on every rules constant.

Declined, recorded: pinning the resolver's service set to `SECRET_BEARING_REFERENCE_PREFIXES` (module-private constant in a leaf module; a fourth service needs a resolver arm anyway); the lane test reviewer's per-leaf `toBe` on the nested case (take-or-leave); orchestrator test G3, G5, G6, G7 and code m3, n4.

Mutation probes after the review round: nested-reference scan disabled (2 red), path-pass gate removed (1 red), `refuseUncertifiedSubtree` gate removed (1 red). Round 2: the nested scan's `start === 0` guard relaxed (1 red), its vouching guard removed (2 red). Orchestrator round: persisted-text signal forced off (5 red: both replay orders, the drift case, the entry-point case, the producer pin); `resolveValue` routed through the persisted entry (3 red, the needle cases); predicate off (11 red, the events-sink case among them).

## Integ (written, NOT run in this lane)

`secrets-dynamic-ref` gains three phases:
- Phase 1b3, `CDKD_TEST_SERVICE_SPAN=output`: exit 0, `Failed to resolve output ServiceSpanLeak: Refusing to resolve {{resolve:***}}`, no password in the `--verbose` log (the `Outputs:` summary included) or `state.json`, no `ServiceSpanLeak` key.
- Phase 1b4, `CDKD_TEST_SERVICE_SPAN=resource`: an `AWS::IAM::Role` whose `Description` is the token. The phase first PROVES the vehicle with a throwaway role (`cdkd-test-dynref-span-premise-<acct>`: create with `--description '{{resolve:not-a-secret}}'`, read it back equal, delete; also swept by `cleanup` by exact name), so an absent role means cdkd refused. Probed gone before and after, deploy exits non-zero with the refusal, no state record; the rollback journal (when present) and every `deployments/` object are scanned too, with a floor of 2 objects and at least one carrying the masked refusal.
- State is checked by the SET OF PATHS holding the password (`assert_password_paths`), never a whole-file grep: the fixture's own Secret is declared from `unsafePlainText`, so `resources.<Secret>.properties.SecretString` holds it by design. Exactly that path (matched by resource TYPE) must hold it in state.json; none in the journal or events. Failures print paths, never values.
- Phase 1b5 (layer 2, which the refusal pre-empts in the two phases above): seeds `outputs.ServiceSpanLegacy = {{resolve:<password>}}` into `state.json` (premise: that key equals the seeded text), runs a real `cdkd scrub`, asserts the password is gone and the key reads `{{resolve:{{resolve:secretsmanager:...}}}}`, then drops the key. The same flow is unit-pinned in `scrub-deleted-output.test.ts`.

The two modes of the gate cannot share a deploy: a failing output freezes the outputs bag, and a failing resource skips the outputs pass.

Closes #2743


go-to-k:
Real-AWS results (us-east-1, on e61e63d78 — the later commit on the branch is the integ ledger only):

| Fixture | Result |
|---|---|
| `secrets-dynamic-ref` (with the new phases 1b3 output / 1b4 resource / 1b5 scrub) | PASS, 479s, destroy clean, 0 surviving S3 object versions |
| `lambda` (broad) | PASS |

Discrimination proofs, each with one expression neutralised, the binary rebuilt, the fixture run on real AWS, then the source restored byte-exact:

- layer 1 disabled (the arm's `!persistedText && (twin-differs || needle-masks)` forced false): RED at `the probe deploy did not report 'Failed to resolve output ServiceSpanLeak: Refusing to resolve {{resolve:***}}'`.
- layer 2 disabled (`spanNamesResolvableService` forced true): RED at `state.json after 'cdkd scrub' over the seeded {{resolve:<password>}} output: the resolved password sits at an unexpected set of paths`.

Phase 1b4's premise is now proven inside the run rather than assumed: IAM accepts and returns a `{{resolve:...}}` role Description, so "the probe role is absent" means cdkd refused.

An independent review round (spec, code, test, security) ran on 49db321d8 and re-checked the delta on 0d1542cfa: no blocker, no major, and the rollback-replay finding was resolved by exempting persisted text from the refusal. The nits taken after that re-check are in e61e63d78: a content floor on the path scan (an empty document used to pass the negative for free), a literal rather than regex mask when printing a path, a poll around the IAM premise read, and `persistedText` propagated to the sibling delegation.

Orphan audit after the run: no probe roles, no fixture stack, no fixture secret, 0 state object versions under the fixture prefix.

